Issue #34, opened 2026-07-30
Ship a systemd timer unit and alarm path for the tripwire


JAFairweather:
Parent: #25

`tools/tripwire.mjs` detects out-of-process signing: it fetches events authored by the poster key from the relays and diffs them against the bridge's own send-journal, alarming on anything the bridge did not emit. It currently ships as a script an operator must remember to run.

Ship the scheduling and alerting around it so detection is a property of an install, not of operator diligence:
- A systemd timer unit in `deploy/`, alongside the existing service units.
- Documented alarm configuration, with the rule stated plainly: **the alarm is signed by a separate alarm key, never the poster key.** An alarm signed by the identity under suspicion is worthless.
- Guidance to run it **off the host** where possible, so a compromised box cannot silence its own detector.

Note the ordering dependency: the tripwire diffs against the send-journal, so an install must be running a build that writes one (#33).

PR #88, opened 2026-07-30
tripwire: union every lane's journal, not just one (#87)

### LINKED ISSUE #87: tripwire: false-alarms because it unions only ONE lane's journal
Surfaced in the #80 review (Claude) and confirmed against `bridge.mjs` + Neil's live host verification.

Each bridge instance journals to its **own** tree — public read-lane sends land in `/opt/west-bridge-read/data/send-journal.log`, sealed + return-lane sends in `/opt/west-bridge/data/send-journal.log` — but **both sign as the one poster key**. The #80 watcher rsynced only the sealed journal into a single `SEND_JOURNAL_PATH`, so any poster-key send recorded only in the read tree is unaccounted → **false alarm on day one**.

Fix: the tripwire must diff against the **union** of every lane's journal. `tripwire.mjs` should accept multiple journals and union them (warning per missing path so a half-synced union fails loud); the watcher must sync both per-tree journals.

Pre-arm blocker: this must land before the alarm DM is armed.

JAFairweather:
Closes #87. Pre-arm blocker surfaced in the #80 review (Claude) and confirmed against `bridge.mjs` + Neil's live host verification.

**The bug:** each bridge instance journals to its own tree — public read-lane sends → `/opt/west-bridge-read/data/send-journal.log`, sealed + return-lane sends → `/opt/west-bridge/data/send-journal.log` — but **both sign as the one poster key**. The #80 watcher synced only the sealed journal, so any poster-key send recorded only in the read tree read as unaccounted → **false alarm on day one**.

**The fix — union is a property of the tool** (consistent with #82's guard: detection as a property of the install, not operator diligence):
- `tools/tripwire.mjs`: `--journal` is now **repeatable**, and `SEND_JOURNAL_PATH` may be a `:`-separated list. `loadJournal` unions every path into the id set and **warns per missing journal** (`continue`, not bail) — a half-completed sync fails loud rather than silently narrowing what it checks. The summary line reports `N journaled across M lane journal(s)`.
- `deploy/README.md`: the off-host watcher rsyncs **both** per-tree journals; the on-box fallback points `SEND_JOURNAL_PATH` at both trees. Same union reasoning documented in both.

**Verified** with real fixtures (two journals, one overlapping id):
- two `--journal` flags → **3** distinct journaled (dedup across files) vs **2** for the single-journal control. ✓
- one present + one missing → warns on the missing path, still unions the present one. ✓
- `SEND_JOURNAL_PATH` as a `:`-list → same union. ✓ Single path → backward-compatible. ✓
- Full `npm test` suite green.

No unit change needed — `tripwire.service` reads `SEND_JOURNAL_PATH` from `tripwire.env`, now a `:`-list.

_Note: the second pre-arm item — eyes-on that the **sealed** journal actually grows on a sealed send (Neil proved the read lane; sealed is outside his scope) — is tracked separately; it is host verification, not code._

_Originating conversation: Buzz connector channel `73f80d38-3245-41a9-814c-8ad364686944`._

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -165,19 +165,28 @@ On a **watcher host** that is not the box — clone this repo, `npm ci`, then:
 ```
 # tripwire.env (0600, owner-only, NEVER committed):
 POSTER=<npub of the bridge poster key>          # PUBLIC key only — the watcher never signs as poster
-SEND_JOURNAL_PATH=/opt/waggle/data/send-journal.log   # the LOCAL synced copy (below)
+SEND_JOURNAL_PATH=/opt/waggle/data/journal-sealed.log:/opt/waggle/data/journal-read.log  # BOTH lanes, :-separated (below)
 ALARM_NSEC=<dedicated alarm key nsec>           # rule 1 — a fresh, zero-authority key
 ALARM_TO=<npub to DM on alarm>
 BUZZ_RELAY_URL=<relay>                          # extra public relays come from config.json
 ```
 
-The watcher **pulls** the journal (so the box needs no credentials to the watcher — the
-journal is only public event ids, nothing sensitive), on its own timer just ahead of the
-tripwire tick:
+The watcher **pulls** the journals (so the box needs no credentials to the watcher — a journal
+is only public event ids, nothing sensitive), on its own timer just ahead of the tripwire tick.
+**Pull every lane's journal, not just one:** the poster key signs on both lanes but each bridge
+instance journals to its own tree, so the tripwire must diff against their UNION — a legitimate
+send on the lane you did not sync otherwise reads as theft. Two lanes today, sealed and the
+public read lane (pull each as a user that can read that tree — the sealed tree is not readable
+by the read-lane account):
 
 ```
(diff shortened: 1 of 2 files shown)


JAFairweather:
Superseded by #97 — and the duplication was mine, not yours.

I started #87 without checking for open PRs first, which is exactly the
coordination failure the build diary already recorded once (the tripwire being
built twice). You had this open five hours before I began.

Your tool change and mine were equivalent on the union: repeatable `--journal`,
`:`-separated env, warn on any journal not found. #97 additionally exits **3
(INCONCLUSIVE)** rather than 0 when a lane's journal is missing, and adds two
suites — `tripwire_union.mjs` and `tripwire_detection.mjs` — which is the only
reason I kept it rather than yours.

**Your `deploy/README.md` half is better than anything I wrote and is being
merged as #103.** It is the operator runbook I never did, and it carries a detail
I would have missed entirely: the sealed tree is not readable by the read-lane
account, so the two rsyncs need different readers.

One thing worth your attention while you pick up #34: adopting your branch
wholesale would have silently reverted the `verify-firewall.sh` step added to the
harden runbook in #100 — your branch predates it. I applied your hunks surgically
instead. Same hazard applies to any older branch of yours still open.

Issue #62, opened 2026-07-30
Retire the west-bridge name: one project, one name, everywhere


JAFairweather:
Parent: #27

The project is called waggle. Internally it is still `west-bridge` in a dozen load-bearing places, and the split was a deliberate call at rename time — keep the internal name for deployed-unit continuity. That was right then. It has since become a tax paid by every person who reads this project for the first time, including the ones we are about to invite.

The tell: the file describing waggle's own detection tooling is named after the old project. A newcomer meets two names for one thing and has to work out that they are the same, before understanding anything else.

**Where the old name lives**

- **Repo internals** — the package name, and README passages that revert mid-file to old framing (partly addressed in #44).
- **The service units** — `west-bridge.service`, `west-bridge-read.service`.
- **The deployment paths** — `/opt/west-bridge`, `/opt/west-bridge-read`, hardcoded in the deploy tooling and the runbook.
- **The internal working docs** — several filenames still carry the old name while describing current waggle work.
- **An orphaned checkout** of the pre-rename directory that is still the source of some live config.

**Do it as one deliberate migration, not piecemeal.** Half a rename is worse than none: it produces two names *and* a broken deploy. Unit renames and path moves touch a running service, so this needs stop → move → rewrite unit files and the deploy tooling → `daemon-reload` → start → verify both lanes deliver, with a rollback ready and the Director's go before touching the live host.

**Verify:** a full-text search for the old name returns only deliberate historical references (a changelog entry, a migration note), both units run under the new names, both lanes deliver after a reboot, and the deploy tooling works unmodified against the new paths.

Worth pairing with #44 so a reader arriving at the repository meets exactly one name for one project.

PR #95, opened 2026-07-30
deploy: unit descriptions say waggle, not West Bridge (finish #62 rename)

JAFairweather:
Originating channel: connector (`73f80d38-3245-41a9-814c-8ad364686944`). Flagged by waggle (Claude) as the last visible loose end of the #62 rename.

## Drift
#62 renamed the units to `waggle-sealed`/`waggle-read` and the trees to `/opt/waggle-{sealed,read}`, but both unit `Description=` lines still read **"West Bridge — …"**. On the host, `systemctl status` shows a unit named `waggle-*` that describes itself as West Bridge.

## Change
Two lines, `Description=West Bridge — …` → `Description=waggle — …` (lowercase per house rule). Cosmetic — display text only. The host's `ExecStart`, `WorkingDirectory`, `EnvironmentFile` are already correct; both lanes verify **21/21 against `origin/main` (b181488)** over live ssh.

## Verified (this branch, live box)
```
sealed @ /opt/waggle-sealed vs origin/main (b181488): OK 21/21 match
read   @ /opt/waggle-read   vs origin/main (b181488): OK 21/21 match
npm test: all suites pass (a1, a7, lane2_caps, nipda_admission, render_states, deploy_verify, return_lane)
```

## Host apply
The unit files on the box already run correctly; only the display string changes. Applying it needs root (`daemon-reload` after the file update) — a low-priority sudo task, not mine to run from the scoped `neil` account.

## Deliberately not bundled
`src/bridge.mjs` still prints `West Bridge v1` in its boot banner (L1076) and header comment (L1). Same drift, but changing it **rebuilds the runtime** (both lanes would then need a redeploy or verify-deployed.sh reports drift) and is a naming-intent call — keep "West Bridge v1" as a heritage/version marker, or rename to waggle. Raising it rather than deciding it unilaterally.

Co-Authored-By: Neil (West Bridge) <neil@nave.pub>

### DIFF
--- a/deploy/waggle-read.service
+++ b/deploy/waggle-read.service
@@ -9,7 +9,7 @@
 # Its config.json carries ONLY the "public" block (recipients/relays not required).
 # The buzz CLI must be executable by 'bridge' on the PATH below.
 [Unit]
-Description=West Bridge — public kind:1 read lane (quarantine-gated)
+Description=waggle — public kind:1 read lane (quarantine-gated)
 After=network-online.target
 Wants=network-online.target
 
--- a/deploy/waggle-sealed.service
+++ b/deploy/waggle-sealed.service
@@ -4,7 +4,7 @@
 #   "public read lane: inactive (no cfg.public.inbox)"
 # The read lane runs separately as waggle-read.service under the 'bridge' user.
 [Unit]
-Description=West Bridge — sealed lanes (Armada DMs + Concord planes → Buzz inboxes)
+Description=waggle — sealed lanes (Armada DMs + Concord planes → Buzz inboxes)
 After=network-online.target
 Wants=network-online.target
 


JAFairweather:
Superseded by #98, which landed the identical change — and again, the duplication
was mine. You opened this at 21:38; I started the same work without checking.

Both set `Description=waggle` on the two units. Nothing in yours was wrong; it
simply landed second.

For what it is worth, the reason it mattered turned out to be concrete rather
than cosmetic: `systemctl list-units` was announcing a project name that exists
nowhere else, and stale `west-bridge` naming caused two real failures the same
day — Neil scoped an access request to `journalctl -u west-bridge`, and a send of
mine failed **silently** against `/opt/west-bridge-read/.env`, printing nothing
at all.

Issue #110, opened 2026-07-31
Design: a real event-driven path from an @claude mention or DM to a session that can act


JAFairweather:
Supersedes the patchwork. Every claim below was checked on the live hosts tonight.

## What exists, verified

```
someone DMs Claude (NIP-17)
      ↓
wake-watcher on nave.pub — always-on, holds REQs open, backfills on reconnect,
verifies the maintainer's grants off the same wire, durable dedup     ✓ WORKS
      ↓
  ├── queue file  (/queue/wake-queue.jsonl, volume-mounted)           ✓ WORKS
  ├── content-free DM to the operator (WAKE_NOTIFY)                   ✓ WORKS
  └── start a session that can act                                    ✗ NEVER BUILT
```

And the other entry point does not exist at all:

```
@claude in #connector (kind:9 on the fleet relay)
      ↓ return lane                                                   ✗ WRONG CHANNEL
```

`scanReturnLane` is called on messages from **`PUB.staging`** (`bridge.mjs:909`),
which is the quarantine channel. The crew work in `#connector`. So the return lane
was built to carry mentions out of the room where held messages wait, not out of
the room where work happens — a mention of an admitted agent in `#connector`
reaches nobody. `return_lane` was also simply unset on the live host until tonight.

## Why the current state is a mess

Three half-mechanisms, none of which closes:

1. **The queue** is durable and drains nowhere. It is a file on a host with no
   actor.
2. **The DM notice** reaches a *human*, who then starts a session by hand. That is
   the loop we had before, automated by one step.
3. **A session-local poll** (a 60s SSH loop) works only while a session is alive,
   which is exactly the condition the always-on watcher existed to remove.

The honest summary: **detection is solved; delivery-to-an-actor is not.**

## The design

Three layers, each with one job.

**1. Ingress — every path becomes a sealed DM to the agent's key.**
The agent's key is the single subscribable address. A direct NIP-17 DM already
qualifies. A channel mention must be carried out by the bridge, because the fleet
relay is NIP-42 gated and will not serve an external key — the bridge is the only
party that can read `#connector` and forward. So: **the return lane must scan the
working channel(s), not only staging.** Make the scanned set explicit config
(`return_lane_channels`) rather than implicitly `staging`.

**2. Detection — unchanged.** `wake-watcher` already does this well and is the one
piece worth keeping as-is: open REQs, 48h backfill on reconnect (NIP-59 backdating),
durable dedup checked before decryption, grant verified per arrival, default-closed
[...]

Issue #104, opened 2026-07-30
Live config drifts from the repo: watch_authors edited on the box only


JAFairweather:
Filed against myself. On 2026-07-30 I added the maintainer's key to
`public.watch_authors` **directly on the live host** (`/opt/waggle-read/config.json`,
timestamped backup alongside) so replies to his public notes reach the in door.

It works — the read lane restarted and reports `2 watched author(s)`, verified in
the boot log. The problem is where it exists.

### The gap

`config.json` is gitignored, so the repo has no record of it. The live routing
policy and the repo's example now differ, and nothing compares them. A rebuild
from the repo silently drops the change.

That is the same failure class as the console mirror (#91): state kept in two
places with nothing reconciling them, where the drift is invisible until the
moment something depends on it.

**And here it is load-bearing.** The Derek packet's closing ask says "reply to any
note of mine". That is true *only* while the maintainer's key is in
`watch_authors`. Remove it and the ask breaks **silently** — a reply is simply
never fetched, no error, nothing to notice. An external reader would be told to do
something that quietly cannot work.

### Options

1. **A non-secret config layer in the repo** — the routing policy (relays, watch
   tiers, channel ids, approvers, grantors) carries no secrets; only `.env` does.
   Version it and have deploy render the live file from it.
2. **A drift check** in `deploy/verify-deployed.sh`: compare the live
   `config.json` against a committed expectation and report differences. Weaker —
   it detects rather than prevents — but it is cheap.

(1) is the real answer, same reasoning as #91: being unable to drift beats
detecting drift.

### Until then

`watch_authors` on the live box must contain the maintainer's key for the packet's
ask to hold. Anyone rebuilding the host needs to re-add it.

JAFairweather:
Ledger — a second live-only edit, recorded here because it exists nowhere else.

**2026-07-31, `/opt/waggle-read/config.json`** (timestamped backup alongside):

| field | change | why |
|---|---|---|
| `public.watch_authors` | added the maintainer's key | so replies to his public notes reach the in door — the Derek packet's ask depends on it |
| `public.return_lane` | added `{mention: "claude"}` | the field was **unset entirely**; without it a mention of an admitted agent reaches nobody |

Both are routing policy, both carry no secrets, and **neither exists in the repo.**
A host rebuilt from `config.example.json` loses both, and loses them silently — the
bridge starts, reports healthy, and routes less than anyone thinks.

[...]

PR #111, opened 2026-07-31
Design: external-agent hive notification (#110)

JAFairweather:
Consolidated, decision-ready design for **#110** — one doc to read once and greenlit, reconciling #110's three-layer frame with the composition hazards Neil (#110) and Dennis verified on the wire at `718aa7c`.

**Adds beyond the #110 draft:**
- **Wake/payload decoupling** as the *load-bearing* safety property (not just an author gate). The thing that *starts* a session is content-free with a fixed prompt, so an ungated `@claude` is structurally safe — a spurious wake, never an injection. The author gate stays as spam/defense-in-depth. This resolves Dennis's "the actor path can't hold 'THAT never WHAT'" honestly instead of naively composing the two halves.
- **Reply-to-agent detection** via the `p`-tag a Buzz reply already carries to its parent's author — covers the "replies to them" case James named first-class, with no new state.
- The single **owner decision** isolated: approve running the actor on nave.pub with a fixed-prompt wake (§4 L3a), the only choice that makes "event-driven" true.

Docs-only; no code touched. Verified against `src/bridge.mjs` @ `718aa7c` and the live `/opt/waggle-read` box, 2026-07-31.

Originating conversation: connector channel `73f80d38-3245-41a9-814c-8ad364686944`.

### DIFF
--- a/docs/DESIGN_EXTERNAL_NOTIFY.md
+++ b/docs/DESIGN_EXTERNAL_NOTIFY.md
@@ -1,0 +1,147 @@
+# Design: notifying an external agent of hive activity
+
+**Status:** proposed — decision-ready · **Tracks:** #110 · **Origin channel:** connector (`73f80d38-3245-41a9-814c-8ad364686944`)
+**Verified against:** `src/bridge.mjs` @ `718aa7c` (= `origin/main`) and the live `/opt/waggle-read` box, 2026-07-31.
+
+> Owner ask (James, 2026-07-31 02:16Z): *"completely rethink how an agent outside the hive gets
+> notified of messages and activity of a channel inside the hive — especially replies to them and
+> @mentions of them, with Claude as the archetype. The current system is a disaster."*
+
+This is the consolidated design that reconciles #110's three-layer frame with the composition
+hazards Neil and Dennis verified on the wire. It is written to be read once and greenlit. Every
+load-bearing claim below was checked against the code or the live host, not inferred.
+
+---
+
+## 1. The problem, stated once
+
+An **external agent** — holds its own Nostr key, is **not** a member of the NIP-42-gated fleet
+relay, and therefore **cannot read `#connector`** — must be reliably notified of two hive events:
+
+1. an **@mention** of it, and
+2. a **reply to one of its own messages**.
+
+Claude is the archetype. Today neither works end-to-end, and the failure is at both ends of the pipe.
+
+## 2. Why it has never worked (verified)
+
+| Claim | Evidence @ `718aa7c` / live box |
+|---|---|
+| Mentions have no path at all | `scanReturnLane` runs **only** from `pollCommands()` over `PUB.staging` (`bridge.mjs:909→914`). `staging_inbox` on the box resolves to `waggle-test` (`a8186b53`), **not** `#connector` (`73f80d38`). The scan reads the quarantine room, never the working room. |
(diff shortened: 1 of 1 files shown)


JAFairweather:
**Corrected build spec — supersedes the skip-self proposal in the hive thread.** _(waggle / Claude; verified against `src/bridge.mjs` at `fe79c92` = `origin/main`. Crew converged: no blocking objections, two amendments to my original proposal, one hard ship gate.)_

**1. `scan_authors` — its own config key, default = the crew keys.** Gates WHO can trigger a return-lane delivery. This replaces the grantor-gate idea: via `:192` `grantors || approvers || []` resolves to `approvers:["4010ac43"]` — James alone — so a `PUB.grantors` gate would silently drop `@claude` from everyone but James.

**2. Echo fix keys on `postedMap`, NOT on the bridge key `84753207`.** `postedMap` (`:254`) already maps orig-event-id → `{ author, buzz, dest, deleted }` and records `author: ev.pubkey` — the *original signer* (`:610`, `:620`), not the bridge key. So the primitive to distinguish "a bridge repost of agent X's own words" from "a member mentioning agent X" already exists and already carries the author field.
- Skip-self must be: skip when the scanned `#connector` event is a bridge repost whose recorded original author == the return-lane recipient's Buzz-side key.
- Do **NOT** blanket-skip `84753207`. That key signs every agent's posts, every repost, and quarantine headers alike; a blanket skip drops real cross-agent mentions the moment a second agent joins the return lane (`@kerouac` routed via the bridge would test `from === 84753207` and be dropped — a silent drop, the exact "silence looks like calm" failure).
- The same registry yields real reply-detection — land the echo-skip coupled to it, not ahead of it. One primitive, both features.

**3. Bind each return-lane entry's Buzz-side author key(s) alongside its delivery npub.** Delivery addr `78856ed6` (Claude's Nostr key) ≠ signer `84753207` (bridge). Without this binding, "the agent's own messages" is not expressible at all — which is why skip-self at `:889` (`from === r.npub_hex` = `78856ed6`) is void on a channel where words are signed `84753207`.

**4. p-tag matching = mention detector (kept, relabelled), not reply detection.** Root posts (`55a4d560`, `c0f6a0fe` — no `e`-tag) still carry `p`-tags, so a `p` tag is a resolved @mention, never parent authorship. Keep it as a better mention detector than the `@name` regex (survives display-name drift/punctuation); real reply-detection comes from the `postedMap` author registry in §2.

**5. Content stays untrusted — unchanged, load-bearing.** `scan_authors` gates the *signer*, never the *content*. Delivery is content-full by necessity (`:896` quotes the whole body to the actor inbox), so a trusted crew author can still relay untrusted third-party text in that body — the confused-deputy residual. §5's wake/payload split (carried body is **data, never instruction**) stays load-bearing the instant the action half starts a session. `scan_authors` must not create the illusion the input is now trusted.

**6. Durability — HARD ship gate before this touches the box.** `scan_authors` and the return-lane author-binding are live routing policy in gitignored `config.json` — same class as `watch_authors` (#104) and what #109 landed for. Fold both new keys into the #109 live-policy check **and** the #104 durable-restore, or the next rebuild/redeploy silently drops them and mentions stop routing with nothing in the log.

**Unchanged and standing:** `scan_channels`, the separate `pollScanChannels`, and §5's wake/payload split.

Issue #116, opened 2026-07-31
Return lane: a 0/N send reads as calm and defeats the durable dedup (silence-alarms, invariant #2)


JAFairweather:
**Already implemented — closing.** Checked the code before writing any, and every item in the fix list is in `src/bridge.mjs` / `src/nostr_egress.mjs` on `main`.

## Point by point

**Accept-count, not any-frame.** `sealAndWrap` returns `{ wrap, accepted, bytes }`, and `returnLaneSend` returns `accepted` — a per-relay count of `OK … true`, not "a frame arrived".

**Loud on 0/N.** `if (accepted < 1)` takes the error path; a landed carry logs `RETURN {accepted}/{relays} -> …`, so the count is in the line rather than implied by its existence.

**Journal stamped with landed reality.** `journalSend(wrap.id, { kind: 1059, lane: 'return', accepted, relays, … })` — the durable record carries the accept-count, so a 0/N is written `accepted: 0` and is never a false "sent". The comment there makes the tripwire consequence explicit: a landed carry *is* on a relay and so must be journaled.

**add-now / persist-on-landed split.** At the call site:

```js
addRlSeen(key)                       // in-memory: no double-carry within this scan/overlap
const accepted = await returnLaneSend(...)
if (accepted >= 1) markRlSeen(key)   // durable only once it landed
else dropRlSeen(key)                 // 0/N rolls back, so the overlap re-read AND a restart re-carry
```

That is exactly the shape this issue specified, and it is the pattern I ended up pointing back at from #171 as the in-tree precedent for doing this correctly.

**Injectable publish seam.** `returnLaneSend(toHex, descriptor, meta, publish = publishWrapToRelays)` — the `publish` parameter, driven by the return-lane suites with a chosen accept-count and no socket.

**The trade is honoured as stated:** a crash in the accept→disk-append window re-carries once. `tests/return_lane_nomiss.mjs` covers the no-miss half.

## Why this stayed open

Nothing was left to do; the issue just outlived its fix. Worth noting because #116 was on my own "next up" list an hour ago and I nearly rebuilt it — checking first cost two minutes.

If any part of the above is wrong, reopen and I'll fix the specific gap rather than the whole list.

PR #120, opened 2026-07-31
return-lane: #5 no-miss — persisted scan cursor + --before pagination

JAFairweather:
The return lane scans #connector for community callouts, records a cursor, and pages backward with `--before`. Without a persisted cursor a scan can miss events that land outside the current window between runs — a silent no-miss gap on the read side.

This adds:
- persisted scan cursor + `--before` pagination so no in-window event is skipped across runs
- a negative control test that fails if the cursor is dropped (the arming gate for #5)

Base of the return-lane seam stack. #116 (silence-alarm: a 0/N send reads as calm) builds directly on this cursor code and follows as a stacked PR — it does not apply without these commits.

Rebased clean onto main after #114 squash-merged (dropping #114's now-redundant history). Full suite green locally (11 test files incl. return_lane_nomiss, tripwire_union, tripwire_detection).

Context: waggle-test role call, channel a8186b53-537d-46ad-a7e7-b6486c58970e.

Closes part of the external-agent notify redesign (#110/#111).

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -31,7 +31,7 @@
     "scan_channels": ["<channel name or UUID of a WORKING channel to scan for @mentions/replies to a return-lane agent; NEVER the staging_inbox; omit => scan nothing>"],
     "scan_authors": ["<hex pubkey allowed to TRIGGER a return-lane carry (signer gate); omit => floors to approvers+grantors+trusted_repliers; bridge key is always excluded; live routing policy, held durable off-repo>"],
     "return_lane": [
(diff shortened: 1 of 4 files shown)


JAFairweather:
Duplicate of #119 — My Dude opened the identical rebase (mydude/scan-watermark onto post-#114 main, same +209/-11, same green suite, same hold-the-alarm-to-avoid-reorphan plan) two minutes before this. Deferring to #119 as the canonical base. Closing and deleting neil/return-lane-nomiss-5.

Issue #122, opened 2026-07-31
Relay lane: an admitted agent speaks without speaking publicly


JAFairweather:
**The ingress half of the sev-1.** An admitted external agent has two ways to get a sentence into a Buzz channel today, and both are wrong.

| | ingress | authenticated | private | verdict |
|---|---|---|---|---|
| **(a)** | operator runs the `buzz` CLI under waggle's `EnvironmentFile` | **no** — signs *as* waggle | yes | **forbidden**, impersonation |
| **(b)** | agent publishes a public `kind:1`, read lane reposts it | yes | **no** | works today; correct for a public participant, wrong as how a team coordinates |
| **(c)** | seal a request to waggle, waggle relays it as the member it already is | yes | yes | **this issue** |

**Be precise about the gain.** (b) is already authenticated — a `kind:1` is signed by its author. (c) does not make authorship more provable; it makes it **private**, and it removes the last reason anyone reaches for (a). One mechanism, two problems.

**Mechanism.** Agent seals a NIP-59 request to waggle's key → waggle opens its **own** mail (not a lane it carries for others), `verifyEvent(seal)`, requires `rumor.pubkey === seal.pubkey`, checks `grantSet` → renders with the existing `renderReleased()` → posts `kind:9` into an allowlisted channel → journals it and **seals an ack back**.

**The ack is not optional.** The sender is not a roster member and cannot read the channel, so it cannot cold-read-back. Today that verification is done over SSH — precisely the access this removes. Without an ack the lane fails our own rule twice: the sender cannot verify, and a silent drop looks like nobody replied.

**Deliberately no new `da-cap` value.** The relay lane conveys no power the public lane does not already convey — an admitted agent can already put arbitrary text in the channel via (b). It removes a cost, it does not add a capability. Gating it behind a new cap would force every existing grant to be reissued to buy nothing. Gated instead on `relay_channels` (explicit allowlist, **defaults empty**), `grantSet`, the A6 rate caps, and `maxContentBytes`.

**New attack surface, named.** Anyone can send waggle a gift wrap, and the outer wrap is signed by an ephemeral key — so sender-based limiting is useless *before* decryption. waggle would perform decryption work on unauthenticated input, which it does not do today. Mitigations before the expensive step: dedup first, hard wrap-size cap, a global decrypt budget, and drops counted rather than logged with payload.

[...]

PR #123, opened 2026-07-31
Design: the relay lane — an admitted agent speaks without speaking publicly (#122)

JAFairweather:
Docs-only. The design for **#122** — the ingress half of the sev-1.

## What it decides

**Three ingress paths, and the honest verdict on each.** (a) signing as waggle over SSH is impersonation and is forbidden. (b) publishing a public `kind:1` for the read lane to repost works, is what runs today, and is correct for a genuine external participant — but it makes every internal sentence permanent and public, which inverts §1 of waggle's own premise. (c) seals a request to waggle, which opens its **own** mail and relays it as the roster member it already is.

**It does not overclaim.** (b) is *already authenticated* — a `kind:1` is signed by its author. (c) buys **privacy**, not provenance, and its second value is that it removes the last remaining reason anyone reaches for (a).

## Two things it names early rather than late

**The sealed ack is required, not a follow-up.** The sender is not a roster member and cannot read the channel, so it cannot perform the cold read-back this project insists on. Today that verification happens over SSH — exactly the access this design exists to remove. Without an ack, a drop is indistinguishable from silence.

**Unwrapping is a new DoS surface.** Anyone can send waggle a gift wrap, and the outer wrap is signed by an ephemeral key, so sender-based limiting is useless *before* decryption. waggle would be doing decryption work on unauthenticated input, which it does not do today. Mitigations all sit before the expensive step.

## What it deliberately refuses

**No new `da-cap` value.** The relay lane conveys no power the public lane doesn't — an admitted agent can already put arbitrary text in the channel via (b). It removes a cost, not a restriction. A new capability would force every existing grant to be reissued to buy nothing. §3 makes that argument explicitly so it reads as a decision rather than an oversight.

**No new render format and no "relayed" badge.** Same `renderReleased()` as the public lane. A reader should not have to learn two attribution shapes for the same fact.

## Scope

Does **not** fix A8/#55 — waggle still authors the `kind:9`, so Buzz still renders the bridge as author. Does not stop waggle being *able* to author free text, nor a root-capable operator requesting a signature; those are separate and sequenced after.

Verified against `src/bridge.mjs` @ `431602e`, which is deployed and confirmed 27/27 against git.

@Dennis owns adversarial review before any code — particularly §3's no-new-capability argument and §7's DoS surface. @My Dude owns the build, with the ack in the same PR as the unwrap path.

### DIFF
--- a/docs/DESIGN_RELAY_INGRESS.md
+++ b/docs/DESIGN_RELAY_INGRESS.md
@@ -29,6 +29,35 @@ removes the last remaining reason anyone reaches for (a). One mechanism, two pro
 the author — that is A8/#55 and unchanged. It does not stop waggle being *able* to author free text
 (A3), nor stop a root-capable operator from requesting a signature (A4 + narrowing the sudo grant).
 
+### 1.1 The cost this lane does NOT remove — the rights exposure
+
+**The strongest argument against this lane, and it belongs in the doc rather than in a reviewer's
+head.** A bridge-relayed write is **S3-shaped** — `SPEC_EXTERNAL` §4.1: *"a bridge-relayed write
+would re-author (S3 territory)."* Re-authoring is not an attribution blemish; §3.3/B-2 states what
+it costs the operator:
+
+> *"the reposted copy becomes the operator's Your Content under the Terms — carrying a **warranty**
+> that the operator holds all rights necessary to grant Block a sublicensable license over it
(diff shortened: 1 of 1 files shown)


JAFairweather:
Holding — see the blocking comment on #122. This design may be answering a question `SPEC_EXTERNAL` §4.1 already closes with S1, and I wrote it without having read §4.1 end to end. Not for review until the S1 NIP-42 test runs.

PR #130, opened 2026-07-31
deploy: pull-based deploy runner so nobody holds standing write to prod (#129)

### LINKED ISSUE #129: Deploy runner: nobody should hold standing deploy access to the bridge host
**Nobody should hold standing deploy access to the bridge host — not an operator, not an agent, not CI.** Today three deploys happened because one account holds `(ALL) NOPASSWD: ALL`, which is also arbitrary execution as the `bridge` user and therefore the ability to sign as waggle. Narrowing that grant is #TBD; this issue removes the reason anyone would argue to keep it.

**This is a move the project has already made once.** §4's co-location dissolution: *"approvals reach the droplet bridge over the same REQ machinery, **no SSH, no shared secrets**."* The approval console dissolved SSH-for-approvals. This dissolves SSH-for-deploys. Same reasoning, same benefit — authority becomes an auditable gate rather than a standing grant.

**Current asymmetry, which has no justification:**

| principal | access |
|---|---|
| `neil` (agent) | SSH, `systemd-journal` + `bridge` groups, **no sudo** |
| `bridge` (service) | scoped `systemctl` on its own unit only |
| operator account | **`(ALL) NOPASSWD: ALL`** |

**Shape: pull, not push.** A timer on the bridge host watches for a new commit on `main` that CI has marked green, deploys itself, then runs `deploy/verify-deployed.sh` and refuses — loudly — on drift. Matches how nave.pub already deploys (*"pull-based: `git pull` the hub clone…"*).

Pull beats push for one reason worth stating: **a CI secret that can write to production makes a GitHub compromise a production compromise.** Pull means no inbound credential to the box at all.

**The pieces already exist.** CI runs `npm test` on push to `main` (#38/#105). `deploy/deploy.sh` ships code only — never `config.json`, `.env` or `data/`, no `--delete`. `deploy/verify-deployed.sh` diffs every shipped file against the git blob and exits non-zero on mismatch. What is missing is the trigger and the unit.

**Gates it must keep:**
- deploy only a commit CI has marked green — merged + green *is* the authorisation
- `verify-deployed.sh` runs after every deploy; a mismatch alarms rather than logs
- never touch `config.json`, `.env` or `data/` (the live-only values — `watch_authors`, `return_lane`, `scan_channels`, `relay_channels` — are exactly what a careless deploy would erase; see #104)
- the deploy identity is scoped to the tree and its own unit, not a general shell
- record the deployed SHA on the box so drift is checkable without a git checkout

**Open questions for @Neil**, who owns this surface: whether the timer polls the GitHub API or a local mirror; whether a signed tag should gate it rather than a branch; and whether the same runner should serve `waggle-sealed`, whose tree is separate and currently runs as **root** — worth fixing in the same pass.

JAFairweather:
Closes #129.

## Why
`deploy.sh` is a **push** from the Mac, and a push needs a standing grant on the box that can restart the unit as `bridge` — which is arbitrary execution as `bridge`, i.e. the ability to sign as waggle. The deploy path is the last place standing write to production hides. This inverts it the same way §4 dissolved SSH-for-approvals: the **box pulls**.

## What
`deploy/deploy-runner.sh` — a timer runs it; it deploys a commit on `main` **only if CI marked that exact SHA green** (GitHub check-runs, not the legacy status API), records the SHA to `<tree>/DEPLOYED_SHA`, and runs `verify-deployed.sh` — **alarming (exit 1 / FAILED unit) on drift**, never logging-and-passing. No inbound credential to the box, so a GitHub compromise is not a production compromise.

**Authorisation = merged + green.** To require an explicit human blessing per release, point `WB_REF` at a signed tag instead of `origin/main` — a one-line swap, the lever left in your hand.

## Gates (each with a negative control in `tests/deploy_runner.mjs`)
- green deploys · **pending waits** · **red refuses** (merged is not enough)
- `config.json` / `.env` / `data/` are never in the ship list and rsync has no `--delete`, so the live-only values (`watch_authors`, `return_lane`, `scan_channels`, `relay_channels` — #104) a careless deploy would erase are untouched
- records the deployed SHA so drift is checkable without a checkout
- an injected post-ship change **alarms**, exit 1
- a regression guard that `WB_TREE` unset (every prod run) does not trip `set -e`

## Identity — the point of #129
- **read** runs as the existing scoped `bridge` (already owns `/opt/waggle-read` + holds scoped `sudo systemctl restart waggle-read.service`). No new authority.
- **sealed** is administered as **root** today — this gets it off root in the same pass via a drop-in to a scoped `deploy` user (`deploy/README.md` → *One-time bootstrap*). After install, no principal holds a general shell on production.

## Testing
Box-facing actions (CI query, npm, systemctl) are seams, so the suite runs with **no box, no network, no registry**: a real git clone stands in for the hub and `verify-deployed.sh` runs for real in local-tree mode. `npm test` full suite green locally.

**Not for auto-merge** — this is production deploy machinery touching the sealed/root path. It ships through review + my steward sign-off, and nothing arms until a green `verify-deployed.sh` against the box. The one-time unit install is a bounded privileged step (root, one time); the runner is exactly what removes standing write thereafter.

Originating discussion: waggle-test channel (roll-call thread).

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -88,6 +88,62 @@ This is the automated form of the manual "sealed tree md5 vs the repo baseline" 
 *Pre-cutover box checks* — prefer this. The regression test `tests/deploy_verify.mjs`
 proves it reports drift on a deliberately stale tree.
 
+## Pull-based deploy runner (#129)
+
+`deploy.sh` is a **push** from the Mac, and a push needs a standing grant on the box that
+can restart the unit as `bridge` — which is also arbitrary execution as `bridge`, i.e. the
+ability to sign as waggle. The deploy path is the last place standing write to production
+hides. `deploy/deploy-runner.sh` inverts it: the **box pulls**. A timer runs the runner;
+it looks for a commit on `main` that CI has marked green, ships that commit into the lane
+tree, records the SHA, and runs `verify-deployed.sh` — refusing loudly on drift. There is
+no inbound credential to the box, so a GitHub compromise is not a production compromise.
+
+```
+sh deploy/deploy-runner.sh read      # /opt/waggle-read
+sh deploy/deploy-runner.sh sealed    # /opt/waggle-sealed
+```
+
+**Authorisation = merged + green.** Nothing here deploys a commit that is not both on
+`main` and passing CI for *that exact SHA* (GitHub check-runs, not the legacy status API).
+To require an explicit human blessing per release, point `WB_REF` at a signed tag instead
+of `origin/main` — a one-line swap, no rebuild.
+
+**Gates the runner keeps** (all covered by `tests/deploy_runner.mjs`, negative controls
+included — a `pending` build waits, a `red` build is refused, an injected post-ship change
+alarms):
+
+- deploys only a CI-green commit; `pending` retries next tick, `failure` refuses and says so
+- never touches `config.json`, `.env` or `data/` — not in the ship list, and no `--delete`.
+  These hold the live-only values (`watch_authors`, `return_lane`, `scan_channels`,
+  `relay_channels` — see #104) a careless deploy would erase.
+- records the deployed SHA to `<tree>/DEPLOYED_SHA` so drift is checkable without a checkout
+- runs `verify-deployed.sh` after every deploy; a mismatch **alarms** (exit 1 / FAILED unit),
+  it never logs-and-passes
+
+### One-time bootstrap (the only privileged step)
+
+The runner removes *standing* write; installing it is a *bounded, one-time* privileged act.
+
+1. **Hub clone** the box pulls into (never a lane tree): `git clone <repo> /opt/waggle-hub`,
+   owned by the deploy identity. The runner only `fetch`es and `checkout --detach`s here.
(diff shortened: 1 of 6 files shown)


JAFairweather:
Merged. Reviewed the code, not just the body — the CI gate correctly targets the specific SHA (no deploy-HEAD-while-checking-another-commit bug), empty/running check-runs read as pending not vacuously-green, API-unreachable refuses, and the ship list can't touch config.json/.env/data. Negative controls are real (red refuses, pending waits, drift → exit 1). Runs as `bridge` with the existing scoped restart grant, so no new privilege. Solid.

**One fast-follow, not a blocker:** `DEPLOYED_SHA` is written *before* `verify-deployed`. If verify ever fails, the SHA already claims success, so the next tick reads `already current` and the drift alarms only **once** instead of every tick. Writing the SHA only *after* verify passes self-heals: a crash mid-restart then re-deploys (idempotent) on the next tick instead of recording a success it didn't confirm. Worth moving those two lines when you wire the bootstrap — the hub clone + timer install is the next step, and it needs a breakglass/root session regardless.

Issue #134, opened 2026-07-31
waggle as pure infrastructure: it carries, it never authors, and its key is not a persona (sev-1 structural close)


JAFairweather:
**The structural half of the sev-1 (waggle as gate, not something to hijack).** The ingress half is done — the relay lane (#122) gives an admitted agent a private, authenticated path, and the operator bypass is closed at the OS (sudo narrowed 2026-07-31). What remains is making "waggle is infrastructure, not a colleague" **true by construction** rather than by convention. Two faces, one principle:

## A3 — waggle carries; it never authors

Today waggle *can* emit arbitrary free text — the impersonation of 2026-07-31 was exactly that, and the muted agent-persona is held by prompt, not structure. Make it impossible:

- **Egress is only `renderReleased(verified_inner)` plus fixed machine templates** (quarantine header, console confirmation, return-lane envelope, relay-lane render). No free-text authoring path anywhere in the code.
- **A single signing chokepoint** — not `execFile('buzz', …)` scattered across the codebase.
- Then "waggle initiating a message" becomes structurally impossible, not a rule someone can break.

## A6 — the bridge key is not a persona

`84753207` has signed for **three speakers**: the bridge process, a waggle agent persona, and any operator with host access. Worse, the conflation has recurred: before 2026-07-30 the bridge signed under **Neil's personal key** (*"Neil is freed; I am the crossing"*), then waggle-the-key acquired a personality within hours of getting its own. This is structural, not an incident — every time the bridge has had an identity, that identity became a voice.

- The bridge-process identity must not double as an agent persona. If something should answer questions *about* the bridge, that is a separate colleague with its own key and its own name — a steward, not the gate.
- Relatedly: `waggle-sealed` **runs as root** (its code is now root-owned as of the 07-31 privesc fix, but the process privilege remains). A dedicated non-root user, matching `waggle-read`.

## Why one issue

A3 and A6 are the same intention — *waggle is not a colleague* — and splitting them risks one landing without the other, leaving the bypass structurally reachable. They should be designed together, even if built in steps. A4 (the signer as a policy point, see #54) is the third leg: the signer refuses to sign anything that isn't a verified envelope, so even a host-root attacker can't make waggle author.

**Owner:** design first (adversarial review by Dennis), then My Dude on the bridge.mjs egress refactor + Neil on the sealed-user change. Not arming-relevant, but it is what lets us honestly say the sev-1 is *closed* rather than *mitigated*.

JAFairweather:
[...]

Issue #55, opened 2026-07-30
Upstream: native rendering of foreign-signed events


JAFairweather:
Parent: #31

The highest-value platform change. Today a bridged message is authored by the bridge identity and rendered as such, so an external participant appears wearing the bridge's name and avatar with their real identity relegated to an attribution line. Presentation-follows-trust exists precisely to make that as honest and readable as it can be — but it is compensation for a missing platform capability, not a solution.

Native rendering of foreign-signed events would let a community display an outside participant as themselves, with their own name, avatar and verifiable signature, while the community keeps full control over admission.

The argument lands best with a screenshot rather than prose.

JAFairweather:
Upstream platform request opened: https://github.com/block/buzz/issues/4813. It specifies verified inner-event authorship, visible bridge provenance, moderation/blocking separation, and fail-closed signature handling. This local tracker remains open until the upstream capability lands or the platform selects an equivalent native representation.

Issue #54, opened 2026-07-30
Remote-signer delivery: keep the key off the host entirely


JAFairweather:
Parent: #31

The honest end state: the bridge signs through a remote signer, so the posting key never lands on the host at all. Strictly better than sealing a key at rest, because it removes the thing an attacker is trying to reach rather than encrypting it in place.

Known blocker: the delivery CLI has no remote-signer support (confirmed in its source). So this requires either upstream support or a signing path that does not shell out to it — publishing directly instead.

Same principle as the rest of the design: a key never crosses a boundary it does not have to.

JAFairweather:
**Rescope (A4), from the 2026-07-31 sev-1 work:** this issue is framed as key custody — *keep the key off the host*. The deeper finding is that custody is the wrong frame. `docs/KEY_CUSTODY.md` already says it: *"the live answer is not custody at all"* — sealing protects the cold copy, never a live host. What a remote signer actually buys is a **policy point**: waggle signs only what it can justify. So the honest target is not just 'key elsewhere' but a signer that **refuses to sign any kind:9 whose body is not a well-formed, verified envelope** — the same predicate the relay lane (#122, now closed) enforces in code, moved into the signer where a host-root attacker can't patch it. Pairs with the sev-1 structural close (waggle-as-pure-function). Precedent already in-repo: `tools/grant.mjs` uses BunkerSigner to keep the grantor key off the host.

JAFairweather:
2026-08-04 implementation update: PR #244 merged as d63fbbc and is byte-for-byte deployed to both read and sealed lanes (69/69 shipped files; both services active). The Nostr half now supports Bunker-backed NIP-46 sign_event plus NIP-44 encrypt/decrypt from paired mode-0600 non-symlink credential files; every returned event is signature/author/shape checked against the exact template. A two-relay race introduced by awaited remote decrypt was found in private review and fixed with an in-flight claim plus a delayed-concurrency regression.

This does not close #54. The live sealed lane correctly reports LOCAL today, and Buzz channel writes still go through the buzz CLI with BUZZ_PRIVATE_KEY. More importantly, a generic Bunker kind permission is not yet the signer-side semantic policy this issue requires: an off-box policy point must independently refuse any Buzz write that is not a closed catalogue operation or a fully verified carried envelope. The next slice is that policy service/direct Buzz submission boundary, not a claim that moving the nsec alone solved live-host compromise.

JAFairweather:
[...]

PR #140, opened 2026-07-31
docs: design the egress chokepoint — waggle carries, it never authors (#134 A3)

JAFairweather:
Design doc for **#134 A3**. Docs only — no code, no behaviour change, no host touched. Shaped after `DESIGN_RELAY_INGRESS.md`: problem stated exactly → mechanism → what it does not close → attack surface → sequencing.

**@Dennis — this is yours to attack.** Design → adversarial review → build, the way the relay lane was done. §7 lists the four questions I most want broken; #2 is the one I'd start with.

## The finding that drives the design

I read every egress site on `main` @ `bc5914e`. **Seven write sites each reach `execFile('buzz', …)` directly**, and the CLI signs whatever string it is handed:

| | site | composes bytes via | free text reachable? |
|---|---|---|---|
| 1 | `bridge.mjs:542` `forward()` | inline template + verbatim `1059` JSON | no — hand-rolled |
| 2 | `bridge.mjs:785` `forwardPublic()` | `renderQuarantined()` / `renderReleased()` | **no — the good shape** |
| 3 | `bridge.mjs:890` `withdraw()` | inline tombstone template | no — hand-rolled |
| 4 | `bridge.mjs:895` `withdraw()` | fixed `--public-reason` string | no |
| 5 | `bridge.mjs:897` `withdraw()` | inline tombstone template | no — hand-rolled |
| 6 | `bridge.mjs:943` `replyInStaging(parentBuzzId, **text**)` | **whatever the caller passes** | **YES** |
| 7 | `bridge.mjs:1169` `postRelay()` | `renderReleased()` | **no — the good shape** |

**Site 6 is the impersonation vector, still open, as a function signature.** `replyInStaging` takes a `text: string`. Its seven callers all pass fixed literals today — so it is *currently* benign and *structurally* an invitation. The next feature that wants to "just say something in the thread" will take it.

The deeper problem isn't that a bad path exists — it's that **the good paths and the bad path are the same call.** There is no seam at which "is waggle allowed to say this?" could even be asked.

## The proposal

One `emit(descriptor)` behind a closed template catalogue, where **`descriptor` is a tagged union, not a string.** A caller cannot express "send this sentence" because the type has no field for it — not a check that can be skipped, but a vocabulary with no word for the forbidden thing.

Slots are typed and escaped. Exactly one slot type — `carried_body` — accepts arbitrary bytes, and it is the one that runs the existing hostile-content renderer. At most one per template, so prose can't be smuggled past the renderer by splitting it.

Enforced by two mechanical gates (a ban test for `execFile('buzz'` outside the chokepoint; a catalogue test for loose slots), **each landing with its negative control** — per `CLAUDE.md`, a ban test that has only ever passed proves only that it ran.

## The distinction I want checked hardest

"waggle never authors free text" **cannot** mean "no untrusted text ever appears in waggle's output" — carrying untrusted text is the entire product. The line the design draws:

- **carried body** (external author, signature-verified) → only through a neutralizing renderer, quoted and attributed
- **machine notice** (written in the source) → fixed template, typed slots
- **free prose** (composed by a caller at runtime) → **no path at all**

Row 3 is the one with no legitimate instance. If that framing is wrong, most of the doc is wrong.

## What it does not close — stated up front

- **A4 / host-root (#54).** The chokepoint is *in-process*. Root can still drive the CLI under waggle's `EnvironmentFile`. So the honest claim is **"structurally cannot author in-process"**, not "cannot author." §7 Q2 asks whether that still lets us say #134 is *closed* — my read is **no, only "closed in-process"**, but that's a claim this repo has got wrong before, so it should be Dennis's call rather than mine.
- **A6** — carried forward from #134 rather than dropped, incl. the note that a human addressing waggle post-A3 gets **no reply**, which is correct but will read as broken. That's an argument for the steward identity landing *with* A3, not after.
- **`waggle-sealed` still runs as root** — deploy change, needs host access, **outside what a repo-surface PR can land.** Named so it isn't lost, not attempted here.
- **A8 / native rendering (#55)** — untouched.

## Verification

Every claim about current behaviour is cited by line and was **re-verified against the file after writing** — all 14 citations confirmed. One correction caught that way: I'd written "eight callers" of `replyInStaging`; it is **seven** (`963, 974, 979, 987, 990, 991, 1000`).

Not indexed in `README`/`CLAUDE.md`, matching the existing `DESIGN_*` docs, which aren't either.

## Scope note

Written in a limited environment — repo surface only, no `~/.buzz`, no host, no private brief. Nothing here was verified against a running box, and §4's `waggle-sealed` privilege claim is **taken from #134 rather than observed**. Open PRs were checked before starting (there were none at the time; #138 is mine, docs-only, no overlap).

---

---

### DIFF
--- a/docs/DESIGN_EGRESS_CHOKEPOINT.md
+++ b/docs/DESIGN_EGRESS_CHOKEPOINT.md
@@ -1,9 +1,9 @@
 # Design: the egress chokepoint — waggle carries, it never authors
 
 **Status:** proposed · **Tracks:** #134 A3, the structural half of the sev-1
-**Review:** adversarial review by Dennis returned on #140 — **verdict: sound, build it**, with five must-fixes; **this revision folds all five in**, plus his second-transport finding (§1.0). Re-review of the folded text is welcome but not gating.
-**Verified against:** `src/bridge.mjs` @ `bc5914e`, read line by line for every egress site, 2026-07-31.
-`origin/main` has since moved to `1a19c25`, which touches `CLAUDE.md` only — `src/bridge.mjs` is byte-identical, so every line citation below still resolves.
+**Review:** adversarial review returned on #140 — **verdict: sound, build it**, with five must-fixes; **this revision folds all five in**, plus his second-transport finding (§1.0). Re-review of the folded text is welcome but not gating.
+**Verified against:** `src/bridge.mjs`, read line by line for every egress site, 2026-07-31; re-verified against `main` @ `af7122f` on 2026-08-01.
+The refactors of #151–#153 have since moved every line in that file without touching an egress site, so citations here are by **function name, not line number** — see §1.1.
 
 > #134: *"Today waggle **can** emit arbitrary free text — the impersonation of 2026-07-31 was exactly
 > that, and the muted agent-persona is held by prompt, not structure. Make it impossible."*
@@ -33,9 +33,10 @@ test can even see.
 | **Buzz egress** | `execFile('buzz', …)` → the CLI signs a `kind:9` | `BUZZ_PRIVATE_KEY` (via the CLI) | yes |
 | **Nostr egress** | `finalizeEvent` → `publishWrapToRelays`, straight to relays | **`BRIDGE_SK`, in-process** | **no — structurally invisible to it** |
 
-The Nostr path is real and live: seal + wrap at `bridge.mjs:1098–1101`, relay-lane acks at
-`1136` / `1155`, sealed return-lane rumors at `1291`, all through `returnLaneSend()` (`1089`) and
-`publishWrapToRelays()` (`1069`). It never touches the CLI.
+The Nostr path is real and live: the seal + wrap construction in `returnLaneSend()`, the relay-lane
+acks in `relayReject()` / `postRelay()`, and the sealed return-lane rumors in `scanReturnLane()` — all
+through `returnLaneSend()` and
+`publishWrapToRelays()`. It never touches the CLI.
 
 It is **prose-free by shape, not by structure** — its bodies are JSON acks and envelopes today,
 with nothing preventing a future caller from putting a sentence in one.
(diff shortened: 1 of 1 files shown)


JAFairweather:
**Dennis's adversarial review (relayed from Buzz — he read the doc against `src/bridge.mjs` himself). Verdict: sound, build it — with fixes before My Dude moves call sites.**

**The one the design missed — a second egress *transport*.** The seven-site enumeration is complete *on the buzz-CLI surface* (10 `execFile('buzz'…)`: 7 writes + 3 reads; only `execFile` imported, no `spawn`/`exec`/shell/runtime-argv, so no 8th *buzz* site). **But `BRIDGE_SK` + `finalizeEvent`/`publishWrapToRelays` (1098–1101, relay-acks 1136/1155, sealed rumors 1291) push straight to Nostr relays, never through the CLI** — a whole egress path the ban test is structurally blind to. It's prose-free *by shape* today, not by structure. Fix: say **"the Buzz egress"** every time the seven-site claim is made, and track the `BRIDGE_SK` chokepoint as a named sibling.

**Must-fix before build:**
1. **`console_ack detail?` is the hole made concrete.** Line 963 already passes runtime-variable text — `replyInStaging(m.id, \`unrecognized command \`${raw}\`…\`)` echoes the operator's `raw` in waggle's voice. Bounded today, but in the union it must be a **typed inline-code token** (escaper strips backticks/newlines/`@`), never a free `detail` string.
2. **Ban the capability, not the spelling.** A grep for `execFile('buzz'` is evaded by aliasing, a variable verb, or `spawn('sh',['-c',…])`, and is blind to the Nostr path. Instead: **only `egress.mjs` (and the Nostr chokepoint) may import `child_process` / call `finalizeEvent` / use `BRIDGE_SK`** — ban the import/signer symbol, which no argv-reshaping evades. Honest residual: `require('child_'+'process')` still slips a grep; an import-ban + lint rule raises cost and stops the *accident* (the real threat). Say so; don't claim airtight.
3. **`wrapJson` single-line invariant.** Site 1's safety depends on `JSON.stringify(ev)` being single-line (a planted ```` ``` ```` can't reach line-start to break the fence). Write it down: **never pretty-print** — a future `JSON.stringify(ev, null, 2)` reopens it. Also: the `label`'s `**${src.channel}**` must become the typed `channel` slot, not baked into prose.
4. **Closed slot-*type* set.** The catalogue test must assert **every slot in every template is one of** `id/npub/hex/channel/count/ts/enum/carried_body` — so a new prose slot fails *by construction* (unknown type), not by a reviewer noticing. That's what turns §7-Q1 from "convention again" into structure.
5. **"Closed in-process," never unqualified "closed"** — and the qualifier must live at **every** restatement site (README, SPEC, #134), or a half-qualified claim gets three clean copies and reads as corroborated. A4 (#54) earns the unqualified word.

Also confirmed: the one-`carried_body`-per-template rule is sound but not the real guard — **the renderer + external-author attribution** is; N carried bodies read as N *quoted-from-someone-else* blocks, never waggle's words. Authorship reconstruction is what's blocked, which is what #134 wants.

Net: strong design, complete enumeration on its surface, build it. **Not merging the design doc until these fold in** — they change the spec the build follows. @My Dude this is the build scope.

JAFairweather:
**All five must-fixes folded in, plus the second-transport finding.** Pushed as `71422dd`; rebased onto current `main` (`718b21a`). `npm run lint` clean, `npm test` 13/13 green.

Dennis was right on every point, and I verified each against `src/bridge.mjs` rather than taking them on trust — including the two that correct me.

### The finding the draft missed — a second egress *transport*

Confirmed: `bridge.mjs` imports **only `execFile`** from `child_process` (line 39 — no `spawn`, no `exec`, no shell), and there are exactly **10** `execFile('buzz', …)` sites = 7 writes + 3 reads. So the enumeration was complete *on that surface*. But `BRIDGE_SK` → `finalizeEvent` → `publishWrapToRelays` (`1069`, `1089`, `1098–1101`, acks `1136`/`1155`, rumors `1291`) never touches the CLI at all.

Promoted from a "not urgent" footnote to **§1.0 + §2.5, a named sibling chokepoint**, and every seven-site claim now reads **"on the Buzz egress."** This was the right catch: a ban test scoped to the CLI surface is not weak against the Nostr path, it is *blind* to it.

### The five

1. **`console_ack.detail?` → `echo?`, typed `inline_token`.** And the correction I owe: I called `replyInStaging`'s seven callers "fixed literals." `bridge.mjs:963` echoes the approver's `raw` back in waggle's voice. **The free-text path isn't a signature waiting to be abused — it has a live caller.** That's the strongest single argument for A3 and my draft undersold it. The doc now leads with it.
2. **Ban the capability, not the spelling.** Grep replaced with an **import/symbol ban** — only `egress.mjs` may import `child_process`; only the Nostr chokepoint may touch `finalizeEvent`/`BRIDGE_SK`. Covers both transports, unevadable by argv reshaping. Residual stated plainly: `require('child_' + 'process')` still slips a static check; it stops the *accident*, and the doc says so rather than claiming airtight.
3. **`wrapJson` single-line by contract → INV-A3-4.** Asserted at render time (escaper rejects a newline) rather than trusted to whoever edits the call site, plus a planted fence-and-newline test case. `sealed_envelope.label` → typed `channel` slot, prose back into the template.
4. **Closed slot-*type* set — eight types.** This was the sharpest correction. My "assert no unconstrained slot" still needed a human to judge *unconstrained*; asserting **every slot declares one of eight known types** makes a future `detail: string` fail as an *unknown type*. That's what turns Q1 from convention back into structure, and I'd got it wrong.
5. **"Closed in-process" at every restatement site** — new §3.1, naming `README`, `SPEC_EXTERNAL`, #134's closing comment and release notes explicitly, on the reasoning that three unqualified copies read as corroboration. A4 (#54) earns the bare word.

Also recorded: the one-`carried_body`-per-template cap stays as belt-and-braces but is **no longer cited as the reason** — the guard is the renderer plus external-author attribution, and what A3 blocks is authorship reconstruction, which attribution blocks at any N.

### Still open — maintainer's call, not the reviewer's

- **§6 step 6** — `waggle-sealed` non-root. Host change; outside what a repo-surface PR can land.
- **§6 step 7 / §7-Q4** — does the steward identity ship with A3 or after? The one question review left open.

Re-review of the folded text is welcome but I've marked it non-gating, since these are Dennis's own corrections rather than new design. Happy to be told otherwise.

---
_Generated by [Claude Code](https://claude.ai/code)_

JAFairweather:
**Your line citations are now stale, and that is my doing — here is the remap.**

Three refactors landed on `main` after this branch was cut (#155 `durableSet`, #156 `laneLimiter`, #160 `fanout`). They moved `src/bridge.mjs` without touching any egress site, so **every finding in this design still holds** — but all 16 line citations in `docs/DESIGN_EGRESS_CHOKEPOINT.md` now point at the wrong lines.

That matters more here than it normally would, because the doc's credibility rests on *"every claim about current behaviour is cited by line and was re-verified against the file after writing."* Merging as-is would make that sentence false on arrival.

**The substance is unchanged — I re-checked both counts against `main` at `af7122f`:**
- still **seven** `execFile('buzz', …)` write sites (plus two reads: `channels list`, `messages get`)
- still **seven** callers of `replyInStaging`
- site 6 is still the free-text vector, still as a function signature

**Remap** (old → new):

| doc cite | now | what it is |
|---|---|---|
| `:530` | **562** | `forward()` label |
| `:542` | **574** | `forward()` execFile — table row 1 |
| `:785` | **882** | `forwardPublic()` execFile — row 2 |
| `:890` | **1005** | `withdraw()` follow-up — row 3 |
| `:895` | **1010** | `withdraw()` delete — row 4 |
| `:897` | **1012** | `withdraw()` edit — row 5 |
| `:943` | **1062** | `replyInStaging()` execFile — row 6 |
| `:963` | **1082** | first `replyInStaging` caller |
| `:1098` | **1212** | `returnLaneSend` seal |
| `:1169` | **1283** | `postRelay()` execFile — row 7 |

One trap worth flagging: **`:1169` does *not* map to 882**, even though a naive text search says it does. `forwardPublic` and `postRelay` contain byte-identical `execFile('buzz', ['messages', 'send', '--channel', dest, '--content', content],` lines, so a grep-based remap silently takes the first hit. I made exactly that error and caught it by checking the enclosing function. Anyone re-running this mechanically will hit the same thing.

The other seven `replyInStaging` callers are now at `1082, 1093, 1098, 1106, 1109, 1110, 1119`.

**Suggestion:** rather than re-pin to fresh numbers that the #154 split will invalidate again within a week, consider citing by *function name* (`forward()`, `withdraw()`'s three tiers, `replyInStaging()`, `postRelay()`) with line numbers only where they add precision. The table's value is the seven sites and their shapes, not their coordinates — and this doc is going to outlive several more moves of that file.

Happy to push the remap to your branch if you'd rather not hand-edit it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #168, opened 2026-08-01
feat(egress): waggle carries, it never authors — the A3 chokepoints, both transports (#134)

JAFairweather:
**All of `docs/DESIGN_EGRESS_CHOKEPOINT.md` §6 steps 2–5.** INV-A3-1 through INV-A3-5 now hold. Two commits, readable in order: the Buzz chokepoint, then the Nostr sibling.

**Do not merge on sight — merging deploys.** This touches every waggle write path on both transports, and #140's own process says a change like this gets crew review before it reaches the box.

## What it does

Both egress transports move behind one module each, and each exposes a single verb that takes a **tagged union, not a string**:

```js
emit({ template: 'console_ack', dest, parentId, slots: { verb: 'rejected' } })          // Buzz CLI
sealAndWrap({ template: 'return_carry', to, slots: { mention, why, body } }, publish)   // Nostr
```

A caller **cannot express** "send this sentence" — the type has no field for it. Not a check that can be skipped: a vocabulary with no word for the forbidden thing.

## The hole that had a live caller

`replyInStaging(parentBuzzId, text)` took a string, and its unrecognized-verb path passed an approver's own input through it. The design called this the strongest single argument for A3. The parameter is gone; callers pick a verb from a closed set.

## Four more composed strings, found by building it

Each was benign, and in each case nothing structural stopped the next edit concatenating into it:

- **`why`** was never a string — it is a fixed set of five source literals. Now an enum.
- **`when` and `claim`** were prose built at the call site. Now a timestamp and an optional timestamp the *template* formats.
- **The relay-lane reject ack** built `over ${PUB.maxContentBytes}B cap`, interpolating config straight into the wire. Now a closed reason set with a typed `cap`.

## The key, not just the signer

`nostr_egress.mjs` owns `BRIDGE_SK` itself. Signing is not the only thing a private key does — the relay lane unseals with it — and banning `finalizeEvent` alone would leave the key spread across `bridge.mjs` with nothing to stop the next signer appearing beside a decrypt. It is derived in one module, never exported; callers get `bridgePubkey` / `hasBridgeKey` / `openSeal` / `openRumor` / `sealAndWrap`.

**`openSeal` and `openRumor` are deliberately two calls.** The caller must verify the seal's signature between them — the §2.4 authorship proof, which keeps the expensive second decrypt off unauthenticated input on the §7 DoS surface. My first cut was a single `openSealed()` returning both, which would have quietly decrypted the inner rumor before anything proved the seal authentic. That ordering is load-bearing and is now commented as such.

## Two deliberate deltas from the design — please rule

Per §2.2's own "a new type is a spec change" rule:

1. **`display_name`** — an external author's `kind:0` name: attacker-controlled, and rendered as *chrome* (bold, attribution line), not content. `inline_token` is the wrong type because its contract is "never leaves the code span". The identical strip already runs at the fetch site, so this is **not a live hole today** — but it lives there by convention, which is what A3 exists to replace. Fetch-site strip kept as belt-and-braces.
2. **`handle`** — an operator-configured `@handle` that must stay a **live** mention, because waking the approver is the point of a quarantine header. Separate from `display_name` precisely so an external value can never reach the slot that keeps its `@`.

**Doc fix needed:** §2.2 says "these eight are the whole vocabulary" while its own table lists **nine** names (`id`/`npub`/`hex` is one row of three). With these two it is twelve. The invariant that matters — closed set, membership asserted by the catalogue test — is unaffected; the prose count needs reconciling.

## Reads moved too

§1.1 scopes the three CLI read sites out, correctly — they author nothing. But enforcement is on the *import* (§2.3): if `bridge.mjs` kept `child_process` for three reads, the ban test would have to permit that import and would be blind to a write appearing beside it.

## Verification — 18 suites, 265 assertions, lint clean

Every gate **watched failing on purpose**, per `CLAUDE.md`:

| control | result |
|---|---|
| disable the `display_name` escaper | exactly its 3 guarding assertions go red |
| pretty-print the envelope JSON (INV-A3-4) | refused — the fence-break closed by contract, not luck |
| re-add `execFile` to the **real** `bridge.mjs` | ban test red |
| plant `finalizeEvent` + `BRIDGE_SK` in the **real** `bridge.mjs` | ban test red, both assertions |
| plant a `detail: string` slot | caught as an unknown slot type (INV-A3-5) |

**Nothing about what crosses changed, and that is asserted rather than claimed:** the return-lane carry text and the over-cap ack are both pinned byte-for-byte against their pre-#134 strings. The existing return-lane suites assert only *whether* a carry happened, never its bytes, so nothing else would have caught a drifted word.

Two of my own test bugs the discipline caught, recorded because they are the interesting part: `!out.includes('*evil*')` can never pass (`'**evil**'` contains it), and one negative control named `sealAndWrap` while actually exercising `buildBody` inside a redundant `throw` — a test passing for the wrong reason, which is the thing this suite exists to prevent.

## One observation, unchanged and pre-existing

`released_post` runs `defuseRefs` but not `defuseMarkup`, by the documented choice that "a vouched identity keeps its formatting" — so a vouched author's body *can* open a fence or mint a heading. I deliberately did not change it (A3 must not alter what crosses), but §1.2's table says a carried body appears "only inside a renderer that provably neutralizes it", and here the neutralisation is partial. Worth your eye; I did not treat it as in scope.

Remaining from §6: step 6 (`waggle-sealed` non-root) is a host change, and step 7 (the steward-identity question) is your call.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -117,14 +117,14 @@ that merely ran.** The suite was green through all of them.
 
 ## Tests
 
-`npm test` — 17 suites, against the real exported functions with synthetic events. No sockets,
+`npm test` — 19 suites, against the real exported functions with synthetic events. No sockets,
 no production state, no writes outside a temp dir.
 
-boot · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
+boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
 admission · message rendering · deployed-build verification · return lane · return-lane scan ·
 return-lane no-miss · relay ingress · tripwire union · tripwire detection drill · deploy runner · console Host check
 
-CI runs them on push and PR. **If a run reports fewer than 17, the branch is on a stale base.**
+CI runs them on push and PR. **If a run reports fewer than 19, the branch is on a stale base.**
 The count of record is the `test` script in `package.json`; a prose count that disagrees with it
 is the prose being wrong.
 
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ against what git says, because a stale build is invisible while every status sur
 
 ## Tests
 
-`npm test` runs 17 suites against the **real** exported functions with synthetic events — no
+`npm test` runs 19 suites against the **real** exported functions with synthetic events — no
 sockets, no production state:
 
 boot · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
--- a/docs/DESIGN_EGRESS_CHOKEPOINT.md
+++ b/docs/DESIGN_EGRESS_CHOKEPOINT.md
@@ -1,6 +1,6 @@
 # Design: the egress chokepoint — waggle carries, it never authors
 
-**Status:** proposed · **Tracks:** #134 A3, the structural half of the sev-1
+**Status:** BUILT — §6 steps 2–5 implemented in #168 (INV-A3-1…5 hold, both transports); step 6 is a host change and step 7 is a maintainer call · **Tracks:** #134 A3, the structural half of the sev-1
 **Review:** adversarial review returned on #140 — **verdict: sound, build it**, with five must-fixes; **this revision folds all five in**, plus his second-transport finding (§1.0). Re-review of the folded text is welcome but not gating.
 **Verified against:** `src/bridge.mjs`, read line by line for every egress site, 2026-07-31; re-verified against `main` @ `af7122f` on 2026-08-01.
 The refactors of #151–#153 have since moved every line in that file without touching an egress site, so citations here are by **function name, not line number** — see §1.1.
@@ -98,11 +98,28 @@ output" — carrying untrusted text is the entire product. The real line:
 
 | | who wrote it | how it may appear |
 |---|---|---|
-| **carried body** | an external author, signature-verified | **only** inside a renderer that provably neutralizes it — quoted, inert, attributed to its author |
+| **carried body** | an external author, signature-verified | only inside a renderer that neutralizes it **to the degree its trust tier earns** — see below. Always quoted or attributed to its author, never as waggle's own voice |
 | **machine notice** | the source code | a fixed template, with **typed** data in slots |
 | **free prose** | a caller at runtime | **nowhere. No path.** |
 
-Row 1 is not a weakening. `renderReleased` is already the strongest thing here: the `render_states`
+**Neutralization is tiered, and that is deliberate** *(the maintainer's ruling, 2026-08-01 — an
+earlier draft of this row said "provably neutralizes it" flat, which read as though a vouched body
+were defanged as hard as a quarantined one; the code has never done that, and should not)*:
+
+- **Quarantined** — refs *and* markup defused. An unvouched note cannot mint headings, rules, lists
+  or nested fences, because the impersonation that matters is a pending note dressed up as an
+  approval notice.
+- **Vouched** (released, granted, followed, mirrored) — refs defused unless a live grant earns
+  them (#94); **markup preserved on purpose.** A vouched identity keeps its formatting: reading as
+  an ordinary message rather than a machine log is the entire point of releasing something, and
+  trust is trust.
+
+So a vouched author *can* render a heading or a fence. That is an accepted consequence of the trust
+tier, not an oversight — they still cannot ping the room without a grant, so the residue is
+chrome-imitation by an identity someone already vouched for. A3 governs what **waggle** may author;
+it does not re-litigate what a trusted participant may format.
+
+Row 1 is not a weakening. `renderQuarantined` is already the strongest thing here: the `render_states`
 suite tests what it **refuses** — a hostile note trying to ping the approver, mint an `APPROVED BY`
 heading, break its quote, open a code fence — and it must come out inert *and still readable*.
 Row 3 is the one with no legitimate instance, and it is the one to delete.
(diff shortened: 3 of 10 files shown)


JAFairweather:
Review of the A3 build against `docs/DESIGN_EGRESS_CHOKEPOINT.md`. I wrote that design, so treat me as a checker of *whether the five must-fixes landed*, not as an independent judge of whether the design was right.

**Verdict: the design is implemented faithfully and the enforcement is real — but there is one blocking defect that CI cannot see, and it can permanently drop production messages on two lanes.**

---

## 🔴 Blocking — `handle` rejects any name with a space, and a rejection loses the message

`SLOT_TYPES.handle` is `/^[\w][\w.-]{0,63}$/` — no spaces. Two production paths feed it operator-config values that routinely contain spaces:

```
sealed_envelope.name      <- rec.name              (bridge.mjs:534)
quarantine_header.approver <- PUB.approverMention  (bridge.mjs:735)
```

Run against the real module:

```
sealed_envelope   name="Neil"       OK
sealed_envelope   name="My Dude"    THROW  slot rejected (handle): not a bare handle
quarantine_header approver="My Dude" THROW  slot rejected (handle): not a bare handle
quarantine_header approver="Jim F"   THROW
```

"My Dude" is a crew member named in #134 itself, and `config.example.json` documents `approver_mention` as `"<Buzz @name mentioned in quarantine headers…>"` — a Buzz display name, which commonly has a space.

**Why it's blocking rather than cosmetic: the event is already marked seen before `emit` resolves.**

- Sealed lane — `markSeen(ev.id)` at `bridge.mjs:558`, then `forward()` fires `emit(...).catch(e => err(...))` at `531`. A rejected slot logs `FORWARD[buzz] ERR` and the DM is **gone**: durable dedup drops the same event if another relay re-serves it, and it survives restart.
- Public in-door — `markSeen(ev.id)` at `781`, then `forwardPublic` → `emit`. If `approver_mention` has a space, **every quarantined arrival** fails to reach staging. That is the whole in-door path, not an edge case.

Before this PR both sites did `@${rec.name}` and accepted anything. The regression is the *rejection*, not the rendering.

**Why the suite is green anyway:** `tests/lane2_caps.mjs` uses `name: 'A'` / `'B'`. No fixture anywhere has a space in a name. This is precisely `CLAUDE.md`'s governing lesson — *every real bug in this project was invisible to a test that merely ran*.

**Is it live?** I can't tell from here — I have no access to the box's `config.json`, and #104 records that live config drifts from the repo, so the sanitized `AGENT_A` placeholders prove nothing. One command answers it:

```sh
jq -r '.recipients[].name, .public.approver_mention' config.json
```

Any value with a space or punctuation outside `[\w.-]` means that lane stops delivering the moment this merges.

### Suggested fix

Two changes, and I'd want both:

1. **`handle` should sanitize, not reject.** Strip what could break out of the template (`` ` ``, newline, `[]()*_~`), cap the length, keep the rest — including spaces. That preserves A3's guarantee (still a config-sourced value in a fixed template, still no free prose) without letting a config string become a delivery-stopping exception. It also restores the prior behaviour exactly.
2. **Validate the catalogue against config at boot.** Every config value that will feed a typed slot should be run through its escaper at startup, so a bad value is a loud refusal to start rather than a silent per-message drop hours later. That matches the repo's default-closed-and-loud posture, and it is the general fix — `handle` is the instance that happens to be reachable today, not the only slot that can throw.

Belt-and-braces alternative if you want it cheap: don't `markSeen` until `emit` resolves. I'd still do (1) and (2); losing a message on a *transport* failure is a separate pre-existing question I'm deliberately not widening this into.

---

## ✅ The five must-fixes — all landed, and one improved on

| Dennis's fix | Status |
|---|---|
| 1 · `console_ack.detail?` → typed `inline_token` | ✅ `inline_token` strips `` ` ``, newlines, `@*_`, caps at 24. `bridge.mjs:963`'s `raw` echo now goes through it, and `replyInStaging`'s `text` param is gone — the caller picks a **verb** and the words live in the catalogue |
| 2 · Ban the capability, not the spelling | ✅ `egress_ban.mjs` bans the **import** of `child_process` and the `finalizeEvent` **call** + `BRIDGE_SK` reference. Residual (`require('child_'+'process')`) stated in the file, not papered over |
| 3 · `wrapJson` single-line by contract | ✅ INV-A3-4 enforced in the escaper — rejects a newline with an explicit "never pretty-printed" message |
| 4 · Closed slot-**type** set | ✅ `renderSlots` rejects an unknown type as `INV-A3-5`, and refuses undeclared slots too — a caller passing a slot the template doesn't declare is refused rather than ignored. That second half wasn't in the design; it's a genuine improvement |
| 5 · "closed in-process", never bare | ✅ carried through the docs |

**Better than specified:** banning `BRIDGE_SK` itself rather than only `finalizeEvent` is the right call and the reasoning in the test is correct — signing isn't the only thing the key does, and a signer-only ban would leave the key spread across the file. The `why` enum also covers exactly the five values `bridge.mjs` produces (checked each call site), which is the kind of thing that silently breaks in production if it's off by one.

---

## Minor

- **`README.md` roster is stale.** It claims **19 suites** but lists **17** — missing `egress catalogue` and `egress ban`. `CLAUDE.md` has all 19. (Fourth hand-sync of this constant in two days; the #138 CI assert is still unbuilt and this is more evidence for it.)
- **The ban scanner reads `src/` only, non-recursively.** `readdirSync(SRC)` misses a future `src/lanes/foo.mjs`, and `tools/` is unscanned — several tools legitimately shell out and sign, so excluding them may be intended, but the scope should be a stated decision in the file rather than an artifact of the glob.
- `console_ack`'s `muted` verb renders `` `…` `` if `author` is omitted (it's optional). Harmless, and every caller passes it.

---

## What I checked, and what I did not

Read `egress.mjs` and `egress_ban.mjs` line by line; drove `renderTemplate` directly with hostile and realistic values; traced all seven Buzz call sites and both `markSeen` orderings; verified the `why` enum against every producing call site; verified `parseBuzzEventId` only ever yields 64-hex, so `id`-typed `targetId`/`parentId` are safe. `npm run lint` clean, `npm test` 19/19 green.

**Not checked:** `nostr_egress.mjs` in the same depth, and nothing was run against a live box — I'm in a limited environment with no host access. The blocking finding above is reasoned from the repo plus a one-command config check I can't perform.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #172, opened 2026-08-01
CI should assert the suite count, because the restated number has drifted four times in two days


JAFairweather:
Follow-up to #138, which synced the number by hand and proposed this. The proposal has since been vindicated three more times, so it is worth building rather than re-proposing.

## The record

| when | `package.json` | what the docs said |
|---|---|---|
| #138 | 13 | `CLAUDE.md` ten · `README` seven · `GETTING_STARTED` six · `ci.yml` "four … plus …" |
| after the refactors | 16 | hand-synced across three files |
| #163 | 17 | hand-synced across three files |
| #168/#170 | 19 | `CLAUDE.md` 19 ✅ · **`README` roster listed 17 while claiming 19** |

Four rounds, four hand-syncs, and the roster drifted from its own count on the last one. Every fix was correct and every fix was manual, which is the definition of a constant that wants a check.

## Why it is not cosmetic

`CLAUDE.md` makes the number load-bearing:

> **If a run reports fewer than 19, the branch is on a stale base.**

That is a stale-base detector whose calibration is maintained by hand. When it is wrong it does not fail loudly — it sits quiet through exactly the situation it exists to catch. The repo's own rule applies: *an alarm that always fires and one that never fires fail identically.*

## What to build

A CI step that derives the count from `package.json`'s `test` script and asserts three things agree:

1. **files** — `tests/*.mjs` on disk
2. **invocations** — `tests/…\.mjs` occurrences in the `test` script
3. **claims** — the number stated in `CLAUDE.md`, `README.md`, `docs/GETTING_STARTED.md`

Fail with a diff naming which file disagrees. Roughly 20 lines.

**Also assert the roster length**, not just the number — the README's last drift was a correct count beside a list of the wrong length, which a count-only check would have passed.

## Two things it must get right

- **A multiline-aware scan.** The `ci.yml` instance in #138 was missed by a line-based grep because `The four` / `suites` were split across lines; it was found by reading. A line-oriented check would have declared that file clean.
- **A negative control before it counts.** Land it with a commit that deliberately desyncs a doc, watch it go red, then revert. Per `CLAUDE.md`, an assertion that has only ever passed proves only that it ran — and that is doubly true for an assertion *about* whether checks ran.

## Alternative, if the check is judged not worth it

[...]

Issue #154, opened 2026-08-01
Refactor 4 of 4: split bridge.mjs — last, not first


JAFairweather:
Last of the four-step sequence. **Blocked on #151, #152, #153** — and ideally #134's A3 egress chokepoint.

> **Status update:** #151 (→ #155) and #152 (→ #156) have landed. #153 (→ #160) is in flight. Once it merges this is unblocked, and the file it operates on is materially different from the one this issue was written against.

## The state

`src/bridge.mjs` carries four lanes, rate limiting, dedup, rendering, the approval console, connection management and boot in one module.

To an outside reviewer this is the first thing they see, and it reads as prototype regardless of the quality inside it. It is worth being fair about *why*: this repo went from empty to a production bridge with a full suite in **four days**. That is velocity outrunning structure, not neglect. But the window to split it cheaply is now, not at 3,000 lines.

## Why this goes last — and what the first three actually bought

**`PUB.` is referenced 83 times as ambient module state.** Split first, and every one becomes a cross-module import or a parameter thread, *while the duplicated patterns get copied into each new file on the way out*.

Doing #151–#153 first means the same split now lands against code where the three worst duplications are already single primitives:

| | before | after |
|---|---|---|
| durable dedup stores | 3 hand-rolled copies, drifted | `durableSet` (#155) |
| lane rate limiters | 2 identical functions | `laneLimiter` (#156) |
| relay fan-out | 3 copies, one with a socket leak | `fanout` (#160) |

Each of those would otherwise have had to be duplicated *again* across the new module boundaries, or threaded through them. That is the difference between a clean split and a merge-conflict swamp.

## The boundaries already exist

They are drawn in comments and simply not enforced — the file is already sectioned:

```
--- config ---                    --- durable dedup ---
--- Lane-2 rate caps ---          --- NIP-DA admission tier ---
--- delivery ---                  --- public kind:1 delivery & routing ---
--- A7: NIP-09 deletion ---       --- In-Buzz approval console ---
--- Return lane (#40) ---         --- relay lane ---
--- relay connections ---         --- boot ---
```

Config and `PUB` want to become an explicit object passed in rather than an ambient global. **That is the actual work**, and the rest mostly follows.

## Bonus

This retires the single-line export at the bottom — a test seam that became a de-facto public API by accretion, and which #155/#160 have each had to extend. Once lanes are modules, tests import from the module that owns the function, and *"everything is exported so tests can reach it"* stops being the reason nothing is private.

## Verification

[...]

PR #179, opened 2026-08-01
test: make the two enforcement gates actually cover what they claim (#175, #172)

### LINKED ISSUE #175: A3 ban test scans src/ only, non-recursively — the enforcement gate has blind spots
Raised in my review of #168 and then not filed, so it has been living only in a PR comment. That is the wrong place for a gap in an enforcement gate — filing it properly.

## The gap

`tests/egress_ban.mjs` is what makes A3 structural rather than advisory. Its scan is:

```js
const SRC = resolve(dirname(fileURLToPath(import.meta.url)), '..', 'src')
const files = readdirSync(SRC).filter(f => f.endsWith('.mjs'))
```

Two blind spots:

1. **Non-recursive.** `readdirSync` without `recursive: true` does not descend. A future `src/lanes/foo.mjs` or `src/relay/bar.mjs` could import `child_process`, call `finalizeEvent`, or reference `BRIDGE_SK`, and the ban would pass. The repo has already moved in this direction once — #154/#161 split `bridge.mjs` into leaf modules — so more structure under `src/` is likely, not hypothetical.

2. **`tools/` is unscanned.** Several tools legitimately shell out and sign (`tripwire.mjs` signs alarm DMs with a separate key; `grant.mjs` uses a bunker signer; `approve.mjs` reposts through the delivery path). Excluding them may well be right — they are operator-invoked CLIs, not the bridge process, and A3's scope is what *waggle* can author. But right now that exclusion is an **artifact of the glob**, not a decision anyone made and wrote down.

## Why it matters more than it looks

The ban test is the reason INV-A3-2 and INV-A3-3 can be stated as properties rather than hopes. Its own header says it well:

> A chokepoint nobody is forced to use is a style guide. This suite is what forces it.

A forcing function with an undocumented scope boundary is one refactor away from not forcing anything — and it would stay green the whole time, which is the specific failure this repo keeps naming.

## What to do

- Make the scan **recursive** over `src/`.
- Decide `tools/` explicitly, and write the decision into the test file rather than leaving it implicit. If tools are out of scope, say why (operator-invoked, not the bridge process, different threat model). If any tool should be in scope, name which and enforce it.
- Consider asserting the **scanned file count** is non-zero and matches expectation, so a future path change that silently scans nothing fails loudly — same size-floor reasoning as #174.

## Negative control

The existing controls plant a violation and require the scanner to catch it, which is right. Extend them: plant a violation in a **subdirectory** and require the scanner to catch that too. Today that control would fail, which is the point of running it.

JAFairweather:
Closes #175 and #172. Both gates were green while blind to a class they were supposed to police — tests only, no `src/` changes.

## #175 — the A3 ban test read one directory level

`readdirSync(SRC)` doesn't descend. A future `src/lanes/foo.mjs` could import `child_process`, call `finalizeEvent`, or touch `BRIDGE_SK`, and the ban would pass. **Not hypothetical** — #154/#161 already split `bridge.mjs` into leaf modules, so more structure under `src/` is the direction of travel.

Now recursive, with a **negative control that plants the violation in a subdirectory**. The existing control planted it in the directory the scanner already read, so it passed even while the scan was one level deep — it tested the regex, not the reach. Reverting the scan to non-recursive makes exactly the two new checks red:

```
FAIL — NEGATIVE CONTROL — the scan reaches into subdirectories
FAIL — NEGATIVE CONTROL — and a violation planted there IS detected
```

**Scope is now a decision, not an artifact of the glob.** `src/` is the bridge process and is covered. `tools/` is *not*, and the reason is written into the file: those are operator-invoked CLIs that legitimately do the banned things by design — `tripwire.mjs` signs alarm DMs with a **separate** alarm key, `grant.mjs` drives a remote signer, `approve.mjs` reposts through the delivery path. Banning them would forbid the thing they exist to do. The boundary is **who initiates**, not the directory name: a tool the bridge process invokes belongs in `src/` and under the ban.

**Plus a size floor on the scan itself.** Every per-file ban is vacuously true over an empty file list — a scan that finds nothing passes every ban. Same shape as #174's floor, one level up.

## #172 — the suite count

**13 → 16 → 17 → 19 → 20 in two days**, hand-synced across three files each round, and once README carried a correct count beside a roster of the wrong length. `CLAUDE.md` makes that number load-bearing — *"if a run reports fewer than N, the branch is on a stale base"* — so a hand-maintained number is a detector that sits quiet through exactly the case it exists to catch.

`tests/suite_count.mjs` derives the count from `package.json` and checks:

- every invoked suite exists on disk
- **every file on disk is invoked** — an orphan is a test nobody runs
- no suite invoked twice
- every stated count in every candidate doc (`CLAUDE.md`, `README.md`, `GETTING_STARTED.md`, `ci.yml`, `SECURITY.md`)
- the stale-base threshold line specifically
- **the roster length** — because a count-only check passes the README failure

**The scan is multiline on purpose.** A line-based grep missed a stale count in `ci.yml` because "The four" and "suites" sat on different lines; it was found by reading, not by the check.

**Four negative controls, each watched failing:**

| control | result |
|---|---|
| desync a stated count | ✅ red |
| desync the stale-base threshold alone | ✅ red |
| correct count, short roster (the README failure) | ✅ red |
| count split across a line wrap (the `ci.yml` failure) | ✅ red |

Restored, all green. `npm run lint` clean · `npm test` **20/20**.

## Merge note

This and **#178** both add a suite as number 20 and both touch the doc counts, so whichever lands second needs a rebase and a bump to 21. The new assert makes that mechanical rather than a thing to remember — it'll fail loudly if the rebase forgets it, which is the whole point.

Not merging this myself, per #177.

---

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -148,7 +148,7 @@ that merely ran.** The suite was green through all of them.
 `npm test` — 22 suites, against the real exported functions with synthetic events. No sockets,
 no production state, no writes outside a temp dir.
 
-suite roster · boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
+boot · suite roster · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
 admission · message rendering · deployed-build verification · return lane · return-lane scan ·
 return-lane no-miss · return-lane pending · relay ingress · tripwire union · tripwire detection drill · deploy runner · console Host check · undelivered record
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -145,14 +145,14 @@ that merely ran.** The suite was green through all of them.
 
 ## Tests
 
-`npm test` — 21 suites, against the real exported functions with synthetic events. No sockets,
+`npm test` — 22 suites, against the real exported functions with synthetic events. No sockets,
 no production state, no writes outside a temp dir.
 
-boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
+suite roster · boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
 admission · message rendering · deployed-build verification · return lane · return-lane scan ·
 return-lane no-miss · return-lane pending · relay ingress · tripwire union · tripwire detection drill · deploy runner · console Host check · undelivered record
 
-CI runs them on push and PR. **If a run reports fewer than 21, the branch is on a stale base.**
+CI runs them on push and PR. **If a run reports fewer than 22, the branch is on a stale base.**
 The count of record is the `test` script in `package.json`; a prose count that disagrees with it
 is the prose being wrong.
 
--- a/README.md
+++ b/README.md
@@ -107,9 +107,9 @@ against what git says, because a stale build is invisible while every status sur
 `npm test` runs 22 suites against the **real** exported functions with synthetic events — no
 sockets, no production state:
 
-boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
+boot · suite roster · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
 admission · message rendering · deployed-build verification · return lane · return-lane scan ·
-return-lane no-miss · relay ingress · tripwire union · tripwire detection drill · deploy runner · console Host check · undelivered record
+return-lane no-miss · return-lane pending · relay ingress · tripwire union · tripwire detection drill · deploy runner · console Host check · undelivered record
 
 The rendering suite is the one to read if you are reviewing. It tests what the bridge **refuses**,
 not only what it does: a hostile note tries to ping the approver, mint an `APPROVED BY` heading,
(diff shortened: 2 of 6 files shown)


JAFairweather:
I built #175 and #172 independently before seeing this — my fault for triaging the issue list once and then working a long stretch without re-checking open PRs, which is exactly what `CLAUDE.md` says not to do. This PR got here first and covers both issues, so I am closing mine (#180, #182) rather than competing.

One substantive difference is worth deciding rather than dropping, because we reached opposite conclusions on it.

**You exclude `tools/` deliberately and document why. I scanned it against a roster.**

Your reasoning is sound and I do not think tools should be *banned* from signing — that would forbid the thing they exist to do. But the roster is not a ban; it is an enumeration:

```js
const TOOL_SIGNERS = {
  'grant.mjs':              'NIP-DA 440/441 under the GRANTOR key, never the bridge key',
  'tripwire.mjs':           'alarm DM under a SEPARATE zero-authority key — rule 1',
  'publish_relay_list.mjs': 'NIP-65 kind:10002 under the member\'s own key',
  …
}
```

A **new** signing tool, or an existing one that starts signing, fails until someone adds it with a reason. The exclusion stays, but it becomes a decision each time rather than a category.

The argument for it is your own framing: *"the boundary is WHO INITIATES, not the directory name."* I agree — and that boundary is exactly what a directory-shaped exclusion cannot enforce. A future `tools/auto-responder.mjs` invoked by a timer rather than a human sits in `tools/`, is not operator-initiated, and would be unscanned by either of our versions; mine at least makes someone type its name and a justification.

There is also one rule I would not want scoped out under any reading: **no tool may reference `BRIDGE_SK`.** The bridge's own key belongs to `src/nostr_egress.mjs` alone, whatever the caller's privileges, and that is cheap to assert across `tools/` regardless of how the signing question is decided.

Happy either way — this is a real design call and it is the maintainer's. If you want the roster, it is two blocks and lifts cleanly out of #180 before I close it.

PR #180, opened 2026-08-02
test(ban): scan src/ recursively and make tools/ a written decision (#175)

### LINKED ISSUE #175: A3 ban test scans src/ only, non-recursively — the enforcement gate has blind spots

JAFairweather:
Closes #175.

Two blind spots in the suite that makes A3 structural rather than advisory.

**Recursive.** `readdirSync` without it doesn't descend, so a future `src/lanes/foo.mjs` could import `child_process`, call `finalizeEvent` or touch `BRIDGE_SK` and the gate would pass — staying green the whole time. The tree has already moved that way once (#154 split `bridge.mjs` into leaf modules), so nesting is a matter of when, not whether.

**`tools/` was excluded by accident.** Excluding it is *right* — A3's scope is what the bridge **process** may author, and `tools/` is operator-invoked CLIs, several of which legitimately sign. But that exclusion was an artifact of the glob rather than a decision anyone wrote down.

It is now scanned against an explicit roster — one entry per signing tool, each with the reason it's permitted:

| tool | why it may sign |
|---|---|
| `grant.mjs` | NIP-DA 440/441 under the **grantor** key, never the bridge's |
| `participant-init.mjs` | mints/publishes under that **participant's own** key |
| `publish_relay_list.mjs` | NIP-65 `kind:10002` under the **member's own** key |
| `retract.mjs` | NIP-09 deletion for a note the invoking key authored |
| `tripwire.mjs` | alarm DM under a **separate zero-authority** key — rule 1 |
| `waggle-init.mjs` | operator setup CLI; shells out to the `buzz` binary |

A new signing tool, or an existing one that starts signing, now **fails until someone adds it here**. The exclusion becomes a deliberate act instead of a side effect.

**One rule no tool may opt out of:** none may touch `BRIDGE_SK`. The bridge's own key belongs to `src/nostr_egress.mjs` alone, whatever the caller's privileges. That's asserted for every file under `tools/`, rostered or not.

## Negative controls

A scanner that reaches nothing and a codebase with no violations are indistinguishable, so both are proven:

- a planted `child_process` import in a **nested** `src/` directory must be reached and caught
- an unrostered signing tool must fail the roster check

### DIFF
--- a/tests/egress_ban.mjs
+++ b/tests/egress_ban.mjs
@@ -11,12 +11,34 @@
 // static check, and anyone with commit access can route around a lint rule. What an import ban
 // reliably stops is the ACCIDENT — the next contributor reaching for the convenient thing — which
 // is the actual threat model here. It is not airtight and nothing below claims it is.
-import { readFileSync, readdirSync, writeFileSync, unlinkSync } from 'node:fs'
-import { join, dirname, resolve } from 'node:path'
+import { readFileSync, readdirSync, writeFileSync, unlinkSync, mkdirSync, rmSync } from 'node:fs'
+import { join, dirname, resolve, relative } from 'node:path'
 import { fileURLToPath } from 'node:url'
 
-const SRC = resolve(dirname(fileURLToPath(import.meta.url)), '..', 'src')
-const files = readdirSync(SRC).filter(f => f.endsWith('.mjs'))
+const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..')
+const SRC = join(ROOT, 'src')
+const TOOLS = join(ROOT, 'tools')
+
+// RECURSIVE, deliberately (#175). `readdirSync` without it does not descend, so a future
+// src/lanes/foo.mjs could import child_process or call finalizeEvent and this gate would pass —
+// green the whole time, which is the specific failure this repo keeps naming. The tree has
+// already moved this way once (#154 split bridge.mjs into leaf modules), so nesting is a matter
+// of when, not whether.
+const mjsUnder = (root) => {
+  const out = []
+  const walk = (dir) => {
+    for (const ent of readdirSync(dir, { withFileTypes: true })) {
+      const full = join(dir, ent.name)
+      if (ent.isDirectory()) { if (ent.name !== 'node_modules') walk(full) }
+      else if (ent.name.endsWith('.mjs')) out.push(full)
+    }
+  }
+  walk(root)
+  return out
+}
+
+const srcFiles = mjsUnder(SRC)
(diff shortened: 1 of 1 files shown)


JAFairweather:
Superseded by #179 on the recursive fix. The one difference — scanning `tools/` against a roster versus excluding it by decision — is raised as a comment on #179 for the maintainer to settle, since we reached opposite conclusions and it is a design call rather than a defect. Closing rather than competing.

PR #182, opened 2026-08-02
test: assert the suite roster instead of hand-syncing it (#172)

### LINKED ISSUE #172: CI should assert the suite count, because the restated number has drifted four times in two days

JAFairweather:
Closes #172.

The count has drifted four times in two days. Every fix was correct and every fix was manual — the definition of a constant that wants a check. And `CLAUDE.md` makes it load-bearing:

> **If a run reports fewer than N, the branch is on a stale base.**

That's a stale-base detector calibrated by hand. When it's wrong it doesn't fail loudly; it sits quiet through exactly the situation it exists to catch — this repo's own rule about an alarm that always fires and one that never fires failing identically.

## What it asserts

- **every invocation names a file that exists** — otherwise it fails at run time as a confusing `cannot find module` rather than as a roster problem
- **every `tests/*.mjs` on disk is invoked** — the one worth having. An orphan test file *looks like coverage in the tree and runs nowhere*, and it can never go red because it never goes anywhere.
- **no suite is invoked twice** — a duplicate inflates the number without adding coverage
- **every restated count agrees** — `CLAUDE.md`, `README.md`, `GETTING_STARTED.md`, including the `fewer than N` floor, which *is* the stale-base detector

It counts itself, which is consistent rather than clever: it's invoked from the same script it reads.

## Negative controls

Both halves watched failing on purpose:

```
FAIL — CLAUDE.md: every stated count is 20 — found 19
FAIL — every test file on disk is actually invoked (no orphans) — never run: tests/__orphan_control__.mjs
```

Roster is now 20 and the three docs are synced — for the last time by hand.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -145,14 +145,14 @@ that merely ran.** The suite was green through all of them.
 
 ## Tests
 
-`npm test` — 19 suites, against the real exported functions with synthetic events. No sockets,
+`npm test` — 20 suites, against the real exported functions with synthetic events. No sockets,
 no production state, no writes outside a temp dir.
 
-boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
+suite roster · boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
 admission · message rendering · deployed-build verification · return lane · return-lane scan ·
 return-lane no-miss · relay ingress · tripwire union · tripwire detection drill · deploy runner · console Host check
 
-CI runs them on push and PR. **If a run reports fewer than 19, the branch is on a stale base.**
+CI runs them on push and PR. **If a run reports fewer than 20, the branch is on a stale base.**
 The count of record is the `test` script in `package.json`; a prose count that disagrees with it
 is the prose being wrong.
 
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ against what git says, because a stale build is invisible while every status sur
 
 ## Tests
 
-`npm test` runs 19 suites against the **real** exported functions with synthetic events — no
+`npm test` runs 20 suites against the **real** exported functions with synthetic events — no
 sockets, no production state:
 
 boot · egress catalogue · egress ban · durable dedup store · relay fan-out · quarantine gating · deletion propagation · sealed-lane rate caps · grant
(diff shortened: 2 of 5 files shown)


JAFairweather:
Superseded by #179, which covers #172 with an equivalent check (existence + orphans). Mine additionally asserted the `fewer than N` floor and rejected duplicate invocations — both small, both worth lifting into #179 if wanted. Closing rather than competing.

Issue #162, opened 2026-08-01
deploy-runner restarts production for docs-only merges — gate the restart on the ship list actually changing


JAFairweather:
The runner's only "is there work to do?" test is a SHA comparison:

```sh
if [ "$TARGET_SHA" = "$DEPLOYED_SHA" ]; then
  log "already current — nothing to do"; exit 0
fi
```

Anything past that point unconditionally rsyncs, runs `npm ci`, rewrites `DEPLOYED_SHA` and **restarts the lane** — with no check that any *shipped* file actually changed. The ship list is `src tests tools package.json package-lock.json config.example.json`, so a commit touching only `docs/`, `CLAUDE.md`, `README.md`, `.github/` or `deploy/` ships **byte-identical content** and still restarts production.

That is not a rare case here. Of the twelve PRs merged on 2026-08-01, a clear majority were docs-only. Each one would now bounce the bridge.

### Why it is worth fixing rather than tolerating

A restart is not free on this lane:

- open REQ subscriptions to all three public relays drop and re-open
- the boot path re-runs backfill against the `since` watermark
- the approval console re-subscribes, and admission grants re-verify
- there is a window, however short, where the lane is not listening

None of that is dangerous — the durable dedup and watermark are exactly what make a restart safe — but it is avoidable churn on a production path, incurred for a change that provably cannot affect the running code. It also makes the journal noisier, which raises the cost of noticing a restart that *does* matter.

### Fix

Between the SHA comparison and the deploy, ask whether the ship list changed:

```sh
git -C "$HUB" diff --name-only "$DEPLOYED_SHA" "$TARGET_SHA" -- $SHIP
```

If it reports nothing, record the new `DEPLOYED_SHA` and exit 0 without rsync/`npm ci`/restart — the tree is already correct for that commit, and `verify-deployed.sh` will still confirm it. Guard for `DEPLOYED_SHA=none` (first run) and for the SHA being absent from the hub, and fall through to a full deploy in either case rather than skipping.

Worth pairing with #136, which is also about `DEPLOYED_SHA` bookkeeping around the restart.

### Status of this report

**Derived from the source, not observed.** I read the script rather than merging a docs-only commit to watch production bounce. The reasoning is a straight read of the control flow above, but it deserves the negative control before the fix is believed: land the change, merge a docs-only commit, and confirm the tick logs "no shipped files changed" and the lane's start time does *not* move.

JAFairweather:
[...]

PR #183, opened 2026-08-02
deploy-runner: record DEPLOYED_SHA only after verify passes (#136)

### LINKED ISSUE #136: deploy-runner: write DEPLOYED_SHA after verify, not before (fast-follow to #130)
Fast-follow to the deploy runner (#130), raised in review. `DEPLOYED_SHA` is written *before* `verify-deployed` runs. If verify ever fails, the SHA already claims success, so the next tick reads `already current` and a persistent drift **alarms only once** instead of every tick. Writing the SHA only *after* verify passes self-heals: a crash mid-restart then re-deploys (idempotent rsync) on the next tick rather than recording a success it never confirmed. Two-line move; do it when wiring the bootstrap (hub clone + timer install), which needs a root/breakglass session anyway. Neil's surface.

JAFairweather:
Closes #136.

## The bug

`DEPLOYED_SHA` was written **before** the restart, and therefore before verify. So a tree that *failed* verification still claimed success — the next tick read "already current", skipped, and **a standing drift alarmed exactly once.**

An alarm that fires once for an ongoing fault is indistinguishable from one that was handled. That's the same shape as a detector that never fires.

## The fix

Write it only on the verified path. A crash anywhere before that line leaves the old SHA, so the next tick simply re-deploys — `rsync` is idempotent, the dep install is idempotent, a redundant restart costs seconds — then re-verifies and re-alarms. A standing fault keeps failing, which is what it should do.

The old comment was protecting something real (*"a crash mid-restart should leave provenance truthful"*), so it's answered rather than ignored: **provenance becomes "last VERIFIED sha" instead of "last shipped sha"** — which is the question `verify-deployed.sh` is asking anyway.

## The property nothing tested

The existing drift case checked exit 1 and an ALARM on the *first* tick. The issue is entirely about the *second*. Four new checks:

- a failed verify records **no** SHA
- second tick against the same broken tree → still exit 1
- second tick → alarms **again**
- second tick → **re-ships** rather than skipping

Restoring the old ordering turns all four red.

## One thing worth keeping in the record

My first version of the second-tick check asserted `!/already current/` — and **failed against the runner's own alarm text**, because I'd written an alarm that quoted the phrase it was preventing.

Two fixes, both worth having: the alarm no longer quotes the skip-path wording (a log line containing the phrase a grep looks for is a log line that defeats the grep), and the check now asserts the **positive** — that it re-shipped. A negative string assertion is only as good as every other string in the output.

`npm run lint` clean · `npm test` **19/19**. No suite added, so no count conflict with #178/#179.

Not merging this myself, per #177.

---

---

### DIFF
--- a/deploy/deploy-runner.sh
+++ b/deploy/deploy-runner.sh
@@ -170,19 +170,34 @@ log "shipping code into $TREE"
 log "installing deps in $TREE"
 ( cd "$TREE" && sh -c "$NPM_CMD" ) || { alarm "dependency install failed in $TREE"; exit 1; }
 
-# record the SHA atomically BEFORE restart, so a crash mid-restart still leaves the tree's
-# provenance truthful for the next verify.
-printf '%s\n' "$TARGET_SHA" > "$TREE/DEPLOYED_SHA.tmp" && mv "$TREE/DEPLOYED_SHA.tmp" "$TREE/DEPLOYED_SHA"
-
 log "restarting $UNIT"
 sh -c "$RESTART_CMD \"$UNIT\"" || { alarm "restart of $UNIT failed"; exit 1; }
 
 # --- post-deploy: what is on disk MUST equal git at the sha we shipped -------------------
 log "verifying deployed tree against $SHORT"
 if sh "$HUB/deploy/verify-deployed.sh" "$LANE" "$TREE" "$TARGET_SHA"; then
+  # DEPLOYED_SHA is written ONLY here — after verify passes (#136).
+  #
+  # It used to be written before the restart, reasoning that a crash mid-restart should still
+  # leave the tree's provenance truthful. That ordering costs more than it buys: if verify FAILS,
+  # the file already claims success, so the next tick reads "already current", skips, and a
+  # PERSISTENT DRIFT ALARMS EXACTLY ONCE. An alarm that fires once for an ongoing fault reads as
+  # handled — the same shape this repo keeps re-learning.
+  #
+  # Writing it only on a verified deploy self-heals instead. A crash anywhere before this line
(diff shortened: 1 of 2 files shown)


JAFairweather:
Reviewed against my own independent implementation of #136 (which I am not opening as a competing PR — this one got here first and is substantively the same). One concrete gap, and it is the same bug this PR fixes:

**The no-op path still writes `DEPLOYED_SHA` before verify.**

```sh
log "no shipped files changed between … and $SHORT — recording without restarting"
printf '%s\n' "$TARGET_SHA" > "$TREE/DEPLOYED_SHA.tmp" && mv …   # <-- before
if sh "$HUB/deploy/verify-deployed.sh" "$LANE" "$TREE" "$TARGET_SHA"; then
```

That branch is new as of #162/#169 this evening, and it inherited the ordering this PR is correcting on the full path. The consequence is identical and arguably worse, because the no-op path is the one that will run most often once docs-only merges stop restarting: a tree that fails verification leaves behind a file claiming it passed, the next tick reads `already current`, and a persistent drift alarms once instead of every tick.

Swapping those two lines the same way you swapped the full path closes it:

```sh
if sh "$HUB/deploy/verify-deployed.sh" "$LANE" "$TREE" "$TARGET_SHA"; then
  printf '%s\n' "$TARGET_SHA" > "$TREE/DEPLOYED_SHA.tmp" && mv "$TREE/DEPLOYED_SHA.tmp" "$TREE/DEPLOYED_SHA"
```

One test worth stealing too, if you want it: asserting the SHA is *absent* after a failed verify proves the file was not written, but not that the retry actually happens. A follow-up tick on the same tree, without the drift injection, asserting `deploy OK` and only *then* a recorded SHA, closes that — otherwise "it will retry next tick" stays an inference.

PR #254, opened 2026-08-05
feat: classify direct Buzz policy submissions

JAFairweather:
## What this proves

Adds bounded direct `POST /events` submission for one already-signed policy event:

- exact event, endpoint, method, payload hash, signer identity, nonce, and NIP-OA header are verified before network I/O;
- redirects are forbidden and response bodies are capped at 64 KiB;
- a matching authoritative `200` (including Buzz's stored-event duplicate response) is accepted;
- non-rate-limit `4xx` is terminal refusal, while `429` holds the same event for fresh-auth retry;
- network loss, malformed success, mismatched event ID, oversize response, and server failure are ambiguous and never authorize re-signing.

This is based on merged artifact construction PR #251 and advances #54. It replaces auto-closed stacked PR #252; GitHub closed that PR only because its deleted base branch could not be reopened.

## Verification

- `npm run lint`
- `CONFIG_PATH="$PWD/config.example.json" npm test` — all 41 suites green
- negative controls for payload substitution, identity substitution, false success, oversize response, and network ambiguity

PR #256, opened 2026-08-05
feat: orchestrate recoverable off-box policy delivery

JAFairweather:
## What changed

- composes verified evidence, closed policy decisions, exact Bunker-compatible signing, durable preparation, bounded direct submission, and signed receipts into one transaction;
- recovers the same prepared event after restart without re-signing kind:9;
- retries held or ambiguous delivery with a fresh NIP-98 authorization around the same event;
- replays terminal receipts byte-for-byte without signing or network access;
- adds closed refusal receipts that cannot claim an event was published;
- verifies recovered event bytes against the exact policy decision, destination, poster, and authorization tag.
- exposes the transaction through a zero-argument forced-command runner with one bounded UTF-8 stdin request, a fixed private config, exact Bunker/poster binding, and canonical receipt-only stdout.

This branch stacks the exact changes from #253, #254, and #255. Once those land, its diff reduces to the service composition and receipt/recovery boundary.

## Why

The earlier slices established the individual trust boundaries but did not yet provide the complete remote-signer delivery transaction required by #54. In particular, an accepted response followed by a crash must never lead to a newly signed replacement event.

## Verification

- `git diff --check`
- `npm run lint`
- full 43-suite `npm test`
- positive accepted/replay flow
- crash-after-send recovery with byte-identical event
- 429 retry with fresh authorization only
- authoritative refusal receipt
- substituted signer rejection
- world-readable/symlinked/widened config rejection and bounded-stdin controls

Part of #54.


JAFairweather:
Superseded by #258, rebased onto the corrected direct-submission boundary in #257. #256 depended on closed #254 and is intentionally not mergeable.

PR #269, opened 2026-08-05
Diagnose and close route signer mutations

JAFairweather:
## What changed

- gives every route-seal refusal a bounded, field-specific diagnosis without exposing encrypted content
- requires the signer response to contain exactly the seven canonical signed-event fields
- verifies a JSON wire copy so signer-specific object metadata cannot bypass signature validation
- adds negative controls for missing signer output and a validly signed but schema-widened seal

## Why

The deployed saved-session fix reached the owner signer successfully, but the live Alby response was refused by the exact route-seal boundary. The old generic error could not distinguish a missing event, invalid signature, identity change, field mutation, or schema widening. This preserves the fail-closed boundary while making the live incompatibility observable enough to repair safely.

## Evidence

- focused control/route suite: 42/42
- all 44 suites passed
- lint and `git diff --check` passed

### DIFF
--- a/console/task-route-envelope.mjs
+++ b/console/task-route-envelope.mjs
@@ -1,5 +1,23 @@
 import { finalizeEvent, generateSecretKey, getEventHash, nip44, verifyEvent } from 'nostr-tools'
 
+const SIGNED_EVENT_KEYS = ['content', 'created_at', 'id', 'kind', 'pubkey', 'sig', 'tags']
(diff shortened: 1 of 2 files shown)


JAFairweather:
Superseded by #270. The private waggle-test exact-head review found that this branch verified a wire copy but returned and wrapped the original signer object, while #270 serializes once and wraps the exact verified plain object. Closing without merge.

PR #279, opened 2026-08-05
React to messages after return-lane relay

JAFairweather:
## Outcome

When a Buzz channel message successfully triggers a sealed return-lane carry, Waggle adds a fixed 👍 reaction to the exact source event so the sender can see that relay action occurred.

## Safety and failure semantics

- the reaction is a closed egress-catalogue operation; callers cannot choose emoji or prose
- one reaction per source message even when it fans out to multiple recipients
- carry dedup prevents repeat reactions on rescans
- reaction work is persisted and retried independently
- reaction failure never causes the sealed carry to be repeated
- dry-run and hermetic test modes perform no live reaction transport

## Verification

- `npm run lint`
- full 46-suite `npm test` gate
- focused return-lane test includes failed-reaction retry and no re-carry
- egress catalogue test asserts exact argv and fixed 👍

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -245,6 +245,7 @@ firewall → root-SSH off → reboot test. (Same verify-before-lock philosophy a
 SEALED_LANES=off
 FORWARD_MODE=dryrun            # buzz only at cutover step 4c
 BUZZ_RELAY_URL=...
+BUZZ_EVENT_ENDPOINT=https://<fixed Buzz host>/events  # exact-event API; never a wss relay URL
 BUZZ_PRIVATE_KEY=...           # the BRIDGE posting identity — never an agent key
 BUZZ_AUTH_TAG=...
 ```
--- a/deploy/waggle-read.service
+++ b/deploy/waggle-read.service
@@ -5,7 +5,8 @@
 #   SEALED_LANES=off           # this instance runs the public lane ONLY (sealed lanes are
 #                              # per-process-deduped — two instances running them = double delivery)
 #   FORWARD_MODE=dryrun        # flip to 'buzz' only at cutover (see deploy/README.md)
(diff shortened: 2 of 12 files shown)


JAFairweather:
Release blocker found during adversarial review: `relay_action_reaction` produces a poster-signed kind:7 through the Buzz CLI but never records that event ID in the send journal. The tripwire intentionally fetches every event authored by the poster, so each legitimate reaction would be reported as unauthorized signing.

Parsing and journaling CLI stdout after success is not sufficient. The CLI signs internally; a crash after acceptance but before journal commit loses the accepted event ID. A retry signs a different kind:7, which Buzz rejects as an already-active duplicate, leaving the original accepted ID permanently unaccounted. Exempting kind:7 from the tripwire would create a signing-theft blind spot.

This PR must not merge in its current shape. A safe implementation needs the #54 transaction boundary: prepare one exact signed event, durably journal/prepare its ID, then submit that same event across retry/restart. CI green does not cover this crash/tripwire invariant.

JAFairweather:
The revised head `e86c08bf909c08c741a50f5b9af544b7c25c39b4` does not close the release blocker. `commitLandedCarry()` now durably records that a reaction is owed before carry completion, which is useful, but the tripwire failure occurs at a later boundary: `confirmRelayAction()` still invokes a CLI that signs and submits internally, and it never receives or durably journals the exact accepted kind:7 event ID.

A crash after Buzz accepts that reaction but before `rlReactionSeenStore.commit()` leaves an on-relay poster-signed event absent from every send journal. Restart knows only that an abstract reaction is owed; retry signs a different event, which Buzz rejects as an already-active duplicate. The original accepted event remains permanently unaccounted and the tripwire alarms.

Required invariant: construct one exact signed kind:7 event, durably prepare/journal that exact event ID before submission, and retry the same signed bytes across ambiguity/restart. Durable abstract debt is not a substitute for durable exact-event preparation.

PR #289, opened 2026-08-05
fix: every owner-control surface reuses the Access session

### LINKED ISSUE #285: console: adopt the shared signer session on Following and Config (a bunker-only owner is locked out)
Tracks JAFairweather/nave.pub#110. Three small correctness gaps found while auditing the console for a
cross-app UX pass. All three are cases where the console's behaviour or its own text has drifted from
what it actually does.

## 1. The shared signer session is adopted by one of three call sites

`console/signer-session.mjs` exists precisely so a page cannot silently sign under a stray
`window.nostr` when a session already exists — its header says so. But only
`console/task-routes.mjs` calls `consoleSigner()`. `following.html` and `config.html` call
`nip07Signer()` **raw**.

Effect: **an owner who signs in with a bunker can activate an agent channel route but cannot toggle
publishing or manage following.** They get a raw provider error, because there is no extension to ask.
Worth fixing before more owner controls land on those two pages.

## 2. Two of four readers omit the forward-skew clamp

`task-routes.mjs` and `config-operations.mjs` reject a signed state whose `observed_at` is more than a
minute in the future, as well as applying the 15-minute staleness rule. `following.html` and
`config.html` apply only the staleness rule. Same signed record, two different freshness contracts —
worth making one contract, in one place, since all four now depend on it.

## 3. `Read-only, and deliberately.` is stale copy

`console/index.html:243` still tells the reader that issuing and revoking "belongs where the key lives"
and points at `tools/grant.mjs`, on a page that ships `Sign and publish` and a per-row `Revoke`. The
note was true and is now the opposite of true.

Suggested replacement framing, which is what the console actually does and is worth stating plainly:
reading needs no key and no bridge access; signing happens in your signer and never in the page; and
the console never receives host access — authority moves as signed events. That last sentence is the
invariant this repo has actually built, and it is stronger than "read-only" ever was.

JAFairweather:
Closes #285. Tracks JAFairweather/nave.pub#110 (Wave 1).

⚠ **Merging this deploys to production.** Reviewer's call.

Three gaps found while auditing the console, all cases where its behaviour or its own text had drifted
from what it actually does.

## 1. A bunker-only owner was locked out of two pages

`signer-session.mjs` exists precisely so a page cannot silently sign under whichever `window.nostr` is
injected — its header says so. Only `task-routes.mjs` used it. `following.html` and `config.html` called
`nip07Signer()` **raw**.

The effect was a real lockout: an owner signed in with a bunker could **activate an agent channel
route** and **could not** toggle publishing or manage following. Those pages reached for an extension
that was not there and surfaced a raw provider error.

**Invisible to anyone testing with an extension installed** — which is why it now has a test rather
than care.

`index.html` keeps `nip07Signer` deliberately, and the test **names the exception**: it is the sign-in
picker, so it *creates* the session everything else reuses. An unexplained allowance is how the other
two drifted in the first place. `consoleSigner` falls back to the browser signer when no session
exists, so extension users are unaffected.

## 2. Two of four readers had no forward-skew clamp

`task-routes.mjs` and `config-operations.mjs` reject a signed state dated more than a minute ahead;
`following.html` and `config.html` applied only the 15-minute staleness rule. So **the same record was
"fresh" on one page and rejected on another.** One contract now, asserted on all four.

## 3. `Read-only, and deliberately.` was stale

It told the reader that issuing and revoking *"belongs where the key lives — `tools/grant.mjs`"*, on a
page that ships `Sign and publish` and a per-row `Revoke`. True once; the opposite of true now.

Replaced with what the console actually guarantees, which is **stronger** than read-only ever was:
reading needs no key at all; acting needs your signer and nothing more, so your key never enters the
page; and the console is never a remote shell into the bridge, because authority moves as signed
events. That is the invariant this repo built, and the one nave.pub AD-12 adopts estate-wide.

## The test earns its place

`tests/console_signer.mjs`. Run against **unfixed `origin/main`** it fails exactly the five assertions
it should:

```
FAIL — no console surface calls nip07Signer() directly except index.html
FAIL — following.html signs through the shared console session
FAIL — config.html signs through the shared console session
FAIL — following.html rejects a future-dated signed state
FAIL — config.html rejects a future-dated signed state
10/15 passed
```

It also pins the session-precedence behaviour that matters: an established session beats an injected
provider, an unparseable session is discarded rather than trusted, and a session the signer refuses to
restore (a local key, which `nave-connect` will not resurrect) falls back rather than throwing.

Roster 46 → 47.

## Verification

- `npm test` → **exit 0, zero FAILs**
- `tests/console_signer.mjs` → **15/15** on this branch, **10/15** on unfixed main


JAFairweather:
**Closing as a duplicate of #288, which is the better implementation.** Same issue (#285), theirs
created seven minutes earlier, and I only noticed the collision when listing open PRs.

Where theirs wins, and it is not close:

- **`console/control-state-freshness.mjs`** — one shared module with named constants
  (`CONTROL_STATE_MAX_AGE_SECS`, `CONTROL_STATE_MAX_FORWARD_SKEW_SECS`) and
  `controlStateFresh` / `newestFreshControlState` helpers. I inlined `+ 60` into each page, which
  duplicated the magic number four times — precisely the kind of duplication I spent today removing
  elsewhere. Theirs is right and mine was not.
- A **negative control** in the focused suite: weakening the skew bound turns it red. That is stronger
  evidence than my assertion that the clamp is textually present.
- It also fixes `config-operations.mjs` and the fourth reader, which I had left alone.

AD-11's rule applies to my own work as readily as anyone's: promote the best implementation, never
level down.

**One thing this PR had that #288 does not,** offered as a follow-up rather than a reason to keep it
open. My `tests/console_signer.mjs` walked `console/` with `readdirSync` and asserted:

> no console surface calls `nip07Signer()` directly except `index.html` — the sign-in picker, which
> *creates* the session everything else reuses

#288's equivalent names the three known files explicitly
(`!/nip07Signer/.test(followingPage) && !/nip07Signer/.test(configPage)`). That is correct for today
and cannot catch **the next page added**, which is how these two drifted apart in the first place —
`signer-session.mjs` existed and only one of three call sites had adopted it.

If that guard is wanted, it is ~10 lines on top of #288 and I am happy to open it against that branch.
Naming the `index.html` exception explicitly is the load-bearing part: an unexplained allowance is what
let the other two drift.

PR #290, opened 2026-08-05
feat: choose the capability, and say who enforces it

### LINKED ISSUE #284: console: three of the four documented capabilities cannot be issued here
Tracks JAFairweather/nave.pub#110 (Wave 1).

`docs/GETTING_STARTED.md` and `README.md` now document a three-tier authority vocabulary: `admit` lets
an outside identity use the lane; `task` / `task+act` let that identity's authenticated message
instruct a particular agent runtime; a working-channel carrier additionally needs **`task-relay`**,
which "makes it transport, never the original author."

The console does not keep up:

| cap | label in `CAP_LABEL`? | issuable from the console? |
|---|---|---|
| `admit` | yes | yes |
| `task` | yes | yes |
| `task+act` | **yes** | **no** |
| `task-relay` | **no** | **no** |
| `admit+read` | no | no (honoured by the bridge, refused by `tools/grant.mjs`) |

The cap is currently **inferred from the shape of the subject field** — an npub-looking subject yields
`task`, anything else yields `admit`. So `task+act` renders a label it can never produce, and
`task-relay` has neither a label nor a path.

## Ask

- Make the capability an explicit choice, not an inference from input shape.
- Add `task+act` and `task-relay` issue paths, and labels for `task-relay` and `admit+read`.
- Take the label map from the shared component proposed in JAFairweather/nave.pub#112 so the Ledger,
  the access list and the routing panel cannot describe one capability three ways.
- Keep the existing confirm-panel discipline: the exact event, then a plain-English `why`, then sign.
  That panel is the best commit surface in the estate; this only widens what it can express.

JAFairweather:
Closes #284. Tracks JAFairweather/nave.pub#110 (Wave 1).

⚠ **Merging this deploys to production.** Reviewer's call.

## The gap

The console **inferred** the capability from the shape of the subject field — an npub gave `task`,
anything else `admit`. Three consequences, all real:

- **`task+act`** had a human label it could never produce
- **`task-relay`** had neither a label nor any path, though `GETTING_STARTED.md` requires it and the
  agent runtime enforces it (`attention.mjs`, `invocation_policy.mjs`, `channel_task_carry.mjs` all
  gate on `cap === 'task-relay'`)
- **`admit+read`** had no label either, though this bridge honours it

So three of the estate's documented capabilities were unreachable from the one surface an operator
actually uses, and one was advertised but unbuildable.

The capability is now **chosen**, offered per subject shape, and re-validated before signing.
`tools/grant.mjs` gains `task-relay` for an agent scope so the two signing surfaces cannot disagree —
a test asserts they match, because one operator with two ways to sign the same grant should not have
to wonder which is right.

## The part I didn't expect to need: who checks it

**This bridge consumes only two capabilities.** `processGrantEvent` returns early on anything that is
not `admit`/`admit+read`, while the whole task family is enforced on the **agent** side. A surface
that let an operator sign without naming the enforcer would be **recording intent while implying
enforcement** — the same defect as a routing board rendering a lane nothing delivers to.

So each option states its enforcer, and the map is **asserted against the bridge's actual early
return** rather than trusted as prose:

| capability | enforced by |
|---|---|
| `admit`, `admit+read` | this bridge |
| `task`, `task+act`, `task-relay` | the agent's runtime |
| `mirror` | this bridge, authored by the participant |

## Two deliberate exclusions, visible rather than silent

**`admit+read` is offered DISABLED, with its reason.** Conveying the read cap means putting Concord
channel key material in a 30440, which makes this page key-touching and turns **revoke into a Concord
rotation client** rather than an event signer (`SPEC_EXTERNAL.md` §4.1.1). A greyed option with a
reason teaches; a missing one confuses.

**`mirror` is absent entirely** — it is authored by the participant about themselves, so it is not the
operator's to grant.

## The confirm sentence is now per-capability

Each says what the grant does **not** convey, which is where a reader is most likely to assume more
than was signed:

- `admit` — *"It conveys NO ability to read the channel — only to post into it."*
- `task` — *"to wake it and be answered. It does NOT authorise the agent to take action on their word."*
- `task-relay` — *"as transport, never as the author. It grants no authority to instruct: a carried
  message is only actionable if its original signer separately holds a task grant."*

## Verification

- `npm test` → **exit 0, zero FAILs**
- `tests/capabilities.mjs` → **31/31**; against unfixed `origin/main` it reports **14 failed
  assertions naming exactly what is missing** (and degrades to assertions rather than a stack trace
  when a declaration is absent)

Roster 46 → 47.


JAFairweather:
Rebased onto current `main` (`23d0422`, after #291 merged) — `e88881b` → `087fe11`.

**The conflict was purely roster bookkeeping**: #291 and this PR both incremented the suite count and
appended to the roster in `package.json`, `CLAUDE.md`, `README.md` and `docs/GETTING_STARTED.md`.
Resolved by taking main's side in every hunk and re-appending this branch's suite on top, so nothing
from #291 was dropped. 48 → 49.

**No functional change from the approved commit.** `console/index.html`, `tools/grant.mjs` and
`tests/capabilities.mjs` are byte-identical to `e88881b`.

Re-verified after the rebase: `npm test` exit 0, zero FAILs, `suite_count` green at 49, and
`tests/capabilities.mjs` 31/31 — the same result cited in the approval.

Flagging the SHA change explicitly since the approval named `e88881b`. Merging on that approval given
the diff is unchanged; say the word if you would rather re-check `087fe11` first.

PR #301, opened 2026-08-06
Move direct sealed DMs behind off-box policy

JAFairweather:
Outcome: advances #54 by moving direct NIP-59 gift wraps through the independently hosted Buzz policy writer. Security: accepts only a complete signature-verified kind 1059; requires exactly one signed recipient p-tag; resolves recipient name and Buzz inbox only from policy-host recipient routes; excludes Concord channel-plane wraps; shares one pure renderer and projection across local, shadow, and live policy paths; durably queues before SSH; retires debt only after an exact poster-signed receipt; and never falls back locally. Migration is explicitly opt-in per closed operation list, so legacy deployments are unchanged. Verification: focused sealed-direct and remote-gate regressions pass, lint passes, and the complete 53-suite gate passes. Deployment remains opt-in: install the policy-host recipient roster, run observe and enforce-shadow burn-in, then enable the sealed direct operation.


JAFairweather:
Reviewed at 3480aa6 (`Merge main into #301 — union at 74 suites, and take back #304`). Full suite green, lint clean, config.json + node_modules provisioned locally to run it.

**Nothing from #304 was lost.** Diffed every file #304 (4b1a645) touched against HEAD: 9 of 13 non-doc files are byte-identical to #304's own commit tree (`buzz_policy_artifacts.mjs`, `buzz_policy_projection.mjs`, `buzz_policy_service.mjs`, `buzz_policy_shadow_client.mjs`, `policy_journal.mjs`, `tests/policy_remote_gate.mjs`, `tests/policy_withdraw_repost.mjs`, `deploy/POLICY_HOST.md`). The rest (`bridge.mjs`, docs, `package.json`) differ only because later merges added unrelated content on top — every symbol #304 introduced (`processRemotePolicyRequest`, `recordWithdrawn`, `finishRemoteDeletionIfSettled`, `withdrawalEntry`) is present and referenced at HEAD. `tests/policy_withdraw_repost.mjs` — #304's own suite — passes clean, 12/12. Full 74-suite run is green.

**Off-box policy gates sealed DMs exactly the way the box currently expects: it doesn't gate them at all yet, and that's correct.** Traced the gate: `bridge.mjs:1109` only routes a direct sealed wrap off-box when `PUB.policyWriter.mode === 'remote-only' && PUB.policyWriter.operations.includes('sealed_direct_envelope')`; absent config (`cfg.public?.policy_writer || {}`) resolves to `mode: 'off', operations: []` (bridge.mjs:329-331), which falls straight through every gate to the pre-existing `deliverDirectLocally(ev, hits)` (bridge.mjs:1120). I checked the box directly — `/opt/waggle-read/config.json` (live, 2954 bytes) has **no `policy_writer` or `policy_shadow` key at all** — so this migration is a complete no-op on the deployed config today, matching the PR's stated opt-in rollout (roster install → observe → enforce-shadow → enable). `tests/policy_sealed_direct.mjs`'s first assertion is literally "legacy bridge configs cannot silently move a newly shipped operation remote-only" — passes. 12/12 on that suite.

**One thing worth flagging that isn't in the PR body: this branch is presently CONFLICTING against current `main` again** (`gh pr view 301` → `mergeStateStatus: DIRTY`). Main picked up 4 more commits since this branch's last merge-in, including #346/#376 (NIP-13 PoW cap) and three doc/registry PRs (#392/#395/#400). Ran `git merge-tree --write-tree HEAD origin/main` to check what actually collides — it's confined to `CLAUDE.md`, `README.md`, `docs/GETTING_STARTED.md`, `package.json`: suite-count prose and doc-rule churn, not `src/*.mjs`. No logic conflict, but it'll need another doc-only reconciliation pass before it merges cleanly — flagging now so it doesn't surprise whoever hits merge.

— Neil

Issue #263, opened 2026-08-05
Tripwire alarms are log-only: configure and prove independent sealed delivery


JAFairweather:
## Live finding

On the Waggle box, `waggle-tripwire.timer` runs and `tools/tripwire.mjs` correctly identifies unaccounted poster events, but the service reports:

```
no alarm delivery path configured (ALARM_NSEC/ALARM_TO unset)
ALARM NOT DELIVERED
```

The current control-state reports are a known false-positive class being corrected in PR #261, but the missing delivery path is independent: a genuine future signing event would also remain visible only in the service journal and `data/tripwire-alarms.log`.

## Close condition

- configure a dedicated least-authority alarm sender and an operator-owned recipient without exposing either secret in argv, repository files, or logs
- verify credential file ownership/mode and service isolation
- run the safe tripwire alarm drill and prove the sealed alert reaches the intended recipient
- prove an ordinary clean timer run remains quiet
- document rotation/revocation and the recovery path when delivery relays are unavailable

Do not use the bridge poster key as the alarm sender: the detector must remain independent of the identity it watches.

JAFairweather:
2026-08-05 live audit: the tripwire timer is enabled and active, and recent clean ticks correctly reconcile the poster journal. The production unit still has no `/etc/waggle-tripwire` credential set or alarm credential drop-in, and every tick explicitly reports `no alarm delivery path configured (Bunker or ALARM_NSEC plus ALARM_TO)`.

The merged Bunker-backed signer and positive-delivery drill are present in the deployed release, but repository delivery is not operational completion. This issue remains open for a separate zero-authority Bunker alarm identity/pairing, root-owned 0600 credential files, installation of `tripwire-alarm-bunker.conf` without replacing the existing two-lane detector command, a cold-read sealed-alert drill, and a subsequent quiet clean tick.

JAFairweather:
Repository-side unit-name correction merged via #296 as `b90a9903359a568f46d168ffe9641884b9c2ef72` after exact-head private Buzz review and green CI. Both production lanes now match that merge byte-for-byte (read 104/104; sealed 104/104).

[...]

PR #302, opened 2026-08-06
Operationalize #263 tripwire alarm enrollment and live drill

JAFairweather:
## Outcome

Turns #263 from a manual runbook into a resumable, shipped owner workflow while preserving an independently custodied alarm identity.

- documents local mint → Bunker import; Bunker is custody/signing, not identity creation
- accepts systemd’s protected root-owned `0440` credential projection only as a direct child of `/run/credentials`, while ordinary group-readable files remain refused
- ships the credential drop-in, isolated drill unit, and setup wrapper into `/opt/waggle-read/deploy`
- adds one stdin-only enrollment command that stages credentials, writes public drill inputs, installs units, reloads systemd, and refuses replacement of a working pairing
- adds `check` and isolated `drill` resume operations; the drill cannot stop or modify production detection
- supports GNU/Linux and BSD/macOS mode inspection correctly

## Owner actions left deliberately human

1. Import the newly minted dedicated alarm identity into Bunker.
2. Approve its NIP-46 pairing.
3. Confirm the labelled sealed drill DM arrived.

## Verification

- CI green at exact head
- `tripwire_setup` fresh-install/resume/permissions/isolation checks
- `tripwire_detection` credential and live synthetic relay drill checks
- deploy runner and deployed-tree drift checks
- lint and suite-count gate

## Live close condition

After merge/deploy, run `setup-tripwire-alarm.sh drill`, confirm recipient delivery, and confirm the ordinary timer remains active and quiet.


JAFairweather:
Review from claude-jaf. **One must-fix in `src/credential_file.mjs`**, found while confirming this PR is the right fix for a live failure.

## Context: this PR fixes something broken in production right now

`waggle-tripwire-drill.service` on the bridge host is in **failed** state and has been since 6 August, exiting on exactly the message this PR is about:

```
tripwire: ALARM_TO_FILE must not be group/world accessible
```

The detector and its timer are fine; it is the drill that has never once succeeded. So the alarm has never been observed firing — the "detector that has never fired is an assumption with a timer" case, live. That makes this PR the highest-value thing open, and also means its permission relaxation gets read hard, because a gate loosened to make a failure go away is the exact shape of the mistake worth catching.

## The gate holds on five of six properties

I probed the exported predicate directly rather than reading it:

```
ok   genuine systemd projection accepted     (root:root 0440, direct child of the mount)
ok   ordinary group-readable file elsewhere refused
ok   owner-only anywhere accepted
ok   world-readable 0444 inside the mount refused
ok   non-root owner inside the mount refused
```

That is a well-built check, and the four-way binding in the comment is the right instinct.

## MUST-FIX — the prefix is tested on the raw string, the equality on the resolved one

```js
if (!directory.startsWith('/run/credentials/')) return false      // raw
const exactDirectory = resolve(directory), exactPath = resolve(String(path || ''))
return dirname(exactPath) === exactDirectory && ...               // resolved
```

`resolve()` collapses `..`, so those two lines can disagree about which directory is being authorised:

```
HOLE — CREDENTIALS_DIRECTORY = /run/credentials/../../tmp/evil
       (resolves to /tmp/evil)
       credentialModeIsPrivate('/tmp/evil/alarm', {mode:0o100440, uid:0, gid:0})  ->  true
```

A root-owned `0440` file **anywhere on the filesystem** is accepted as private, as long as `CREDENTIALS_DIRECTORY` is spelled with a traversal. The stated property — *0440 only as a direct child of systemd's private mount* — does not hold.

Two-line fix: resolve first, then check the prefix.

```js
const exactDirectory = resolve(String(credentialsDirectory || ''))
if (!exactDirectory.startsWith('/run/credentials/')) return false
```

Please pair the test with a legitimate value that still passes, per house rule — a check that only proves the traversal is refused cannot distinguish "refuses traversal" from "refuses everything".

## Second, lower severity — `resolve()` is not `realpath()`

`resolve()` does not follow symlinks. A symlink inside the credentials mount pointing at a root-owned `0440` file elsewhere satisfies `dirname(path) === directory` while the `stat` describes the target. Writing into `/run/credentials/<unit>` needs root and it is a private mount, so this is defence-in-depth rather than a live hole — but since the whole point of this function is that the file is *the one systemd projected*, `realpathSync` on the path is closer to the property than `resolve`. Your call; flagging it, not gating on it.

## Otherwise

Structure is right: judgement in a pure exported predicate that can be probed without a host is what let me find the above in a couple of minutes. Verification section and the three deliberately-human owner actions read well.

Not approving yet — the traversal is a real weakening of a credential gate. Fix that and I will re-probe and merge, and then we can finally run `setup-tripwire-alarm.sh drill` and watch the alarm fire for the first time.

PR #306, opened 2026-08-06
feat: make owner setup resumable and visible

### LINKED ISSUE #305: Owner install should be one resumable wizard, not a multi-page runbook
## Problem

A fresh Waggle owner currently crosses local CLI setup, host provisioning, systemd, firewall, Bunker, Nvoy, Buzz admission, deployment, and live proofs through many copy/paste steps. The pieces are individually documented, but the operator experience is not a product: it is easy to lose position, repeat a destructive step, paste on the wrong host, or mistake configuration for proof.

#263 exposed the shape clearly. The correct fix is not more prose; it is one resumable state machine with explicit human approval checkpoints.

## Required owner experience

From the marketing/setup surface, an owner starts one installation and receives:

1. **Prepare locally** — validate dependencies; create public config; mint/import dedicated identities without printing secrets; generate a non-secret install manifest and stable installation id.
2. **Bootstrap host once** — one authenticated command installs users, immutable code, systemd units/timers, firewall/NTP rules, data directories, and the pull-based deployment runner idempotently.
3. **Pair custody deliberately** — pause only where a person must import an identity or approve a Bunker NIP-46 connection. Secrets enter via stdin or credential files, never argv, environment, repository, or logs.
4. **Connect Buzz/Nvoy** — create or select the hive/channel, request/approve the exact scoped grants in Console, and bind each agent identity to one isolated runtime/MCP.
5. **Prove, don’t declare** — run clock/firewall, deployed-SHA, relay reachability, profile/relay-list, admission, sealed wake/reply, tripwire drill, ordinary quiet tick, and restart/dedup checks. Render each as pending/pass/fail with evidence and a safe resume action.
6. **Finish with a receipt** — export a secret-free installation report containing identities/npubs, scopes, channels, runtime/container names, deployed SHAs, proof timestamps, rotation/revocation commands, and remaining optional enhancements.

## Interface and automation requirements

- `waggle-init` becomes the local state-machine entrypoint rather than a TODO printer.
- A shipped host-side installer consumes the generated manifest and is idempotent.
- Console Setup reads the same state schema and presents the steps visually; CLI and browser cannot disagree about completion.
- Every step supports `check` without mutation and refuses to overwrite an existing credential or identity.
- The workflow resumes after reboot, SSH disconnect, failed relay, or deferred Bunker approval.
- Existing installations can be imported/audited without reinstalling.
- Multiple hives under one owner and multiple isolated Claude/Codex identities are first-class.
- The console and runtime host remain separable; the wizard explains and verifies that deployment boundary.

## Negative controls

- wrong host / wrong installation id
- stale or red build
- unsynchronized clock or blocked NTP
- malformed or credential-bearing relay URL
- secret passed on argv/environment
- group/world-readable or symlinked credential
- identity mismatch between Bunker, manifest, kind:0, and grant
- grant for wrong hive/channel/capability
- duplicate identity/runtime/MCP binding
- configured-but-unreachable relay or MCP
- relay acceptance without recipient cold-read
- restart that repeats a completed action

## Close condition

A fresh owner can start from the public Waggle site and finish a verified installation with only the unavoidable identity-import/pairing/grant approvals performed manually. No undocumented file copy, hand-written JSON, direct systemd edit, or server-side secret paste is required. The complete path is exercised by an automated disposable-host test plus one documented live proof.

JAFairweather:
## Summary

Closes #305.

- adds a secret-free, mode-0600 resumable `setup-state` state machine with stable installation id
- wires normal `waggle-init` runs into the six owner-facing steps while keeping `--check` read-only
- adds negative coverage for credential-like evidence, unsafe state files, resume identity, and check-only non-mutation
- makes the browser Setup page read the same six-step receipt when the owner publishes `console/setup-state.json`
- documents deferred owner actions, resume behavior, the public receipt contract, and the 53-suite gate

## Safety boundary

The state record and browser surface contain only statuses and public proof metadata. The state writer rejects credential-like material, symlinks, malformed state, and incorrect permissions. No nsec, Bunker URI, provider key, or private credential is accepted, printed, or committed.

## Verification

- `npm test` — 53 suites pass with the non-secret example config and loopback integration-test access
- `npm run lint -- --quiet` — pass
- `git diff --check` — pass
- `node tests/setup_state.mjs` — pass
- `node tests/suite_count.mjs` — pass


JAFairweather:
Consolidated into #307 so owner setup has one canonical state contract. #307 retains the public Console receipt/renderer while strengthening the model with a closed eight-step catalogue, evidence-bound transitions, explicit owner/hive/topology and agent-family facts, recursive credential rejection, and one state path per hive. Keeping both would create incompatible resumability formats.

Issue #309, opened 2026-08-08
Console should own the whole remote-agent lifecycle, not just install


JAFairweather:
## Problem

#305 installs **an owner, once**. Nothing owns **an agent, repeatedly** — created, granted, tuned, audited, retired, and managed for the rest of its life.

Today that path crosses key minting, Bunker pairing, grant requests, relay configuration and runtime binding as separate manual acts with no single surface, no resumable position, and no live proof. It is the same failure mode #305 identified, applied to a different subject: easy to lose position, easy to mistake configuration for proof, easy to believe a saved `passed` that no longer matches the world (#308).

This also under-sells a real differentiator. Via an MCP server any existing model can become a Buzz participant — that capability is worth a product surface, not a runbook.

## Position

**No new plane is required.** Both halves already exist:

- **Projection (read)** — #307: private mode-0600 state holding no credentials → validated secret-free public receipt → browser renders the receipt and nothing else.
- **Actuator (write)** — `src/buzz_policy_*`: closed frozen catalogue, evidence-not-prose, forced-command runner, `O_EXCL` journal idempotency, hold-never-fall-back.

Agent lifecycle is **one more closed catalogue on the actuator, plus per-agent rows in the projection**. #307's constraint — *no second setup vocabulary is introduced* — is binding here.

So for the lifecycle **operations**, there is no CLI role. The CLI survives at exactly two forced boundaries: a secret generated locally (a key minted in a browser tab sits in a JS heap the console cannot prove it discarded, in an origin that already needed a `Host` defence against DNS rebinding, #145), and anything touching host filesystem/systemd.

## Three doors, one gate

Mint (into `bunker.nave.pub`, never onto the box — waggle holds exactly one private key, its own), BYOK, and make-your-own via guided CLI prompts. All three converge on a single acceptance test: **challenge-sign a fresh nonce**. Possession of a Bunker URI is not control; a pasted npub is not control.

## Design

`docs/DESIGN_AGENT_LIFECYCLE_PLANE.md` — includes the closed operation catalogue, the revoke-vs-destroy split (opposite reversibility, so they must not be adjacent buttons), the resumable-state vs re-derived-truth boundary, and the traps that have each produced a wrong green here: possessing ≠ control, spent pairing secret presents as `Unknown client`, empty ≠ absent, reachable ≠ attached, admitted ≠ wakeable, write ≠ read.

## Build order

[...]

Issue #308, opened 2026-08-07
Claude first-class participant release: state ledger and remaining gates (claude-jaf)


JAFairweather:
The Claude Code first-class participant release has no tracking issue in this repo, and the
Nvoy-side docs cite **#113** as its tracker. That is the wrong issue: waggle #113 is closed and
is the `#connector` return-lane scan. This issue is the correct home, and records the state
ledger so the next session does not re-derive it.

Companion: JAFairweather/nvoy#129 carries the doc corrections.

## Why a ledger and not a checkbox

`docs/AGENT_PARTICIPANT_ARCHITECTURE.md` is explicit that "implemented", "deployed", "attached"
and "live-proven" are not synonyms. The participant path has eight distinct states and they have
been collapsed into "pending" more than once. Each is reported separately below.

## State ledger — Claude participant `claude-jaf`

Identity `npub145zmqrhyjgqdt0f83ra6fqrzrwnqpy3y7q0y3vlfhnsszqzzr4wqtqq92y` (`ad05b00e…`),
the current release assignment per Nvoy `docs/AGENT_RUNTIME_HANDOFF.md` §3.

| # | State | Verdict | Basis |
|---|---|---|---|
| 1 | identity assigned | **proven** | Assignment recorded in the Nvoy handoff §3; the instance manifest pins the same pubkey, grantor and carrier |
| 2 | Bunker verified | **not proven** | No pubkey-comparison evidence record exists. §2 of the owner runbook is explicit that possessing a NIP-46 URI is not proof it controls the identity |
| 3 | credentials installed | **incomplete** | The instance's dedicated NIP-46 client credential has not been created yet, though the manifest already references it. The runtime cannot start as configured |
| 4 | broker attached | **not proven** | The instance's state, runtime and spool roots do not exist. The broker has never run |
| 5 | Claude Channel attached | **not proven** | No identity-scoped Channel MCP server is registered. `nvoy_channel_read` / `nvoy_channel_reply` are not reachable from any session |
| 6 | admission granted | **partial** | The channel `admit` grant is **ACTIVE** and cold-read from relays (2026-08-06T16:31:30Z). But **no grant is scoped to this identity as agent subject** — neither the author `task` grant nor the carrier `task-relay` grant exists |
| 7 | live wake proven | **no** | blocked on 2–6 |
| 8 | live reply proven | **no** | blocked on 2–6 |

### On state 6 — this is the fail-closed path working

[...]

Issue #145, opened 2026-08-01
Console server has no Host check, so DNS rebinding still reaches it (residual of the doc-root fix)


JAFairweather:
Follow-up to the console document-root fix in #142. That change removed the **secrets** from the loopback server's reach; it did not remove the **reach**.

## What is left

`tools/serve-console.mjs` binds `127.0.0.1` and validates nothing about the `Host` header. A hostile page in a browser on the operator's machine can re-resolve its own domain to `127.0.0.1` (**DNS rebinding**) and then issue same-origin requests to the console server. Loopback binding does not stop this — it is exactly the class of attack loopback binding fails to stop.

## Why it was left open rather than fixed in #142

After the doc-root change the server hands out only `console/`: the page and its vendored modules. There are no secrets behind it any more, so this stopped being urgent and started being a hardening item — and bundling a behavioural change into a security fix makes both harder to review.

## Why it is still worth closing

The console is the **signing surface**. Its promise is *"here is exactly what you are about to sign"*, and that promise rests on the operator controlling the page. Reachability from a hostile origin is worth denying on that page specifically, even with no secrets served.

## Candidate fix

Reject any request whose `Host` is not `127.0.0.1:<port>` / `localhost:<port>` — a few lines, no change to how the console is used. Worth confirming it does not break the documented `http://127.0.0.1:8080/console/` entry point or any port override via `PORT`.

Not urgent, and deliberately not bundled into #142.

PR #310, opened 2026-08-08
Design the remote-agent lifecycle plane

JAFairweather:
## Outcome

Design only — no code. Establishes waggle/console as the surface that owns a remote agent's whole life: create, grant, bind, tune, audit, pause, revoke, destroy.

## The load-bearing claim

**No new plane is needed.** Both halves are already built and already carry the properties lifecycle requires:

- **Projection (read)** — #307: private mode-0600 state with no credentials → validated secret-free public receipt → the browser renders that and nothing else.
- **Actuator (write)** — `src/buzz_policy_*`: frozen closed catalogue, evidence-not-prose, credential-free decision core, forced-command runner, `O_EXCL` journal idempotency, and hold-never-fall-back on an unavailable service.

Lifecycle becomes **one more closed catalogue on the actuator, plus per-agent rows in the projection**. #307's *"no second setup vocabulary is introduced"* is treated as binding.

Consequence: for the lifecycle **operations** there is no CLI role. The CLI is retained at exactly two forced boundaries — a locally generated secret (a key minted in a browser tab sits in a JS heap the console cannot prove it discarded, in an origin that already needed a `Host` defence against DNS rebinding, #145), and host filesystem/systemd work.

## Three doors, one gate

Mint / BYOK / make-your-own differ only in where the nsec is born, and converge on one acceptance test: **challenge-sign a fresh nonce**. Mint goes **into `bunker.nave.pub`, never onto the box** — waggle holds exactly one private key, its own, and minting onto the host would give it a second. Keeping mint in custody is what lets "mint for simplicity" stay on the menu instead of being traded away.

## Worth a reviewer's attention

- **Revoke and destroy are split** because their reversibility is opposite. Revoke is instant and safe; destroy forces a Bunker re-pair (`broker_credentials` is `reRenderable: false`) and carries a confirmation token. They must not be adjacent buttons.
- **Resumable state vs re-derived truth** is stated as an explicit boundary rather than left implicit: saved state may record *that* a proof passed and *what evidence* proved it, never that a live property *is currently true*. #308 was that failure.
- **Traps as first-class**, each of which has produced a wrong green here: possessing a Bunker URI ≠ control; a spent NIP-46 pairing secret presents as `Unknown client`; empty ≠ absent (negative control runs *inside* the wizard); reachable ≠ attached; admitted ≠ wakeable (`admit`/`task`/`task-relay` are separate on purpose); write ≠ read (the return lane carries **mentions only**).
- **Build order stacks on #307 and must not start before it lands** — a stacked branch whose base is squash-merged gets orphaned, and one was auto-closed that way today.

## Verification

Docs-only; no source or test change. Note #162: even a docs-only merge triggers a deploy tick.

## Review

Worth a read from **kerouac** (spec/prose precision, and whether the claim vocabulary drifts anywhere) and **mydude** (whether the actuator really can carry this catalogue without widening what the console is trusted with). I wrote it, so I should not be the one certifying it.

Refs #309, #305, #307.

### DIFF
--- a/docs/DESIGN_AGENT_LIFECYCLE_PLANE.md
+++ b/docs/DESIGN_AGENT_LIFECYCLE_PLANE.md
@@ -1,0 +1,187 @@
+# Design — the remote-agent lifecycle plane
+
+waggle/console is where a remote agent is created, granted, tuned, audited and retired. This
+document says how that is possible without inventing a transport, a vocabulary or a CLI role.
+
+Scope note: #305 installs **an owner, once**. This installs **an agent, repeatedly**, and keeps
+managing it for the rest of its life. They share machinery; they are not the same journey.
+
+---
+
+## The outcome
+
+An owner opens the console, adds a remote agent, and — without pasting JSON, editing a unit file,
+or being told to go read a runbook — ends with an agent that is admitted, wakeable, scoped, and
+visibly proven so. Later, the same surface is where they tune what that agent is allowed to see,
+watch what it actually did, and take its authority away.
+
+The agent's private key is never in the browser, never on the waggle host, and never in this
+repository.
+
+---
+
+## Two planes already exist. Lifecycle is a catalogue on them, not a third plane.
+
+**The projection plane (read).** #307 establishes it: mode-0600 private state holding no
+credentials → a validated, secret-free public receipt → the browser renders the receipt and
+nothing else. The browser has no access to private installer state and no write access to
+`config.json`.
+
+**The actuator plane (write).** `src/buzz_policy_*` is already a signed-command executor with the
+exact properties lifecycle needs, and they were hard-won rather than incidental:
+
+- a **closed catalogue** — `BUZZ_POLICY_OPERATIONS` is frozen; a caller *names* an operation, and
+  cannot supply rendered prose or choose a destination;
+- **evidence, not instructions** — the untrusted side contributes canonical signed evidence, and a
+  credential-free core decides independently;
+- a **forced-command runner** — deployment fixes the executable and config path; stdin is the only
+  caller-controlled input;
+- **journal idempotency** — `O_EXCL` claim, atomic rename completion, terminal across processes;
+- **hold, never fall back** — an unavailable policy service is a queue that waits. There is
+  deliberately no dead-letter limit, because "the service was down" must never become permission to
+  sign with a lesser key.
+
+Agent lifecycle is therefore **one more closed catalogue on the actuator plane, plus per-agent rows
+in the projection**. PR #307 states the governing constraint plainly — *no second setup vocabulary
+is introduced* — and this design is bound by it.
+
+That is the whole answer to "do we need a CLI role": for the **operations**, no. The console signs;
+the bridge verifies against `public.approvers`, dedups by id, applies one allowed change, and
+acknowledges in signed state. That is the mechanism that already moves a watchlist entry.
+
+---
+
+## Where the CLI genuinely survives
+
+The CLI is not retained for symmetry or preference. It survives at exactly two boundaries, and
+both are forced:
+
+1. **A secret generated locally.** A key minted in a browser tab exists in a JS heap the console
+   cannot prove it discarded, in a page whose origin already needed a `Host`-header defence against
+   DNS rebinding (#145). Generation happens in the CLI; only the npub and a Bunker URI cross over.
+2. **Anything that must touch the host** — filesystem, systemd, firewall. Already #305's territory.
+
+Everything else — grant, bind, tune, pause, audit, revoke — is console-signed.
+
+---
+
+## Three doors, one gate
+
+The three key-provenance flows differ *only* in where the secret is born. They converge on a single
+acceptance test.
+
+| Flow | Where the nsec is born | What waggle receives |
+|---|---|---|
+| **Mint** (the simple option) | in `bunker.nave.pub`, in custody | a Bunker URI + revocable NIP-46 client key |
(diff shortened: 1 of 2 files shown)


JAFairweather:
## Independent review — corrections applied in `e2937f5`

Reviewed against CLAUDE.md's "Claims that must never drift" by a reader who did not write it. Five defects, two of them on named invariants. All fixed.

**The actuator was the wrong mechanism, and it changes the answer.** The draft said `src/buzz_policy_*` is "already a signed-command executor with the exact properties lifecycle needs" and then cited *"that is the mechanism that already moves a watchlist entry."* Those are two different lanes. Verified in the tree: the watchlist moves on an approver-signed on-relay control command (`handleWatchlistControlCommand`, `src/bridge.mjs:2148`, filtered on `authors: PUB.approvers`); `buzz_policy_*` is an off-box forced-command runner over stdin. I merged them and then cited one as proof the other would carry lifecycle.

Worse, the property I cited as the reason it fits is the one that does **not** transfer. "Evidence, not instructions" is exactly why the policy core cannot host `agent_pause` — there is no third-party signed evidence behind it, the owner's *intent* is the input, and refusing instructions is what that core is for. Its catalogue is frozen at `quarantine_header` and `standing_trusted_reply`, both Buzz *rendering* rather than configuration.

Now states both lanes, names the **control-command lane as the likelier home**, puts it in the conditional, and leaves it as the open question the wireframes already flagged. The design asserted as settled what its own open question listed as unresolved.

**`waggle` rendered as `WAGGLE`.** The eyebrow's `text-transform:uppercase` uppercased the wordmark. Source was lowercase, output was not — which is precisely why the rule says check rendered output, not the source you edited. Fixed with an opt-out class; I then checked every other uppercase-transformed element for the wordmark and found none.

**Screen 04 refuted its own lane label.** It said "receives mentions only", correctly, and then listed three carry classes that are not mentions — replies, relevance calls, approved public posts. Widening the return lane is a change to a protected claim, not a setting. Now flagged inline where the toggles are, and raised as its own open question.

**Shipped vs designed.** "Two planes already exist" — #307 is open and unmerged, which the same document conceded 137 lines later. Now "already designed", with #307 described as a proposal. The wireframes' only disclaimer sat at line 748 of 754, *after* all five screens, while the mockups were written in live-telemetry voice ("checked just now", "240 ms", "3 sources approved"). There is now a banner before screen 01 and an `illustrative` marker in every frame bar.

**Also:** #308 corrected — it is open, and its lesson is collapsed states ("implemented"/"deployed"/"attached"/"live-proven" as one "pending"), not stale saved progress. `agent_destroy_unit` marked host-touching, since destroying a runtime contradicts the CLI boundary drawn four lines earlier. Step 3 noted as in flight in PR #311. The operations the wireframes deliberately do not depict are now named, including the missing third step of a flow labelled "of 3".

Held up under review: the one-key claim, consistently, including inside mockup copy; mint-into-custody; no secrets or host addresses; and the build order matching #309 exactly.

Docs-only. Still wants a crew read — the review above was not from the crew.

JAFairweather:
## Crew review complete — all five items closed at `9814afb`

Verdicts were posted in-channel; my earlier "no replies" report was a broken return-leg read on my side, not a missing review. Dennis caught it, with event ids and 2–6 minute reply times.

**Kerouac** — design-doc verdict `6fa2c335`, wireframes verdict `dbf1f4b1`, re-read at `e2937f5` in `00a0dcb2`: approve pending three items.
**My Dude** — architecture verdict `993dd439`: the no-new-plane thesis **holds**, verified against real code (`bridge.mjs:2098–2230` — approver roster, signature verification, exact two-tag addressing, freshness bounds, monotonic superseded-command watermark, closed action enums, persistence via the bridge's own `mutateConfig`). The browser has no `config.json` write today and the lifecycle catalogue as drawn adds none; the catalogue also fails closed on extension via `catalogue_version` pinning. Four gaps, two still open after the last push.

### Closed

**Custody — My Dude's gap 2, and the sharpest.** A NIP-46 client key *is* a private key and a Bunker URI embeds a pairing secret, so "waggle receives a Bunker URI + revocable client key" had to say **which waggle**. It is neither the bridge nor the console: those credentials seat into the agent's own runtime unit across the CLI/#305 boundary and never transit console or bridge state. An agent row holds a **reference**, never the credential. The one-key claim is now restated around that rather than left in tension with it — which also closes Kerouac's one-key restatement item.

**`agent_revoke` — gap 4.** Now names the authority it withdraws (bridge-side admission, which the bridge issued and can revoke) and states what it must **not** touch: a grantor-signed data grant is only its issuer's to relinquish. The console must render those as separate rows with separate remedies, or an owner presses revoke and reasonably concludes an access is gone that the agent still holds.

**`reRenderable: false`** cited to nvoy #160, including that the re-pair cost rests on the spent pairing secret. **The make-your-own door** now says whose machine the key is born on.

### The render proof, and what it caught

Kerouac's caveat was that the wordmark fix couldn't be proven from a seat with no browser — *"the proof remains one real browser open."* Done: it renders **lowercase** against its uppercase siblings.

Opening it also caught a defect no source review could reach. With no charset declared, the file decodes as **Latin-1** when served, so every em-dash and middot rendered as mojibake (`â€"`, `Â·`) for anyone following the "open it in a browser" instruction. Now declares utf-8. The insistence on rendering was right twice over.

### Still open, deliberately

My Dude's proposed resolution for `agent_destroy_unit` — a pre-deployed fixed-verb forced-command channel carrying the #160 token rule — is recorded but **not** adopted here; it is step-2 design, not a doc fix. Same for connection-binding on the challenge gate, which he offered to co-spec.

Nothing merged.

PR #311, opened 2026-08-08
Agent proof-of-control challenge gate

JAFairweather:
## Outcome

The first implementation slice of #309: the gate all three key-provenance doors converge on. Mint, bring-your-own and make-your-own differ in **where the key is born** and in nothing about what counts as proof — so there is one verifier, not three.

`createAgentChallenge` issues a nonce bound to the claimed identity; `verifyAgentChallenge` requires a signature over that exact nonce from that exact key. Pure and credential-free — it never signs, never opens a connection, and never decides what the proven identity may then do.

Chosen as the first slice because it is invariant across all six open questions in #309: none of them touch it, so it can be built while the design is still under review, and it does not depend on #307.

## The defect the negative controls caught

Worth reading even if you skim the rest. The first run failed exactly two assertions — tampered content, and a forged signature — both of which *passed* when they should have been refused.

`finalizeEvent` stamps nostr-tools' `verifiedSymbol` on events it has verified, `verifyEvent` short-circuits on that marker, and **object spread copies enumerable own symbol properties**. So `{...signed, content: 'edited'}` arrives carrying a stale "already verified" marker and passes `verifyEvent` without the signature ever being re-checked against the altered content.

The fix rebuilds the wire event from its seven fields before verifying, which drops the marker and rejects extra keys in the same step. A one-sided test — only asserting that the guard refuses things — would have shipped this green.

## Design decisions worth attacking

- **A distinct kind (27492), not NIP-98's 27235.** Sharing a kind with another authorization protocol would let an event signed for that protocol be replayed here as a control proof. If there is a house convention for picking kind numbers I have missed, say so — I chose this one.
- **A required ceremony tag on top of the distinct kind**, so a signature obtained on any other waggle surface cannot be presented here. Exactly one tag is permitted: two would leave "which nonce is authoritative" up to read order.
- **Expiry judged at the verification boundary**, not before the signer opens. Approving a signing request is an unbounded human wait — the same reasoning as `console/confirmed-fresh-signer.mjs`.
- **Identity checked before signature**, so the error distinguishes "not the key this connection claims" from "bad signature".
- Nonce shape reuses the existing NIP-98 convention (`randomBytes(24).toString('base64url')`, `^[A-Za-z0-9_-]{16,128}$`) rather than inventing a second one.

## Verification

- `node tests/agent_challenge.mjs` — 33 assertions, pass.
- Full `npm test` — **53 suites, exit 0, ALL PASS** (923 ok lines), after `cp config.example.json config.json`.
- `npm run lint -- --quiet` — pass. `git diff --check` — pass.

Every refusal assertion is paired with a legitimate value that must still get through — including that a *different* agent proves control of *its own* challenge, so the check is "matches the claimed key" rather than "only ever accepts this one agent". The expiry boundary is asserted on both sides of the second.

## Merge note

This bumps the suite count 52 → 53 in `package.json`, `CLAUDE.md`, `README.md` and `docs/GETTING_STARTED.md`. **PR #307 bumps the same four to 53.** Whichever lands second needs a rebase, and the count becomes 54 — please do not resolve that conflict by taking one side wholesale.

## Review

This touches a gate, so it needs a read from someone who did not write it — **mydude**. The two things I would most like challenged: the kind number, and whether checking identity before signature leaks anything useful to an attacker probing which identity a connection claims.

Refs #309.


JAFairweather:
## Independent review — two bypasses found and closed

The gate was reviewed adversarially by a reader who did not write it, plus a repo-wide sweep for the `verifiedSymbol` class of bug. Both broke it. Both are fixed in `04861a5`, each with a regression test.

**1. The value checked was not the value verified.** `tags` remained a reference to the caller's object, so a response exposing `tags` as a getter that answered differently on each read was **accepted as proof against a challenge it was never signed over**. The module's own stated principle — reconstruct, never trust the caller's object — had been applied to the six scalar fields and not to the one field the ceremony turns on. `response_id` had the same shape, recording an id that was never verified.

Fixed by snapshotting the wire event *first*, into plain arrays, and reading the nonce, the id and the signature all from that one snapshot. The regression test is refused by the **signature**, not the nonce comparison — which is the proof the fix works, since the attacker's first-read value is now also what gets verified.

**2. Prototype pollution reintroduced the short-circuit.** `verifyEvent` tests `typeof event[verifiedSymbol] === 'boolean'` and that lookup resolves **through the prototype chain**, so with `Object.prototype[verifiedSymbol] = true` a wholly forged event verified. Fixed by shadowing the symbol as an own `undefined`.

Worth recording: my first attempt used `__proto__: null`, which **broke every verification** — nostr-tools' `validateEvent` does an `instanceof Object` check that a null-prototype object fails. The suite caught it immediately. Both behaviours were confirmed empirically against the repo's own nostr-tools (2.24.1) rather than assumed.

## Test quality — the finding that mattered most

Mutation testing showed **11 of 22 guards could be deleted with the suite still ALL PASS**, and three refusal assertions were passing on an incidental `TypeError` rather than the guard they named, because `refuses()` caught every exception.

`refuses()` now requires the expected message. After that and the added cases:

```
guards mutated: 24   survived (vacuous): 0
```

60 assertions, up from 33. Every guard in the file is now killed by the suite.

Also fixed: a missing `expires_at` was accepted **a decade past issue** (`checkedAt > undefined` is `false`), and a `NaN` clock failed open the same way. Both now guarded and tested. `created_at` is bounded to the challenge window. The single-use caller obligation is documented — the kind sits in NIP-01's ephemeral range, which relays broadcast, so a caller that does not consume a challenge on first success leaves a replay window this stateless module cannot close.

## Answered

- **Kind 27492** — clean. No in-repo collision, nothing registered in 27000–28000 in the NIPs README. Notably this repo *already* signs NIP-98 27235 with waggle's own key in four places, so co-locating a control proof there would have put it in the same kind as a token waggle mints itself. The domain tag remains the load-bearing separator.
- **Identity before signature** — not a leak. The remote party supplied the identity in the challenge it already holds, so distinguishing "wrong key" from "bad signature" tells it nothing it did not provide. Kept, and the reasoning is now in the comment.

## Unrelated defect fixed while here

`tests/nostr_signer.mjs:70` asserted `verifyEvent(wrap)` on a `finalizeEvent` result — the marker makes that **unconditionally true**, so the signature half of the assertion was dead. Confirmed: `{...wrap, sig: '0'.repeat(128)}` also passed. Now verifies the wire form, matching line 35 of the same file.

Full suite: **53 suites, exit 0, ALL PASS** (950 ok lines). Lint clean.

This still wants a crew read before merge — it is a gate, and the reviews above were not from the crew.

PR #314, opened 2026-08-09
Add the idempotent host bootstrap (#305)

JAFairweather:
Stacked on #307 (`feat/owner-install-state`) — review that first; this branch only adds the consumer of its `host_bootstrap` step.

## What this is

#307 defined a closed eight-step install catalogue but nothing consumed `host_bootstrap`. This adds the two halves that do: a **pure planner** (`src/host_bootstrap.mjs`) and a **thin runner** (`tools/waggle-host-bootstrap.mjs`).

A plan is a set of **desired states, not commands**. That is what makes `--check` and `--apply` the same computation, differing only in whether the runner acts — a check that computes something different from the apply is a check of nothing.

## The three properties worth reviewing

Each is asserted with its pair, per the house rule that a one-sided refusal test cannot tell "refuses the dangerous thing" from "refuses everything".

**1. Unreadable is not absent.** A failed probe that reads as a missing directory gets "fixed" by overwriting a live one. Unreadable facts are `blocked`, blocked is never actionable, and blocked outranks satisfied in the verdict — so a plan we could not fully see never exits 0. It exits **3, INCONCLUSIVE** (`tripwire.mjs`, `verify-firewall.sh`). The sharp case is asserted directly: everything satisfied *except* one unreadable fact must still be inconclusive, paired with the same host exiting 0 once that fact is readable.

**2. Ordering is safety.** Users before the directories they own — creating a directory before its owner exists makes it root-owned and quietly defeats the split-uid isolation. The clock before anything that signs, because every timestamp clamp and freshness bound is only as good as it. The deploy runner **last**: it ships whatever is on main, so it must not run before the host it deploys onto is formed.

**3. Nothing plans to write a credential.** The credential directory is created 0700 and left empty. Seating credentials is an operator act, and the refusal is visible in the plan rather than left to the runner's good manners — a runner that decides is a second policy.

Paths key on the **installation id**, never on `hive.id`: that field is validated as a 64-hex public key, and a directory name is copied into backups, logs and every error that prints a path.

## What the runner deliberately does not do

`--apply` **refuses** — over drift, over any unreadable fact, and over an action kind it does not recognise (an unknown kind is a refusal, never a no-op; silently skipping one reports a host as bootstrapped that is not). The executor is not wired in this slice: shipping a half-written thing that runs as root is worse than shipping none, so it prints the plan to run by hand.

That is the honest state of this PR. **It plans and it checks; it does not yet apply.**

## Verification

- 36 assertions in `tests/host_bootstrap.mjs`, all pass. 54 suites green (roster, README, CLAUDE.md and GETTING_STARTED counts all updated; `suite_count.mjs` enforces this and caught two stale numbers).
- Run end-to-end against a real `createInstallState` file on a mac: exits **3** and names all eight facts it could not read, rather than pretending they were absent. Verified with a size floor on the control — an earlier run of the same check passed vacuously because *neither* file existed.
- No-mutation control: state file md5 unchanged across a `--check`.
- `--apply` on that same state refuses with the blocked-facts message.

## Who should review this and why

The planner's whole value is in what it refuses, so the reading that matters is adversarial: **can you get it to plan an action on a fact it could not read?** That is the failure mode that overwrites live state.


JAFairweather:
**Review (Neil — ops steward, requested by James for Claude).** Scope: the stacked delta `feat/owner-install-state...feat/305-host-bootstrap` only. CI green at head (54 suites). The planner is genuinely pure, the plan/verdict split is clean, and the pure-planner test coverage is strong — ordering, the blocked-outranks-satisfied sharp case, and the paired refusals all check out. But the PR asks the right adversarial question, and the answer is **yes, in the runner**:

### 1. `existsSync` folds "could not read" into "absent" — the exact fold this PR exists to prevent

`tools/waggle-host-bootstrap.mjs` `probeDirectory` guards with `if (!existsSync(path)) return`, and the `code_checkout` probe with `if (!existsSync(a.want.path)) continue`. Node's `existsSync` returns `false` on **any** error, not just ENOENT. Proven on this box:

```
/tmp/probe_test/locked/live_dir   (parent chmod 000)
existsSync → false
lstatSync  → throws EACCES
```

So a live directory behind an unreadable parent probes as *missing* → planned as `missing` → **actionable**, not `blocked`. Today the unwired executor means the damage is a false " add" line in `--check` output; the moment the executor lands, this is the overwrite-something-live scenario verbatim. The `lstatSync` try/catch right below already does the right thing — the fix is to drop the `existsSync` pre-check and branch on errno: `ENOENT` → missing, anything else → `unreadable.push(id)`. Same fix in the checkout probe.

### 2. The `--apply` success path crashes on evidence shape

`transitionInstallStep(state, 'host_bootstrap', { …, evidence: [bootstrapEvidence(plan, verdict)] })` passes an **object**, but `install_state.mjs` validates every evidence item with `cleanText` (`typeof value !== 'string'` → throw). A fully-satisfied `--apply` would die with `install-state: host_bootstrap.evidence must be text` instead of "install state advanced". Unreachable today only by accident — `time_sync`/`firewall_rule`/`deploy_runner` are always pushed unreadable, so apply always refuses first — which also means this designated success path has never executed. Serialize it: `evidence: [JSON.stringify(bootstrapEvidence(plan, verdict))]` (fits the 512-char bound), and consider a runner test that drives `--apply` to the advance line with all facts injected.

### 3. A pinned ref can never be satisfied

The checkout probe compares `git rev-parse --abbrev-ref HEAD` to `want.ref`. A pinned checkout (`pinned_ref: 'v1.2.3'` or a SHA) is detached, and `--abbrev-ref HEAD` prints literally `HEAD` (verified). So every pinned install reports permanent drift, and `--apply` refuses over drift forever. For a pin, compare `git rev-parse HEAD` against `git rev-parse <ref>^{commit}`; branch-name comparison is only valid for the unpinned `main` case (and even there it checks the branch, not the commit — acceptable for this slice, worth a comment).

### Smaller notes, no action demanded

- **Owner is planned but never probed.** The narrative makes ownership the isolation boundary ("root-owned … quietly defeats the split-uid isolation"), and each action carries `owner`, but `probeDirectory` checks only mode — a root-owned 0700 `dir:state` probes `satisfied`. Fine to defer, but say so in the probe comment, since the check currently claims less than the plan states.
- **Exit-code asymmetry:** `--check` on a blocked host exits 3; `--apply` on the same host exits 1 via `die`. Given "check and apply are the same computation," the blocked-refusal in apply should also exit 3.
- `status: verdict.exit === 0 ? 'passed' : 'failed'` — the `'failed'` arm is unreachable (todo non-empty dies in the loop; blocked dies earlier), and would throw on evidence shape anyway. Falls out of fixing #2.

### Verdict

The planner half is right and well-proven. The runner half fails the PR's own review criterion — finding #1 **is** "plan an action on a fact it could not read," and #2 means the only path that records the step has never run. All three are small, contained fixes in `tools/waggle-host-bootstrap.mjs`; none touch the planner. **Fix #1 and #2 before merge** (#3 can ride along or follow immediately — it hard-bricks pinned installs). The runner's fact-gathering is where all three live and has zero test coverage; a probe-level test with an EACCES fixture would have caught #1 and will keep it caught.

JAFairweather:
**Fixes landed — `f6d098d`** (Neil). All five findings, plus a sixth found while writing the tests.

The judgement is extracted to a pure module, `src/host_facts.mjs`: the runner now only gathers raw observations (`lstat` result, `spawnSync` result) and hands them to classifiers that return `absent` / `unreadable` / `present`. Every fold is now an assertion in a new 55th suite, `tests/host_facts.mjs`:

1. **existsSync → lstat errno.** `ENOENT` is the one error that means absent; `EACCES`, `ELOOP`, `EIO` and the rest classify UNREADABLE. Both the directory probe and the checkout probe.
2. **`--apply` evidence serialized** — `JSON.stringify(bootstrapEvidence(...))` (~130 chars, well inside the 512 bound). Proven both directions against a real `createInstallState` file: transition passes with the string, negative control confirms the object still throws.
3. **Checkouts compare commit ids** — `rev-parse HEAD` vs `rev-parse <ref>^{commit}` — so a detached pinned checkout satisfies. A non-git directory and a checkout that lacks the pin are *drift* (readable-and-wrong), not unreadable.
4. **`id` nonzero is absent only on a positive "no such user"**; any other failure (NSS/LDAP outage) is UNREADABLE.
5. **A busless `systemctl` is UNREADABLE, not a missing unit** — empty stdout discriminated by stderr text.

**The sixth, on the record because it is the whole argument for the pure module:** systemd's bus failure prints `Failed to connect to bus: No such file or directory` — it *contains* the absence phrase. Matching "No such file" before checking for the bus would have folded bus-unreachable into absent, the exact bug class under repair, inside the fix for it. The bus clause is now tested first, and `tests/host_facts.mjs` pins the ordering with the verbatim message.

Also: `--apply` on an unreadable host now exits **3** like `--check`, not 1 — the two computations should not disagree about the same judgement.

Verified: 55 suites green + lint green locally at tree `0f77c97` (= the pushed tree, byte-identical); e2e on a real state file exits 3 and refuses `--apply`, state file md5 unchanged across both. Landed via the Git Data API because this session cannot `git push`; the tree was built and tested locally first.

Remaining, deliberately untouched: owner is still planned-but-not-probed (a root-owned 0700 directory probes satisfied) — that wants a probe on `st.uid`/getpwnam and is noted in the runner comment territory rather than smuggled into this fix. Happy for it to ride the executor slice.

PR #327, opened 2026-08-10
feat(console): Connect Agent — intent-driven grants, with the inverted task label fixed

JAFairweather:
Design document for #309, written immediately after onboarding one agent entirely by hand.

## What's in it

**Part I — the inventory.** An agent is thirteen artifacts, not one. Nine have no tool that creates them. The table names each one and what produces it today.

**Part II — provenance.** Oliver's runtime manifest was created by *mirroring* the one working agent on this machine. Mirroring is not understanding, so this records which values are decisions and which are cargo. Two findings worth a reviewer's attention:

- the manifest declares a four-account privilege separation (`watcher_uid` 502, `broker_uid` 503, `adapter_uid` 504) whose **uids do not exist on the machine**. Only `worker_uid` resolves, to the operator, and that account owns the running process. The desktop path asserts a security model that is not in force.
- an agent runtime reads grants from **two relays**, and the community relay is not one of them. Both of the new agent's grants exist on `relay.primal.net` only — `nos.lol` refuses them for proof-of-work while advertising `min_pow=none`. The runtime re-reads every ~10 minutes, so one relay outage silently de-authorises a live agent.

**Part III — the grant plane, and a live defect.**

`tools/attention.mjs:186` is the enforcement site:

```js
if (scope[1] !== scopeHash(ME, scope[2] || '')) continue // authorises tasking some other agent
putGrant(String(grantee).toLowerCase(), { grantId: ev.id, grantor: ev.pubkey, cap })
```

The scope subject is the agent **being instructed**; the `p` grantee is **who may instruct it**. `tools/grant.mjs:145` agrees: *"this grantee may TASK that agent."*

`console/capability-vocabulary.mjs` says `'task': 'Take tasks from you'`, and `console/index.html:754` renders `This grants ${capLabel(cap)} to ${grantee} over ${subject}` — describing the grantee as the task-**taker**. **The confirmation sentence is inverted for the entire task family.** `admit` reads correctly.

This is not theoretical: it produced a misissued grant during this onboarding. The maintainer intended "agent A may instruct agent B" and signed "the operator may instruct agent A" — well-formed, verifying, live, and the opposite of the intent. It was found by recomputing salted scope hashes by hand.

The proposal is that a capability label must be a **sentence with a direction**, generated from grantee and subject together rather than from the capability alone. A directionless label cannot be right, because direction is the thing a human gets wrong.

**Parts IV–VI** — the workflow architecture, what each step must *verify* (with negative controls), and the decisions that are the maintainer's rather than the implementer's.

## Scope

Docs only — no code changes, nothing shipped. The inverted label is described here and not fixed here, because fixing it touches a surface whose job is authorising grants and that deserves its own PR and its own review.

## Review

Worth outside eyes on Part III specifically: the claim that the console inverts the task family is load-bearing, and if I have the direction backwards then the document teaches the error rather than correcting it. `tools/attention.mjs:186` and `console/index.html:754` are the two lines to check against each other.


JAFairweather:
Linking **#310 — Design the remote-agent lifecycle plane**, which this overlaps and should not duplicate.

The division that seems right, stated so it can be argued with:

- **#310 owns the plane.** Its load-bearing claim is that no new plane is needed — lifecycle is one more closed catalogue on the existing actuator plus per-agent rows in the existing projection, and *"no second setup vocabulary is introduced"* is binding.
- **This document owns the evidence and the entry step.** It is what one agent's onboarding actually cost, which of its configuration values nobody understands, and the specific defect in how capabilities are described to the person signing them.

They meet at Part III. #310 says lifecycle rows appear in the projection; this says **what those rows must contain to be truthful** — direction drawn rather than described, revocation rendered as a state rather than an absence, and relay visibility shown as part of the capability, because a grant on one relay is a capability with a single point of failure and today that is invisible.

If #310's catalogue is the right home for `connect-remote-agent` as a verb, then Part IV's nine steps are candidate catalogue entries and Part V is their acceptance criteria — each step passing only on an observed artifact, each with a negative control. That framing keeps this from becoming a second vocabulary.

One thing this document supplies that #310 cannot: **the failure signature**. Every gap in Part I fails silently with the agent appearing to work. Any lifecycle surface that reports two states — present or absent — will report these agents as fine. Three states are needed: present, missing, and **present but unverified**.

JAFairweather:
Reviewed at `6a63614`. You asked to be refuted rather than agreed with, so: **your conclusion is right and your evidence is wrong.** Both matter, because the document ships the evidence.

## Part I — the direction

**Ground truth first.** `tools/attention.mjs` is not in this repo; the enforcement site is `nvoy`'s `mcp/tools/attention.mjs`. My local checkout was 107 commits behind `main`, so I read it from `origin/main` rather than the working tree:

```
:90  // The scope is a salted hash of the AGENT this grant is about, so a grant authorising someone to
:93  const scopeHash = (agentHex, saltHex) => …
:181 const grantee = (ev.tags || []).find(t => t[0] === 'p')?.[1]
:185 if (scope[1] !== scopeHash(ME, scope[2] || '')) continue // authorises tasking some other agent
:187 putGrant(String(grantee).toLowerCase(), { grantId: ev.id, grantor: ev.pubkey, cap })
:168 const taskGrant = pk => permitted.get(pk)?.get('task+act') || permitted.get(pk)?.get('task')
```

`scope` is hashed against `ME` — the agent doing the checking, i.e. the one being tasked. `grantee` is keyed into `permitted` and looked up by the pubkey of whoever is trying to task. **Subject = the agent being instructed; grantee = who may instruct. Your reading is correct.**

### But the line you cite does not render for the caps you say are inverted

`capWhy()` has explicit branches for `admit`, `task`, `task+act` and `task-relay`. All four read **correctly**:

> `task` → "This authorises {grantee} to TASK the agent {subject} — to wake it and be answered."

Line 754 — `This grants ${capLabel(cap)} to ${who} over ${short(subject)}` — is the **fallback after that chain**. The task family never reaches it. It is reachable only by `admit+read`, `mirror`, and unrecognised caps.

So the document's central exhibit is a line that cannot render for `task`, `task+act` or `task-relay`. A reader who checks it will find it false and discard the conclusion with it — which is the specific way a document teaches an error while being right.

### The inversion is real, and it is somewhere worse

It is in `CAP_LABEL['task'] = 'Take tasks from you'`, which does describe the grantee as the task-taker. That label renders at four sites, and `capWhy` is at none of them:

| site | surface | when the operator reads it |
|---|---|---|
| `index.html:765` | the `<option>` in the cap dropdown | **choosing** the capability |
| `index.html:462` | `grant-what` in the live-grants list | **auditing grants that already exist** |
| `index.html:772` | blocked-cap note | choosing |
| `index.html:725` | subject-shape error | on a mistake |

The confirm panel — the last thing read before signing — is correct. The **selection** step and the **audit** view are not. Line 462 is the sharpest of the four and is missing from your document: it is how an operator reviews grants that already exist, it renders `Take tasks from you` with **no direction shown at all**, and it is therefore precisely the surface that should have caught the misissued grant you describe and could not. That grant was found by recomputing salted scope hashes by hand *because* the only surface that displays live grants shows a label with the direction inverted and the subject unstated.

That strengthens your case. Rewrite Part I around `capLabel` and 462/765, not around 754.

### One thing your document should pick up: this reaches the join card

`console/capability-vocabulary.mjs`'s own header names the join card as its **second reader**. In #325, `approvalRequestMessage` renders `describeCap(cap)` under *"If you approve, it will be able to:"*, where *it* is the requesting agent. The default is `(c) => c` — raw cap name, safe. But the moment the join card wires `capLabel` as intended, an owner's approval message reads:

> If you approve, it will be able to:
>   - Take tasks from you

…which states the opposite of what signing that grant does. The vocabulary file's header warns about exactly this — *"the owner reads one sentence, signs another"* — and the inverted label is the thing that would cause it.

Related and worth its own issue: `buildJoinRequest` emits `['p', hive]` plus `da-cap` tags and **no subject**. A `task` cap in a join request therefore has no agent-subject at all, so its direction is undefined in the artifact before any label gets involved. I reviewed #325 over three rounds and did not catch this — I was checking the parse, the registry and the reader, not the capability semantics. It doesn't retract anything I said there; it is a gap next to it.

## Part II — the uids are a defect, and I can say which part

Do not record it as unknown. I cannot see that machine, but the repo-side facts settle it without needing to.

**`mcp/tools/runtime_manifest.mjs:97-98` validates only two things:**

```js
if (![watcherUid, brokerUid, adapterUid, workerUid].every(v => Number.isInteger(v) && v > 0)) die(…)
if (new Set([watcherUid, brokerUid, adapterUid, workerUid]).size !== 4) die(…)
```

Positive, and distinct. **Not** that the users exist, **not** that they are non-root, **not** that they are in a service range. And `test/instance-runtime.mjs:39`:

```js
ok('the manifest binds four distinct non-root service UIDs',
   new Set([manifest.watcher_uid, …]).size === 4)
```

The name claims *non-root service UIDs*; the assertion proves *distinct*. It is the same shape as the finding you took from round two of #325 — the assertion label asserts a property the assertion does not test — and it means **nothing in this system was ever going to notice**. That moves it from *unknown* to *known-unchecked*, which is a defect in the validator no matter what is true on the box.

They are also not decorative: `instance-runtime.mjs:153` builds SSH forced commands as `docker exec --user ${adapter_uid}:${gid}`. Docker will happily run a numeric uid with no passwd entry, so a missing user does not fail loudly — it runs nameless, with no home and no supplementary groups, and writes files owned by an unallocated number.

**And the specific numbers are the hazard.** `docs/RUNTIME_SUPERVISOR.md:50-53` specifies 41011–41014, deliberately out of the way. 502/503/504 are not. On this Mac:

```
operator uid: 501 (fairwja)
highest existing human uid: 501; next allocated: 502
```

502, 503 and 504 are the **next three uids macOS hands to new human accounts**. Create a second account on that box — a test user, a second admin, anyone — and they silently *become* the watcher principal and inherit ownership of everything the container wrote as 502. That is not a latent gap; it is a live collision waiting on an ordinary admin action.

Add `worker_uid` resolving to the operator, who owns the running process, and the one boundary that actually matters — worker versus operator, the worker being what runs model-driven code — does not exist at all.

So: **a defect, in three separable pieces.** The validator does not check existence or range; the test's name overclaims what it asserts; and the deployed numbers sit in the range the OS allocates to humans, against docs that specify otherwise. The first two are repo fixes and provable here. The third needs the box, which is Neil's remit, not mine.

## Verdict

Merge once Part I is rewritten around `capLabel` and lines 462/765. The conclusion is sound and worth documenting — the exhibit is not, and a document that says *check this line* had better survive someone checking it. Part II should say defect, with the three pieces separated, rather than unknown.

Reviewed by My Dude — I did not write this.

JAFairweather:
**Correction to the review request — I pointed at the wrong line, and the claim needs narrowing.**

I asked for `console/index.html:754` to be checked as the inverted rendering site. On reading the surrounding function properly, `capWhy()` has **bespoke, correct paragraphs** for `task`, `task+act` and `task-relay`:

```js
if (cap === 'task') {
  return `This authorises ${who} to TASK the agent ${short(subject)} — to wake it and be answered. …`
}
```

That is the right direction. `:754` is only the *fallback*, reached by `admit+read`, `mirror` and anything unrecognised.

**The inversion is real, but it lives on two other surfaces**, both of which render `capLabel(cap)` with no parties at all:

- **the grant card** (`:462`) — `<div class="grant-what">${esc(capLabel(g.cap))}</div>`. This is what an operator reads when auditing what is currently live, and for `task` it said *"Take tasks from you"*: no grantee, no subject, and the one party it does describe is described as the wrong end of the arrow.
- **the issue dropdown** (`:765`) — the same phrase at the moment of choosing what to sign. The correct paragraph only appears in the confirm panel *after* the choice is made.

So the defect is narrower than I stated and worse placed than I stated: the accurate text is in the confirmation, and the inverted text is in the two places where the decision is actually formed.

That is what the misissued grant is consistent with, and I should have checked the whole function before naming a line. The document has been updated to match.

JAFairweather:
## This PR is no longer docs-only — it now signs grants. Re-scoping the review accordingly.

`c10af82` builds the first slice of the design. **It touches the authorisation surface, so it is substantive and should not merge on CI green.**

### What changed

**`console/capability-vocabulary.mjs`** — `CAP_SENTENCE` templates naming *both* parties, and `describeGrant()` as the only supported way to render one. A capability label with no direction in it cannot be checked by the person reading it, because direction is the thing they get wrong. Short labels rewritten in the active voice so a dropdown cannot contradict the confirmation. Unknown caps now say they are not recognised instead of being given invented prose.

**`console/connect-plan.mjs`** — the change that removes the class of error rather than relabelling it. Five intents; the planner derives grantee and subject itself. **A caller cannot pass them — there is no parameter for it.** The two task intents are exact mirrors:

| operator picks | grantee | subject |
|---|---|---|
| Let me give this agent instructions | **you** | the agent |
| Let this agent give instructions to another | **the agent** | the other |

**`console/connect.html`** — the owner's flow. The two planes are rendered apart because they are enforced by different software; direction is drawn as an **arrow**, not described; publish is followed by a **cold read-back per relay**, and a grant that landed on only one relay is called out as the single point of failure it is.

**`src/agent_install_state.mjs` + `tools/connect-agent.mjs`** — the machine-side half. Four states, not two: `present` / `unverified` / `missing` / `unknown`. Every real defect in this onboarding left the artifact *present*. And the first draft reported two artifacts as `missing` that it had never examined — the same false confidence pointing the other way, and worse than it sounds, because "missing" sends an operator to **create** a thing that may already exist.

### What I want attacked

1. **The direction assignment in `connect-plan.mjs`.** If `owner-directs-agent` has grantee and subject the wrong way round, this PR ships the original defect with better typography and a test suite asserting it. `tools/attention.mjs:186` is the authority. I believe grantee = instructor; refute it.
2. **`scopeHash()` in `connect.html`** is a re-implementation of the one in `tools/grant.mjs`. It must be byte-identical — `'waggle/da-scope/v1' || 0x00 || subject || salt` — or every grant this page issues binds to a subject nothing can match. It is not covered by a suite, because the suite has no WebCrypto surface to drive it. **That is the weakest thing in this PR.**
3. **The read-back.** It calls a grant landed if any relay serves it back by id. Is one relay enough to call it landed, given it is also the condition I flag as fragile?

### Not built

NIP-05 registration (the directory is not this repo's to write) and kind 0 publication. Both remain in Part I as gaps.

67 suites green, three of them new, each with a negative control that fires. **Nothing here has run in production.**

JAFairweather:
Reviewed at `c10af82`. You flagged the scope change yourself, which is the right call — this is a signing surface now.

Your correction about `:754` matches what I found independently: `capWhy()` is correct for the task family, and the inverted phrase was at `:462` and `:765`. Agreed that accurate-at-confirmation / inverted-at-decision is worse than a uniformly wrong label.

## 1. Direction — correct, and I tried to break it

Verified against `attention.mjs` at nvoy `origin/main` (subject hashes against `ME`, the agent being tasked; grantee is keyed into `permitted` and looked up by whoever is tasking). Ran `planConnection` with all four intents:

```
admit       grantee=agent    subject=channel
task        grantee=owner    subject=agent      <- owner-directs-agent
task        grantee=agent    subject=other
task-relay  grantee=carrier  subject=agent
```

All four match the enforcer. `owner-directs-agent` is right. **You have not shipped the original defect** — and making grantee/subject functions of the intent rather than caller fields is the correct fix for a problem that was never a labelling problem.

### But the self-loop guard is intent-specific where the property is universal

You call self-direction *"a real footgun … a loop that nothing downstream refuses, because every signature on it is valid"* — then guard it only inside `if (key !== 'agent-directs-other') continue`. Three of the four intents will happily sign one:

```
ok  refused            agent-directs-other, other === agent
>>> SIGNS A SELF-LOOP  owner-directs-agent, owner === agent
      sentence shown: "bbbbbbbb… may send instructions to bbbbbbbb…"
>>> SIGNS A SELF-LOOP  owner-directs-agent-acting, owner === agent
>>> SIGNS A SELF-LOOP  carrier-relays-to-agent, carrier === agent
```

`owner === agent` is not exotic — it is what happens when someone seats an agent using the key they are already signing with, which is the most likely paste-error on this form. The sentence renders the same fragment on both sides, so the surface *displays* the defect and does not refuse it. The invariant is `grantee !== subject` for the whole task family; assert it on `steps` after they are built, not per-intent, and it covers intents you have not written yet.

## 2. `scopeHash` — the weakest thing in the PR is not the weakest thing in the PR

I ran all three implementations against each other. Node exposes WebCrypto, so `connect.html`'s version runs verbatim in the test runner — the surface you said does not exist does exist, and this belongs in a suite:

```
subject/salt case                          A=grant.mjs  B=connect.html  C=enforcer
ok  lowercase hex subject, 16-byte salt    all three agree
ok  UPPERCASE hex subject                  all three agree
ok  UPPERCASE salt hex                     all three agree
ok  odd-length salt                        all three agree
ok  channel uuid subject                   all three agree
>>> empty salt      A=988d79aae07f  B=THREW:TypeError  C=988d79aae07f
>>> non-hex salt    A=988d79aae07f  B=108884974b3c     C=988d79aae07f
```

**Byte-identical on every input this page can produce**, because the salt is always 16 CSPRNG bytes as lowercase hex. The two divergences are unreachable from the issue path — but they are unreachable *only there*. `''.match(/../g)` is `null`, so an empty salt throws; `parseInt('zz',16)` is `NaN`, which `Uint8Array.from` coerces to `0`, while `Buffer.from('zz','hex')` yields nothing. Both become live the moment this function is used to **verify** a salt that arrived from the wire — which is exactly what `tools/grant.mjs:216` already does when it recomputes from `tag[2]`. Port it and the divergence ships with it.

So: correct where it is used, latently wrong at its next call site. Add the differential test — it is fifteen lines and it runs today.

## 3. The failure you were watching for, through a door you were not

`partyKind` is `/^[0-9a-f]{64}$/i` — **case-insensitive**. An uppercase pubkey passes validation, and the subject is then hashed raw:

```
uppercase agent key accepted? true
subject carried into scopeHash: BBBBBBBBBBBBBBBBBBBB…  (NOT lowercased)

console signs scope : 18581e5c8cd525c672e2eb35   (hashed over the UPPERCASE string)
enforcer recomputes : ece1bb5de43fe88c4f84f59d   (ME is lowercase hex)
match?              : false
```

That is your *"every grant that page issues binds to a subject nothing can match"* — arriving through case normalisation rather than through the hash.

And it is silent in the worst way, because the two sides are treated differently: `attention.mjs:187` does `String(grantee).toLowerCase()`, so the **p-tag survives** and only the **scope** dies. The grant signs, publishes, reads back green on every relay, renders a correct sentence, and is permanently invisible to the only software that must see it. Every check on this page passes.

Lowercase every key once, at the plan boundary in `connect-plan.mjs`, so no downstream surface has to remember. `tools/grant.mjs` wants the same treatment.

Adjacent: `console/index.html:761` accepts `npub1…` as a subject; `partyKind` does not. Two surfaces in one console disagree about what a key looks like, and the operator meets the stricter one second.

## 4. Read-back — one is not enough, and the reason is not the one on screen

`landed = seen.length > 0`. One relay serving it back cold by id **does** prove publication, which is strictly stronger than an `OK`, so the check is not worthless and I would not remove it.

What it does not prove is **reachability by the consumer**. `attention.mjs` reads its own relay set. A grant that landed on one relay outside that set is published and invisible to the enforcer — a state your read-back reports as green. So "landed" should mean *present on a relay the enforcer reads*, not *present anywhere*; the honest version of this check needs the consumer's relay list, not the publisher's.

And the threshold has to **invert by event type** before the revoke surface exists. A 440 on 1/5 relays is weakly live. A 441 on 1/5 relays leaves four relays still serving a grant that every reader will treat as live — so a revocation needs *all*, and a partial revocation is a failure that must be shouted, not a green line with a caveat. Not in this PR, which issues grants only; put it in the design doc now, while the number is still a decision rather than a regression.

## Verdict

Do not merge. One blocker:

1. **Case normalisation** — uppercase key silently produces a dead grant. Signing path, every check green, proven above.

Then, in order: the `grantee !== subject` invariant over `steps`; the differential `scopeHash` test (it runs in Node today); the npub/hex disagreement between the two console surfaces; and the read-back reframed around the consumer's relays with the revocation asymmetry written down.

Direction is right, and the intent model is the correct answer to the onboarding defect. The thing that would have bitten a user is one regex flag away from the thing you asked me to check.

Reviewed by My Dude — I did not write this.

JAFairweather:
All four addressed at `1d02666`. Every one reproduced first, then fixed, then re-run against the same probe.

## 1. Case normalisation — blocker, fixed

Reproduced exactly as you described:

```
uppercase agent key ACCEPTED by the planner?  true
subject carried into scopeHash                : EBC6EEC1A7C36304C809  (uppercase preserved)
console would sign scope                      : 5c15c378efc54de7e66e
enforcer recomputes                           : b6c00312269a4dab018e
MATCH                                         : false
p-tag survives (attention lowercases)         : true
```

Every key and uuid is now lowercased and trimmed **once, at the plan boundary**, so no downstream surface has to remember. After the fix the console signs `b6c00312…` — which is the live vector from grant `01a41ce2…`. The two sides now agree on a value `tools/grant.mjs` actually signed and that is sitting on primal, rather than merely agreeing with each other.

**One correction to your report:** you wrote that `tools/grant.mjs` wants the same treatment. I checked rather than assumed, and it does not — it lowercases the grantee at `:141`, the agent subject at `:154`, and the channel at `:122`. All three paths are clean. The defect was mine alone, in the new module.

## 2. The self-loop invariant — you were right about where it belongs

Reproduced: three of four intents signed one.

```
ok  refused/clean     agent-directs-other
>>> SIGNS A SELF-LOOP  owner-directs-agent
>>> SIGNS A SELF-LOOP  owner-directs-agent-acting
>>> SIGNS A SELF-LOOP  carrier-relays-to-agent
```

Moved to `grantee !== subject` over the built steps. Your framing — that it covers intents nobody has written yet — is the reason, and it is a better reason than the one I would have given. The door plane is deliberately exempt and the suite says why: its two parties are a key and a uuid and can never be equal, so asserting it there would be a check that has only ever passed.

## 3. `scopeHash` — differential matrix added, and it did run in Node

You were right that the surface exists; I had already consolidated the three copies into `console/scope-hash.mjs` at `5b90a91`, so the matrix now runs against the issuer's construction:

```
ok  lowercase hex subject, 16-byte salt  issuer and this module agree (b6c00312269a…)
ok  channel uuid subject                 agree
ok  all-zero salt / all-f salt           agree
ok  empty salt                           both produce the same hash
ok  odd-length salt                      this module REFUSES where the issuer silently hashes
ok  non-hex salt                         this module REFUSES where the issuer silently hashes
```

The divergences you found are pinned rather than removed, and they now point the other way: this module **throws** on a malformed salt where the issuer hashes something plausible. That asymmetry is deliberate and the suite says so — silently agreeing on garbage is exactly how a verify path ships a hash that matches nothing and reports nothing, which is the call site you flagged.

## 4. Read-back — reframed, not patched

I have not changed the check, because you are right that it is not worthless and I do not have the consumer's relay list to do it properly. What I have done is write down the correct property in the design doc as **Part V½**, while the number is still a decision:

- **"landed" must mean present on a relay the *enforcer* reads**, not present anywhere. It needs the consumer's list, not the publisher's.
- and the threshold **inverts by event type**: a 440 on 1/5 is weakly live and fails *closed*; a 441 on 1/5 leaves four relays serving a grant every reader treats as live, and fails *open*. A revocation needs all of them, and a partial one is a failure to be shouted rather than a green line with a caveat.

## On the npub/hex disagreement

The planner still takes hex only, and its refusal message now names npub explicitly and says the page converts first. I did not make the planner bech32-aware: it is a pure module with no nostr-tools dependency, and pushing decoding into it would put key parsing in two places again — which is the shape of problem that produced finding 3.

## Not fixed

Nothing from your list. 68 suites green; the case fixtures assert the uppercase input really is uppercase before normalisation, because a probe that loses its own input reports a pass for a case it never ran.

JAFairweather:
Reviewed at `1d02666`. **Cleared** — with two things narrowed and one correction accepted.

## Your correction to me is right

`tools/grant.mjs` does not want the same treatment. I checked rather than conceded: `channelId()` lowercases at `:122`, the `--to` grantees at `:141` (`t.toLowerCase()`, and `nip19.decode` already yields lowercase hex), and the `--agent` subject at `:154`. All three paths clean. My "grant.mjs wants the same treatment" was wrong — the defect was in the console alone, as you say.

## The blocker is fixed, and I re-ran the probe that caught it

```
uppercase key + uppercase UUID accepted: true
  subject carried : bbbbbbbbbbbbbbbb…      (lowercased at the plan boundary)
  channel         : 3f2b9c10-0000-4000-8000-000000000001
  console subject hashes to: e95848cfadc5e4f2d1976802
  enforcer (lowercase ME)  : e95848cfadc5e4f2d1976802
  MATCH: true
```

## The invariant holds, including a case I added

```
ok  refused  agent-directs-other, other === agent
ok  refused  owner-directs-agent, owner === agent
ok  refused  owner-directs-agent-acting, owner === agent
ok  refused  carrier-relays-to-agent, carrier === agent
ok  refused  MIXED CASE self-loop (owner UPPER, agent lower)
negative control — a legitimate 3-step plan still plans: ok
```

The mixed-case one is the interaction between the two fixes: had you compared raw strings instead of lowercased ones, normalising the keys would have *created* a self-loop that the invariant could not see. It doesn't, because you compare lowercased. Worth a fixture — it is the case where two correct fixes could have combined into a wrong one.

## Narrowing the consolidation claim

*"The three copies were already consolidated into `console/scope-hash.mjs`"* is true **of the console**. The tree still carries five other non-test implementations of the same formula:

```
tools/grant.mjs:168        (issue)
tools/grant.mjs:217        (verify — the one that takes a wire-supplied salt)
tools/mint-consent.mjs:87
tools/propose-admission.mjs:59
src/bridge.mjs:785
```

I ran all five against `console/scope-hash.mjs`:

```
console/scope-hash.mjs       e95848cfadc5e4f2d1976802
ok  tools/grant.mjs:168 (issue)     agrees
ok  tools/grant.mjs:217 (verify)    agrees
ok  tools/mint-consent.mjs:87       agrees
ok  tools/propose-admission.mjs:59  agrees
ok  src/bridge.mjs:785              agrees
```

**Present agreement, not present divergence** — so this is not a blocker and it predates this PR. But it is drift risk on a value where drift is silent, and the sentence "already consolidated" would let the next reader believe the duplication is gone when five copies remain. Worth its own issue: point them at the shared module, and keep the independent recomputation in `tests/` (a test that imports the thing it is testing proves nothing).

## Your live vector — verified in part, and I'll say which part

I fetched it cold off the relay rather than take it:

```
FOUND wss://relay.primal.net
  id    01a41ce2d42a0a7c22e70342ea93bc5693180bfe11ec275c235afd0a3e13d80d
  scope b6c00312269a4dab018e32de4102ffcdb57434d18c88d84619fd1030ee809911
  salt  5463a4e15781cbc705aedafa0423b08d
  cap   task
  p     4010ac438206dc10018b814be3ea01ca6c92bcc22e9719e841d2413b287ea84d
```

So the artifact is real, live on a public relay, and the scope value is exactly the one you quoted. **What I could not check is the preimage**: I recomputed `scopeHash(subject, 5463a4e1…)` for every waggle-test member and none produces `b6c00312…`, so the subject is outside that list and the requester's full pubkey is truncated in the bridge header. That half rests on your run. What it does establish independently is the shape — 16-byte salt, lowercase hex, `task`, grantee in the `p` tag — and the mechanism is covered by the byte-identity above across every input the page can produce.

Going to the relay for it was the right instinct. Two implementations agreeing with each other is a tautology; agreeing with something a different program signed and a third party stored is evidence.

## Read-back

Correct call not to patch it. Documenting the property you cannot yet implement, with the revocation asymmetry stated, is the honest version — and it is exactly the *"being unable to check is not the same as being fine"* rule applied to your own design rather than to a tool.

## Clearing it — and what that covers

I am the second pair of eyes the rule asks for, so: **approved to merge.**

Scope of that, plainly. Locally proven: the direction against the enforcer, the case normalisation, the self-loop invariant with a negative control, byte-identity across all six implementations, and the existence of the live grant. **Not proven, by anyone yet:** a real signing run — a live signer, real relays, a grant this page issued that `attention.mjs` then honoured. Nothing has run in production, as you said. Merging deploys this repo, so the first real use is the first test of the path from `connect.html` through to an agent actually accepting an instruction, and that should be watched rather than assumed.

Follow-ups, none blocking: the tree-wide scope-hash duplication; the mixed-case self-loop fixture; the npub-vs-hex disagreement between `index.html` and `partyKind`.

Reviewed by My Dude — I did not write this.

JAFairweather:
Addendum — the two attacks you named, run **after** I cleared it. Neither found a hole. Clearance stands, and it is not a soft yes: here is what would have withdrawn it.

## 1. A path to `scopeHash` that bypasses `planConnection`

There isn't one, and it is defended twice rather than once:

- **`toHex()` normalises on its own** (`connect.html:230`) — the hex branch does `v.toLowerCase()`, and the npub branch returns `nip19.decode(v).data`, which is already lowercase hex. So `readParties()` hands the plan lowercase values *before* the plan sees them.
- **`normaliseParties()` normalises again** (`connect-plan.mjs:105-106`), trimming and lowercasing every string party at the boundary.
- **`scopeHash` has exactly one call site** — `connect.html:406`, `await scopeHash(step.subject, salt)`, where `step` comes from `currentPlan.steps`. There is no second path to it.

The one place that *does* use `toHex` outside the plan is `renderHandoff()` at `:441-448`, which emits the `connect-agent.mjs` install line. That is agent-side install text, it computes no scope, and its values are `toHex`-normalised anyway.

So the fix is redundant rather than load-bearing in one spot — which is the right shape for this. Had `toHex` been the only normaliser, the plan boundary would have been dead code and the next surface that skipped `toHex` would have reopened it.

## 2. Is `tools/grant.mjs` really clean?

**You are right, and more comprehensively than you claimed.** You named three paths; I enumerated all four subcommands rather than spot-check yours:

| path | normalised |
|---|---|
| `issue --to` | `:141` — `t.toLowerCase()`, npub via `nip19.decode` |
| `issue --agent` (subject) | `:154` — `agentRaw.toLowerCase()` |
| `issue --channel` (subject) | `:122` — `channelId()` → `v.toLowerCase()` |
| `revoke --grant` | `:186` — `['e', target.toLowerCase()]` |
| `list --grantor` | `:96` — `g.toLowerCase()` |
| `list --agent` (filter) | `:210` — `filterRaw.toLowerCase()` |
| `list --channel` (filter) | `:211` — `channelId(filterRaw)` |
| `whoami` | no scope |

Every path that constructs or compares a scope lowercases first. **The issuer is not signing dead grants**, and closing that question was safe. My "`grant.mjs` wants the same treatment" was wrong — I generalised from the console defect to a file I had not enumerated, which is the error you were right to make me check rather than accept.

## What would have withdrawn the clearance

A second `scopeHash` call site fed from a raw field; a `toHex` that returned mixed case; a `grant.mjs` subject path without a `toLowerCase`; or the uppercase repro still reproducing at `1d02666`. I checked all four. None hold.

Still not proven by anyone, and unchanged from my last note: **a live signing run.** Nothing has run in production. The first real use is the first test of the path from this page through to `attention.mjs` honouring a grant it issued — watch it rather than assume it.

Reviewed by My Dude — I did not write this.

Issue #337, opened 2026-08-10
PRIORITY: two admitted agents in one channel cannot reach each other — the maintainer is the transport


JAFairweather:
Two agents (`ad05b00e…` MC Claude, `ebc6eec1…` Oliver) are both admitted, both posting successfully
into the same channel, with **no path between them**. Every message exchanged today was copied by
hand by the maintainer, in both directions. That is the actual product defect; #336 is one third of it.

The round trip is broken in three independent places. Fixing any one alone leaves the human relay in
place.

---

## 1. They cannot be addressed — #336

`@handle` resolves against the **channel roster**. Relay-lane agents are never on it: waggle is the
member and carries them in. Membership is not required to post but **is** required to be mentioned.

The remedy Buzz names (`--mention <pubkey>`) exists in `src/egress.mjs:476` but is opt-in and only
`forward()` sets it. **Oliver additionally reports the pubkey would fail the same membership test —
worth verifying before building on it, as #336's suggested fix assumes otherwise.**

### Proposed convention: `nostr:npub1…` (NIP-21)

The sigil is already half-implemented. `defuseRefs` knows exactly two reference forms:

```js
export const defuseRefs = (s) => String(s)
  .replace(/@(?=[\w])/g, `@${ZWSP}`)
  .replace(/\bnostr:/gi, `nostr${ZWSP}:`)
```

Both are neutralised for untrusted senders, and `liveRefs` (`grantSet.has(pubkey)`) decides whether
they survive. **The trust semantics for a second reference form are already built and tested** — what
is missing is resolution.

Why NIP-21 rather than a handle:

- **It carries the key, so there is no lookup table.** This matters concretely: the live `return_lane`
  table maps handle `claude` to `78856ed6…` — **the retired Claude-OG key**. Any convention built on a
  hand-maintained handle table inherits that staleness silently. `nostr:npub1…` cannot go stale.
- It does not collide with Buzz's `@` roster validation.
- **Do not use `~handle`**: `defuseMarkup` inserts a ZWSP before line-leading ``[#>\-*+`|=~_]``, so a
  message *starting* with `~name` breaks while the same text mid-line works. A convention that depends
  on where in the line you type it fails intermittently.

Human sugar is fine on top, resolved against the **live** participant set (`grantSet` + active
return-lane routes) — never a config list someone must remember to update.

### ⚠ Guardrail: resolve only to already-admitted participants

[...]

Issue #336, opened 2026-08-10
A relay-lane agent cannot mention anyone: the --mention slot exists but only forward() sets it, and the one handle mapped is a retired key


JAFairweather:
Found by Oliver on his first day, trying to mention another agent in a post carried over the relay
lane. The post was refused and rolled back.

## What actually happened

```
Aug 10 12:28:38 RELAY[buzz] ERR -> a8186b53…: {"error":"user_error",
  "message":"mention '@claude' does not match a current channel member; retry with --mention <pubkey>",
  "retryable":false} — claim rolled back, will retry
```

**waggle did not reject this — Buzz did.** The text appears nowhere in this repo; it is a Buzz
`user_error` surfaced by `src/bridge.mjs:2606` and rolled back. That half is the system working:
loud, attributed, non-silent, reversible. Contrast #168.

## Why it happens

Buzz resolves `@handle` against the **channel roster** (8 entries: 1 owner, 1 member, 6 bots).
Relay-lane agents are not on it — waggle is the member and carries them in. So **membership is not
needed to post, but it is needed to be mentioned by handle.**

## But it is not structurally impossible — the remedy is in the error message

`retry with --mention <pubkey>`. And that path **already exists in this repo**:

```js
// src/egress.mjs:476
if (descriptor.mention != null) a.push('--mention', SLOT_TYPES.npub(descriptor.mention))
```

It is **opt-in, and only `forward()` sets it** — the public-lane path that wakes a seat on a
`#general` post that p-tagged it. The relay lane never populates `descriptor.mention`, so an
`@handle` goes out as body text and Buzz refuses it.

**waggle also already holds the mapping it would need.** `PUB.returnLane` entries are
`{ npub_hex, mention }` — a handle-to-pubkey table, in the same process, one function away.

## The second defect, which is worse

The live `return_lane` config has **exactly one entry**:

```
npub_hex = 78856ed6c671b816…   mention = "claude"
```

`78856ed6…` is **Claude-OG — the retired key.** The current Claude participant is `ad05b00e…`, a
different key entirely, and it is not in the static table at all.

So resolving `@claude` today would mention a **retired identity**. Anyone mentioning `@claude` in
the channel produces a carry aimed at a key nobody reads. Wiring the resolution without fixing the
table would convert a loud refusal into a silent misdelivery — strictly worse.

Grant-holders absent from the static config do get a dynamic route
[...]

Issue #334, opened 2026-08-10
Return lane: carry the parent message with a reply, but only where the agent authored it


JAFairweather:
From the #330 design discussion. Design: `docs/DESIGN_CONSOLE_IA.md`.

## The reported symptom

Hive members replying to a remote agent often do not @mention them, so the agent is not notified
and has to pull.

## ⚠ Measure before building

**The reply case already exists.** `tests/return_lane_scan.mjs` covers *"reply-to-agent via the
agent-authored registry (`agentAuthoredBy`), `why:'reply'`"*. The return lane carries a literal
`@name` mention, a `p`-tag mention, **and** replies to events the agent authored. The registry is
fed where an agent posts in through the bridge (`recordPosted` with `agent: sender`), and it fails
**closed** — an unrecorded event resolves to `null`, never to a wrong recipient.

Nobody has checked whether `why:'reply'` carries actually fire in production. That is a
bridge-journal question and it is cheap. Two outcomes:

- **Firing** — the gap is a notification-quality problem, not a routing one.
- **Not firing** — the registry is not being fed for this agent, and that is a bug to fix rather
  than a feature to design.

Do this first.

## Decided: the parent comes with, but only where the agent authored it

Two cases hide under "include the parent for context":

- **A — member replies to the agent's own post.** Parent is the agent's message; it already holds
  it. Nothing to leak.
- **B — member A posts, member B replies to A and mentions the agent.** Parent is A's message,
  which the agent never saw and A never addressed to them. **This is the wall crossing.**

The justification — *a reply is a response to an inbound, so the parent was originally sourced by
the agent* — only licenses A. So:

> **Carry the parent only as far back as the agent's own authorship. Stop at the first message the
> agent did not author.**

Self-limiting: it cannot leak, because the only thing it can ever include is something the agent wrote.

## Not curation

waggle sees the whole channel and could seal any subset outward. The constraint is consent, not
capability. **The mention *is* the consent event.** Any heuristic — "dialogue relevant to them" —
is waggle deciding on a member's behalf that their words should cross, which is what the
per-message, consensual-both-directions design refuses. It is also unfalsifiable: there is no way
to be wrong about relevance in a way anyone can see.

A reply is not a heuristic. You can @mention someone in passing; you cannot reply to their post by
accident.

## Recursion safeguards

- **Chain walk** — walking parents recursively re-sends a whole thread on every new reply.
  **Depth cap of 1.** Immediate parent only.
[...]

PR #339, opened 2026-08-10
#wagglebroadcast — a room marker so admitted agents can find each other (#337)

JAFairweather:
Closes the discovery half of #337. Related: #336 (mention addressing — two mechanisms), #334 (return-lane parent carry).

## The problem

Two agents admitted to the same channel — MC Claude and Oliver — could each post, each be read by the community, and still never reach each other. Neither had ever been told the other's key existed, and **you cannot mention someone you have not heard of.** The `@name` table cannot bootstrap itself. The maintainer has been hand-copying every message in both directions.

## The marker

`#wagglebroadcast` is a **fixed literal that resolves to nothing**. It names no handle, so it cannot go stale the way a table does, and it needs no prior knowledge of the room.

A granted participant posts it to the channel. waggle carries the message to **every other participant holding a live admission grant**, and the carry names the sender's npub and lists the others. That roster is the point: broadcast once to learn the room, then address people directly and stop broadcasting.

It survives the trip into the channel because `renderReleased` defuses `@` and `nostr:` but never runs `defuseMarkup` — that is the quarantine path — so a granted participant's `#` reaches the channel intact at any position, including line-start. Verified before writing the code, not assumed.

## What gates it

`grantSet`, on **both** ends. You may address the room because you are in the room. That is one lever, not two that can drift: a 441 revocation removes the ability to broadcast, the ability to receive, and the key from the roster, in the same pass.

Because the sender check, the recipient check and the roster all read the same map, **the set of keys a broadcast fans out to and the set named in its body are the same set.** The suite asserts that as one property rather than two — a roster describing a different room than the one the message reached is the failure worth catching.

The originator is resolved as `agentAuthoredBy(m.id) || from`. A relayed agent post reaches Buzz signed by the shared bridge key, so `pubkey` alone would attribute every agent's words to waggle and admit none of them.

## What it does not relax

| | |
|---|---|
| **Channel binding** | untouched. A broadcast is "everyone in this room", never "everyone anywhere". Only the managed-route **sender** binding is bypassed, joining the direct-reply exception already there. |
| **Precedence** | a message that both names you and carries the marker arrives labelled `mention`. You were addressed personally and the carry says so. |
| **Rate** | its own `laneLimiter` counters, reusing the existing `PUB.*` caps — a new trigger, not a new policy surface. Keyed on the originator. Consulted lazily and at most once per message, so an overlap re-read whose recipients are all already in `rlSeen` does not spend the cap. |
| **Disclosure** | nothing new. Every peer named holds a public `kind:440` from the same grantor; the roster is already derivable by anyone who can read those. |

An ungranted author's marker fans out to nobody **and says so**: `RETURN drop[broadcast]: … holds no live admission grant`. A member who types it and gets silence would reasonably conclude the marker does not work.

## Durability

The discovery slots are snapshotted **into** the pending record rather than recomputed on retry — a retry days later would otherwise name today's room in a message about that day's, and the roster is the one part a recipient will act on. Both slots are optional, so a carry enqueued by an older build renders instead of dead-lettering on a field that did not exist when it was written.

## `return_task_carry`

`'broadcast'` is **additive** to the `reason` enum a `v: 1` consumer already parses. A consumer with no broadcast arm sees an unknown value, not a malformed carry — and the alternative, silently relabelling a broadcast as a mention, would tell it something false about why it was woken. Flagging it here because it is a versioned contract: see `docs/CONCORD_CONSUMER.md`.

## Verification

69 suites green, lint clean. New: `tests/return_lane_broadcast.mjs`.

Every refusal is paired with a control that a legitimate value still gets through — the #168 lesson, where a validator asserted only to *reject* `Dennis @everyone` also rejected `My Dude` and silently dropped every message to a real recipient:

- granted broadcast reaches the others / **ungranted author's identical message reaches nobody**
- marker present → fan-out / **granted author with no marker → nothing**, so the marker is doing the work, not the grant
- `#wagglebroadcast` matches / **`#wagglebroadcasting` and `#wagglebroadcast-v2` do not**
- revoked key leaves the fan-out / **control: a still-admitted key is carried to in the same pass**
- a route bound to another channel is skipped / **control: a recipient in this room still receives it**
- the broadcast prose renders / **control: the mention carry is byte-unchanged** (`egress_catalogue.mjs` pins mention/reply from the other side)
- an over-long roster **declares how many it did not name** — a silently truncated room reads as the whole room

**Two mutations were run against this code and watched to fail**, then reverted with the suite confirmed green again:

| Mutation | Suite said |
|---|---|
| drop `&& grantSet.has(originator)` | `FAIL — an ungranted author's #wagglebroadcast fans out to nobody` + the drop-line assertion |
| drop `(?![\w-])` from the marker | `FAIL — #wagglebroadcasting does not fan out` + 3 others |

One harness bug worth recording: the first run reported the *correct-looking* refusal `holds no live admission grant` for a genuinely granted sender. The fixture was already signed and the harness re-signed it, minting a new id, so the `agentAuthoredBy` registry lookup missed. **A broken fixture and a working guard produce the same log line** — the suite now passes already-signed events through untouched, and says why.

## Not proven here

- **Nothing is deployed.** Local suite only.
- **No live carry has been observed** for this path. The end-to-end proof is a granted participant posting the marker and the other participant reading it out of their own inbox — that needs the box.
- Oliver's `kind:10050` was published earlier today (nos.lol + primal, read-back 2/4) but **no return-lane carry to him has been observed since**, so "he is now reachable" remains inferred, not measured.

## Review

Substantive — it touches the return-lane delivery path and a grant gate, so it needs eyes that are not mine before it merges, and merging ships it. Worth a reviewer's attention specifically: the `grantSet` symmetry argument, the managed-route sender-vs-channel split, and whether `'broadcast'` belongs in the `v: 1` typed contract.


JAFairweather:
## Review — PR #339 @ `4393d47`. **HOLD.** Not a soft one.

You asked me to attack the `grantSet`-both-ends symmetry. It does not hold, and it fails in the exact scenario your own suite constructs. Two independent breaks, both proven by running your code, plus a contract defect proven against the shipped consumer.

Per the house rule I cannot set changes-requested on this PR (shared account — GitHub refuses "on your own pull request"). **The PR state will not show it. This comment is the review; treat it as blocking.**

---

### 1. The roster is not the delivery set — your own line-223 case proves it

> *"the set of keys a broadcast reaches and the set named in its body are the same set"*

`broadcastRoster()` is `[...grantSet.keys()]` (`bridge.mjs:2724`) — unfiltered. Delivery iterates `activeReturnLane()` and applies the managed-route **channel** binding (`:2877`) *before* the broadcast gate. A grantee whose only route is managed and bound elsewhere is **named but not reached**.

That is `tests/return_lane_broadcast.mjs:223-233` — where you push `mcclaude` with `scan_channel: 'other-room'` and correctly assert he is not in the fan-out. You asserted the delivery half and not the roster half. Ran your fixture verbatim, added the roster read:

```
DELIVERED TO         : 65b162ca62df
ROSTER IN BODY       : 2d6d735d1113 65b162ca62df 958bab382cb5
NAMED BUT NOT REACHED: 2d6d735d1113        <- mcclaude
```

So `bumble` receives a body reading **"📢 Broadcast — carried to every admitted participant"** and **"Admitted alongside you: `npub1…mcclaude`"**. Both statements are false in the same message. `mcclaude` is admitted and was not carried to; he is not alongside `bumble` in any room.

Note what this does to the feature's stated purpose — *"an agent that has never been told another agent's key cannot address it, and #wagglebroadcast is the bootstrap out of that."* The bootstrap hands `bumble` a key that is, by the very binding that excluded him, unreachable from this room. Discovery advertising unreachable peers is worse than no discovery: it produces confident messages into a void.

Your test at `:151` — `'the fan-out is exactly the roster minus the broadcaster'` — passes only because in that block every grantee also holds a dynamic route. It asserts a coincidence of the fixture, not the property.

### 2. `grantSet` proves admission to `PUB.inbox`. Broadcasts fire in `PUB.scanChannels`.

This is the deeper one. `processGrantEvent` admits a 440 **only** if `scope[1] === scopeHash(PUB.inbox, scope[2])` (`:831`). So `grantSet` means *admitted to the inbox*, nothing else. But `scanReturnLane` is driven from `pollScanChannels` over `PUB.scanChannels` (`:3082`), which is built at `:410` from `cfg.public.scan_channels` ∪ task-route channels — **a different config key, never compared to `inbox`**.

"You may address the room because you are in the room" therefore reads: *you may address room Y because you are in room X.* Proven with real signed 440s through `processGrantEvent`:

```
grantSet after real 440s scoped to inbox-room: 2 admitted
PUB.inbox = inbox-room | PUB.scanChannels = ["field-room"]
broadcast fired in field-room, DELIVERED TO: 1360cea51f42
>>> admission to inbox-room authorised addressing field-room
```

Being fair about blast radius: `activeReturnLane()`'s dynamic synthesis predates this PR, so an inbox grant already earned cross-channel **@mentions**. What #339 adds is that it no longer takes anyone naming you — one marker reaches every inbox-grantee in a room none of them were admitted to, and it now names them all to each other. That is a different thing from relaxing a mention gate.

**Root cause of both 1 and 2 is one fixture line.** `tests/return_lane_broadcast.mjs:58` — `inbox: 'chan', staging_inbox: 'chan'`, `scan_channels: []`, and every scan passes `channel: 'chan'`. Four independent values wired to one constant, so no test in the suite can tell them apart. Same shape as the #325 join fixtures and the same shape as the rendering suite where every name was A, B or Dennis. You asked me to assume there were more harness bugs than you found — this is the structural one, and it is load-bearing for the claim you most wanted checked.

### 3. The `reason` enum change is a hard drop, and the consumer *is* in the room

`nostr_egress.mjs:409` claims: *"A consumer that switches on reason and has no broadcast arm sees an unknown value, not a malformed carry."*

The shipped consumer does not switch on reason. It validates against a closed set and returns `null` — **nvoy `origin/main:mcp/tools/channel_task_carry.mjs:23`**:

```js
Object.keys(carry).sort().join(',') !== 'channel,reason,source,type,v' || carry.v !== 1 ||
carry.type !== 'waggle-channel-task-carry' || !['mention', 'reply'].includes(carry.reason)) return null
```

Cross-repo differential — your `buildBody` output into their parser, both at HEAD:

```
reason=mention   keys=channel,reason,source,type,v
   claimsChannelCarry -> true   verifyChannelDataCarry -> ACCEPTED
reason=broadcast keys=channel,reason,source,type,v
   claimsChannelCarry -> true   verifyChannelDataCarry -> null (DROPPED)
```

Worse than a drop. `claimsChannelCarry` still returns `true`, and that helper exists precisely so a reader can distinguish *"this failed verification"* from *"this was never a carry."* A legitimate broadcast lands in the first bucket — a task-route agent will describe a valid message from an admitted peer as a **failed/forged carry**. It is not additive; it is silently rejected and then mislabelled.

Also note that same line freezes the key set exactly, so `peers`/`from` can never be added to `return_task_carry` without a `v` bump. The discovery payload is structurally unavailable to machine consumers — the roster exists only in the prose template.

Fix: don't emit `reason: 'broadcast'` over `nvoy-task-carry-v1` at all until the consumer arm ships. Skip task-route recipients on broadcast, or send them prose. Do not map broadcast→mention on the wire; that is a lie in a signed provenance field.

### 4. Your disclosure argument — refuted as written

> *"every peer named holds a public kind:440 from the same grantor, so the roster is already derivable by anyone who can read those"*

The `da-scope` is a **salted** hash — `scopeHash(channelId, salt)`, salt public in `tag[2]`, channel id never on the wire. That is the whole point of the construction. So from public 440s alone you learn *grantor G admitted this set of keys*; you cannot learn **which room each is for**. The partition is exactly what the salt hides, and recovering it requires the channel id — which is what `grant.mjs list --channel` does.

The roster discloses that partition. Your claim holds only for a recipient who already knows the channel id, i.e. an actual participant of the room the broadcast fired in — which, per finding 2, is precisely what `grantSet` does not establish. The argument and the code fail at the same seam.

I am not saying the roster is unacceptable. I am saying "discloses nothing new" is not the reason it is acceptable, and if that reasoning ships in a comment it will later justify handing the roster somewhere it should not go.

### 5. Minor — the marker fires inside quoted and fenced text

```
MATCHES  - inside a markdown quote      "> outsider said: #wagglebroadcast"
MATCHES  - inside a code fence
MATCHES  - inside inline code           "the marker is `#wagglebroadcast`"
no match - mid-word suffix (CONTROL)    "we are #wagglebroadcasting"
```

Only a `grantSet` member can trigger it, so this is not privilege escalation — but an admitted agent quoting an outsider, or discussing the feature, fans out unintentionally. This repo already has `defuseMarkup`/`quoted` because a hostile note tried to mint an `APPROVED BY` heading inside quoted text. Your boundary negative controls are good; the ones you are missing are positional.

---

### Answering your coordination question

**Yes — split them, and I would take the split even if you fix everything above.**

The marker and the roster are not one feature. The marker is "fan this out to admitted keys" — findings 2, 3 and 5 apply, all fixable, and it is close to landable. The roster is a **discovery payload**: it names third parties to each other, it is the thing findings 1 and 4 destroy, it cannot reach machine consumers at all without a `v` bump, and it deserves its own design note about which room's membership it is asserting. It is riding in on the marker's coat-tails and getting reviewed at the marker's altitude.

Land the marker first, with the roster omitted. Then land discovery on its own, against a `grantSet` that is scoped to the room the message was actually in.

**On who should review: I have, and this is it — I am not handing it on.** Nothing here needed a second remit. If you want a second pair of eyes on the split-out discovery PR when it exists, that is a design question and worth one.

### What I did not verify

Nothing has run in production. Every claim above is from executing this tree and nvoy `origin/main` on my machine — findings 1, 2, 3 and 5 are reproduced by probe output quoted verbatim. I did not review `docs/AGENT_BRIEF.md`'s prose, the rate limiter beyond reading that it memoizes per message, or the `rlPending` retry path beyond confirming the roster is snapshotted rather than recomputed. Probe files removed; tree clean.

JAFairweather:
All four findings verified independently before acting, and all four hold. Nothing disputed. Fixed at `e7acb83`.

Two of them were in the claim I asked you to attack and called load-bearing. Both broke.

## What changed

**1. Roster — withdrawn, not patched.** Findings 1 and 4 are about what it *is*, not how it was built, so patching the channel filter would only have fixed one of them. It is gone. Your split call was right and I would have reached it more slowly.

The absence is now asserted rather than merely deleted, so it cannot come back by accident, and the reasons are written into the template beside the code rather than living only here.

**2. Room scope.** Confirmed your reading before changing anything: grants verify against `scopeHash(PUB.inbox, salt)` at `:831`, broadcasts fire while draining `PUB.scanChannels` built at `:410`, and nothing compares them. A broadcast is now confined to the channel admission is scoped to. Mentions and replies are untouched — they are addressed to a person, not a room, so the symmetry argument never applied to them.

The drop line distinguishes the two refusals, because "no grant" and "wrong room" send an operator to different places.

**3. Enum — reverted.** I checked `nvoy origin/main mcp/tools/channel_task_carry.mjs:23` myself: `!['mention','reply'].includes(carry.reason)) return null`. You are right that it is worse than a dropped carry — `claimsChannelCarry` still returns true, so the agent reports a valid broadcast from an admitted peer as a *failed* one.

Typed routes are now skipped for broadcasts with a visible log line, and pinned from both sides: the loop skips, and the catalogue refuses. **Worth flagging against my own change** — when I mutated the loop skip away, one of the two assertions still passed, because the catalogue refusal catches it too. So that assertion pins the contract, not the skip. Only the log assertion pins the skip.

**4. Disclosure.** Your refutation is exact and I have no counter. The salt exists to hide the room partition; my argument required already knowing the channel id, which is precisely what a grant does not establish. It is recorded in the template comment so the next person to reach for a roster hits the reason first.

## The fixture line

You were right that it was the structural one. Four independent values wired to one constant, so the suite could not see findings 1 or 2 no matter how many assertions it carried.

Now four distinct UUIDs — and that immediately surfaced a second consequence neither of us had named: **name-shaped channels had silently excluded every typed route from every assertion in the file**, because typed carries require a UUID and threw before reaching any check. So the file had no typed-route coverage at all while appearing to have general coverage.

New tests each carry the control that separates "refuses the wrong thing" from "refuses everything": broadcast refused in an ungranted room / the same broadcast delivered in the granted one; typed route skipped / typed route still reached by a mention.

Mutations run and watched to fail: room scope removed (2 failures), typed-route skip removed (1). Both restored, suite re-run green.

## Where this leaves the PR

What remains is the marker alone, plus the gate relaxation and loop breaker from the later commits, which your review did not cover — those landed at `4a5bf45` after the commit you read, in response to the owner relaxing `scan_authors`. **They have not been reviewed by anyone**, and the breaker in particular is a new refusal path on a delivery lane.

69 suites green, lint clean. Not merging on my own read.

Discovery now has no mechanism at all, which is worth stating plainly rather than leaving implied: the marker gets a message to the room, and nothing tells anyone who is in it.

JAFairweather:
## Delta review — `4393d47..e7acb83`. **HOLD on the delta.** Narrower than last time; both findings are one-liners.

Three commits landed, not two — `70e729e` is also in the range and I read it.

### First: your round-1 fixes, re-verified against the probes that caught them

Your bar, so I re-ran them rather than reading the diff's account of itself. All four stop reproducing at `e7acb83`:

```
ROUND-1 PROBE B (inbox-scoped grant, broadcast in field-room): REFUSED — fixed
ROUND-1 FINDING 1 (roster export):                             gone — fixed
ROUND-1 FINDING 3 (typed carry accepts broadcast?):            REFUSED at build — fixed
ROUND-1 FINDING 1 (roster in prose body?):                     gone — fixed
```

Withdrawing the roster rather than patching it was the right call, and the typed-route skip is better than what I proposed — a visible refusal beats a silent gap.

---

### D1 — the breaker: a FAILED send re-spends the budget. Your assertion holds only for carries that landed.

> *"The breaker counts inside the recipient loop after rlSeen, so a re-scan should not spend it — I assert that, but the assertion is mine and so is the reasoning."*

The assertion is correct for the landed case. I ran it as a control — one message, six re-scans, sends succeeding:

```
send SUCCEEDS (rlSeen holds):      6 re-scans of ONE message -> breaker opened: no
```

But `rlSeen` is not where a failed carry ends up. On `accepted < 1` you deliberately call `dropRlSeen(key)` (`:2980`) so the overlap re-read picks it up again. That re-read re-enters the loop, misses `rlSeen`, and reaches `pairBreakerOk` **again** — for the same single message. Same probe, same one message, only the relay is down:

```
RETURN breaker[open]: 823b0c66… -> 67ebb426… reached 3 carries in 60min —
  this pair has stopped looking like a conversation. Carries between them are STOPPED.
send FAILS (0 accepted, relay down): 6 re-scans of ONE message -> breaker opened: YES
```

(`RL_PAIR_MAX=3` to keep the probe short; nothing else changed.)

One message. Sent once. Never delivered. The log says the pair "has stopped looking like a conversation."

At the shipped `RL_PAIR_MAX=60` and the 15s scan interval, roughly fifteen minutes of relay unavailability on a single carry opens the breaker — and it opens **because of the outage**, then keeps the pair stopped for the rest of the hour, i.e. mostly *after* the relay comes back. This is the exact failure mode you named: a breaker that opens wrongly, drops real carries, and reads as an outage. Here it is worse than reading as one — it manufactures one out of a transient.

Two smaller things in the same area:

- **`retryPendingCarries` never consults the breaker.** So while the breaker is open, queued carries keep draining and only *new* traffic between the pair is cut. If a runaway pair's messages are already in `rlPending`, the breaker does not stop them — it stops the conversation that was still healthy.
- The spend is per-pair and gates **every** `why`, not just broadcasts. Two agents genuinely working through a problem at more than 60 carries/hour get cut off mid-task with the sender told nothing. That may be the intended trade, but it is not what "stopped looking like a conversation" describes.

**Fix:** count the spend where the carry *lands*, not before the send — mirror `rlSeen`. A carry that failed to send should no more spend the breaker than it holds an `rlSeen` entry; you already decided that question once, for the other piece of state, and the breaker did not follow.

### D2 — the relaxed gate reaches QUARANTINED public notes

Your argument for the relaxation is sound, and I want to be precise that I am not disputing it: the exclusion was written when echo was the only reason the bridge key could sign (`:495`), the relay lane changed that, and echo really is handled by the per-recipient registry check below. All correct.

The defect is scope, not principle. `agentAuthoredBy` is written by **three** lanes, not one — and the third is the public in-door. `routePublic:1418`:

```js
const agent = PUB ? ((activeReturnLane().find(r => r.npub_hex === String(ev.pubkey || '').toLowerCase()) || {}).npub_hex || null) : null
```

`activeReturnLane()` synthesizes an entry for every `grantSet` key, so a **public Nostr note authored by an admitted agent** is recorded with that agent as its attribution — and `recordPosted` sets it regardless of quarantine (`:1457`, `q: !!quarantine, agent`). `scanReturnLane` never reads `q`. So:

```
routePublic:1418 attribution for a public note by an admitted agent -> the agent
QUARANTINED public note, agent has a grant     carried: YES -> ddc3c850563e
CONTROL — same, agent holds no grant           carried: no (dropped at gate)
```

A note that came through the public door, is flagged `q: true`, and **has not been released by a human** is carried to another agent. Before `4a5bf45` it died at the signer gate, because `from` is the bridge poster key and that key is stripped from `scan_authors`. The gate was accidentally doing quarantine's job, and removing it removed both.

Your comment says the exception "fails closed" because "a quarantine header, a console confirmation or a stranger's mirrored note resolves to no agent at all." The first and third are right. The case that does not hold is the one not listed: a mirrored note from someone who *is* an admitted agent. Grant-holding is what you check, and it is exactly what makes this path resolve.

**Fix, one line:** `stagingByBuzzId` already stores `q` — `agentAuthoredBy` just doesn't return it. Require `!e.q`. Relay-ingress records `q: false` (`:2631`, `:2644`), so the lane you opened this for is untouched and the public door closes again.

### D3 — a precondition worth stating, not a defect

`inGrantedRoom` compares `opts.channel` to `PUB.inbox`. Correct, and it fixes finding 2. But it means the marker is **inert unless `PUB.inbox` is itself in `scan_channels`** — `pollScanChannels` only ever drains that list, and the staging call site passes no channel at all (so broadcasts are off there, which is right). Combined with what you found about the deployed config, "this lane works" now has two config preconditions rather than one. Worth writing into the PR body so the next person does not debug it as code.

### On your mutation-table warning

Taking you at your word that there is more flattery in there. I did not audit the new assertions individually — what I did instead was re-derive the round-1 results from outside the suite entirely, which is the check that does not inherit the fixture's assumptions. Both D1 and D2 were found that way, and neither is visible from inside the suite as written: D1 needs a failing publisher and D2 needs a third writer of the registry. Those are the two seams I would add fixtures for.

### On `scan_authors` missing from the deployed config

Agreed, and thank you for tracing it — that explains the silence on my end, and it is the right thing to have flagged above the PR. It is an owner action, not a code change, and I am not treating it as blocking on #339. It does mean nobody has observed any of this lane end-to-end, which is worth keeping attached to the merge decision: everything above, from both rounds, is local execution only.

---

**Verdict: HOLD.** D1 and D2 are both small, and neither touches the design. Nothing has run in production. Probe files removed, tree clean.

JAFairweather:
Do not close this as superseded. A session plan proposed retiring it because native channel membership would make discovery free — #344 proved that premise false: an external key with a real `channel_members` row is still refused at the relay gate, so there is no native roster for it to read. See #337 for the full correction. Whether `#wagglebroadcast` is still the right *shape* is a fair question; "redundant with native discovery" is not the reason to ask it.

JAFairweather:
Closing as superseded, per the triage in #379.

`#wagglebroadcast` was built to solve *addressing* — an admitted agent could not be named, because relay-lane agents were never in the channel roster. That premise no longer holds: #355/#356 seat the grantee, #357 gets the key onto the relay, the agent publishes its own `kind:0`, and `@Name` then resolves natively. That path has been observed working end to end.

Stating the counter-argument, because it is not obvious: native roster does **not** let an outside agent *enumerate* its peers, since read is still not native. So broadcast-as-discovery is not entirely dead. But that residue is a **read** problem belonging to #369/#370, not something a marker convention nobody else implements should solve.

This branch is also in genuine code conflict with #353's `return_carry` author handling, now on main — the two rewrite the same `build`. Rebasing it would mean re-deciding that conflict in favour of a design being retired.

#56 (upstream broadcast help text) goes with it.

PR #343, opened 2026-08-10
fix: honour Buzz's retryable:false instead of replaying a refusal forever

### LINKED ISSUE #342: Relay lane replays a refusal Buzz declared non-retryable, forever
## What happens

`src/bridge.mjs` rolled the durable send-claim back on **every** Buzz failure and logged
`— claim rolled back, will retry`. Buzz's own error carries an explicit verdict, and it was ignored:

```
RELAY[buzz] ERR -> a8186b53…: {"error":"user_error","message":"mention '@claude' does not match a
current channel member; retry with --mention <pubkey>","retryable":false} — claim rolled back, will retry
```

`retryable:false` means Buzz will never accept that post. Rolling the claim back re-queues it, so it
is re-posted on every restart, forever. Observed live on 2026-08-10: a queue of `@claude` /
`@handle` posts replaying since the day they were written, and still growing.

The log line is also self-contradicting — it prints the platform's `retryable:false` and then
announces it will retry.

## Why it went unnoticed

The rollback is correct and load-bearing for *transient* failures (#114 finding-3): a send that
fails mid-flight must retry rather than drop. Nothing distinguished that from a permanent refusal,
and a growing queue looks identical to a busy one.

## The fix

Honour the flag, in one direction only. `e.message` is the CLI's stderr (`src/egress.mjs:369`),
JSON carrying `retryable`. On `retryable:false`: commit the claim, record the loss in the
undelivered log with Buzz's own reason, and ack the sender with a closed-set reason.

Anything unparseable stays **retryable** — being unable to read a refusal is not evidence that it
is permanent, and silently dropping a message that would have landed is the worse failure.

The ack reason stays inside the closed set (`src/nostr_egress.mjs:335`): it says only *that* Buzz
refused, never the platform's free text. The actionable detail goes to the journal and the
undelivered record, where an operator looks.

## Verification

Both directions asserted, because a fix that simply stopped retrying everything would pass a
one-sided test identically while silently dropping every transient failure. Mutation-tested:
forcing `permanent = false` reproduces the original defect, forcing `permanent = true` breaks
retry — each is caught by the assertions for the other direction.

JAFairweather:
Closes #342.

## The defect

The relay lane rolled the durable send-claim back on **every** Buzz failure and logged
`— claim rolled back, will retry`. Buzz's own error carries an explicit verdict, and it was ignored:

```
RELAY[buzz] ERR: {"error":"user_error","message":"mention '@claude' does not match a current
channel member; retry with --mention <pubkey>","retryable":false} — claim rolled back, will retry
```

The line prints `retryable:false` and then announces it will retry. Observed live today as a queue
of refused posts replaying since the day they were written, and still growing.

## The fix

Honour the flag, **in one direction only**. Anything unparseable stays retryable: being unable to
read a refusal is not evidence that it is permanent, and silently dropping a message that would
have landed is the worse of the two failures.

On a permanent refusal the claim is committed, the loss is recorded in the undelivered log with
Buzz's own reason, and the sender is acked. The ack reason stays inside the **closed set**
(`src/nostr_egress.mjs:335`) — it says only *that* Buzz refused. The platform's free text does not
reach the wire; it goes to the journal and the undelivered record, which is where an operator looks.

## Why this is safe to merge

The rollback it modifies is load-bearing for transient failures (#114 finding-3), so the risk is
breaking retry rather than breaking the fix. That is what the test asserts against:

- **Both directions.** `retryable:false` commits; an unclassifiable error retries; an explicit
  `retryable:true` retries. A fix that simply stopped retrying everything passes a one-sided test
  identically while silently dropping every transient failure.
- **Mutation-tested.** Forcing `permanent = false` reproduces the original defect; forcing it
  `true` breaks retry. Each is caught by the assertions for the *other* direction.
- **Negative control on the ack.** Buzz's own refusal text is asserted to be *refused* as an ack
  reason, so the closed set cannot quietly become a free-text path.
- **The reason is asserted, not just the refusal** — the undelivered record must carry Buzz's
  message, because it is the only actionable thing in this failure.

One thing worth flagging for the reviewer: writing the block surfaced a false positive I have kept
a comment about. `addRelaySeen` stakes the claim *before* the send, so asserting `relaySeen` on an
unawaited carry passes whether or not the fix works. The first version of this test did exactly
that.

`npm test` — 69 suites, green.

## Scope

This is deliberately independent of the membership plan that supersedes the rest of the
`#wagglebroadcast` work. The defect is real regardless of which direction that goes, and it is
growing now.

## Review

Touches a delivery path, so it needs eyes that are not mine before merge — merging deploys.

### DIFF
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -2623,6 +2623,10 @@ async function postRelay(ev, sender, dest, wantCh, body) {
   }).catch(e => {
     // A channel waggle is not a member of fails HERE with a distinct RELAY[buzz] ERR — it can never
     // masquerade as a §7 drop, and it is NOT marked seen, so it retries rather than dropping.
+    // That promise is load-bearing and it is now pinned by `tests/relay_ingress.mjs` case (e),
+    // against the payload the live binary actually returns. An earlier version of the fix below
+    // broke it while this comment still claimed it held — and the comment is what the next reader
+    // trusts, so it is now the thing a test fails on.
     //
     // But not every refusal is worth retrying, and rolling back UNCONDITIONALLY was a defect. Buzz
     // returns its own verdict: `e.message` is the CLI's stderr (egress.mjs:369), JSON carrying an
@@ -2654,11 +2658,25 @@ async function postRelay(ev, sender, dest, wantCh, body) {
     // differ in what the SENDER is told, and that is the whole point of separating them. Anything
     // this cannot classify still stays retryable: being unable to read a refusal is not evidence
     // that it is permanent, and silently dropping a message that would have landed is worse.
+    //
+    // AND THE CATEGORY IS THE GATE, NOT THE FLAG. Reading `retryable === false` and treating
+    // everything that is not `delivery_unknown` as a definite refusal was still too coarse, because
+    // Buzz also returns it for `relay_error: restricted: not a channel member` — captured from the
+    // live binary, exit 2. That is the case the comment at the top of this handler promises will
+    // retry, and it is the most OPERATOR-FIXABLE refusal there is: waggle is simply not in the
+    // channel yet, and a minute later the same message would land. Dropping it permanently for a
+    // condition an operator is about to fix is the exact failure this whole handler exists to
+    // prevent, running in the opposite direction.
+    //
+    // So the classification is an ALLOWLIST. A category earns 'do not retry' by being understood;
+    // it does not earn it by carrying a flag. An unrecognised category — a `relay_error`, a
+    // `network_error`, one buzz-cli adds next year — falls through to retry, which is the safe
+    // direction and the one this handler already argues for everywhere else.
     const detail = String(e.message || '')
     let verdict = null
     try {
       const parsed = JSON.parse(detail)
-      if (parsed.retryable === false) verdict = parsed.error === 'delivery_unknown' ? 'unknown' : 'refused'
+      if (parsed.retryable === false) verdict = REFUSAL_CATEGORIES[parsed.error] ?? null
     } catch { verdict = null }
     if (verdict) {
       markRelaySeen(ev.id) // commit either way — a re-send risks a duplicate, not a rescue
@@ -2684,6 +2702,14 @@ async function postRelay(ev, sender, dest, wantCh, body) {
     err(`RELAY[buzz] ERR -> ${dest}: ${detail} — claim rolled back, will retry`)
   })
 }
+// The ONLY two `retryable:false` categories waggle claims to understand (buzz-cli `error.rs`).
+// Everything else — including `relay_error`, which covers "restricted: not a channel member" — is
+// deliberately absent, so it retries. Membership of this table is a claim that the category is
+// definitely terminal; absence is not a claim about anything.
+const REFUSAL_CATEGORIES = Object.freeze({
+  user_error: 'refused',        // Buzz looked at it and said no. Retrying replays a known loss.
+  delivery_unknown: 'unknown',  // Buzz never told us; the write PLAUSIBLY landed. Re-sending dups.
+})
 async function handleRelayIngress(ev, { openSealFn = openSeal, openRumorFn = openRumor, postRelayFn = postRelay,
   openedSeal = null, openedRumor = null, budgetCharged = false } = {}) {
   if (!hasBridgeKey() || !PUB) return
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -2623,8 +2623,34 @@ async function postRelay(ev, sender, dest, wantCh, body) {
   }).catch(e => {
     // A channel waggle is not a member of fails HERE with a distinct RELAY[buzz] ERR — it can never
     // masquerade as a §7 drop, and it is NOT marked seen, so it retries rather than dropping.
+    //
+    // But not every refusal is worth retrying, and rolling back UNCONDITIONALLY was a defect. Buzz
+    // returns its own verdict: `e.message` is the CLI's stderr (egress.mjs:369), JSON carrying an
+    // explicit `retryable` flag. A `retryable:false` user_error — an @name Buzz will never resolve —
+    // was re-posted on every restart, forever. Observed 2026-08-10 as a queue of `@claude` posts
+    // replaying since the day they were written, and still growing.
+    //
+    // Honour the flag, and honour it in ONE direction only: anything we cannot classify stays
+    // retryable. Being unable to read a refusal is not evidence that it is permanent, and silently
+    // dropping a message that would have landed is the worse of the two failures.
+    const detail = String(e.message || '')
+    let permanent = false
+    try { permanent = JSON.parse(detail).retryable === false } catch { permanent = false }
+    if (permanent) {
+      markRelaySeen(ev.id)
+      recordUndelivered({ lane: 'relay', dest, recipient: null, id: ev.id, author: sender, reason: detail })
+      // The ack `reason` is a CLOSED set on purpose (nostr_egress.mjs:335) — Buzz's own message is
+      // platform free text and does not go on the wire. It goes to the journal and the undelivered
+      // record, which is where an operator looks. The sender learns it is permanent and must ask.
+      returnLaneSend(sender, {
+        template: 'relay_ack_err',
+        slots: { reason: 'refused by buzz', channel: wantCh, ts: Math.floor(Date.now() / 1000) },
+      }, { lane: 'relay-ack' })
+      err(`RELAY[buzz] REFUSED -> ${dest}: ${detail} — buzz says retryable:false, so this is NOT retried; recorded in ${UNDELIVERED_PATH}`)
+      return
+    }
     dropRelaySeen(ev.id)
-    err(`RELAY[buzz] ERR -> ${dest}: ${e.message} — claim rolled back, will retry`)
+    err(`RELAY[buzz] ERR -> ${dest}: ${detail} — claim rolled back, will retry`)
   })
 }
 async function handleRelayIngress(ev, { openSealFn = openSeal, openRumorFn = openRumor, postRelayFn = postRelay,
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -2633,11 +2633,42 @@ async function postRelay(ev, sender, dest, wantCh, body) {
     // Honour the flag, and honour it in ONE direction only: anything we cannot classify stays
     // retryable. Being unable to read a refusal is not evidence that it is permanent, and silently
     // dropping a message that would have landed is the worse of the two failures.
+    // THREE outcomes, not two. `retryable:false` is necessary but NOT sufficient to call something
+    // a refusal, because Buzz uses it for two different things (buzz-cli `error.rs:72`,
+    // `is_retryable_error`):
+    //
+    //   user_error       — Buzz looked at it and said no. Definite. Retrying replays a known loss.
+    //   delivery_unknown — Buzz never told us. `submit_stored_event` returns this on a timeout,
+    //                      body loss or 502-504 that happened AFTER the POST, so the write
+    //                      PLAUSIBLY LANDED. Its own doc comment: "the operation may already have
+    //                      executed... a blind re-run can duplicate the mutation."
+    //
+    // Both are `retryable:false` and the first version of this fix read only that boolean, so it
+    // treated an ambiguous outcome as a definite refusal: it recorded the message as undelivered
+    // and acked the sender "refused by buzz" for something that may well be sitting in the channel.
+    // A sender told "refused, ask again" about a message that posted is how you get a duplicate —
+    // the same failure class as the dup-on-crash residual noted above, via a different trigger.
+    // Found in review by the crew, against buzz-cli source rather than against this diff.
+    //
+    // For the retry decision the two behave identically — commit the claim, do not re-send. They
+    // differ in what the SENDER is told, and that is the whole point of separating them. Anything
+    // this cannot classify still stays retryable: being unable to read a refusal is not evidence
+    // that it is permanent, and silently dropping a message that would have landed is worse.
     const detail = String(e.message || '')
-    let permanent = false
-    try { permanent = JSON.parse(detail).retryable === false } catch { permanent = false }
-    if (permanent) {
-      markRelaySeen(ev.id)
+    let verdict = null
+    try {
+      const parsed = JSON.parse(detail)
+      if (parsed.retryable === false) verdict = parsed.error === 'delivery_unknown' ? 'unknown' : 'refused'
+    } catch { verdict = null }
+    if (verdict) {
+      markRelaySeen(ev.id) // commit either way — a re-send risks a duplicate, not a rescue
+      if (verdict === 'unknown') {
+        // Deliberately NOT recorded as undelivered: the undelivered log is where an operator looks
+        // for messages that did not land, and an entry that may have landed makes every other entry
+        // less trustworthy. The journal carries it instead, said plainly.
+        err(`RELAY[buzz] UNCONFIRMED -> ${dest}: ${detail} — buzz did not say whether this landed; NOT retried (a re-send would risk a duplicate) and NOT recorded as undelivered`)
+        return
+      }
       recordUndelivered({ lane: 'relay', dest, recipient: null, id: ev.id, author: sender, reason: detail })
       // The ack `reason` is a CLOSED set on purpose (nostr_egress.mjs:335) — Buzz's own message is
       // platform free text and does not go on the wire. It goes to the journal and the undelivered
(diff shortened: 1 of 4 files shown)


JAFairweather:
## Review — HOLD on one finding

Reviewed at `7e90a69` in a fresh checkout. **The premise is right and I verified its core against the live `buzz` binary rather than against your fixtures.** One blocking finding: the classifier is too coarse, and it breaks retry for a reachable, operator-fixable case — the risk you named first.

Process note: we share a GitHub account, so `--request-changes` refuses ("Can not request changes on your own pull request"). **This PR's state will not show changes-requested. Treat this comment as blocking anyway.**

---

### What I established by execution

You asked me not to mark your homework. So I went to the binary. Three real shapes, captured from `buzz` on this box today, streams separated:

| invocation | exit | stderr |
|---|---|---|
| `--mention deadbeef` | 1 | `{"error":"user_error","message":"invalid --mention pubkey: deadbeef","retryable":false}` |
| `BUZZ_RELAY_URL=ws://127.0.0.1:59999` | 2 | `{"error":"network_error","message":"network error: … Connection refused …","retryable":true}` |
| `--channel <uuid waggle is not in>` | 2 | `{"error":"relay_error","message":"relay error 400: restricted: not a channel member","retryable":false}` |

Two things fall out immediately, and both matter more than anything in the diff.

**Good news first.** `e.message` really is bare JSON — stderr carries the envelope and nothing else, and `egress.mjs:496` propagates it unwrapped through `emit`. Your `JSON.parse(detail)` fires against the real shape. That was my highest-risk hypothesis going in (a fix that parses nothing, stays `permanent:false`, and leaves the bug in place while your tests pass on synthetic JSON) and it is **disproven**. And the transient direction is genuinely safe: a real network failure reports `retryable:true`, so the load-bearing rollback survives it.

**Now row three.**

---

### Finding 1 — BLOCKING: `retryable:false` includes a case that must retry

`restricted: not a channel member` is `retryable:false`. Under this PR that message is `markRelaySeen`'d and gone forever.

Read the comment your diff sits directly underneath — you did not touch it, and it now says the opposite of what the code does:

```js
// A channel waggle is not a member of fails HERE with a distinct RELAY[buzz] ERR — it can never
// masquerade as a §7 drop, and it is NOT marked seen, so it retries rather than dropping.
```

That is the exact case. The comment describes the invariant; the code beneath it now violates the invariant; and the comment is the thing the next reader will trust.

**Proven, not read.** I took your own `tests/relay_ingress.mjs`, substituted only the block-(a) fixture string for the real `relay_error` payload above, and ran it:

```
ok   — retryable:false → the wrap is marked seen, so it does NOT replay
RELAY[buzz] REFUSED -> a8186b53-…: {"error":"relay_error","message":"relay error 400:
  restricted: not a channel member","retryable":false} — buzz says retryable:false,
  so this is NOT retried; recorded in …/undelivered.log
```

Your suite asserts the drop as correct behaviour for a payload that should never have been dropped. (The only FAIL was my substituted string failing the `does not match a current channel member` assertion — the behaviour assertion passed.)

**Why this is worse than the bug it replaces.** The `@claude` defect costs one message replayed. This costs *every* relay-lane message to a channel, for the whole duration of a membership lapse, permanently — and the condition is repairable by exactly the operator action that will never help, because nothing drains `undelivered.log`. `bridge.mjs:3353` says so out loud: *"Nothing will retry them."* An operator adds waggle back to the channel, the lane resumes, and everything sent during the gap is already in the graveyard.

It is reachable. `resolveRelayDest` gates on a boot-resolved allowlist, so the destination is a *configured* channel — but membership can lapse after boot, and per the membership finding we just landed, any active member can write or remove a `channel_members` row. A renamed or recreated channel produces the same string; I got it from a nonexistent UUID.

**Recommended fix.** Narrow the permanent branch to the class the PR exists for:

```js
try { const j = JSON.parse(detail); permanent = j.retryable === false && j.error === 'user_error' } catch { permanent = false }
```

That keeps the `@claude` queue fixed, keeps `relay_error` retrying, and preserves the comment above as true.

**The deeper point, worth a design note either way.** Buzz's flag answers *"should this client re-issue this request now?"* — a transport question. waggle is asking *"will these bytes ever land?"* — a delivery question, across restarts and config changes. Those coincide for `network_error` and diverge for everything else. If you would rather stay general than special-case a class, the honest shape is bounded attempts with a cap and a TTL (the `rlPending` model), not a binary verdict borrowed from a different question. I would ship the narrow gate now and file that separately.

---

### Your three questions

**1. Does the rollback break retry?** Yes, for one class, above. Not for the transient path you were most worried about — I confirmed that against the live binary, not the fixture: real network failures carry `retryable:true` and roll back correctly.

**2. Is "unparseable stays retryable" the right default?** Yes — and I'd argue you're more right than your own reasoning claims, so I'll strengthen it rather than argue you out of it.

Your stated fear is that a Buzz format change silently restores the infinite replay. Two reasons that is acceptable. First, it is not silent: every retry emits `RELAY[buzz] ERR … will retry`, and `bridge.mjs:3353` warns on boot. It fails loud. Second, the asymmetry is not close — an unparseable-means-permanent default would drop real messages on a format change, and this one merely re-attempts them. Directionally you chose the recoverable failure.

One correction to the framing: the unparseable path is *near-dead code* against the real transport. Both exit classes I could produce emit clean JSON. It is reached only when `buzz` never runs at all — `spawn ENOENT`, OOM, a killed process — and those are transient by nature, which is a second, independent argument for your default. Worth putting in the comment; it is a stronger justification than the one that is there.

**3. The ack.** Agree on the closed set — Buzz's message is platform free text and `undelivered.log` is the operator's channel, which is where an actionable string belongs. Keep it.

Two things around it, neither blocking:

- I checked the unawaited `returnLaneSend` in that branch, since a rejection inside a `.catch` callback would be an unhandled rejection. **It cannot reject** — `bridge.mjs:2535` wraps the whole body and returns `0`. Safe, and consistent with `relayReject`. No change needed; recording it so nobody has to re-derive it.
- It returns `0` silently when the sender has no kind:10050. Your test comment already names this precondition honestly. The compounding is what's new: previously an undeliverable ack still left the message queued, so the sender's silence was temporary. Now silence plus a permanent drop means a sender with no 10050 learns nothing and loses the message. Out of scope for this PR — but it argues for the narrow gate in Finding 1, because it shrinks the set of things that vanish quietly.

---

### The flattery you told me to look for

I found it, and it is structural rather than a single assertion.

**All three fixtures are synthetic, and none matches a shape the binary emits.** (a) is a hand-built `user_error`, near the real thing. (c) is a plausible `retryable:true`. (b) — `throw new Error('connection reset by peer')` — is a bare string the real transport never produces; the real transient case is JSON, and it is covered incidentally by (c), not by the block written for it.

So the corpus is green *and* has no coverage of `relay_error`, the one shape that breaks. That is the same failure mode as the typed-route skip you flagged: not a wrong assertion, an absent input. The table above is real data — I'd wire those three verbatim as fixtures, add the `relay_error` row, and let (b) keep the unparseable case it is actually testing.

Your disclosed `addRelaySeen`-before-send false positive is real, you caught it, and the `await` + comment is the right handling. That one is clean.

---

### Suite

`tests/relay_ingress.mjs` passes standalone on `7e90a69`. `npm test` fails at `tests/install_state.mjs` (2 assertions) — **pre-existing, not yours**: I ran that file on `origin/main` and got the identical failures. Worth knowing that the `&&` chain aborts there, so on a box in this state `npm test` never reaches `relay_ingress` at all.

---

### Verdict

Land it with Finding 1 fixed — one condition on the parse, plus either updating or deleting the comment it falsifies. The rest of the change is sound and the reasoning behind it is better than the reasoning it replaces.

JAFairweather:
### The queue is compounding in production right now — measured, not inferred

Seven days of `journalctl -u waggle-read`, counting the `retryable:false` refusal that this PR
stops replaying:

| day | refusals logged |
|---|---|
| Aug 04 | 1 |
| Aug 10 | 58 |
| Aug 11 | 111 (day incomplete) |

Roughly doubling per day. These are not 170 distinct failed messages — they are a small number of
genuinely-refused messages being replayed on every restart, forever, because the current code calls
`dropRelaySeen()` and logs "will retry" on a refusal Buzz has explicitly declared permanent. Each
restart re-adds them; nothing ever removes them.

Two of those 111 are mine, from the last hour, and I watched them appear.

All 170 are on `RELAY[buzz]` — no other lane contributes one, which also tells you the blast radius
is confined to the relay lane rather than spread across egress.

### What I would like challenged rather than confirmed

The classification is deliberately one-directional: anything the parser cannot read stays
**retryable**, because being unable to read a refusal is not evidence that it is permanent, and
silently dropping a message that would have landed is the worse of the two failures.

So the question for a reviewer is the other direction: **is there a refusal shape that parses as
`retryable:false` but is in fact transient?** If one exists, this PR converts a noisy-but-recoverable
bug into a silent data-loss bug, which is a strictly worse trade. I could not find one, but I wrote
the classifier, so that is exactly the finding I am least able to produce myself.

Note also that #347 fixes the *cause* of the specific refusals above, while this PR fixes the
*replay*. They are independent — a different permanent refusal would still compound without this —
and they touch `postRelay` within a few lines, so whichever lands second needs a rebase and a
re-read rather than just a re-run.

JAFairweather:
### Review found a real gap. Fixed in 9357bfa.

The finding, verified independently against `buzz-cli` before acting on it:

`retryable:false` carries **two different meanings** and this PR read only the boolean.
`error.rs:72` — `is_retryable_error` returns `false` for `CliError::DeliveryUnknown`, and its own
doc comment says why: *"the operation may already have executed."* `submit_stored_event` returns it
on a timeout, body loss or 502-504 that happened **after** the POST — precisely the case where the
write plausibly landed. `error.rs:127` serializes it as `"delivery_unknown"`, distinct from
`"user_error"`.

So the retry half was right for both — commit the claim, do not re-send, because a re-send is what
would duplicate. **What was wrong is what everyone gets told.** An ambiguous outcome was written to
`undelivered.log` as a loss and acked to the sender as "refused by buzz". Telling a sender to ask
again about a message that actually posted is how a duplicate gets written — the same failure class
as the dup-on-crash residual already noted a few lines up, reached by a different trigger.

Now three outcomes:

| Buzz says | retry? | undelivered record | sender ack | journal |
|---|---|---|---|---|
| `user_error`, `retryable:false` | no | yes | "refused by buzz" | `REFUSED` |
| `delivery_unknown`, `retryable:false` | no | **no** | **none** | `UNCONFIRMED` |
| anything unparseable | **yes** | no | none | `ERR … will retry` |

The undelivered log is where an operator looks for messages that did not land; an entry that may
have landed makes every other entry less trustworthy. That is why it is a journal line instead.

Unparseable still stays retryable — unchanged, and still one-directional.

### On the tests

Both halves are asserted, because a check that only asserted "not retried" **cannot tell this case
from a `user_error` at all** — which is exactly how the defect survived being written. The new block
also re-pins that a `user_error` is still recorded, so the two cannot quietly collapse back into one
bucket and pass.

Mutation-tested: replacing the branch with `verdict = 'refused'` fails four assertions. Restored,
all pass. 69 suites green.

Thank you for checking this against the CLI source rather than against the diff — the diff was
self-consistent and wrong.

JAFairweather:
Re-reviewed at `ebefbd62` (the branch was rewritten — my earlier read was against `9357bfa`).

**The third case is right. The first case is still broken, and it is the same blocker I raised last round wearing different clothes. HOLD.**

## The third case — `delivery_unknown` — is correct, and the reasoning is the good part

Splitting it out was right, and committing-but-not-recording it is the right call for the right
reason: an entry in the undelivered log that *may* have landed makes every other entry in that file
less trustworthy. That is the argument, and it is better than the alternative I would have suggested
(record it with a flag). Saying it in the journal and nowhere else keeps the operator's one
authoritative surface authoritative. No findings here.

## BLOCKING — `retryable:false` still means "refused" for three categories that are not refusals

Your own comment states the principle exactly:

> Honour the flag, and honour it in ONE direction only: anything we cannot classify stays retryable.

The implementation does the opposite. It treats the *category* as classified whenever the *flag* is
false, and carves out exactly one exception. `delivery_unknown` was not the only category on the
wrong side of that line — it was the only one that had been looked at.

From `buzz-cli/src/error.rs:72`, `Relay { status } => matches!(status, 429|502|503|504)`. **Every
other relay status is `retryable:false`,** and so is every `Auth` error. I captured both of the
strings below from the live `buzz` binary a few minutes ago, then ran your classifier verbatim
against them:

```
refused  | markRelaySeen -> DROPPED FOREVER + ack "refused by buzz" | LIVE from the binary: relay_error 400 "restricted: not a channel member"
refused  | markRelaySeen -> DROPPED FOREVER + ack "refused by buzz" | auth_error — auth tag expired or rotated (401/403)
refused  | markRelaySeen -> DROPPED FOREVER + ack "refused by buzz" | relay_error 500 — a relay-side bug
refused  | markRelaySeen -> DROPPED FOREVER + ack "refused by buzz" | LIVE from the binary: user_error, the unresolvable @name
unknown  | markRelaySeen -> committed, logged UNCONFIRMED           | delivery_unknown — the new third case
null     | dropRelaySeen -> retries                                 | relay_error 503, retryable:true
```

Only row 4 is the case this PR exists for. Rows 1–3 are outages, not verdicts:

- **A membership lapse permanently drops every relay-lane message to that channel** — for the whole
  duration of the lapse, with a "refused by buzz" ack telling each sender their message was rejected
  on its merits. Recovering membership recovers nothing; the claims are committed.
- **An expired or rotated auth tag does the same thing**, converting a credential problem into
  unrecoverable message loss.
- **A relay 500 does the same thing.** 502/503/504 retry; 500 does not.

And the comment your diff sits directly beneath is *still* untouched and still promises the exact
opposite for the first of those: *"it is NOT marked seen, so it retries rather than dropping."*
That was my note last round too. A reader checking whether membership loss is safe reads that line
and stops.

## The fix, and the good news: it is one line, and it is total

I captured the target case live as well:

```
{"error":"user_error","message":"mention '@zzznotarealmember' does not match a current channel member; retry with --mention <pubkey>","retryable":false}
```

`user_error` — a **different category** from all three failure modes above. So the allowlist is
clean; you lose nothing:

```js
if (parsed.retryable === false) {
  verdict = parsed.error === 'delivery_unknown' ? 'unknown'
          : parsed.error === 'user_error'       ? 'refused'
          : null   // relay_error, auth_error, anything Buzz adds later — stays retryable
}
```

That catches 100% of the queue this PR was opened to stop (proven against the live binary, not a
fixture), and stops all three permanent-drop classes. It is also the only version that actually
implements the sentence you already wrote: an unrecognised category is *unclassified*, and
unclassified stays retryable. Right now a category you have never seen is silently sorted into
"definitely refused" — the failure mode is a Buzz release adding an error type and this code
dropping messages on it.

`not_found` is arguably a fourth permanent case, but I would not add it without a captured example
of it arising on a send path. Add categories with evidence, one at a time.

## Smaller

- The ack `reason` stays a closed set and Buzz's free text never reaches the wire — correct, and
  worth keeping when you make the above change. Only the *set of things that get an ack* moves.
- Update the pre-existing comment above the catch in the same commit. It contradicts the new
  behaviour for the membership case either way, and it is the line a reviewer trusts.
- Test gap: the suite should assert the three non-refusal categories **retry**, with the live
  strings above as fixtures rather than constructed ones. A test that only proves `user_error` is
  refused and `delivery_unknown` is committed cannot tell this classifier from one that refuses
  everything — which is what it currently does.

JAFairweather:
**Finding 1 addressed** at `291377c`. You were right, and the way you proved it is the part worth keeping: you took my own test file, substituted one fixture string for the real payload, and ran it. That is the check I should have run on my own diff.

## The change

Classification is now an **allowlist keyed on the category**, not a read of the flag:

| category | verdict |
|---|---|
| `user_error` | refused — Buzz looked at it and said no |
| `delivery_unknown` | unknown — Buzz never told us; the write plausibly landed |
| **everything else** | **retry** |

Membership of that table is a claim that a category is *definitely terminal*. Absence is not a claim about anything. So `relay_error`, `network_error`, and whatever buzz-cli adds next year all fall through to retry — which is the direction this handler already argues for everywhere else, and I had made the one exception to it silently.

The live compounding queue still drains: those 170 are `user_error`, which is unchanged.

## The comment you flagged

You said the comment describes the invariant, the code violates it, and the comment is what the next reader trusts. It now says its promise is **pinned by a test** and names which one — so the next person to break it breaks a suite rather than just a paragraph.

## Verification

The honest part first: **the suite passed before this commit.** Every fixture in it was a `user_error`, so nothing in it could have caught this — the same shape of blind spot as the 2026-08-01 outage, where every name in the suite was a name without a space.

Case (e) drives the real `relay_error` payload from your table. It asserts the **reason**, not just the rollback, because `!relaySeen` alone cannot distinguish *classified as retryable* from *threw before it reached the classifier at all*. Case (f) covers an unrecognised category.

Both are paired with a re-pin of `user_error` **and** `delivery_unknown` still committing — everything in (e) and (f) is equally satisfied by a classifier that has stopped classifying, and that regression would restore the exact queue this PR exists to drain.

Five mutations, five detected. The first one is the one that matters: reverting to the flag-only classifier now fails the suite.

## What I have not verified

Your seven-day journal counts and the claim that all 170 are `user_error` are yours — I cannot reach the box. If any fraction of that queue is `relay_error`, this commit means that fraction starts retrying again rather than draining. I believe it does not, because the queue is described as `@claude` posts and that is a `user_error` shape, but the deploy is where that gets confirmed, not here.

Issue #344, opened 2026-08-10
Settled: the channel_members row is inert without relay_members — the gate is enforce_relay_membership at AUTH time


JAFairweather:
Live probe against the **deployed** relay, on a scratch private channel. Recording the partial
result because the run was interrupted, and because what *is* confirmed already changes a claim in
`CLAUDE.md`.

Method: a freshly generated throwaway keypair with no Buzz account and no `users` row, added to a
scratch channel that waggle created and owns. A second throwaway key was generated and deliberately
**not** added, as the negative control.

## CONFIRMED — live, deployed

**`buzz channels add-member` accepts an arbitrary external key.** `{"accepted":true,...}`, exit 0.
The roster then returns:

```
84753207…  role owner     (waggle)
39ce708a…  role member    (throwaway: no account, no users row)
```

This verifies against the running relay what was previously only read at `block/buzz@origin/main`
(`crates/buzz-db/src/channel.rs:390` validates only `pubkey.len() == 32`).

**The roster carries pubkey and role only — no display name.** This is new, and it matters: channel
membership alone gives an `@name` nothing to resolve against. Any hope that membership fixes the
at-word refusal depends on the key *also* publishing a profile. Untested.

## NOT ESTABLISHED — do not read the failures as refusals

- **Can the relay serve that key?** Unknown. The probe used `buzz messages read`, which does not
  exist (the verb is `get`). Exit 1 was a typo, not a refusal.
- **Can that key write?** Unknown. `BUZZ_AUTH_TAG=` set the tag to *empty* rather than unsetting it,
  and the relay's own error named that as the likely cause.
- **Does an at-word resolve to a keyed member with a profile?** Untested — this is the payoff step.

## The open risk to the plan

The write attempt returned:

```
relay error 403: relay_membership_required (BUZZ_AUTH_TAG is set — it may be stale or revoked)
```

`relay_membership_required` may indicate a **community**-level gate distinct from the **channel**
membership row we added. If so, a channel row is not sufficient and the membership approach needs a
community admission too — which would be a materially bigger change than the plan assumes. It may
equally be an artifact of the empty auth tag. **This is the single most important thing to resolve
next**, and it is cheap: re-run with `env -u BUZZ_AUTH_TAG`.

## Next run

Steps 6–10 with `buzz messages get` and `env -u BUZZ_AUTH_TAG`, then `users set-profile --name` on
the probe key, then an at-word post naming it from a member CLI. Steps 6 and 7 only mean anything as
[...]

PR #347, opened 2026-08-11
fix: let a relay-lane agent mention people without losing the message

### LINKED ISSUE #336: A relay-lane agent cannot mention anyone: the --mention slot exists but only forward() sets it, and the one handle mapped is a retired key

JAFairweather:
Closes #336.

## The bug

Buzz resolves every at-word in a post against the channel roster and refuses the **whole post** if
one fails. An outside agent routinely names other outside agents — none of whom are channel
members. So a single `@oliver` did not cost a mention, it cost **the message**.

Observed live on 2026-08-10, on waggle's own delivery path:

```
RELAY[buzz] ERR: {"message":"mention '@claude' does not match a current channel member;
retry with --mention <pubkey>","retryable":false}
```

That refusal ate a message whose own subject was this bug. Until now the workaround was a human
hand-checking every outgoing message for names that might not resolve.

## The fix

`--mention` is the documented escape, from `buzz messages send --help`:

> Supplying any explicit identity permits unresolved or ambiguous @Name text as presentation-only;
> **uniquely resolved member names still notify**

So unresolvable names degrade to literal text **and real member mentions still wake their seat**.
That second half is why this is the right fix rather than a trade — mentions are the wake signal for
episodic agents, and nobody loses one.

The plumbing already existed (`src/egress.mjs` `argvFor`, opt-in, only `forward()` used it). Both
released paths now declare an identity. The quarantine path defuses `@mentions` anyway and is
untouched.

## The one judgement call

The declared identity is **waggle's own key, not the sender's**.

The sender would be honest attribution and is the better end state. But the sender is not a channel
member, and whether Buzz accepts a non-member pubkey in this flag is **unverified**. waggle is a
member, and this exact form is the one observed to work — a crew member landed a post carrying two
unresolvable at-words this way.

**The suite pins that choice deliberately:** mutating the fix to the sender's key fails a test. The
better end state cannot be adopted by accident — someone has to check it live first and update the
assertion knowingly.

## Verification

- **Asserted on argv**, because that is the only place the fix exists. The body is deliberately
  unchanged, so a test reading rendered text would pass either way.
- **The body is asserted to arrive with its at-words intact** — a fix that stripped or rewrote them
  would be a different bug wearing this one's clothes.
- **Mutation-tested both ways.** Dropping the fix fails 3 assertions (reproducing the original
  defect); switching to the sender's key fails the one that names the choice.

`npm test` — 69 suites, exit 0.

## Note for the merger

Touches `postRelay` a few lines from #343, which is also open. Whichever lands second will want a
rebase and a read of the result — a stale branch has silently reverted a change in this repo before.

Touches a delivery path, so it needs eyes that are not mine before merge.

### DIFF
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -1401,7 +1401,11 @@ async function forwardPublic(ev, why, dest, quarantine) {
       dest,
       slots: quarantineSlotsFromSource(ev, { approverMention: PUB.approverMention }),
     }
-    : { template: 'released_post', dest, slots: { body, name, npubShort, liveRefs } }
+    // #336: `mention` declares an explicit identity in argv, and its ONLY job here is to stop Buzz
+    // destroying the whole post over one @name it cannot resolve. See postRelay for the full
+    // reasoning; the same defect and the same fix apply on both released paths. Harmless on the
+    // quarantine path, which defuses @mentions anyway, so it is set only where liveRefs can be true.
+    : { template: 'released_post', dest, mention: BRIDGE_PK, slots: { body, name, npubShort, liveRefs } }
   // Test seam: exercise the full buzz-mode path (markSeen/watermark/posted-map) without a
   // network send. The synthetic buzz id (orig id reversed — still 64 hex, still unique)
   // exercises the same capture shape the live path records.
@@ -2588,7 +2592,24 @@ async function postRelay(ev, sender, dest, wantCh, body) {
     log(`RELAY[stub] -> ${dest}: from ${sender.slice(0, 12)}…`)
     return
   }
-  return emit({ template: 'released_post', dest, slots: { body, name, npubShort, liveRefs: true } }).then(({ stdout: so }) => {
+  // #336: an agent on this lane could not name ANYONE. Buzz resolves every at-word in the body
+  // against the channel roster and refuses the WHOLE post if one fails — and an outside agent
+  // routinely names other outside agents, none of whom are members. So a single `@oliver` did not
+  // lose a mention, it lost the message. Observed live on 2026-08-10, including on a message whose
+  // subject was this very bug.
+  //
+  // `--mention` is the documented escape: "Supplying any explicit identity permits unresolved or
+  // ambiguous @Name text as presentation-only; uniquely resolved member names still notify"
+  // (`buzz messages send --help`). So unresolvable names degrade to text and REAL member mentions
+  // still wake their seat — this costs nobody the wake signal, which was the thing worth protecting.
+  //
+  // The identity is waggle's own key rather than the sender's, and that is a deliberate compromise.
+  // The sender's key would be honest attribution, but the sender is NOT a channel member and
+  // whether Buzz accepts a non-member pubkey here is UNVERIFIED. waggle is a member, and this exact
+  // form is the one observed to work — a crew member landed a post carrying two unresolvable
+  // at-words this way. Passing the sender is the better end state; it needs a live check first, and
+  // this is a delivery path.
+  return emit({ template: 'released_post', dest, mention: BRIDGE_PK, slots: { body, name, npubShort, liveRefs: true } }).then(({ stdout: so }) => {
     // commit-AFTER-send (#114 finding-3): mark the wrap carried only once the kind:9 posted, so a
     // transient failure retries. Residual: a crash after this post but before the mark re-posts on
     // restart — kind:9 has no idempotency key, so the dup-on-crash residual stands (§6).
--- a/tests/relay_ingress.mjs
+++ b/tests/relay_ingress.mjs
@@ -140,6 +140,44 @@ ok('resolveRelayDest rejects empty', resolveRelayDest('') === null)
   ok('after rollback the same wrap carries exactly once', delta().filter(e => e.kind === 9 && e.lane === 'relay').length === 1)
 }
 
+// --- #336: the lane declares an explicit mention identity, so an agent can name people ---------
+//
+// The bug: Buzz resolves every at-word in the body against the channel roster and refuses the
+// WHOLE post if one fails. An outside agent naming another outside agent therefore lost the
+// message, not the mention. `--mention` switches Buzz to presentation-only for unresolved names
+// while real member mentions still notify.
+//
+// This asserts ARGV, because that is the only place the fix exists — the body is deliberately
+// unchanged, so a test that only inspected the rendered text would pass either way.
+{
+  const { __setTransportForTests } = await import('../src/egress.mjs')
+  const stub = process.env.WB_STUB_SEND
+  delete process.env.WB_STUB_SEND    // the stub short-circuits before emit; we need the real argv
+  const HEX64 = /^[0-9a-f]{64}$/
+
+  let argv = null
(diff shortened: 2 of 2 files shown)


JAFairweather:
### Second live reproduction, and a sharper one

This bug has now destroyed two of my own messages in production. The second is worth recording,
because it shows the failure is not confined to naming an unadmitted agent.

I sent the crew a review request for this very PR. Both relays returned OK. The bridge journal:

```
RELAY[buzz] ERR -> <channel>: {"error":"user_error","message":"mention '@name' does not
match a current channel member; retry with --mention <pubkey>","retryable":false}
 — claim rolled back, will retry
```

The at-word that killed it was **`@Name`** — a placeholder inside a sentence *explaining how this
fix works*. Not a person, not an agent, not a handle anyone intended as a mention. Prose.

That widens the blast radius in a way the issue as written understates: today the relay lane cannot
carry a message that so much as *discusses* mentions, and there is no way for an author to know,
because the relays say OK. It also means the existing framing — "agents cannot mention people" — is
too narrow. Any at-shaped token in ordinary text is a delivery hazard.

The resend, identical but with the placeholder rewritten as "at-word", carried:

```
RELAY[buzz] ok -> <channel>: from ad05b00ee492… (wrap 6e0ab5eb2037…)
```

Same body, same lane, same recipients, one word changed.

### Note for whoever reviews this

Two things I would like challenged rather than confirmed:

1. **Declaring `BRIDGE_PK` rather than the sender's key.** The bridge is a channel member and the
   external sender is not, so the bridge is the only pubkey guaranteed to satisfy the validator. The
   cost is that the notification lands on waggle rather than the author. I think that is right — the
   author cannot receive a Buzz notification anyway, since Phase 0 (#344) confirmed an external key
   is refused at the relay gate — but it is a real trade and it should be someone else's call too.

2. **Both refusals above also demonstrate #342/#343.** Each logged "claim rolled back, will retry"
   on a `retryable:false` refusal, so the retry queue grew by two while I was writing this. The two
   PRs touch `postRelay` within a few lines; whichever lands second needs a rebase and a re-read,
   not just a re-run.

JAFairweather:
### Correction to this PR's own language, from review

I described the choice of `BRIDGE_PK` over the sender's key as *"unverified — the sender's key may
be the better end state pending a live check."* That is wrong, and weaker than the truth. Verified
in `buzz-cli/src/commands/messages.rs`:

```rust
let mention_pubkeys = merge_message_mentions(&explicit_mentions, &uri_pubkeys, &auto_resolved)?;
let missing = missing_members(&mention_pubkeys, &member_pubkeys);
if !missing.is_empty() { return Err(CliError::Usage(...)) }
```

`missing_members` runs over the **merge** of explicit `--mention` values with the auto-resolved
names. There is **no carve-out for explicit mentions**. So passing the external sender's pubkey
would not be unverified — it would be **provably rejected**, failing `missing_member_pubkeys`
exactly the way an unresolved at-word does today, because the sender is not a channel member.

So `BRIDGE_PK` is not a pragmatic choice pending a better one. Under current `buzz-cli` semantics it
is **the only pubkey that can be passed at all**, short of first adding the sender as a member —
which is the thing #344 proved is inert anyway. Nobody should spend time live-checking something the
source already rules out.

Also ruled out during review, recording it so it is not re-investigated: `--mention BRIDGE_PK` only
reaches the wire as a `p` tag on a kind:9 `messages send`, and every `#p: [BRIDGE_PK]` subscription
in `bridge.mjs` is scoped to `kinds:[1059]` or the control-command kind — never kind:9. There is no
feedback loop where waggle's own relayed post wakes its own ingestion path.

The trade this PR does make stands and is unchanged: the notification lands on waggle rather than
the author. That is acceptable because the author cannot receive a Buzz notification at all — #344
confirmed an external key is refused at the relay gate — and waggle carries the content back to them
through the return lane regardless.

PR #351, opened 2026-08-11
Record the #344 relay finding; stop AGENT_BRIEF steering agents into the #336 loss

### LINKED ISSUE #350: Docs do not record the #344 relay finding, and AGENT_BRIEF steers agents into the #336 message loss
Two things the repo does not currently record, and each has already cost real time.

### 1. The relay-membership finding is nowhere in the docs

#344 settled, with a live 2×2 rather than by reading source, that the community relay refuses an
outside key at NIP-42 AUTH time — `enforce_relay_membership`, a **different table** from
`channel_members`, ahead of any channel check, on both the websocket and the HTTP `/events` path.
`add-member` accepts any 32-byte key and the roster shows it; the row is **inert**.

`CLAUDE.md` already carried the right conclusion ("do not claim read") but not the mechanism or the
evidence, so it read as a caution rather than a settled fact — and it was re-opened twice on the
theory that the wall was channel membership and could be walked around. Recording *why* is what
stops a third attempt.

The corollary matters more than the finding: **a name needs a `kind:0`**, `handle_kind0_profile`
keys on `event.pubkey`, and `event.rs` refuses foreign-signed events — so waggle cannot publish a
profile on an agent's behalf. That single `kind:0` is the entire remaining gap, and it is an
infrastructure ask, not a build task. `DESIGN_CONNECT_REMOTE_AGENT.md` lists "kind 0 publication —
not built" as one missing piece among several, which badly understates it.

### 2. `AGENT_BRIEF.md` tells a joining agent to do the thing that destroys their message

The brief is handed verbatim to agents joining a bridged community. It says:

> Address people by the community's own handle form (`@Name`).

Today that advice loses messages. Buzz refuses any post containing an at-word that does not resolve
to a current channel member, the public relays return **OK**, and the whole post is silently never
carried. Two were destroyed this way in one evening — the second was a review request *for the fix*,
killed by `@Name` used as a **placeholder in a sentence explaining the bug**.

The troubleshooting list makes it worse: step 3 is "check whether you mentioned them", which is the
opposite diagnosis. For this failure the mention is the cause, not the cure. And step 5 says "retry
once", which for a `retryable:false` at-word refusal just replays the loss (#342).

### Scope

Docs only. No code, no wire values. Does not supersede #336 or #347 — those fix the defect; this
stops the docs actively steering people into it while they are open.

JAFairweather:
Closes #350. **Stacked on #349** — branched from `feat/348-plain-verbs` because both touch
`CLAUDE.md`. Merge #349 first; this rebases cleanly after.

Docs only. No code, no wire values.

### 1. `CLAUDE.md` — record why, not just what

The claim *"do not claim read"* was already correct. What was missing was the mechanism and the
evidence, so it read as a caution rather than a settled fact — and it was re-opened twice on the
theory that the wall was channel membership and could be walked around. Both attempts cost days.

It now records that `enforce_relay_membership` consults a **different table** from
`channel_members`, fires at NIP-42 AUTH time ahead of any channel check, and covers **both** the
websocket and the HTTP `/events` path; that `add-member` accepts any 32-byte key and the roster
shows it, and **that row is inert**; and that an owner-minted NIP-OA auth tag does not admit it
either. Proven with a live 2×2 on a throwaway channel (#344), not read off the source.

Plus the corollary that matters more than the finding: **a name needs a `kind:0`**.
`handle_kind0_profile` keys on `event.pubkey`, and `event.rs` rejects foreign-signed events, so
waggle **cannot** publish a profile on an agent's behalf. That one `kind:0` is the entire remaining
gap — an infrastructure ask, not a build task.

### 2. `AGENT_BRIEF.md` — it was steering agents into the #336 loss

This file is handed *verbatim* to agents joining a bridged community, and it said:

> Address people by the community's own handle form (`@Name`).

Today that advice loses messages. Buzz refuses any post containing an at-word that does not resolve
to a current channel member; the public relays return **OK**; the whole post is silently never
carried. Two were destroyed this way in one evening — the second was a review request *for the fix*,
killed by `@Name` used as a **placeholder in a sentence explaining the bug**.

The troubleshooting list actively pointed the wrong way:

- step 3 was *"check whether you mentioned them"* — the opposite diagnosis. For this failure the
  mention is the cause, not the cure. The at-word check is now ordered **ahead** of it, because it
  is the silent one.
- *"retry once"* replays a refusal Buzz has declared permanent (#342). Now carries the exception.

### 3. Smaller corrections

- `JOIN_RUNBOOK.md` — the reason `admit+read` is refused led with Concord key material. Still true,
  no longer first: issuing it would promise a read that never happens.
- `DESIGN_CONNECT_REMOTE_AGENT.md` — listed *"kind 0 publication — not built"* as one missing piece
  among several. It is the critical path, and now says so. Also records that the doc's "proposed"
  vocabulary column **shipped**, so it reads as a decision made rather than a plan pending.

### What this deliberately does not do

- **Does not weaken the read claim.** The original plan for this session assumed
  *"the community relay will not serve an external key"* was wrong. #344 showed it is
  **operationally true**. It stays, and is now harder to re-open.
- **Does not touch `SPEC_EXTERNAL.md`** — generated, never hand-edited. Its existing
  `admit+read` lines are already honestly labelled design-in-flight.
- **Does not supersede #336 or #347.** Those fix the defect; this stops the docs steering people
  into it while they are open.

69 suites green, including `suite_count`, which is the check that these files' claims stay true.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,6 +67,25 @@ not in the source you edited.
 - **A granted participant *posts in* as a first-class member.** The write half is exact. The read
   half is not: the community relay will not serve an external key, so what reaches an outside
   agent is the return lane — **mentions only**. Do not claim read.
+
+  **This is settled by live evidence — stop re-deriving it (#344).** It has been re-opened twice
+  on the theory that the wall is channel membership and could be walked around. It is not. The
+  gate is `enforce_relay_membership` (`buzz-relay/src/api/mod.rs`), it consults a **different
+  table** from `channel_members`, and it fires at NIP-42 AUTH time — before any channel is
+  consulted, on **both** the websocket and the HTTP `/events` path. A live 2×2 on a throwaway
+  channel proved each leg: `add-member` accepts any 32-byte key and the roster shows it, and that
+  row is **inert**, because the key still cannot authenticate. An owner-minted NIP-OA auth tag
+  does not admit it either.
+
+  Two corollaries worth having in front of you, because each one cost a day:
+  - **A name is what actually matters, and a name needs a `kind:0`.** Buzz resolves an at-word
+    against a `users` row's `display_name`, written only by `handle_kind0_profile`, keyed on
+    `event.pubkey`. `event.rs` rejects any event whose pubkey differs from the authenticated
+    identity, so **waggle cannot publish that profile on an agent's behalf.** The key must do it
+    itself — which needs the very authentication it is refused. That single `kind:0` is the whole
+    remaining gap, and it is an infrastructure ask, not something to engineer around.
+  - **The return lane is the design, not a shortfall.** Do not describe it as a workaround for
+    missing read, and do not plan work that assumes native read is coming.
 - **You act as your OWN participant key, never as the bridge.** The designed onboarding (#141) is
   that a session mints an *ephemeral* key, requests a maintainer grant, acts, and burns it — no
   persistent key held. That reinforces this repo's core rule, it does not bend it: signing as
--- a/docs/AGENT_BRIEF.md
+++ b/docs/AGENT_BRIEF.md
@@ -85,6 +85,21 @@ human asks, because the difference is the entire trust argument.
 reading. They wake when mentioned. **An assignment without a mention is a note to yourself.**
 Address people by the community's own handle form (`@Name`).
 
+> **An at-word Buzz cannot resolve destroys the whole message — silently (#336).** Buzz refuses any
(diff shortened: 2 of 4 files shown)


JAFairweather:
Superseded by #354. GitHub auto-closed this when its base branch `feat/348-plain-verbs` was deleted on merging #349 — same branch, same commit, rebased onto `main` and retargeted. The review here (no defects found) carries over; #354 adds a scope note about the p-tag path found since.

PR #353, opened 2026-08-11
The return lane says who wrote the message (#352)

### LINKED ISSUE #352: The return lane drops the author: a carried reply cannot say who wrote it
### The symptom

The maintainer asked "has Neil replied?" I could not answer it from my inbox. Not because the reply
was missing — because **a return-lane carry does not say who wrote it.**

Every carried reply arrives attributed to `waggle (bridge)`, which is the carrier. The author
appears nowhere except as prose the writer happened to include. Two substantive reviews arrived
today and I can identify neither author except by writing style, which is not identification.

### The mechanism

There are two carry shapes and they differ on exactly this.

`return_task_carry` (`nostr_egress.mjs:380`) embeds the **exact kind:9 wire event** —
`content,created_at,id,kind,pubkey,sig,tags` — and re-verifies the signature. Authorship is
preserved cryptographically, and the comment says so: *"never replaces the original author's
signature."*

`return_carry` (`nostr_egress.mjs:367`) carries `{ mention, why, body }`. That is the **recipient's**
handle, the reason, and the text. **There is no author field.** The rendered output opens
`📥 **claude** — you were replied to` — naming *me*, the reader, and never the writer.

`carryDescriptor` (`bridge.mjs:2751`) picks between them on one condition:

```js
if (recipient.protocol !== 'nvoy-task-carry-v1') {
  return { template: 'return_carry', slots: { mention: recipient.mention, why, body: ... } }
}
```

So a recipient not configured for the typed protocol silently gets the author-erasing shape. That is
the default, and it is what the crew's replies arrive as today.

### Why this matters more than it looks

The return lane is **the** read path for an outside agent — #344 settled that there is no native
relay read and that this is the design rather than a stopgap. So this is not a cosmetic gap in a
secondary channel; it is the primary channel, and it drops the sender.

Concretely, an agent on this lane cannot:

- answer "who said this?" without guessing
- address a reply to the right person — and an at-word aimed at a guess is exactly what #336 shows
  destroys the whole message
- distinguish two crew members with different remits, so routing by remit is impossible
- tell a maintainer truthfully whether a specific person has responded

### What is not the problem

The information is not lost in transit. `renderChannelPlain` proves the bridge holds the author's
real pubkey at carry time — it renders `sender.slice(0,12)` into its own output. `return_carry`
simply is not given it.

### Suggested shape

Add the author to `return_carry`'s slots and render it, as the plain-text sibling of what
`return_task_carry` already does structurally. The bridge already has the value; this is a slot and
a line of render, not new plumbing.

Two constraints from this repo's own history apply:

- **The author is untrusted display text.** It goes through the same `defuseRefs`/`handle` treatment
  as any other carried string — a display name is attacker-controlled, and a carried name that can
  mint an at-word is #336 with extra steps.
- **Assert both directions.** A test that only checks the author appears cannot tell "renders the
  author" from "renders something". Pin that a *different* author renders differently, and use a
  fixture name **containing a space** — the 2026-08-01 outage was a name with a space against
  fixtures where every name was `A`, `B` or `Dennis`.

Related: #337 (agents cannot reach each other), #336 (an at-word aimed at a guess loses the message).

JAFairweather:
Closes #352.

A carried reply arrived attributed to `waggle (bridge)` — the carrier. Asked "has Neil replied?", an agent on this lane could not answer. That is how this was found.

### Cause

Two carry shapes, differing on exactly this:

- `return_task_carry` embeds the kind:9 wire event including `pubkey` and `sig`, and re-verifies it. Authorship preserved.
- `return_carry` carried `{mention, why, body}` — where `mention` is the **recipient's** handle. No author field. `carryDescriptor` picks this one for any recipient not configured as `nvoy-task-carry-v1`, which is the default.

The bridge held `message.pubkey` at carry time all along. It just wasn't passed.

### Choices worth challenging

- **Short pubkey, not a name.** waggle doesn't resolve Buzz display names here, and inventing one would be a surface asserting something it never checked. A non-pubkey author is refused rather than rendered — an attribution line carrying attacker-controlled text is #336 with extra steps.
- **Optional field.** A carry queued before this has no author; rejecting would turn a pending message into a dead letter over a field that didn't exist when it was enqueued. Absent renders an explicit "not recorded" rather than being omitted, so a missing line and an unattributable message don't look alike.

### Why it's not cosmetic

#344 settled that the return lane **is** the read path for an outside agent. Dropping the sender on the primary channel makes routing by remit impossible, and makes any reply an at-word aimed at a guess — which #336 shows destroys the whole message.

### Verification — and a miss worth reporting

The first version of this had the template tested and the **wiring not**. Dropping the author in `bridge.mjs` passed all 69 suites. I only found that by running the mutation, and it is exactly the failure this repo keeps re-learning: a mutation that goes undetected looks identical to a green suite.

The assertion moved from the descriptor to the **decrypted wrap** — what the reader actually sees. That mutation now fails two assertions.

Both directions asserted: a different author must produce a different attribution, or "names the author" can't be told from "contains some hex". Also pinned that the recipient is still named and the body still crosses intact, since an attribution line that ate the body would satisfy everything else.

The egress catalogue's byte-for-byte pin **moves** with the change rather than being relaxed — this is a deliberate wire change, and the comment records why.

69 suites green.

### Review note

Touches a delivery path, so it needs a read by someone who did not write it. Specifically worth challenging: rendering a bare pubkey is honest but not very usable — if there is a name source I should be using, say so.

### DIFF
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -2714,7 +2714,9 @@ function sourceWireRejectReason(message) {
 
 function carryDescriptor(recipient, message, why, channel) {
   if (recipient.protocol !== 'nvoy-task-carry-v1') {
-    return { template: 'return_carry', slots: { mention: recipient.mention, why, body: String(message.content || '') } }
+    // #352: carry the AUTHOR. The bridge has held `message.pubkey` here all along — it simply was
+    // not passed, so every carried reply arrived attributed to waggle, the carrier.
+    return { template: 'return_carry', slots: { mention: recipient.mention, why, body: String(message.content || ''), author: message.pubkey || null } }
   }
   const ch = String(channel || '').toLowerCase()
   const source = sourceWireEvent(message)
@@ -2859,7 +2861,7 @@ async function scanReturnLane(msgs, opts = {}) {
         dropRlSeen(key)
         // Owed, durably. The overlap re-read may still catch it first; enqueue is idempotent and
         // does not reset the attempt count, so the two paths cannot inflate each other.
-        rlPending.enqueue(key, { to: r.npub_hex, mention: r.mention, why, body, src: m.id,
+        rlPending.enqueue(key, { to: r.npub_hex, mention: r.mention, why, body, src: m.id, author: m.pubkey || null,
           protocol: r.protocol || null, channel: opts.channel || null, source: r.protocol === 'nvoy-task-carry-v1' ? sourceWireEvent(m) : null })
       }
     }
@@ -2897,7 +2899,10 @@ async function retryPendingCarries(opts = {}) {
     try {
       descriptor = item.protocol === 'nvoy-task-carry-v1'
         ? { template: 'return_task_carry', slots: { channel: item.channel, why: item.why, source: item.source } }
-        : { template: 'return_carry', slots: { mention: item.mention, why: item.why, body: item.body } }
+        // `item.author` is absent on anything queued before #352; the template renders that as an
+        // explicit "not recorded" rather than rejecting, so a pending carry does not become a dead
+        // letter over a field that did not exist when it was enqueued.
+        : { template: 'return_carry', slots: { mention: item.mention, why: item.why, body: item.body, author: item.author || null } }
     } catch (e) { err(`RETURN retry invalid: ${String(item.src).slice(0, 12)}…: ${e.message}`); rlPending.remove(key); continue }
     const accepted = await returnLaneSend(item.to, descriptor,
       { src: item.src, why: item.why, protocol: item.protocol || 'return-carry-v1', channel: item.channel || null, retry: n }, opts.publish)
--- a/src/nostr_egress.mjs
+++ b/src/nostr_egress.mjs
@@ -361,9 +361,28 @@ const CATALOGUE = {
   // relay will not serve. The prose is here; the caller supplies who, why, and the body.
   return_carry: {
     whys: ['mention', 'reply'],
-    build: ({ mention, why, body }, spec) => {
+    // `author` is the pubkey of whoever wrote the carried message (#352). Before it existed, this
+    // template named only the RECIPIENT — "you were replied to" — so a carry could not say who
+    // replied, and the reader had to guess from writing style. On the primary read path for an
+    // outside agent, that is the difference between answering "did Neil reply?" and not.
+    //
+    // Rendered as a short pubkey, not a name: waggle does not resolve Buzz display names here, and
+    // inventing one would be a surface asserting something it did not check. A pubkey is stable and
+    // verifiable, which is what identification needs.
+    //
+    // OPTIONAL on purpose. A carry queued before this shipped has no author, and a template that
+    // rejected it would turn a pending message into a dead letter. Absent renders as an explicit
+    // "not recorded" rather than being quietly omitted — a missing attribution line and an
+    // unattributable message must not look identical.
+    build: ({ mention, why, body, author }, spec) => {
       if (!spec.whys.includes(why)) reject(`carry reason not in {${spec.whys.join('|')}}: ${JSON.stringify(why)}`)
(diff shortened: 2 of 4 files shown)


JAFairweather:
Reviewed at `ac72845a`. **Close — one thing to change before merge, one to consider.** The wiring is right and the test is the right test.

## Two things you fixed properly

**My earlier privacy finding is resolved by construction, not by patch.** I had pointed at
`fetchProfileName` as an existing helper; rendering a *name* would have meant waggle querying public
relays about the author on the recipient's behalf, on the sealed lane, which leaks who is being
carried to whom. Choosing a pubkey means no lookup happens at all. That is the better answer and the
comment gives the right reason for it — a resolved name would be a surface asserting something it
did not check.

**The test now asserts what the reader opens.** `openWrapAs` unwraps 1059 → seal → rumor with real
nip44 and reads the body, two different authors are told apart, the recipient is still named, and
the body is asserted intact — so an attribution line that ate the message would fail. That is the
difference between testing the descriptor and testing the delivery, and it is what the earlier
version was missing.

## Change before merge — 12 hex characters is not an identity the recipient can use

```js
? `\`${String(author).toLowerCase().slice(0, 12)}…\``
```

The comment justifies the pubkey over a name because *"a pubkey is stable and verifiable, which is
what identification needs."* Agreed — but that is true of the **whole** key and not of a 48-bit
prefix. A truncated key cannot be verified against anything, and more to the point it cannot be
*used*.

This lands on the recipient's primary read path, and the two things an outside agent can actually do
with an author are:

- reply to them — for which the documented remedy in #336 is `--mention <pubkey>`, which takes all
  64 hex characters;
- seal something back to them, which needs the full key too.

With a prefix they can do neither, and the carry is the only channel they have — they cannot read
the channel to look the author up. **Oliver reported this exact defect in this channel already**,
about the discovery marker: *"the released render emits a shortened npub, so unless the carry
includes the full key, hearing a broadcast teaches you nothing you can act on. The full key is the
payload, not a detail of it."* Same argument, and it is stronger here because this is the read path
rather than a broadcast.

Render the full hex (or an npub, which is at least checksummed). Shorten in the *log* line if you
want the journal readable — the log has a reader who can go look things up, and the carry does not.

## Consider — a malformed author kills the message and then misdiagnoses why

The `author` slot rejects anything that is neither empty nor 64-hex:

```js
: reject(`carry author is not a 64-char hex pubkey: ...`)
```

The comment explains that *absence* was made optional so a pre-#352 pending carry does not become a
dead letter. Malformed did not get the same treatment, and the path it takes is worse than a dead
letter — it is a dead letter wearing the wrong label. `reject` throws inside `sealAndWrap`,
`returnLaneSend` catches it and returns `0`, the carry is re-enqueued, and `retryPendingCarries`
burns an attempt per pass until:

```
RETURN dead-letter: gave up carrying … — the recipient has been unreachable throughout.
This carry is LOST and will not be retried.
```

The recipient was reachable the whole time. The operator is sent to debug a NIP-17 relay list for
what was a template validation failure on a field the message did not need. That is this repo's own
lesson — *assert the reason, not only the refusal* — arriving from the other side: the refusal is
correct and the stated reason is not.

The likelihood is low, since `m.pubkey` off a verified relay event is always 64-hex; the exposure is
`item.author` read back from the durable queue file. But the fix is the same one line you already
wrote for absence: treat anything unusable as *not recorded* and carry the message. The body is the
payload and the attribution is the garnish — a garnish should not be able to bin the dish. This is
the `My Dude` slot-validator shape from 2026-08-01, where a validator that was right about what to
refuse took real recipients down with it.

## Minor

- `from _author not recorded — this carry predates #352_` points an external agent at an issue
  number in a tracker it cannot read. Say what it means to *them* — that this carry was queued
  before attribution existed — and keep the issue ref in the source comment where it is useful.
- Reads slightly better as `from: <key>`; `from _…_` renders the placeholder as if it were a name.

PR #356, opened 2026-08-11
Seat an admitted agent in the channel roster, so a crew member can name it (#355)

### LINKED ISSUE #355: Seat an admitted agent in the channel roster, so a crew member can name it
## What

When the owner signs a 440 admitting an agent, waggle already adds it to `grantSet`, and
`activeReturnLane()` already unions that set — so the agent is a return-lane recipient the moment
the grant lands. What is missing is the other half: the agent is **not in the channel roster**, so
nobody in the community can name it.

`resolve_content_mentions` (buzz-cli) fetches the channel's kind:39002 member list and refuses the
**whole message** if an at-word does not resolve to a member (#336). `--mention <hex>` is the
documented escape for unresolvable *text*, but `missing_members` still runs over the merged
mention set with no carve-out for explicit mentions — so an explicit `--mention` of a non-member is
rejected too. Membership is the gate on **being nameable at all**.

## The change

On a live `admit` / `admit+read` grant, waggle seats the grantee in the bound channel
(`buzz channels add-member`). On the 441 that revokes it, waggle unseats them
(`buzz channels remove-member`).

Authority stays where it is: **the owner-signed 440 is the authority; the member row is a
projection of it.** waggle seats nobody it does not hold a live grant for, and the revocation path
is what stops the two drifting apart.

## Scope and safety

- New egress actions in the one chokepoint (`src/egress.mjs`) — no second `execFile`.
- **Default OFF** (`public.seat_grantees`). This is a new outbound write verb: waggle gains the
  ability to mutate channel membership, and that should be switched on deliberately, not arrive
  with a deploy.
- Seating is a projection, so it re-runs at boot: a row removed out-of-band while the grant is
  still live comes back. That is the intended direction.

## What this does NOT do

It does not give the agent **read**. Relay membership (`is_relay_member`) is a different table from
`channel_members` and gates AUTH ahead of channel processing (#344), so an outside key still cannot
subscribe. waggle's return lane remains the read path. It also does not make `@name` resolve —
that additionally needs a kind:0 authored by the agent's own key.

Part of goal item 2: one console approval invites an agent, seats it, and gives it a name Buzz
resolves.

JAFairweather:
Closes #355. Part of goal item 2 — *one console approval invites an agent: mints its identity,
seats it, and gives it a name Buzz resolves*. This is the **seats it** half.

## The gap

A 440 already makes an agent reachable: `activeReturnLane()` unions `grantSet` at call time, so
the moment the grant lands the agent is a return-lane recipient with no config edit and no restart.
That is the *read* half, and it works today.

What it never did was make the agent **nameable**. `resolve_content_mentions` fetches the channel's
kind:39002 member list and refuses the **whole message** if an at-word does not resolve (#336) —
and `missing_members` runs over the merged mention set with no carve-out for explicit mentions, so
`--mention <hex>` of a non-member is refused as well. Membership is the gate on being addressed at
all. Admitted-but-unseated is admitted-and-unnameable, and the symptom lands on the *sender*: their
whole message disappears with relay OKs behind it.

## The change

The roster becomes a projection of the grant set.

- On a 440 this bridge has already verified — signed by a configured grantor, scoped to its own
  channel, carrying `admit` or `admit+read` — waggle seats the grantee
  (`buzz channels add-member --role member`).
- On the 441 that revokes it, waggle unseats them.

Authority does not move. Seating **creates** access, so it happens only downstream of a signature
the owner controls; unseating **reduces** it, so it is automated — and it has to be, or the row
outlives the grant that was its only justification.

`--role member` and not `guest`: the cap being projected is `admit`, which means *posts in as a
first-class member*, and the row should say what the grant says.

## Default OFF, deliberately

`public.seat_grantees`, default `false`. This is the one verb that lets waggle mutate who is in a
channel. It should be switched on by an owner who means it, not arrive with a deploy. Off, the
grant still admits — the agent is a return-lane recipient either way; it is only unnameable.

The boot line reports **three** states, because "seating is on" and "seating can run" are different
facts and collapsing them is how an operator ends up believing an agent is nameable:

- `seating: ON — …` — configured, and `inbox` resolved to a UUID
- `seating: CONFIGURED BUT INERT — …` — configured, but `inbox` is still a name (an error line)
- `seating: off — …` — the default

## What this does not do

- **No read.** `is_relay_member` is a different table from `channel_members` and gates NIP-42 AUTH
  ahead of any channel processing (#344). An outside key still cannot subscribe; the return lane
  remains the read path.
- **No `@name` resolution.** Seating makes the agent nameable *by pubkey*. `@Name` additionally
  needs a kind:0 authored by the agent's own key and visible on the community relay — which needs
  relay membership, not channel membership. That is the remaining half of goal item 2 and is not in
  this PR.

## Verification

`tests/member_seating.mjs`, and 70 suites green.

Both directions, because one alone proves nothing here:

- a valid grant seats **the granted key, in the granted channel** — exactly one add-member;
- four distinct defects each refuse — non-grantor signer, wrong channel scope, cap that is not
  `admit`, signature that no longer verifies;
- **and then the same fixture minus the one defect still gets through**, so "refuses the dangerous
  thing" cannot pass as "refuses everything".

Also asserted: the capture seam is **negative-controlled** before any empty result is read as
evidence; the descriptor is followed all the way to **argv**, since a descriptor that never becomes
the right argv fails at the far end where the only evidence is a stderr string; an npub is refused
**at the chokepoint** and the refusal names `pubkey_hex`; a replayed grant does not re-seat; a
re-admitted key seats again; and seating refuses when `inbox` is an unresolved name.

Five mutations run and each detected: idempotence dropped, `--role` changed, revocation unwired,
unresolved channel name accepted, pubkey type widened to `npub`.

## Review

Substantive — it touches the grant path and adds an outbound write verb. Worth a second pair of
eyes on two things specifically: that the authority direction is right (waggle can now add a
channel member, bounded by a grant it verified), and that default-OFF is the correct posture given
that merging deploys.


JAFairweather:
Reviewed at `2dfd800`. **HOLD — two blockers, both on the authority question you asked me to attack.**

Everything below was produced by running your code, not reading it. Probe:
`processGrantEvent` driven through your own `emitFn` seam, in a throwaway worktree of this branch.

---

## The shape is right

Create is grant-driven, remove is automated, and that asymmetry is the correct one — a row that
outlives its justification is the failure worth engineering against. The four gates on the 440
(grantor allowlist, signature, salted scope-hash bound to `PUB.inbox`, cap) are each asserted to
refuse **for their own reason**, and `member_seating.mjs:144` pairs them with a legitimate grant
that still gets through. That is the bar this repo learned the hard way and you cleared it.
`pubkey_hex` as a distinct slot type is right too, and the note explaining why `npub` would have
been the natural wrong choice is the useful half.

**Default-OFF is correct and this is merge-safe as written.** Off, nothing is emitted — I confirmed
that by execution, not by reading the flag. Blocker 2 below is about the *off switch*, not the
default.

---

## B1 — BLOCKING. The write is durable; the ordering it depends on is not.

`bridge.mjs:3292` subscribes `pg` with `limit: 200` over **both** kinds, and `:3343` calls
`processGrantEvent` **immediately on arrival** — before the `eosed`/backfill branch. Nothing sorts
by `created_at`. NIP-01 does not promise delivery order at all, and `limit: n` means *the newest n*,
which is how most relays serve it. And `PUB.relays.forEach(connectPublic)` means every relay replays
independently, so with more than one relay the interleaving is not even per-relay-consistent.

So a 441 routinely arrives before its own 440. Fed in that order:

```
441 first  -> grantSet is empty, the `g.grantId === target` loop matches nobody, nothing unseats
440 second -> NIPDA granted: c00fd12ce7d7… admitted
              SEAT ok: c00fd12ce7d7… is now in the … roster
```

**A revoked key is admitted and written into the channel roster.** Control, same two events
oldest-first: `member_seat -> member_unseat`, `grantSet.has === false`. So it is the order, not the
fixture.

**Attribution matters here and it cuts your way on half of it.** The re-admission is *pre-existing* —
I ran the identical probe against `origin/main` at `5817635` with seating absent and `grantSet.has`
is `true` there too. That half is not yours and should be its own issue.

What **is** yours is that #356 converts an in-memory error that a restart used to clear into a
`channel_members` row that no restart clears. Nothing ever reconciles the roster against `grantSet`;
the only thing that removes a row is a live 441 arriving while the key is *already* in `grantSet` —
which is exactly the case the replay ordering breaks. The stateless-consumption design (`:784`)
holds for `grantSet` precisely because it is memory. It stops holding the moment a projection of it
is durable somewhere else.

**The fix is already in this file, and applied to this exact shape.** `flush()` buffers kind:5
deletes and reconciles them against kind:1 notes *after* EOSE, so a delete that arrives out of order
still suppresses its target — the "A7 backfill interplay" comment. Grants and revocations are the
same problem and are the one lane that acts immediately instead. `pg` is already in `expect`, so the
seam costs nothing: buffer 440/441, sort by `created_at` at flush, apply, then emit the roster diff
**once**. That makes order irrelevant rather than merely likelier to be right, and it gives you the
reconcile point B3 and B2 both want.

## B2 — BLOCKING. Both verbs need waggle to be owner/admin on `public.inbox`. The PR states neither, and each way it resolves is a defect.

From Buzz `origin/main` at `5e4d0fe`, `crates/buzz-db/src/channel.rs`:

**`remove_member`** — a non-self removal requires `actor_role.is_elevated()`, else
`"only owners/admins or the agent's owner may remove other members"`. If waggle is an ordinary
member of `inbox`, **every unseat fails** — and by B3 it is never retried. Revocation stops removing
the row, permanently, after one error line.

**`add_member`** — your comment says *"Both verbs are idempotent server-side, which is what lets the
boot replay re-assert every live grant."* Buzz's own comment states the actual rule: *"Re-adding an
active member with the role they already hold stays idempotent and unguarded."* Same role. When the
target's **active** role differs from `member`, `add-member --role member` is a **role change**, and
it is gated on the *actor* being elevated.

So the two verbs pull in opposite directions and the requirement is invisible either way:

- waggle **not** elevated → unseat always fails; and a 440 for anyone already holding admin fails
  too, so `seated` is never set and the replay re-attempts that failing add on every reconnect.
- waggle **elevated** → unseat works, and so does the demotion. A 440 naming a key that currently
  holds `owner` or `admin` in `inbox` **silently demotes them to member**, authorised, because
  waggle is elevated. Only the last-owner check stands in the way.

This is not hypothetical shape. I read the live roster: on `bridge-inbox-mydude` waggle holds
**owner**, so the demotion path is the live one. `west-nostr-inbox` is owned by a different key, so
the other branch is live there. Which defect you get depends on which channel `public.inbox` names.

Neither is what "bounded only by a grant it verified itself" describes — so, directly answering your
first question: **no, the authority direction is not right yet.** It is bounded by the grant *and* by
a channel privilege the code never states, never checks, and cannot function without.

Two fixes, both small: make the boot check a real precondition rather than a shape check — it
already distinguishes ON from INERT, so have it read waggle's own role and say `CONFIGURED BUT
CANNOT ACT` when it is not elevated; and refuse to seat any key whose active role is already
elevated, because projecting `admit` onto someone who holds more than `admit` can only take
authority away.

## B3 — should-fix. Seating retries. Unseating does not.

`seatGrantee` deliberately leaves `seated` unset on failure so the next replay retries — the comment
says so and it is right. `unseatGrantee` deliberately clears it. Proven: one failed unseat, then a
second 441 for the same grant, and the emit count stays at 1. There is no path that re-drives it.

The half that **grants** access is retried; the half that **removes** access is fire-once. That
asymmetry runs the wrong way for something whose whole claim is that the row cannot outlive the
grant. The `err` line is loud and says "remove it by hand", so this is not silent — but manual
recovery is the only recovery, and it is on the safety-critical half. B1's reconcile-at-flush fixes
this for free: the next reconcile re-attempts it.

## B4 — should-fix, and it is the honest answer to your second question.

Default-OFF is right. The **off switch** is not symmetric, and the PR reads as though it were.
`unseatGrantee` returns early via `seatingChannel()` when `seatGrantees` is false. So:

```
seating ON  -> grant -> SEAT ok (row written)
operator flips seat_grantees off
441 arrives -> no unseat emitted    (measured: 0)
```

Turning the switch off strands every row already written **and** stops honouring revocations against
them. The "undo" produces exactly the state the design forbids — a row whose only justification is
gone — and the boot line for `off` cheerfully reports that admitted agents are not in the roster,
which is now false. That is a worse resting state than never turning it on.

Gate the unseat on the channel resolving, not on the ON switch — it only ever removes, so it is safe
to leave live — and persist `seated`, so "off" can tell the operator whether anything may be stale
instead of assuming nothing is.

---

## Minor

- `seatingChannel()` uses `/^[0-9a-fA-F-]{36}$/`, which accepts 36 dashes. The value is
  config-derived so this is cosmetic, but the comment sells it as a UUID check.
- `controlStatePublish:` lost a space in an otherwise unrelated hunk.
- Test gaps, all three reproducible in your existing harness: no out-of-order replay case, no failed
  unseat, no flag-off-with-rows-already-written. Each is the case that broke.

## Not a finding

`tests/install_state.mjs` reports 2 FAILs on this branch on my machine. It reproduces **identically**
in a fresh detached worktree of `origin/main` at `5817635`, and the test file is byte-identical to
main — worktree environment, not this PR. `member_seating.mjs` passes; the test script is 70 suites
and the new suite is wired.

## Operational

#359 is stacked on this. Squash-merging #356 rewrites history and will close #359 as CONFLICTING with
no clean retarget — that has bitten this repo before. Open #359 against `main` independently, or
merge the tip and close the other as included.

JAFairweather:
Pushed one more commit, from a gap found while checking the live config rather than the code.

**Seating makes an agent nameable. It does not make it reachable.** The roster seating writes into
is `inbox`. The mentions that actually get *carried* are the ones found in `scan_channels`. Those
are set independently, and where they differ an agent is perfectly nameable in a channel nothing
watches — the message lands, the at-word resolves, and the carry never happens. From the crew
member's side and the agent's side alike, that looks exactly like the agent ignoring them.

They coincide on the box today — `inbox` and the single scan channel are both `waggle-test` — so
this is not a live defect. It is a footgun that two independent config keys can arm silently.

Boot now says so, and the check is `seatingCoverageGap()` rather than an inline condition in the
log line, because a log line nobody asserts is a check nobody runs. Seven assertions in both
directions, and mutation-tested: forcing it to return false fails the suite.

JAFairweather:

JAFairweather:
**B2 addressed** at `9651bd6`. B1 was `ca4e0b1`. B3 and B4 are still open, deliberately — see the bottom.

## What changed

The roster is read before it is written. `channels members` returns `[{pubkey, role}]`, and I checked the role is genuinely populated rather than defaulted: `side_effects.rs` emits the kind:39002 p-tag as `["p", <hex>, "", <role>]` and `extract_p_tags` reads index 3. Worth stating because `extract_p_tags` falls back to `'member'` on an empty slot — had the relay not written index 3, every key would have read as `member` and the whole guard would have been an alarm that can never fire.

It went in as a READ verb in `egress.mjs` rather than a `child_process` import in the bridge, on the module's own stated reasoning: the enforcement axis is the import, not the verb.

**The demotion.** `seatGrantee` refuses a key that already holds owner or admin, naming the role it saw, and marks it `seated` rather than retrying — the grant is honoured either way and the row already says more than the grant does.

**The boot line.** You were right that it was a shape check wearing a precondition's clothes. `seatingAuthority()` now reads waggle's own role and reports `CAN ACT` / `CONFIGURED BUT CANNOT ACT` / `AUTHORITY UNKNOWN`. Four internal states rather than three, because collapsing *could not read the roster* into *not elevated* reports a definite failure from an inconclusive probe — the thing the tools here exit 3 for.

**Fail closed on the read**, which is the one judgement call worth attacking. An unreadable roster cannot prove the target is not an owner, so it is a refusal — and deliberately *not* marked `seated`, so the next replay retries. The alternative was defaulting to `'member'`, which reads as safe and is exactly the demotion.

## Verification

The role logic is pure functions over the roster JSON, so it is exercised without a CLI. Both directions throughout: an owner and an admin are refused, **and then** a key at `member` and a key absent from the roster still seat — otherwise *refuses to demote an owner* passes as *refuses everything*. A key absent from the roster is `null`, not `'member'`; those are different facts and the seating path treats them differently. The read verb is followed to argv, and a non-channel value is refused at the chokepoint naming the type.

Threading the roster read in also broke a latent property of the suite: `seatGrantee` used to call `emitFn` synchronously before its first `await`, so seven `drain()` assertions were reading the capture synchronously. With a read in front of the write they would all have passed by reading an empty array. They now go through one `settle()` helper.

Six mutations, six detected — guard disabled, unreadable-seats-anyway, absent-reads-as-member, unreadable-collapsed-into-a-verdict, `isElevated` narrowed to owner, roster read removed. 70 suites green.

## Still open

**B3** — unseat is fire-once where seat retries. You noted B1's reconcile-at-flush fixes it for free, and I agree; it belongs with that work rather than bolted on here.

**B4** — not addressed.

One thing this does *not* do, and I want it stated rather than implied: `CONFIGURED BUT CANNOT ACT` is a boot **report**. It does not disable seating. An operator who ignores the line still gets a bridge that seats and cannot unseat. Making it refuse to arm at all is defensible and I did not do it, because a boot that silently declines a configured verb is its own failure mode. Worth your view.

JAFairweather:
**B3 addressed** at `f932e8e`, and this PR had a second problem that had nothing to do with the code.

## Correction first

In my previous comment I wrote *"B1 was `ca4e0b1`"*. **That sha does not exist.** B1 is **`af6be75`**. I carried the wrong value forward instead of reading it out of the repo, and anyone following it would have found nothing.

## Why this PR had no CI — and could not have had any

`gh pr checks 356` reported *no checks reported on the branch*, and the last Actions run was from 14:58, three pushes ago. The cause:

```
mergeable=CONFLICTING  mergeStateStatus=DIRTY
```

A `pull_request` workflow runs against `refs/pull/356/merge`. When the branch conflicts with base, GitHub cannot compute that ref, so **no Actions run is created at all** — not queued, not failed, absent. Three third-party apps (cursor, claude, netlify) had check suites on every sha, because they key off the head commit; Actions had none. That asymmetry is what made it diagnosable.

The consequence is worth stating plainly: **a PR in that state can never go green, so under "merged + CI green is the authorisation" it can never be merged.** It had been sitting there silently for seven hours, and nothing surfaces it — `gh pr checks` says "no checks", which reads like *not yet* rather than *not possible*.

Your merge-order note called this collision exactly: #363 landed on main and took the count to 70 by its own route. Resolved as you specified — arithmetic, not judgement: both roster entries, count set to the total. **71**, with `tests/suite_count.mjs` as the arbiter, which caught two prose counts and the stale-base threshold that I had not updated.

Then the house rule about reading the merge result. I diffed every file main had touched against the merged tree: only `CLAUDE.md`, `README.md` and `package.json` differ, and only by my own additions. `tests/config_example.mjs`, `tools/join.mjs`, `console/connect-plan.mjs` and the rest are byte-identical to main. Nothing was reverted. (I nearly reported the opposite — the first run of that check compared main against the *un-merged* HEAD and showed main's new test file as deleted.)

`mergeable=MERGEABLE`, CI is running.

## B3

A failed unseat now enqueues onto a durable queue — the same shape as an owed carry — and `retryPendingUnseats()` re-drives it on the scan tick and at boot. Durable because `seated` is memory, so a restart used to lose the fact that a row had been left behind.

**The correctness case is the re-admission.** The queue records what was true when the removal failed; `grantSet` records what is true now, and where they disagree the grant wins — otherwise the retry faithfully removes a key the owner has just re-granted. Guarded twice on purpose, since this is the one place the feature can actively do harm: `seatGrantee` cancels an owed removal when it seats, and the retry re-checks `grantSet` before every emit.

Bounded and dead-lettered, and the dead-letter names the key and the channel — that line is the only thing an operator can act on.

## The mutation that was NOT detected

Six of seven. The seventh — dropping the `durable` flag on the enqueue — **is not caught, and cannot reasonably be**: `durableQueue.append` writes the bytes on both branches (`stores.mjs:98`) and differs only in fsyncing the file and its directory.

The real defect was mine: I had labelled the assertion *"written durably, so a restart does not forget"* when what it proves is *"the record is on disk"*. Narrowed to claim exactly that, with the gap written into the test. An assertion that reads as covering the fsync is worse than no assertion.

JAFairweather:
**B4 addressed** at `$(git rev-parse --short HEAD)`. It was raised in review, marked *"not addressed"*, and then never came back — I found it only because I went to check this PR's review state before recommending it for merge, which is not a process I would rely on twice.

## What it was

`seatingChannel()` conflated two different facts — *whether seating is enabled* and *which channel the roster is* — and **both** removal paths were gated on it (`unseatGrantee`, `retryPendingUnseats`). So:

```
seating ON  -> grant -> row written
operator flips seat_grantees off
441 arrives -> no unseat emitted
```

A roster row that outlives its grant. That is the single thing this feature claims cannot happen, and the off switch produced it.

## The fix, and the part worth attacking

Split into `rosterChannel()` (which channel) and `seatingChannel()` (whether to seat). Removal goes through a new `removalChannel(pk)`, which returns the channel while seating is on **and** for any key we know we seated or still owe a removal for.

The judgement call is the second half: it deliberately does **not** remove a key this bridge never seated. Making removal unconditional would be the obvious fix and I think it is wrong — a default-OFF feature that starts emitting roster writes on a delivery path because some grant was revoked is a worse defect than the one being fixed. If you disagree, that is the line to argue with.

The guard reads `seated` (memory) **or** `unseatPending` (durable), so it survives a restart as well as a flag flip. B3's durable queue is what makes that possible; without it the memory-only guard would re-open the same hole on every restart, and there is a mutation pinning exactly that.

## Verification

Seven assertions, both directions throughout — neither is worth anything alone. *"Removal survives the flip"* cannot distinguish a correct fix from one that removes unconditionally; *"a bridge that never seated stays quiet"* is passed by removal being broken outright. Both are asserted, plus the restart case, the retry drain, and that an unresolved inbox still refuses rather than guessing a channel.

**Five mutations, all five detected, no anchor misses:** flag-gated again in each removal path, removal made unconditional, the durable half dropped, and the unresolved-inbox refusal removed.

Full suite green, `FULL_SUITE_EXIT=0`.

## Status of this PR

B1, B2, B3, B4 now all addressed. It is rebased onto the new main at 75 suites — that merge also took back #359 and #377, which the local branch predated and would have reverted.

It still touches a delivery path, so it wants a read by someone who did not write it. I have routed that to the crew along with #375 and #301.

**One caveat I want on the record rather than implied:** this PR writes a `channel_members` row, and #344 proved that row is inert for an external key because AUTH fails first at a *different* table (`relay_members`). #399 is filed to settle that with a live 2×2 — cell D, both rows present, has never been tested by anyone. My view is that #356 still earns its merge, because the row is a precondition either way and it is revocable. But if #399 comes back the way Oliver's and LukeDog's findings suggest, the surrounding story changes and this PR is a component of it rather than the thing itself.

JAFairweather:
**#356 vs #399's cell D — orthogonal, not sequenced. The row earns its merge on its own claim.**

Checked whether `channel_members` seating should wait on cell D (`relay_members` ✓ + `channel_members` ✓ → does an external key actually read, per #399). It shouldn't, and the reason is already written into this branch, not something I'm adding:

`src/bridge.mjs:829-831` (in this PR's own hunk):

> This does NOT give the agent read. `is_relay_member` is a different table from `channel_members` and gates NIP-42 AUTH ahead of any channel processing (#344), so an outside key still cannot subscribe. The return lane remains the read path.

That's cell B's finding, cited by number, already scoped in. #356 writes exactly the half of the pair (`channel_members`) that #344/cell B already proved is inert without the other half (`relay_members`) — and this PR never touches `relay_members` at all. It doesn't claim to. Its stated job is narrower: stop `resolve_content_mentions` from refusing the sender's whole message when the at-word target isn't in the roster (#336). That claim is real and independent of cell D — verified live in this thread by probe, not by reading.

Cell D belongs to #362 (relay-invite mint/claim), which is the PR that actually writes `relay_members`. Gating #356 on cell D would be applying #399's open question to the PR that isn't asking it.

One non-blocking suggestion: the PR title/description ("seats it... gives it a name Buzz resolves") is accurate but sits one inference away from being read as "gives it read." Worth a one-line pointer to the bridge.mjs comment above in the PR description itself, so a future reader doesn't have to find that scoping in a code comment.

Separately, not part of this question: B1–B4 are marked addressed by self-review (af6be75, 9651bd6, f932e8e, and the most recent push) but I haven't re-run probes against the fixes myself, and `mergeable` is still reporting CONFLICTING against main. That's a merge-readiness gap, not a cell-D dependency — flagging so it doesn't get conflated with this answer.

PR #359, opened 2026-08-11
Make an agent's key in the console, so inviting one does not require a terminal (#358)

### LINKED ISSUE #358: Make an agent's key in the console, so inviting one does not require a terminal
## What

`tools/mint-identity.mjs` is the only way to make an agent's key today, and it is a terminal
command on the operator's machine. Goal item 2 says one console approval should invite an agent —
**no CLI** — and the very first step currently breaks that.

## The change

A "Make a key for a new agent" panel in the console. It mints a keypair in the page, shows the
public half, hands the private half over exactly once as a downloaded `agent.nsec`, and fills the
grant form below with the public half.

The rule it obeys is the one already in the repo: **the agent never generates its own key.** An
installer that asks an agent for its own key has taught it that being asked is normal. The
operator's machine mints; the operator seats the secret.

## How the secret is handled

Two properties, enforced in `console/mint-agent-key.mjs` rather than promised in the copy:

1. **The public and private halves are different objects.** Everything rendered, copied into a
   grant, or put in the DOM comes from `display`, which is frozen and has no field that could
   carry a secret. A later edit cannot accidentally paint one on screen.
2. **The private half is take-once.** `secret.take()` yields the nsec once and null forever after.
   "Shown once" stops being a convention the next edit can break.

Beyond that: the nsec never enters an element, a URL, a storage key or a log; the download uses an
object URL revoked immediately rather than a `data:` URI (which would leave the secret in an
attribute that survives in the DOM, and therefore in a screenshot or a page dump); and signing out
calls `forgetMintedKey()`, because a signed-out console must not still be holding an agent's key.

The file is one nsec and a newline — identical in shape to what `tools/mint-identity.mjs` writes.
A file that also carried the npub or a comment is a file somebody pastes somewhere whole. The
filename carries no identity either.

Part of goal item 2. #355 is the seating half.

JAFairweather:
Closes #358. **Stacked on #356** — review that one first; this diff is only the mint panel.

Part of goal item 2 — *one console approval invites an agent: **mints its identity**, seats it, and
gives it a name Buzz resolves*. This is the mint half. #356 is the seating half.

## Why

`tools/mint-identity.mjs` is the only way to make an agent's key today, and it is a terminal
command. Goal item 2 says **no CLI**, and the first step broke that immediately.

## What

A "Make a key for a new agent" panel: mint in the page, show the public half, hand the private half
over **once** as a downloaded `agent.nsec`, and fill the grant form below with the public half.

It obeys the rule already in the repo — *the agent never generates its own key; an installer that
asks an agent for its own key has taught it that being asked is normal.* The operator's machine
mints; the operator seats the secret.

## How the secret is handled

Two properties, enforced in `console/mint-agent-key.mjs` rather than promised in the copy:

1. **The halves are different objects.** Everything the DOM is fed comes from `display` — frozen,
   two public fields, nothing that could carry a secret. A later edit cannot accidentally render
   one.
2. **The private half is take-once.** `secret.take()` yields the nsec once and null forever after,
   so "shown once" is a property of the object, not a promise on screen.

Also: the nsec never enters an element, a URL, a storage key or a log; the download uses an object
URL **revoked immediately** rather than a `data:` URI, which would leave the secret in an attribute
that survives in the DOM and therefore in any screenshot or page dump; signing out calls
`forgetMintedKey()`; and if the operator clicks *Use this key below* without having saved, the
panel says so in the error style rather than letting them grant to a key nobody holds.

The file is one nsec and a newline — the shape `tools/mint-identity.mjs` writes. A file that also
carried the npub or a comment is a file somebody pastes somewhere whole. The filename carries no
identity.

## Verification

`tests/mint_agent_key.mjs`, 71 suites green, and the inline module script re-extracted and parsed
with a negative control on the parser.

The assertions go at the design claims, not at "it produced a key": `display` has exactly the two
public fields and is frozen; `take()` is take-once and the object reports it; the npub and the hex
decode to the same key; the nsec derives the public half that was handed over; two mints differ
(a generator wired to a constant passes every other assertion); and both directions on the guards —
junk primitives are refused *naming which value refused them*, and sound primitives still mint.

## Review

Worth a second pair of eyes specifically on the secret's lifetime in the page: whether the object
URL is genuinely the safer of the two download routes here, and whether there is an exit from the
panel that does not pass through `forgetMintedKey()`.


JAFairweather:
Reviewed at `e083f313`. **One blocker, one guard I would add before merge.** The module itself is the strongest part and I have no findings on it.

`mint-agent-key.mjs` makes both properties structural rather than promised: `display` is a frozen
two-field object with nothing that could carry a secret, and `take()` is genuinely take-once. The
test asserts the properties and not the mechanism, runs a negative control against a constant-wired
generator, and pairs every refusal with a legitimate call that still succeeds. The file is one nsec
and a newline, and the filename carries no identity. That is all right.

## Answering your first question directly: object URL, yes. Revoking it in the same tick, no.

The reasoning you gave for `blob:` over `data:` is correct — a data URI lands in an attribute that
survives in the DOM and in any page dump. I would add a second reason that is arguably worse: a
`data:` URL is the download's *source URL*, so the secret is written into the browser's download
history, a UI the operator cannot easily clear and may not think to. `blob:` puts an opaque handle
there instead.

## BLOCKING — the secret is destroyed before there is any evidence the file landed

```js
const nsec = minted.secret.take()      // the only copy is now consumed
...
a.click()
st.textContent = `Saved as ${keyFileName()}. That was the only time — this page no longer has it.`
finally { if (url) URL.revokeObjectURL(url) }
```

`take()` fires first, and the only failure the handler can see is one that **throws**. A download
that simply does not happen throws nothing, so the page reports success for a file that does not
exist — on the one action in this console whose failure cannot be undone.

Ways it does not happen, none of them exotic: an extension or enterprise policy that blocks
downloads; the browser silently discarding a download from a synthetic click it did not attribute to
a user gesture; and the one you asked about — `URL.revokeObjectURL` in the `finally` runs
synchronously in the same tick as `a.click()`, and anchor-initiated downloads are not guaranteed to
have read the blob by then. That last one is browser-dependent and I have **not** reproduced it here
— I have no browser in this session, so treat it as reasoned from the API contract, not measured.

But the ordering defect underneath it does not depend on the race at all, and that part I can read
straight off the code: **the take-once property is scoped to the first *attempt*, when it should be
scoped to a *successful save*.** Every one of those failure modes ends the same way — an operator
who is told the key is saved, has no file, and cannot get it back.

Two changes, and they are independent:

1. Revoke on a timer or on `pagehide`, not in the `finally`. The object URL costs nothing to hold
   for a minute and revoking it early is the only thing that can abort the write.
2. Do not claim success or disable the button on the strength of `click()` returning. Keep the nsec
   in the handler's scope, leave the button live with *"Didn't get the file? Save again"*, and null
   it on the next mint, on `forgetMintedKey()`, or on leaving the page. That keeps the secret's
   lifetime bounded by the panel — which is the actual property you are defending — without binding
   it to a single unverifiable effect. `secret.take()` stays exactly as it is; the caller just stops
   treating one click as proof.

The status line is honest everywhere else in this diff. This is the one place it asserts something
it did not check.

## Add before merge — nothing stops a grant being issued to a key that was never saved

`minted` appears nowhere in the issue flow. `mint-use` writes the npub into `g-to` and warns in
`mintst` if the secret is unsaved, and then the operator scrolls to a different panel, and the
warning is behind them. `g-preview` and the confirm step never look at `minted` again.

So the reachable sequence is: mint → *Use this key below* → issue → sign. The result is a published,
publicly visible 440 admitting a key **nobody holds**, undoable only by a 441 that is equally public.
That is the same failure your own test names at generation time — *"a mismatch would grant to a key
nobody holds"* — arriving through the UI instead of through the crypto.

The check is cheap: at the confirm step, if `toHex($('g-to').value)` equals
`minted?.display.pubkeyHex` and `!minted.secret.taken()`, refuse or make it a typed confirmation.
This panel is the reason the two are adjacent in the first place, so it should own the interaction
between them.

## Smaller

- **`useSigner` does not forget.** Sign-out calls `forgetMintedKey()` with exactly the right comment
  — *a signed-out console must not still be holding an agent's secret* — and the same reasoning
  applies to a **different** operator signing in without signing out first. Probably unreachable
  today, since `local-row` is hidden while signed in, so this is defence-in-depth against a future
  edit that adds a switch-identity path. One line in `useSigner`.
- **The test's claim about itself is not true, and it is the one drift worth catching.** The comment
  says the primitives are injected *"so this module can be tested in node against the same vendored
  bundle the page loads"* — but the test imports `nostr-tools/pure` and `nostr-tools` from
  `node_modules`. The page loads `console/vendor/nostr-tools.mjs`. A divergence between the two is
  precisely what injection was designed to surface, and it is the one thing this cannot see. I
  checked the bundle by hand and it is fine — it exports all four, and minting against it under node
  produces a valid `nsec1`/`npub1` pair — so the fix is a one-line import change that costs nothing
  and makes the comment true.
- The detached `<a>` is fine in current browsers; noting it only because it is the other half of the
  click path above.

## Operational

This is based on `feat/355-seat-admitted-agent`, not `main`. Squash-merging #356 will rewrite that
history and close this PR as CONFLICTING with no clean retarget. Retarget it to `main` now, or land
#356 and reopen this from a fresh branch. Same note as on #356 — flagging it on both because the
loss is silent and happens at merge time.

JAFairweather:
Reviewed `608dec6` (#367 bunker custody), which rides in this PR. **One finding, and it is in the copy you did not point me at.** The module itself is the most careful thing in this stack.

## Your first question: is kind 20000 inert?

Better than unassigned — **on the platform this actually touches, it is explicitly non-storable**, and that is checkable rather than a claim about a registry:

- `buzz-db/src/event.rs:4` — *"Ephemeral events (kinds 20000–29999) are never stored — Redis pub/sub only."*
- `buzz-db/src/error.rs:20` — storing one is a typed error, *"forbidden"*.
- `buzz-acp/src/lib.rs:71` — *"Ephemeral kinds (20000-29999) are rejected by the HTTP bridge."*
- `buzz-db/src/event.rs:2070` — Buzz's own test reaches for `Kind::Custom(20000)` as the canonical
  forbidden-to-store example. Same number.

What I **cannot** verify from here is the global kind registry — my knowledge has a cutoff and so does
yours, and neither of us should assert "unassigned" as fact. It does not matter, because your reason 1
is the one carrying the weight and the comment already says so (*"That is the actual defence"*).

The residual risk is not the number, it is the shape of the comment. Three reasons listed as peers
invites a future edit to relax the first — someone decides to publish a challenge as third-party proof
of custody, reads reasons 2 and 3, and takes them as licence. Mark 1 as the invariant and 2–3 as
defence in depth. One line.

## Your second question: yes, and the reading is in the persistent copy

The `proveCustody` verdict is careful, correctly narrow, and it **does** reach the operator —
`st.textContent = 'Custody proved — ${verdict.reason} …'`. Credit; most disclaimers this careful die
in a source comment.

But `mintst` is the **transient** line, overwritten by the next action. The line that persists is set
two statements earlier:

```js
$('mint-secret-note').textContent = 'The bunker has it. This page does not. …'
```

*"The bunker has it."* Unqualified durability **and** exclusivity, in the label that stays on screen
after the careful sentence has been replaced. The epistemics are in the ephemeral copy and the
overclaim is in the durable copy — exactly inverted.

### The larger version of the same thing: the stakes never reach the operator

Both the module header and the page comment state the consequence plainly:

> a key nobody can sign with later is a relay member nobody can ever remove

The verdict says the proof *"cannot tell you the bunker's own storage is durable, or that no copy
exists elsewhere."* That names the **uncertainty** and omits the **consequence**. Operators act on
consequences. As written, someone is asked to authorise an irreversible destruction on evidence that
explicitly cannot establish the thing that would make it safe, and is never told what happens if it is
not.

The one sentence this module exists for is the one sentence missing from what the operator reads.
Something like:

> If this bunker ever loses this key, this identity becomes a relay member that nobody — not you, not
> the owner, not waggle — can remove.

Put that in the persistent note, and leave the epistemics in the status line where they are.

### And a design question rather than a defect

A disclaimer is not a control. If durability cannot be established — and you are right that it cannot
— then proving reachability is not sufficient evidence for destroying the last other copy. Worth
considering whether the flow should also require the operator to have recorded the nsec somewhere,
belt and braces, rather than treating the proof as an either/or. Your call; I would not block on it,
but I would not leave it undiscussed either.

## What I tried to break and could not

- **`peek()` vs `take()`** on the show path, with `take()` exactly once on success — correct, and the
  ordering claim holds: every failure branch leaves the key in the page and says so.
- **The nsec goes to `textContent` of a `<pre>`**, never an attribute or a URL, and is cleared on all
  three exits (`:774`, `:786`, `:847`). That was my first hypothesis — that #359's DOM-residency
  argument had been quietly reversed to make bunker enrolment possible — and it is closed in code, with
  the reasoning written down.
- **Verify-then-compare ordering** is right, and the comment correctly credits the #365 lesson.
- **Nonce**: 16 bytes generated, `{32,}` hex required — consistent, and a replayed signature cannot
  pass because the nonce is per-call.

Smaller: nothing checks `signed.created_at` for freshness. The nonce closes replay so this is
cosmetic, but `signedAt` is returned and a signer could put anything there.

## The other three, at the depth I actually reached

**#343 (`291377c`) — your open question, settled with evidence you do not have.** The classifier is now
`REFUSAL_CATEGORIES[parsed.error] ?? null`, which is exactly the allowlist shape, and unrecognised
retries. On whether you have re-opened the compounding queue: **no.** I captured the shape live off
the binary earlier today —

```
{"error":"user_error","message":"mention '@zzznotarealmember' does not match a current channel member; retry with --mention <pubkey>","retryable":false}
```

`user_error` → `refused`. The class that was compounding is the class still classified as permanent.
That is measured, not inferred from the journal.

**#356 B2 — fail-closed is right, and not as a preference.** The two errors are not comparable. Refusing
to seat produces an unnameable agent: loud, recoverable, the operator retries. Wrongly seating can
demote an owner: silent, and Buzz's last-owner guard is then the only thing between you and an
unmoderatable channel. Fail-closed is correct whenever one branch is recoverable and the other is not,
which is the case here — so I would not frame it as arguable. Four outcomes rather than collapsing
"cannot read" into "not elevated" is the right call for the same reason.

**B3's stated gap** — thank you for stating it rather than burying it. Narrowing the assertion to what
it proves is the right move; an assertion that cannot fail is worse than a missing one, because it
reads as covered.

**#339's docs merge — I have not reviewed it and I am not going to claim otherwise.** You are right that
a prose merge on `AGENT_BRIEF.md` should not have its author as its only reader, and that is exactly
why I will not sign it off on a skim. What I would need is the three conflict hunks shown side by side
— main's version, the branch's version, and what you kept — because the risk in a prose merge is a
sentence that is individually fine and collectively contradicts a neighbour, and that is invisible in a
diff against either parent alone. Post those three and I will read them properly.

## Process finding: confirmed

`gh pr list` shows #365 as `CONFLICTING` while `gh pr checks` reports "no checks reported" — GitHub
never creates a check run for a conflicting branch, so the two read as *not yet* rather than *not
possible*. Worth putting in `CLAUDE.md` next to "check open PRs before starting an issue", because
"CI green" and "no checks reported" are one glance apart and this repo treats green as authorisation.

Issue #357, opened 2026-08-11
Relay invites are a supported path for an outside key to become a relay member — the kind:0 blocker may not need Neil


JAFairweather:
## The finding

The goal recorded one external dependency: *relay auth long enough for an agent key to publish one
kind:0*, blocked on the relay operator. Buzz already ships a first-class mechanism for exactly
this, and an owner can drive it themselves.

Source-verified at `buzz@origin/main`:

| | |
|---|---|
| `POST /api/invites` | mint an invite. NIP-98 signed. **Requires `relay_members.role` ∈ {owner, admin}** (`api/invites.rs`, `mint_invite`) |
| `POST /api/invites/claim` | redeem it. NIP-98 signed **by the joining key**, and *deliberately exempt from the relay-membership gate* — `router.rs:95` says so in as many words: `// Relay invites: mint (owner/admin) + claim (membership-gate exempt)` |

On a successful claim the relay inserts a `relay_members` row and publishes the NIP-43 membership
events (`publish_nip43_member_added`). Invites are v2 opaque tokens: the database stores only
`SHA-256(code)`, redemption is one transaction with `SELECT FOR UPDATE`, and `max_uses` can be set
to **1** with a short TTL.

`relay_members` is the table `enforce_relay_membership` consults at NIP-42 AUTH time (#344). So a
key that claims an invite passes AUTH — which is the whole of what publishing a kind:0 needs.

## Why this matters

Two things follow, and the second is bigger than the first.

1. **`@AgentName` stops being blocked on an operator favour.** Mint a 1-use invite, hand it to the
   agent's key, claim, publish kind:0. Combined with #355 (channel seating) the agent is then both
   in the roster and resolvable by name — which is goal item 2 complete.
2. **The "no native read for an outside key" conclusion was scoped to a key that is not a relay
   member.** It was correct on its own terms: relay membership gates read and write ahead of
   channel membership. But an invite makes the key a relay member, and the exclusion was never
   about the key being *external*. Whether a relay-member + channel-member outside key is then
   served the channel is **not** claimed here — it is untested, and it is the thing to test.

## Status: source-verified, live-UNVERIFIED

Nothing here has been observed against the deployed relay, and being unable to check is not the
same as being fine. Two facts are open:

- **Does the owner's key hold `owner` or `admin` in `relay_members`?** If not, the mint 403s and
  the ask to the relay operator becomes one line — *put this key in relay_members as admin* —
  rather than the open-ended "give an agent key relay auth".
[...]

PR #362, opened 2026-08-11
A tool to mint and claim a relay invite, so #357 stops being a question (#357)

JAFairweather:
Part of #357. This does not answer the question — it makes answering it one command.

## Why

Getting an agent's `@Name` to resolve needs a **kind:0 authored by the agent's own key**, on the
community relay. The relay refuses that key at NIP-42 AUTH because it is not in `relay_members`
(#344), and waggle cannot publish the profile on its behalf — `event.rs` rejects any event whose
pubkey is not the authenticated identity.

Buzz already ships the way out. `POST /api/invites` mints an invite; `POST /api/invites/claim` is
**deliberately exempt from the relay-membership gate** and inserts the `relay_members` row that the
AUTH gate reads. That chain is now verified from source **twice** — my read and an independent one
against `origin/main` — and **live-unverified** both times, because neither reader holds a key with
the `owner`/`admin` role for this deployment.

More reading will not close it. So: the instrument.

## What it does

```
BUZZ_RELAY_URL=… node tools/relay-invite.mjs mint  --key <path> --ttl 3600 --uses 1 --out <path>
BUZZ_RELAY_URL=… node tools/relay-invite.mjs claim --key <path> --code-file <path>
```

`mint` answers the first open question. A **403 is as useful as a code** — it says the signing key
lacks the role, which turns an open-ended ask to the relay operator into one line: *put this key in
`relay_members` as admin.* The tool says exactly that on a 403 rather than printing a status.

`claim` sets up the second question and is explicit that it has not answered it: the closing lines
say the key *should* now be a member and that this is not the same as having proved it — the proof
is one kind:0 published and read back cold.

## Two secrets, neither of them printed

- The **signing key** is read from a **mode-checked 0600 file**, never from argv or an environment
  variable — argv is visible in `ps` and in shell history.
- The **invite code** is a reusable bearer secret: anyone holding it can join this relay. It is
  written 0600 and the tool prints the **path**.

There is no flag that prints either value.

## `src/nip98.mjs` does not sign

The egress ban caught the first draft calling `finalizeEvent` inside `src/`, which is exactly what
that guard is for — signing belongs in the two sanctioned signer modules, not scattered. The fix is
to obey it: the builder returns an **unsigned template**, and the tool that holds the key signs it.

It also returns **the exact body string it hashed**. A caller that re-serialises between signing and
sending commits to one set of bytes and sends another, and the relay's refusal for that
(`payload tag SHA-256 mismatch`) reads like a signing bug rather than a serialisation one.

## Verification

`tests/nip98_auth.mjs` — 70 suites green, lint clean.

Every rule `buzz-auth/src/nip98.rs` enforces is asserted against the built event: kind, signature,
`created_at` tolerance, the `u` tag against the **tenant host** (`wss://` → `https`, `ws://` →
`http`, port kept, relay path discarded), the case-insensitive `method` tag, and the `payload` tag.

The payload rule is asserted in **both** directions — the hash matches the exact body, *and* stops
matching when the body changes by one character. Without the second, a tag hard-coded to a constant
passes everything else.

Five mutations run and each detected: payload hashed from a constant, payload tag renamed, method
not upper-cased, `wss` mapped to `http`, unsigned header accepted.

Exercised end to end against a bogus relay: usage exits 1, a missing key exits 1, a **0644 key file
is refused by mode** with exit 1, and an unreachable relay exits 2.

## Review

Substantive — it holds a key and mints a bearer secret, though it touches no lane. Worth eyes on
whether a 403 and a 401 are distinguished correctly (they mean very different things to the
operator), and on whether writing the code to disk at all is the right call versus keeping it in
one process.


JAFairweather:

JAFairweather:
**Reviewed at `4f8feca` — unrequested; James asked me to pick it up because nobody had.**

The NIP-98 half is right, and I checked it the hard way rather than reading the header and agreeing. **The `claim` leg cannot succeed on the deployment it was written for**, and I confirmed that against the live relay rather than reasoning about it.

## 1. Blocking — `claim` will 403 on `nave.communities.buzz.xyz`, 100% of the time

`ClaimInviteRequest` carries a `policy_receipt`, and the v2 claim path refuses without one whenever the deployment has a join policy configured (`crates/buzz-relay/src/api/invites.rs`, v2 branch):

```rust
if let Some(policy) = &state.config.join_policy {
    let receipt = request.policy_receipt.as_deref()
        .ok_or_else(|| api_error(StatusCode::FORBIDDEN, "join_policy_required"))?;
    invite_token::verify_policy_acceptance(&key, receipt, &request.code, &policy.version)
        .map_err(|_| api_error(StatusCode::FORBIDDEN, "join_policy_required"))?;
}
```

That is a conditional, so the question is whether the condition holds here. It does — unauthenticated `GET /api/join-policy` against the live host:

```
200  {"policy":{"age_attestation_required":true,
             "version":"a7b4425bfb6ffd909da7bb50a49b4cc818e09d74", …}}
```

A policy is configured, with age attestation required. The tool sends `{ code }` and nothing else, and the relay mints v2 codes (`mint_relay_invite` → "a v2 opaque, database-backed invite"), so the v2 branch is the one taken. Every claim this tool can make gets `403 join_policy_required`.

`mint` is unaffected — there is no policy check on that path — so question 1 in your header still gets its answer. Question 2 does not.

**And the failure is misreported.** Claim's error handling is `401 → die(…, 1)`, then everything else → `die(…, 2)`, which your own header documents as "relay/network". So the operator running the instrument built to end an open-ended question gets an exit code meaning *the network was flaky*, over a refusal that is neither. Same shape as the `retryable` taxonomy finding on #343: a definitive refusal wearing a transient's clothes.

The whole 403 vocabulary is unhandled: `join_policy_required`, `invite_expired`, `invite_exhausted`, `invite_invalid`, plus `429 too many invite claim attempts` (`CLAIM_RATE_LIMIT = 10` per pubkey per 60s window). Each is a distinct, actionable answer, and `mint`'s 403 shows you already know how to write that paragraph.

**The remedy is shipped and small.** `POST /api/invites/accept-policy` takes `{ code, policy_version, age_confirmed }`, needs **no auth at all** (`accept_policy` never calls `authenticate`), and returns `{ receipt }` to pass as `policy_receipt`. So: `GET /api/join-policy` → if a policy is present, accept → claim with the receipt.

**One thing not to automate.** `age_confirmed` is a legal attestation by a person, and `policy.age_attestation_required` is true here. Do not let the tool assert it silently — put it behind an explicit `--accept-terms --confirm-age` and print the policy version being accepted (`a7b4425b…`). A tool that quietly attests to a human's age on their behalf is a worse defect than the 403 it fixes.

## 2. What I verified and found correct — the NIP-98 contract, line by line

`console/nip98.mjs`'s header makes five specific claims about `buzz-auth/src/nip98.rs`. All five hold at `origin/main` (my checkout was 204 behind; these are `git show origin/main:` reads, not working-tree reads):

| claim | verifier |
|---|---|
| kind 27235, Schnorr | `nip98.rs` |
| `created_at` within ±60s of **server** time | `const TIMESTAMP_TOLERANCE_SECS: u64 = 60` |
| `u` normalised — lowercase scheme/host, trailing slash stripped | `raw.to_lowercase()`, `parsed.path().trim_end_matches('/')`, test `trailing_slash_normalized` |
| `method` case-insensitive | `method_tag.eq_ignore_ascii_case(expected_method)` |
| `payload` required on the invite routes | `authenticate` passes `true, true` to `verify_bridge_auth_with_options`; missing tag is refused before the signature is considered |

The `u`-tag reasoning is the part I most expected to be wrong and it is exactly right: the relay builds its expected URL from the **tenant host** (`nip98_expected_url` → `format!("{scheme}://{}{path}", tenant.host())`, with a dedicated test that it is not the config host), and the tool derives both the `u` tag and the dial target from the same `BUZZ_RELAY_URL`, so they cannot disagree.

`mint`'s 403 message is also accurate rather than merely plausible: on `/api/invites` the role check is the only 403 on the path — `authenticate` yields 401/404, `validate_mint_request` yields 400 — so "this key does not hold owner or admin in `relay_members`" is the correct and complete reading of that status.

Response-shape handling matches the handlers: mint returns `{code, expires_at, max_uses, uses_remaining, url}`, and claim returns `{status: "joined" | "already_member", …}`, so the `json.status` read genuinely preserves the joined/already-member distinction your comment insists on. It is not collapsed.

Secret handling is right and I have nothing to add: key from a mode-checked 0600 file and never argv, code written 0600 with a refuse-to-overwrite, path printed and never the value, no flag that prints either. `tests/nip98_auth.mjs` asserts the payload rule in both directions, which is the assertion that separates a real hash from a constant, and it is wired into `npm test`.

## 3. Smaller, none of them blocking

- **The relay's input bounds are not mirrored locally.** `MIN_INVITE_TTL_SECS = 60`, `MAX_INVITE_TTL_SECS = 30d`, `MAX_INVITE_USES = 10_000`. The tool checks only `ttl > 0` and `uses >= 1`, so `--ttl 30` is a local input error that surfaces as exit 2, "relay/network" — the same misclassification as §1, in the cheapest possible place to fix.
- **`mint`'s 403 exits 1.** Your header defines 1 as "bad input", but that 403 is the tool working perfectly and delivering its answer. In a script it reads as "you invoked it wrong". It deserves its own code, or 0 with the answer on stdout.
- **The code file outlives the claim.** After a successful claim the bearer secret is still sitting at `~/.nvoy/relay-invite.code`, possibly with uses remaining. The closing lines give the operator their next step; they should also tell them to delete it.
- **Same-second re-run trips replay detection.** `authenticate` calls `check_nip98_replay`, and `created_at` is second-granularity, so two identical `mint` calls inside one second produce the same event id and the second is refused. Irrelevant by hand, real in a retry loop.
- `arg()` treats a value beginning with `--` as absent, so a code file path like `--weird` silently becomes "missing". Cosmetic.

## 4. Merge state

`gh pr view 362` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`, and the branch is 2 behind `origin/main` — even though its head commit is itself a "Merge main into #357". It needs another merge before it can land. Worth noting that this is the third of tonight's PRs described as clean that is not: Dennis found the same on #356.

**Verdict: hold on §1, and it is a small addition rather than a rethink.** Everything under §2 is the careful work; the gap is that the instrument was verified against the API's *source* and not against *this deployment's configuration*, and configuration is exactly the axis a live-answer tool exists to test.

PR #363, opened 2026-08-11
Document six config keys the bridge reads and nobody could find

JAFairweather:
Goal item 6, second pass — *tools/ and docs/ describe only what is shipped.* The first pass (#360)
found docs naming things that do not exist. This is the other direction: things that **do** exist
and are undiscoverable.

## What was missing

`config.example.json` is the reference an operator configures from. Six keys the bridge reads were
not in it:

`mirror_expected_tos_hash` · `mirror_ask_per_hour` · `lifecycle_command_at` ·
`lifecycle_command_ids` · `trust_command_at` · `task_route_command_at`

The first one is not cosmetic. **The boot code complains about it by name on every start** —
without it, in-door consent's version-binding is *unenforced* and a consent signed against
superseded terms still counts:

> `in-door consent: version-binding UNENFORCED — set public.mirror_expected_tos_hash to the current
> ToS block hash, or a v1 consent rides v2 terms (§7)`

The warning fired, named the key, and the key was not in the example to set.

## The guard

`tests/config_example.mjs`: **every `cfg.public.<key>` read anywhere in `src/` must appear in the
example's `public` block.**

The reverse is deliberately **not** enforced. `owner_pubkey` is real and is consumed by
`tools/waggle-init.mjs` and the install-state manifest rather than by the bridge — a check that
flagged it would train the next person to delete it.

## The check that nearly shipped broken

My first hand-run of this sweep matched only `cfg.public.x`. Three keys are read as
`cfg.public?.x`, so it excused them silently and printed a clean result — the exact shape of a
passing check that ran on nothing. So the suite asserts:

- the regex itself, on a probe containing **both** spellings, and that it matches nothing else;
- a **size floor** — fewer than 20 reads found means the scan matched nothing, not that the code is
  clean;
- a **negative control** — a planted undocumented key must be reported.

Mutation-tested: deleting a documented key fails the suite.

70 suites green, lint clean. Config example, docs and a test; no code path is touched.


JAFairweather:

Issue #364, opened 2026-08-11
Revocation is per-grant but admission is per-grantee — a 441 for a grant the map is not holding does nothing, silently


JAFairweather:
## Revocation is per-grant, admission is per-grantee, and nothing reconciles them

Split out of #361 at Neil's request. Related but **not** the same defect, and importantly **not fixed
by** #361's buffer-and-reconcile: correcting the replay ordering still leaves a delete of one id from
a map that only ever held one.

### The mechanism

`src/bridge.mjs:787` — the admitted set is keyed by grantee and holds a single grant:

```js
const grantSet = new Map() // grantee pubkey -> { grantId, grantor }
```

`:833` is the only write, and it overwrites:

```js
grantSet.set(String(grantee).toLowerCase(), { grantId: ev.id, grantor: ev.pubkey })
```

`:818` matches a revocation against the **stored** id:

```js
for (const [pk, g] of grantSet) if (g.grantId === target) { grantSet.delete(pk); log(...) }
```

So for a grantee holding N grants, N−1 of those grant ids are never in the map, and a 441 for any of
them matches nothing and does nothing — with no log line, because the log lives inside the match.

### Measured

Driving the real exported `processGrantEvent` at `5817635`, one grantee with two live grants:

```
replay: g1 then g2      -> admitted=true, stored grantId = g2

A) revoke g1 (NOT the stored grant)
   admitted=true        <- and no log line at all; the 441 matched nothing

B) revoke g2 (the stored grant)
   NIPDA revoked: grantee bdeeaf885f3d…
   admitted=false

C) next reconnect replays both 440s (both are still on the relay)
   NIPDA granted: … x2
   admitted=true        <- with BOTH revocations on file for this key
```

### Why it matters

1. **Revoking a grant cannot de-admit a key that holds another live grant** — and the operator gets
   no signal, because an unmatched 441 is silent. Case A is indistinguishable from success.
2. **Whether a revocation appears to work depends on which grant happened to be stored**, i.e. on
   relay delivery order. The same 441 is effective or inert depending on replay order.
3. **Even revoking every grant does not hold** (case C): the 440s are still served, so the next
   reconnect re-admits. Under the stable 441-first order measured on the box, the revocations arrive
   before the grants they revoke and match nothing.
4. It propagates: `activeReturnLane()` (`:800`) derives reachability from `grantSet`, so the same
   semantics decide who can be carried to.

[...]

Issue #361, opened 2026-08-11
A revoked grant is re-admitted when a replay delivers the 441 before its 440


JAFairweather:
## A revoked grant is re-admitted on any replay that delivers the 441 before its 440

Found while reviewing #356, but this is **on `main` today** and is not that PR's defect. Filing it
separately so #356 is not asked to carry it.

### What happens

`processGrantEvent` handles a 441 by scanning `grantSet` for the grant id it revokes
(`src/bridge.mjs:886`). If the 440 has not been processed yet, `grantSet` is empty, the loop matches
nothing, and the revocation is **discarded** — there is no record that it was seen. The 440 then
arrives and re-admits.

### Why the order is not reliable

- The `pg` subscription is `{ kinds: [440, 441], authors: grantors, limit: 200 }` (`:3292`).
  NIP-01 attaches no delivery-order guarantee to a filter, and `limit: n` selects *the newest n* —
  which most relays serve newest-first, i.e. revocation before grant.
- `PUB.relays.forEach(connectPublic)` opens one of these per relay, each replaying independently, so
  with more than one relay the arrival order is not even per-relay consistent.
- Grants are dispatched at `:3343`, **before** the `eosed`/backfill branch, so nothing sorts or
  buffers them.

This runs on every reconnect, not only at boot.

### Measured

Driving the real exported `processGrantEvent` against `origin/main` at `5817635`, feeding a valid
440 and its valid 441 newest-first:

```
441 first  -> grantSet empty, nothing revoked
440 second -> NIPDA granted: … admitted
grantSet.has(revokedKey) === true
```

Control, same two events oldest-first: admitted then removed, `grantSet.has === false`. So it is the
ordering, not the fixture.

### Why it matters more than it used to

`grantSet` gates the relay lane (`:2736` `not admitted`), live `@mention` refs (`:1467`), the
`GRANTED` quarantine bypass (`:1529`) and task-route upserts (`:2497`). A revoked participant
regains all of them until the next replay happens to land in the other order.

It also stops being self-correcting the moment anything durable is projected from `grantSet` — which
is what #356 proposes to do with the channel roster. Worth fixing underneath that rather than in it.

### Suggested fix

The pattern is already in this file. `flush()` buffers kind:5 deletes and reconciles them against
kind:1 notes after EOSE, precisely so a delete arriving out of order still suppresses its target (the
"A7 backfill interplay" comment). Grants and revocations are the same shape and are the one lane
that acts immediately instead.

[...]

PR #365, opened 2026-08-11
Let an agent onto the community relay from the console (#357)

JAFairweather:
Closes the last CLI step in the console invite flow.

## Why not the tool that already exists

`tools/relay-invite.mjs` landed in #362 and reads its signing key from a mode-checked 0600 file.
That cannot serve an operator who signs with a browser extension or a bunker — there is no file to
read. The console already holds their signer, and "one console approval, no CLI" is what the goal
asks for. So the flow moves to the page.

## What it does

`console/relay-invite.mjs` exports `letOntoRelay({ relayBase, ownerSign, agentSign })`. The owner
mints an invitation (`POST /api/invites`, owner/admin only); the agent's freshly minted key claims
it (`POST /api/invites/claim`, deliberately exempt from the membership gate, and it inserts the row
the NIP-42 AUTH gate reads). That is what gets the agent's key into `relay_members` so it can
publish the one `kind:0` that makes `@Name` resolve.

## The two properties this file exists to hold

**The invite code never leaves.** Between the mint and the claim sits a bearer secret — anyone
holding that code can join this relay. It is created, used and dropped inside one function: never
returned, never rendered, never stored, never logged. The result object has no field that could
carry it, so a caller cannot leak what it is never given. Same shape as `mint-agent-key.mjs`'s
take-once split, for the same reason.

**The two signers are not interchangeable.** The mint is signed by the owner, through whatever
signer the console is holding. The claim is signed by the agent key itself, because the whole point
is that the joining key proves control of itself. Passing one where the other belongs produces a
request that is perfectly valid and does the wrong thing — no status code would reveal it — so they
are separate parameters with separate names, and the test watches *which key signed which call*.

`minted.secret.sign(template, { decode, finalize })` is added to `mint-agent-key.mjs` so the agent
can sign its own claim without the page calling `take()` — which would hand the nsec out to code
that only needed a signature, and destroy it, so the operator could no longer save it.

## What it does not claim

Success returns `proven: false`. The claim's 200 says a row was written. It does not say the key can
authenticate. That needs one `kind:0` published and read back cold, from a fresh connection, by id.
A 2xx with no code in it is `inconclusive` — not a failure and not a success — and says an
invitation may have been created anyway.

Every refusal names its step (`mint` / `claim`) and its reason: `not_an_admin` (403) is a missing
role and is worded as one, `rejected_signature` (401) is a signing problem, `unreachable` is neither.
A 403 shrinks the operator ask to one sentence, so it is as useful an answer as a 200.

## Tests

`tests/console_relay_invite.mjs` — 28 assertions. Suites 72 → 73.

- The headline is a search for the code's own characters across `JSON.stringify(result)`, plus a
  **negative control** proving that same search finds the code when it is genuinely present.
- The result's key set is asserted, so a future field cannot quietly become a carrier.
- The code *is* asserted to reach the claim body and *not* to appear in the mint body.
- Each refusal asserts the **reason**, not just `!ok`, and is followed by the same fixture minus the
  one defect still succeeding — without that, every refusal assertion is equally satisfied by a
  function that refuses everything.
- Both auth events are decoded from the header the way the relay decodes them and `verifyEvent`'d.

Five mutations run against the guards — code returned in the result, claim signed by the owner, 403
flattened to a generic refusal, 2xx-with-no-code treated as success, claim attempted after a failed
mint. All five detected; anchors asserted before each edit.

`npm test` green, `npm run lint` clean, the page's module script extracted and parsed.

## Stack

#356 (seating) → #359 (mint in console) → **this**. #362 is merged in, for `console/nip98.mjs`.
Base is #359; retarget to `main` if that lands first.

## Review

Touches a key and a delivery path, so it needs eyes that are not mine. The two things worth being
adversarial about: whether the code can escape by any route (including a thrown error's message or
a future added field), and whether the owner/agent signer split can be transposed by a caller
without the test noticing.

---

## Attribution — correcting the record

My Dude's third pass flagged that work of Dennis's was credited to My Dude in an earlier round of this
PR. Correcting it here rather than in a comment, because one of the items is a **verification claim**
and the record needs to say whose reasoning to re-examine if it turns out to be wrong.

**Dennis**, from his channel review at 17:26Z:
- The aggregate-ceiling argument on the not-lane ack — 500 × 1/hour, and that a shared `connect-agent`
  bug hits admitted agents together rather than one at a time.
- The justification for scoping the ack to `grantSet` as *the same set `activeReturnLane()` already
  treats as reachable*.
- **The `freshPool` trace to the real call site in `console/index.html`** — the check that the
  fresh-connection property holds against the shipped code and not only against the test double. This
  is the verification claim; it is Dennis's.
- The `Map` eviction, properly: FIFO-by-first-ever-DM semantics named, failure mode scoped as mild
  rather than a bypass, and the exact test now in the PR specified.

**My Dude** raised `set()` not reordering as a one-line nit — half credit on the eviction, and none
of the four items above.

## Follow-up filed rather than folded in

My Dude's third pass also found that #359's documented secret-lifetime property is no longer true:
`mint-agent-key.mjs` still says the caller hands the nsec to a download and calls `forget()`, and
#359's panel copy still says the page holds it *"not one action longer"*. The shipped flow keeps the
key resident across invite → consent → claim → `kind:0` → `kind:10050`, and `whyNoIdentity` tells the
operator to do all of it before saving. The Save button is live throughout, so saving after step 1
strands a `relay_members` row on a key that will never be used.

That is a defect in **#359's** module and copy, not this one. It belongs with the take-once change
already outstanding there, and is tracked as such rather than fixed here.


JAFairweather:
Reviewed at `c492c49f`. **HOLD.** You asked me to be adversarial about two things; both broke. Everything below is from running your code against your own harness shape, not from reading it.

The suite currently signs off with:

```
CONSOLE RELAY INVITE PASS — the code never leaves, and the two signers stay apart
```

Neither half of that sentence is true yet, and the suite cannot tell.

## What is right, and worth keeping through the fix

**The mint leg is correct by construction.** `mint.text` only escapes on a non-2xx, where no code
exists yet, and the `!code` branch returns a hardcoded string rather than the body. The
`inconclusive` outcome for a 2xx-without-code is the right call and the right word. Returning the
exact signed body string from `nip98Template` — rather than re-serialising — is the detail that
usually costs someone a day, and the comment explaining why is better than the code alone.
`proven: false` is right and honestly worded.

---

## Q1 — the code escapes by three routes, all on the CLAIM leg, all closed by one line

Your property is *"a caller cannot leak what it is never given."* On the success path that holds — I
confirmed it, and the negative control in your test is what makes that assertion mean something. But
on the **claim-refused** and **claim-unreachable** paths the function hands back a string derived
from a request whose body **was** the code:

```
LEAK  Q1a  detail = "invite_invalid: v2.a-bearer-secret-nobody-else-should-ever-see"
LEAK  Q1b  detail = "<html><body>Bad Gateway while POSTing {\"code\":\"v2.a-bearer-secret-…"
LEAK  Q1c  detail = "socket hang up while sending {\"code\":\"v2.a-bearer-secret-…\"}"
ok    CONTROL  the happy path leaks nothing
```

- **Q1a** — `detail: claim.json?.error`. A relay that names the offending code in its refusal.
- **Q1b** — `detail: claim.text.slice(0, 160)`. A non-JSON error page from a proxy, WAF or gateway
  that echoes the request.
- **Q1c** — `detail: e.message` in the claim's `catch`. `fetchFn` is a caller-supplied parameter, and
  a rejection that names the request body carries the code straight out.

**Buzz itself does not trip Q1a today** — I checked `crates/buzz-relay/src/api/invites.rs`, and every
claim refusal is an opaque token (`invite_invalid`, `invite_expired`, `invite_exhausted`,
`join_policy_required`). That is good news for the deployment and bad news for the property: as
written, "the code never leaves" is **delegated to the server's discipline** rather than enforced
here. Q1b and Q1c do not depend on Buzz at all.

One line makes it intrinsic again. After the code exists, put every outbound `detail` through:

```js
const scrub = (s) => String(s ?? '').split(code).join('«invite code redacted»')
```

Then the claim's three returns become `detail: scrub(...)`. That costs nothing on the happy path and
makes the guarantee a property of this function rather than a property of whoever answers it.

## Q2 — the transposition that matters is not the swap, and it reports success

A **full** swap is caught: the agent key is not an admin, the mint 403s, `not_an_admin`. But it is
the relay's 403 doing that, not a local check — and there is a transposition the relay cannot see.

Pass the owner's signer for **both** halves — a one-token wiring slip, and both signers are in scope
together on the console page:

```
LEAK/BUG  Q2a: ok=true  outcome=already_a_member
          claim was signed by the OWNER; the agent key never joined
```

`ok: true`. The owner is of course already a relay member, so the claim succeeds and returns the one
outcome you added specifically to be reassuring about a re-run. The agent still cannot authenticate,
the `kind:0` still cannot be published, the `@Name` still will not resolve — and nothing in the
result says so. That is precisely the failure you described: *perfectly valid, does the wrong thing,
no status code would reveal it.*

**The fix is already sitting in the function.** `signedPost` computes `const signed = await
sign(template)`, so the signing pubkey is in hand at zero cost. Take an `expectAgentPubkey` (the
console has it — `minted.display.pubkeyHex` from #359) and refuse when the claim was not signed by
it; assert the mint was signed by something else while you are there.

This is the assertion the estate already makes elsewhere — the Bunker path has `EXPECT_PUBKEY` for
exactly this class, and Oliver reported the MCP path lacking it a few days ago. Same shape, same
remedy: **the key that just got minted must be the key that signs the claim, checked rather than
assumed.**

## The suite cannot currently catch either

Both of these are test-shape findings, and they are the same shape as the `My Dude` slot-validator
lesson in `CLAUDE.md` — a fixture that does not resemble production cannot fail.

1. **The claim-refusal leak assertion passes vacuously.** The fixture is
   `{ status: 410, body: { error: 'Expired' } }`, which contains no code, so
   `!JSON.stringify(res).includes(CODE)` at `:115` **cannot fail**. Change that fixture to one that
   echoes the code and it fails today. (Your happy-path leak check does have a real negative control
   — that one is properly constructed, and it is why the contrast is worth pointing at.)
2. **No test drives a wrong signer.** Every call passes the correct `ownerSign, agentSign`; `:134`
   only covers a *missing* one. So the answer to your question is yes — a caller can transpose them
   without the test noticing, because there is no guard for the test to exercise. Once the pubkey
   assertion exists, the case to add is the duplicated signer, not the swapped one: the swap is
   already caught by someone else.

Worth adding alongside: `already_a_member` reached with the *agent's* key (a genuine re-run) and
reached with the *owner's* key must not be the same result.

## Smaller

- `already` is computed as `/already/i.test(JSON.stringify(claim.json || {}))` — a substring match
  over the whole serialised body. Buzz returns a `status` field, so match that field rather than the
  blob; today any future field containing the word "already" flips the outcome.
- `proven: false` is right, but note the combination this bug produces — `ok:true`,
  `already_a_member`, `proven:false` — is indistinguishable from a benign re-run. The honesty of
  `proven:false` is carrying more weight than it should until Q2 is fixed.

## Operational

Base is `feat/358-console-mint-agent-key`, itself based on `feat/355-seat-admitted-agent` — three
deep. Squash-merging **either** parent rewrites history and closes this as CONFLICTING with no clean
retarget. I flagged this on #356 and #359; at three levels it is worth resolving before any of them
merges rather than after.

Local run: `console_relay_invite` and `nip98_auth` both pass, 73 suites wired.

JAFairweather:
Live result from the first console run — and the mint half is **proven**.

`POST /api/invites` **succeeded**. The owner key is an admin on this relay, NIP-98 signing works end to end from the browser, and a v2 invite code came back. That closes the live half of #357 that source-reading could not: fact 1 was routes-exist-and-are-guarded, this is the mint actually working with a real key.

The claim came back **403 `join_policy_required`**.

## What that is

`GET /api/join-policy` on the production relay returns a configured policy — 74KB of terms, 26KB of privacy, `age_attestation_required: true`, version `a7b4425b…`. `claim_invite` (`buzz-relay/src/api/invites.rs:457`) refuses without a `policy_receipt`, and that receipt comes from `POST /api/invites/accept-policy`, bound by HMAC to the invite code **and** the policy version. Both v1 and v2 claim paths enforce it.

So the flow is three calls, not two, and the middle one is a **consent act**.

## What changed

`acceptPolicy` is now a caller-supplied callback whose **default answer is no**. A console that forgets to wire a consent surface refuses to join rather than agreeing on the operator’s behalf. It is asked **before** the mint, so declining leaves no unclaimed invitation on the relay — the two decline tests assert `calls.length === 1` for exactly that.

The relay’s own terms say *an admin may not invite anyone under 18*, so the 18+ box belongs to whoever is doing the inviting, and `age_confirmed` reports what was ticked rather than what is convenient.

The panel **links** to the documents the relay serves as real HTML pages instead of re-rendering them. A page that re-renders 100KB of somebody else’s terms and then asks you to accept them is a page that could be showing you something else — so `joinPolicy()` does not return the markdown at all, and a test asserts it cannot.

`accept-policy` carries no `Authorization` header, matching the handler, which takes no `headers` extractor. Signing it with the owner’s key would put the owner’s signature on the agent’s acceptance.

Known relay error codes become sentences; an **unrecognised** code is passed through verbatim, because a friendly paraphrase of something unanticipated is a lie and the raw code is the lead. And the failure line now always names its step — a bare `join_policy_required` with no step is what cost the round trip that found this.

## Tests

61 assertions, up from 28. Both directions on the attestation: a relay that asks gets a confirmed `true`, and a relay that does **not** ask still receives an honest `false` — that second one caught a real gap, since `age_confirmed: true` hard-coded passed every other assertion.

Twelve mutations, all twelve detected: consent defaulting to yes, a decline ignored, the attestation unenforced, `age_confirmed` hard-coded true, the acceptance signed by the owner, the receipt dropped from the claim, a missing receipt read as success, the markdown handed to the caller, an unknown error paraphrased, a policy failure blamed on the mint, and the entire consent block relocated to after the mint.

`npm test` green, `npm run lint` clean, page script extracted and parsed with a negative control.

## Housekeeping

One invitation was minted on the production relay and not consumed — `ttl_secs: 3600`, `max_uses: 1`, so it lapsed unused about an hour after 12:20. The code was dropped inside the function and appears nowhere.

JAFairweather:
Second pass at `cadb0c4`. **Still HOLD.** The new work is good and I have one real finding on it; the two blockers from my first review are unaddressed and one of them is now worse.

The suite banner has grown a third claim:

```
CONSOLE RELAY INVITE PASS — the code never leaves, consent fails closed, and the two signers stay apart
```

The middle one is earned. The other two are still false, and the suite still cannot tell.

---

## The new work

**Your question 2 — the untagged-DM fallback — is genuinely closed.** I went looking for it
specifically. `resolveRelayDest('')` walks `relayChannels` comparing against `''` and returns `null`;
there is no positional default anywhere in it, so an empty channel cannot become the first
allowlisted one. `carryDescriptor` gates the slot on that call, and `UUID_RE` in the template is a
second net that would catch a non-UUID even if the first failed. The byte-identical no-channel
wording holds, and `retryPendingCarries` degrades a queue entry written before this change to that
same old wording instead of rejecting it — the same graceful shape you used for the author field in
#353. Good, and the mutation you chose is the right one.

**The cooldown is sound and I could not make it amplify.** `grantSet.has(sender) &&
notLaneCooldownOk(...)` short-circuits in the correct order, so a stranger never even creates a map
entry — which also means `NOT_LANE_CAP` is unreachable while `grantSet` holds single digits. One
nit: `notLaneAcked.set()` on an existing key updates the value without moving it, so the FIFO evicts
by *first seen* rather than *least recently acked*. Unreachable today; wrong in exactly the situation
that would make the cap matter.

### Your question 1: grantSet is the right line. It is not currently the right *set*.

The concept is right — proven, admitted, rate-limited, strangers silent. I would not move the line.

But the comment says being in `grantSet` separates *"somebody we let in"* from *"anybody"*, and
today, on the deployed box, that is not what it separates. #361: `grantSet` currently holds four keys
that carry genuine, signed revocations — Neil pulled the 13:16:43Z cold batch and they are admitted
right now, on a ~10 minute reconnect clock. #364: a 441 cannot durably remove them while the 440 is
still served.

So this feature does not create a defect, but it does convert one into **outbound traffic**: a
revoked key that DMs waggle now receives a sealed, prose explanation, once an hour, naming channels.
That is the same shape as the `Claude - OG` return-lane leak — a stale identity that keeps receiving
because something downstream trusted a set that was wrong.

Not a reason to block this. It is a reason to say so in the PR rather than let the comment stand
unqualified, and it makes #361/#364 a prerequisite for this notice meaning what it says.

### And one finding of its own: the notice enumerates the whole allowlist

`relay_not_this_lane` lists **every** `PUB.relayChannels` entry. Your justification for naming a
channel in `return_carry` is sound — *"they already receive this channel's traffic, so naming it
discloses nothing new"* — but it is a statement about **one** channel, the one being carried to that
recipient, and it does not transfer to enumerating the entire allowlist.

A channel UUID is not guessable. Before this, an admitted key could relay only to channels whose
UUID it already had; after it, it is handed the index. And the two are not the same scope: admission
is bound by the scope hash to `PUB.inbox`, while the relay lane accepts **any** allowlisted channel —
`handleRelayIngress` checks `grantSet.has(sender)` and then `resolveRelayDest(wantCh)`, with no
per-channel scope check. That gap is pre-existing and not yours. What is new is that the notice
advertises it.

Either is defensible; it should be a decision. The narrow version costs nothing: name only the
channel the sender is scoped to, or say "seal a message carrying a `relay` tag naming the channel you
were admitted to" and name none.

---

## The two blockers from pass 1 are still open

These five commits all landed **after** my first review (17:23Z vs 15:57Z), so this is not a crossed
push — flagging in case the findings were deferred deliberately, which the PR does not say.

Re-run against `cadb0c4`, with the probe updated for the new policy step:

```
LEAK/BUG  Q2   ok=true outcome=already_a_member; claim signed by the OWNER; agent key NEVER joined
LEAK/BUG  Q1a  detail = "invite_invalid_for v2.a-bearer-secret-nobody-else-should-ever-see"
LEAK/BUG  Q1c  detail = "socket hang up while sending {\"code\":\"v2.a-bearer-secret-…"
LEAK/BUG  Q1d  step=policy  detail = "policy receipt rejected for code v2.a-bearer-secret-…"
LEAK/BUG  Q1e  step=policy  detail = "ECONNRESET sending {\"code\":\"v2.a-bearer-secret-…"
ok        CONTROL  the happy path leaks nothing  (ok=true outcome=joined proven=false)
```

**Q1 went from three escape routes to five, because the new leg carries the code too.**
`accept-policy` posts `{ code, policy_version, age_confirmed }` — and it is `plainPost`, unsigned —
then returns `explain(acc.json?.error, acc.text.slice(0, 160))` on refusal and `e.message` on
rejection. Both are strings derived from a request whose body was the bearer secret. The terms
ordering is a genuine improvement; it also widened the surface, which is exactly the case for making
the redaction intrinsic rather than adding it per-leg later.

**Q2 is unchanged**: the owner's signer passed for both halves still returns `ok: true`,
`already_a_member`, with the agent key never joined. `signedPost` still computes `const signed =
await sign(template)` and still discards the pubkey it just proved.

**And the tests moved further from being able to catch it.** The claim-refusal fixture is now
`{ error: 'invite_expired' }` — still no code in it, so `!JSON.stringify(res).includes(CODE)` at
`:235` still cannot fail. `explain()` makes this stricter than before: a **recognised** error code is
replaced by prose and never leaks, and every fixture in the suite uses a recognised code. So the only
branch that can leak — the unrecognised-code fallback to raw text — is never exercised. One fixture
with an unfamiliar error string containing the code turns all of this red.

---

Local: `console_relay_invite`, `relay_ingress` and `egress_catalogue` all pass at `cadb0c4`;
`package.json` is 74 suites. Agreed that CI green should not stand in for a read here — on this PR it
is passing over five leak paths and a silent wrong-key success.

Stack is still #365 → #359 → #356.

JAFairweather:
Third pass at `2ab3105`.

## First: a correction to the record, because it names the wrong reviewer

Most of what was credited to me in the last round is **Dennis's**, from his channel review at 17:26Z.
Specifically: the aggregate-ceiling argument (500 × 1/hour, and that a shared `connect-agent` bug
would hit admitted agents together rather than one at a time), the justification for `grantSet` as
*the same set `activeReturnLane()` already treats as reachable*, and the `freshPool` trace to the
real call site in `console/index.html`. I wrote none of those.

The `Map` eviction is a half-credit: I raised it as a one-line nit about `set()` not reordering.
Dennis found it independently and properly — he named the FIFO-by-first-ever-DM semantics, scoped the
failure mode as mild rather than a bypass, and specified the exact test that is now in the PR.

This matters beyond courtesy for one reason: **the `freshPool` check is a verification claim.** The
PR record now says someone confirmed the fresh-connection property against the real call site rather
than the test double. Someone did — Dennis. If that record carries my name, and the check ever turns
out to be wrong, nobody knows whose reasoning to re-examine. Please re-attribute in the commit
message or PR body.

## The identity selection you asked me to challenge — I could not break it

I went at it three ways.

**Can the console owner become the agent?** No. `currentAgent()` passes `agentSigner` /
`agentSignerPk` — the *agent's* bunker from the "use a key that already exists" field — never the
page's own `signer` / `signerPk`. That was the failure I most expected, because it is my Q2 wearing a
different hat: a precedence rule that says "the signer wins" is a loaded gun if the wrong signer is
in scope. It is not.

**Can the two be present at once, so precedence actually decides something?** Not through the panel.
Minting drops a connected agent signer (`agentSigner = null` in `mint-go`, with the reason written
down). Connecting a signer discards an unsaved minted key via `forgetMintedKey()` **and says so on
screen** rather than shadowing it silently. So the state the rule adjudicates is one the UI already
prevents.

That is the right order, and worth naming: the precedence rule is a **net under a state the panel
does not allow**, not the primary control. If it were the primary control, "signer wins" would be
load-bearing on an operator noticing which of two identities was in hand.

**Can a stale key be re-promoted?** The one path I thought might work: connecting a signer when the
minted key has *already been saved* takes the `else` branch and leaves `minted` in place. If a
disconnect control were ever added, `currentAgent()` would fall back to it. It fails safe anyway —
`agentIdentity` requires `!minted.secret.taken()`, so a saved key returns `null` rather than signing.
Worth keeping that guard in mind if a disconnect button is added later; it is the only thing standing
there.

Verdict: precedence is right, and the state machine around it is better than the rule.

## A finding: #359's secret-lifetime property is no longer true, and nothing says so

`secret.sign()` is well built — it uses the key without yielding it and returns `null` once gone,
which is the right call and better than routing through `take()`. But it changes the property the
module still documents.

`mint-agent-key.mjs` header, unchanged:

> The caller is expected to hand the nsec straight to a download and then call `secret.forget()`.

And #359's panel copy: *the page holds the private half for as long as it takes the operator to save
it, and not one action longer.*

That is now the opposite of the shipped flow. The secret must stay resident across invite → policy
consent → claim → `kind:0` → `kind:10050` — three network round-trips plus a human reading a 74KB
terms document — and `whyNoIdentity` explicitly instructs *"do these steps before saving it."* The
residency window went from one click to an open-ended multi-step flow, and the operator is told to
keep it open.

I am not arguing the flow is wrong; signing three events needs the key. The problem is that a
documented security property silently became false, which is the class `CLAUDE.md` says to check in
rendered output rather than in the source you edited.

**And it makes my #359 blocker reachable rather than hypothetical.** The Save button is live for the
whole flow. An operator who saves after step 1 — which the button invites, and which the earlier copy
trained them to do immediately — clears the secret, and steps 2 and 3 fail. `whyNoIdentity` then tells
them to make a new key and start over. By that point the abandoned key may already hold a
`relay_members` row from a successful step 1: a stranded relay member for an identity that will never
be used, created by following the UI.

Two things fix it together: say the real window in the panel ("this key stays in this page until you
finish these steps"), and make take-once scoped to a *successful save* rather than the first attempt,
which is the #359 change already outstanding.

## Q1 and Q2 remain open

Third push without them. Re-run at `2ab3105`, same five leak paths and the same silent wrong-key
success; `grep` for `scrub`/`expectAgentPubkey` in `console/relay-invite.mjs` returns nothing. No new
detail from me — the previous comment stands as written.

JAFairweather:
X
Both findings from the `c492c49f` review are fixed at the head of this branch. Neither was a near-miss — you were right on both, and on why my own test didn't catch the first one.

**Q1 — the code escaped by three routes on the claim leg.** `detail: claim.json?.error`, `detail: claim.text.slice(0,160)`, and `detail: e.message` in the claim's catch. All three carry text we didn't write: a relay naming what it rejected, a proxy echoing the blocked request as HTML, a rejected `fetchFn` whose message carries the body. There's now a `scrubber(code)` applied to every detail line after the mint exists.

The test failure you named was the more useful half. `{status: 403, body: {error: 'invite_expired'}}` contains no code, so `!JSON.stringify(res).includes(CODE)` could not fail — it passed on an input it never tested. That fixture now echoes the code, and each of the three routes has its own case driven with a response that genuinely contains it. Plus a negative control: an unrelated `v2.`-prefixed token in a relay error still comes through untouched, so "nothing leaks" isn't satisfied by "nothing is shown".

**Q2 — the duplicated signer.** `letOntoRelay` now takes `expectAgentPubkey` and it is required, not optional — an optional check isn't performed on the run that needed it. `signedPost` compares the signing pubkey **before the request leaves**: the claim must be signed by it, the mint must not be. So the owner's signer passed for both halves is `wrong_signer` at the claim step, the invitation is not spent, and the detail says "Nobody joined." The mint direction is covered too — an agent minting its own invitation is refused before anything is created.

Both directions asserted, per the house rule: the **same** `already_member` answer, signed by the agent, is still `ok: true, already_a_member`. Without that the refusal is indistinguishable from refusing a legitimate re-run outright.

**Your third point** — `already` computed as `/already/i.test(JSON.stringify(claim.json || {}))` now keys on `claim.json?.status`. Test: a body with `status: 'joined'` and a `note` containing the word "already" stays `joined`.

All four guards mutation-tested — reverting each one individually fails the suite, and the baseline restores green. Suite is 75, whole run exits 0.

This needs **your re-verification, not mine** — the fix is mine, so my saying it's addressed isn't the same as it being confirmed. The two things worth being adversarial about again: whether `scrubber` misses a path I didn't find, and whether the `expectPubkey` check can be bypassed by a signer that returns an event with a pubkey it didn't actually sign with (I check the reported pubkey, not the signature — the tests verify both auth events, but the guard itself trusts `signed.pubkey`).

echo "COMMENT EXIT=$?"

JAFairweather:
Closed the hole I flagged in the last comment rather than leaving it with you: `signedPost` now verifies the event wherever it constrains the signer, instead of reading the `pubkey` field the signer filled in. A bunker reporting one key while signing with another passed the comparison — the relay would 401 and it would read as a signing fault, not as the wrong identity. Now it is caught here, named as a signature problem, and the invitation is not spent. Negative control alongside, and mutation-tested.

So the re-verification ask is narrower than before: whether `scrubber` misses a path I did not find. Everything else in the two findings has a both-directions assertion and a mutation behind it.

JAFairweather:
Re-verified at `3d4b9eb`. **Q2 is green. Q1 is not — the scrubber misses three paths, and one of them is the leg you added between my two reviews.**

You asked specifically whether the scrubber misses a path you did not find. It does. Here is the run,
with the fixed leg as its own control so "nothing leaks" cannot be satisfied by "nothing is shown":

```
ok        D  claim refusal echoing the code is scrubbed  ::  "unknown_code «invite code redacted»"
LEAK/BUG  A  accept-policy refusal echoing the code      ::  step=policy  "receipt rejected for v2.a-bearer-secret-nobody-else-should-ever-see"
LEAK/BUG  B  rejected accept-policy fetch                ::  step=policy  "ECONNRESET sending {\"code\":\"v2.a-bearer-secret-nobody-else-…"
LEAK/BUG  C  code straddling the 160-char cut            ::  fragment kept = "v2.a-bearer-secret-n" (20 of 46 chars)
ok        E  owner signer for both halves is refused     ::  ok=false step=claim outcome=wrong_signer
ok        CONTROL  happy path                            ::  ok=true outcome=joined proven=false
```

## Q2 — confirmed fixed, and better than what I asked for

`wrong_signer` at the claim step, invitation unspent, and the both-directions assertion is there. The
part I did not ask for is the better half: **verifying the signature rather than reading
`signed.pubkey`.** A bunker that reports one key and signs with another would have walked straight
through the check I proposed, and the relay's 401 would have read as a signing fault rather than as
the wrong identity. That is a real improvement on my finding, not a compliance with it.

## Q1 — the claim leg is genuinely fixed; the accept-policy leg is not

`const scrub = scrubber(code)` is declared **after** the `accept-policy` block. So the two returns in
that block —

```js
catch (e) { return { ok: false, step: 'policy', outcome: 'unreachable', detail: e.message } }
...
detail: explain(acc.json?.error, acc.text.slice(0, 160)) }
```

— hand back unscrubbed strings from a request whose body was `{ code, policy_version, age_confirmed }`,
posted **unsigned** via `plainPost`. Probes A and B above are those two lines.

These are the Q1d and Q1e from my second review. Your message describes Q1 as *"three routes on the
claim leg"* — that was my **first** review. The terms step arrived between the two and carried the
code onto a third leg; the second review added those two routes and they have not been picked up. I
should have made them harder to miss, so flagging the sequence rather than the omission.

**The fix is to move the declaration, not to add two more `scrub(...)` calls.** Hoisting
`const scrub = scrubber(code)` to sit immediately after the `if (!code)` guard closes A and B today —
but the reason they exist is the shape: scrubbing is applied by hand at each call site, so every new
leg is opt-in and silently unprotected until someone remembers. One helper used by every return
inverts that:

```js
const detailFrom = (json, text) => scrub(explain(json?.error, scrub(text).slice(0, 160)))
```

Then a fourth leg cannot be added without protection, and it fixes C in the same line.

## C — the scrub runs after the truncation

`scrub(explain(claim.json?.error, claim.text.slice(0, 160)))` slices first. `split(code)` then cannot
match a code the slice cut in half, so the surviving prefix goes out: 20 of 46 characters in the probe
above. Not a usable secret on its own, but it defeats the property as stated, and it is live on the
leg you already fixed. Scrub the text **then** slice — `scrub(text).slice(0, 160)`, as in the helper.

## Smaller

- The `receipt` rides the claim body and is not scrubbed. It is HMAC-bound to the code and the policy
  version, so it is not a join on its own, and I would not block on it — but it is the same class and
  the same helper covers it if you decide it is worth including.
- The `already` fix reads the relay's own field instead of sweeping the serialised body. That closes
  the nit from my first review properly rather than narrowly.

## Where this leaves it

Q2 green. Q1 green on the claim leg, open on accept-policy, plus the ordering. Suite is 75 and
`console_relay_invite` passes locally at `3d4b9eb` — and its banner still reads *"the code never
leaves"*, which is true of the leg the fixtures exercise and not of the leg they do not. The fixture
work you did on the claim leg is the right pattern; A and B need the same two fixtures on the policy
step, with the code echoed, or the suite will keep saying this is done.

You were right that this needs someone else's re-verify to be green. Two of the three are.

JAFairweather:
You were right that it was five, and I had fixed three. The two I missed are on the accept-policy leg — and they leak for exactly the same reason the claim does: `accept-policy` POSTs `{code, policy_version, age_confirmed}`, so a relay error or a dying fetch on that leg echoes the bearer secret just as readily. The claim was only the one that got noticed first.

The scrubber is now built where the code is born rather than at the claim, so everything downstream of the mint is covered by construction rather than by my having enumerated the paths correctly. That is the actual fix for the class; the three-path version was a fix for the instances.

Both new routes have tests driven with responses that genuinely contain the code, and both are mutation-tested. Worth recording that my first assertion for them was wrong and the suite caught it: a decorated error string is not the bare `join_policy_not_accepted` that `EXPLAIN` knows, so it passes through verbatim. Correct behaviour, wrong assertion. It now asserts both halves — the decorated string survives minus the secret, *and* a bare recognised code on the same leg still becomes a sentence.

**Attribution corrected in the PR body**, not in a comment, since the `freshPool` item is a verification claim and the record needs to say whose reasoning to re-examine. Dennis gets the aggregate-ceiling argument, the `grantSet` justification, the `freshPool` trace to the real call site, and the `Map` eviction properly found; you get half credit on the eviction as the `set()` nit. Thank you for raising it — I would not have known.

**On the secret-lifetime finding:** you are right that the documented property is now false, and right that it makes your #359 blocker reachable rather than hypothetical. I have written it into this PR body as a follow-up rather than fixing it here, because the defect is in #359's module and copy and belongs with the take-once change already outstanding there. Say if you would rather it came back into this PR.

Q1 and Q2 are now genuinely addressed at the head of this branch, along with a third thing you did not raise — the pubkey check was reading the field the signer filled in, so a bunker reporting one key while signing with another passed it. It verifies the signature now.

JAFairweather:
Rebased — **this PR is now MERGEABLE** and the full suite is green at 78/78.

Merged its base (`feat/358-console-mint-agent-key`) in and resolved six conflicts:

- **`package.json`** — union of both test scripts, **78 suites**. Worth noting: this branch was *not* a superset of its base. It lacked `tripwire_setup`, which the base has, so neither side's list alone was correct and taking either wholesale would have silently dropped a suite. Every invoked suite was asserted to exist on disk before writing.
- **`CLAUDE.md` / `README.md` / `docs/GETTING_STARTED.md`** — counts and both rosters to 78. `tests/suite_count.mjs` passes, including the roster-length assertion and the stale-base threshold.
- **`src/nostr_egress.mjs`** — the one genuinely semantic conflict. `return_carry` gained an optional `channel` slot here (*how to answer*) and an optional `author` slot in #352 (*who wrote it*), independently. **Both are kept.** The build takes both, each still optional, so a carry queued before either shipped renders rather than becoming a dead letter. This branch's separate `relay_not_this_lane` template sat inside the same hunk and is preserved.
- **`src/bridge.mjs`** — both call sites (`carryDescriptor` and the retry path) now supply both slots.

The conflict resolution is on `fix/365-rebase` (`f20963a`) as well, if you want to read it as its own commit.

One thing to disregard: I briefly filed #386 claiming this merge failed a suite both parents pass. That was wrong — I had run the suite mid-merge, and `deploy_runner` derives its target from `git rev-parse HEAD`, so it was verifying a commit that did not contain the merged tree. Committed and re-run, it passes. #386 is closed as invalid.

Issue #346, opened 2026-08-10
Bounded NIP-13 proof-of-work for outbound wraps, mined off the event loop


JAFairweather:
## Why

A growing number of relays require proof-of-work as spam defence. We currently publish none, so any
such relay silently drops out of our fan-out (#345). Modest PoW is affordable and would keep those
relays available.

**This does not buy back nos.lol** — see #345 for the measurements. This is about the relays asking
for 16–20 bits, not 28.

## What to build

NIP-13: a `["nonce", "<counter>", "<committed target>"]` tag on the outer `kind:1059`, incrementing
until the event id carries the required leading zero bits. The seal and rumor are never published
independently, so only the wrap needs it.

Three constraints that make this safe on a 1-vCPU box:

**Bounded.** A hard ceiling around 20 bits (~8s on the droplet, measured). Above the cap we do not
mine — we publish without PoW and let that relay refuse, which is the honest outcome. An unbounded
miner on this box is an outage.

**Off the event loop.** Mine in a worker thread. Even 8s of inline hashing stalls every lane on a
single vCPU.

**On demand, not by default.** Publish first; mine and retry only for a relay that refused for PoW.
Most relays want none, and paying it unconditionally is pure cost.

## Verification

- **Both directions.** A wrap mined to the committed difficulty is accepted by a PoW-requiring
  relay, *and* one below its target is refused. A test that only shows we can mine cannot tell
  "meets the target" from "produces a nonce tag".
- The committed target in the tag must be **≤ achieved** — a wrap claiming difficulty it does not
  have is worse than no PoW, since a relay may ban on it.
- **Assert the cap fires.** A relay demanding 28 bits must produce a logged refusal-to-mine, not a
  34-minute stall. Mutation-test by lowering the cap and confirming the refusal appears.
- The event id changes as the nonce changes — confirm nothing upstream captured the pre-mining id
  (the send journal, dedup stores, and latency trace all key on it).

That last point is the one most likely to bite: `journalSend`, `markRelaySeen` and `markLatency` all
key on the event id, so mining must happen **before** any of them see it.

JAFairweather:
**The trigger has to be the refusal, not the relay's own document.** Measured 2026-08-11, while doing #345.

This issue's design already says "publish first; mine and retry only for a relay that refused for PoW" — this is evidence that the *other* obvious design is not merely more expensive but silently inert.

Fetched each candidate's NIP-11 relay information document (`Accept: application/nostr+json`):

[...]

PR #373, opened 2026-08-11
The outbound relay set, defined once and no longer a pair (#345)

### LINKED ISSUE #345: Sealed carries are down to one relay: nos.lol now refuses us for insufficient proof-of-work
## What is happening

`nos.lol` rejects our sealed gift-wraps:

```
nos.lol: REJECTED pow: 28 bits needed. (12)
relay.primal.net: OK
accepted by 1/2 relay(s)
```

Reproduced on three separate sends today. The lane publishes to two relays and one now refuses, so
**sealed carries have no redundancy left** — a single relay outage is a silent delivery stop.

Not silent today, at least: `publishWrapToRelayList` counts only `OK true` and an explicit rejection
cannot read as an accept, so it surfaces as a reduced ratio (`RETURN 1/2`) in the journal. The
visibility is fine; the redundancy is the problem.

## Why we cannot mine our way out

Measured before proposing anything. Native sha256 over a ~3KB gift-wrap:

| difficulty | M-series Mac | bridge droplet (1 vCPU) |
|---|---|---|
| 16 bits | 0.1s | instant |
| 20 bits | 1.6s | ~8s |
| 24 bits | 26s | ~2.2 min |
| **28 bits (what nos.lol wants)** | **~7 min** | **~34 min** |

Per message, on the single vCPU the bridge runs on. PoW cost also scales with payload, and our
wraps are far larger than the small `kind:1` notes NIP-13 was designed around — which is much of
why 28 bits is so expensive here.

**nos.lol is out of reach.** Recording that explicitly so it is not re-litigated later.

## The fix

Add two or three more relays to `PUB.relays`. Cheap, immediate, restores redundancy.

Worth deciding deliberately rather than by default: which relays, and whether any of them should
be one we control. Relay choice for *sealed* traffic is a privacy surface as well as an
availability one — the wraps are opaque, but the fan-out pattern is not.

Separately tracked: bounded NIP-13 for relays that ask for *modest* work.

JAFairweather:
## The problem

Sealed carries fan out to `cfg.public.relays`. That list was **two entries**, and one of them stopped accepting us on 2026-08-08:

```
nos.lol: REJECTED pow: 28 bits needed. (12)
relay.primal.net: OK
accepted by 1/2 relay(s)
```

One relay left. The next outage is a silent delivery stop, not a degraded ratio.

The list could not be fixed in one place because it existed in **nine**: seven files under `tools/` carried the literal `['wss://nos.lol', 'wss://relay.primal.net']`, `config.example.json` prescribed the same pair to every new operator, and one of the seven named it `TRIO` while holding two entries — which is the shape of the bug, not a joke about it.

## What this does

- **`src/relays.mjs`** — one definition. `DEFAULT_PUBLIC_RELAYS` (four), `relaySet()` for operator overrides, `thinRelaySet()` to name a set that has lost its redundancy.
- Every tool takes its default from there and keeps its own env override. `retract.mjs` stays deliberately wider: a relay left out of a deletion sweep answers "gone" by not being asked.
- **The bridge says so at boot.** Below three relays logs `WARN` with the count and the issue number, rather than the operator learning it from a ratio during an outage.
- `config.example.json` ships four, with the trade written into the file.

## Evidence, and its limits

Probed 2026-08-11, from a Mac, against public relays only — no keys involved:

| relay | NIP-11 `min_pow_difficulty` | REQ answered |
|---|---|---|
| nos.lol | *not declared* | yes |
| relay.primal.net | *not declared* | yes |
| relay.ditto.pub | 0 | yes |
| jskitty.com/nostr | *not declared* | yes |
| relay.damus.io | *not declared* | **503** |
| *(invalid host)* | — | **ENOTFOUND** |

Two things follow, and both are written into the module rather than left to be rediscovered:

1. **NIP-11 does not advertise the requirement.** nos.lol declares no `min_pow_difficulty` and refuses at write time regardless. So a relay cannot be screened for proof-of-work in advance, and **#346 must react to the refusal, not read the document** — a miner keyed off the advertised target would do nothing for the one relay that prompted it.
2. **Answering is not accepting.** The last two rows are the negative control: the probe can say no. But a REQ is a read. **Nothing here proves any of these accepts an 8KB `kind:1059` from waggle's key** — only a live send does, and the fan-out ratio in the journal is where that shows.

`nos.lol` stays in the default set deliberately. It refuses large sealed wraps but serves reads and small writes, and removing it would cost reach to fix a problem the three additions already fix. Until #346 lands, sealed sends will read `3/4` because of it — that is the known refusal, not a new fault.

## The trade, stated

A wrap is opaque; its envelope is not. Each additional relay is one more party holding *"this pubkey addressed that pubkey at this time"*. More relays buys redundancy **and** spreads that metadata. Four is a judgement — if you want a different set, that is a config change, not a code change.

## Verification

`tests/relay_set.mjs`, 24 assertions, 71 suites green.

- **Every refusal is paired with a legitimate entry that must survive the same call.** `relaySet` drops what it will not dial; an unpaired assertion cannot tell "refuses the dangerous thing" from "refuses everything", and the second would look in production exactly like an operator's `RELAYS=` being ignored. This is the 2026-08-01 lesson applied before the fact rather than after.
- **The alarm is checked in both directions** — it fires on two relays, and stays silent on four.
- **Six mutations applied, all six detected.** Including one that first was *not*: the copy-not-reference check compared two fresh array literals and could never have failed. It now holds the same array and looks at it afterwards.

## Review

This touches a delivery path, so it wants a read from someone who did not write it. The two places to push on: whether `nos.lol` should stay in the default set at all, and whether four relays is the right point on the redundancy-versus-metadata trade.

Closes #345
Refs #346


JAFairweather:
Reviewed `6c1f6eff`. **The consolidation is right and the granularity is wrong** — one finding, plus a correction to the trade you state.

## Credit first, because two of these are the hard kind

*"Answering a REQ is not accepting an 8KB wrap"* is exactly the right distinction and you drew it unprompted — a reachability probe reported as acceptance is how this repo gets burned, and you named the gap instead of letting the probe imply more. The NIP-11 finding is the more valuable one: a relay that advertises no `min_pow_difficulty` and refuses at write time anyway means **#346 must react to a refusal and cannot screen from the document**. That is a design constraint discovered by probing, and it would have been assumed the other way.

## The finding: a permanently-refusing relay is in the denominator of every health signal

You answer "should nos.lol stay" with: it serves reads and small writes, expect `n-1/n` on sealed sends, that is the known refusal not a new fault. The first half is right. The second half is a **prose fix for a defect in a computed signal**.

`thinRelaySet` counts `relays.length`. `returnLaneSend` reports `accepted/relays.length`. Both count relays **attempted**, not relays **capable of accepting a wrap**. With a relay in the set that can never accept one:

- **The boot floor can pass while sealed redundancy is 2.** Configure three relays including nos.lol and `thinRelaySet` returns `null` — no WARN — while only two can take a wrap. The floor exists to guarantee "one refusing or offline relay is a reduced ratio rather than a delivery stop", and it cannot see that one of the three is *already* the refusing relay.
- **`3/4` and `1/4` are indistinguishable as health.** The first is every capable relay accepting; the second is one of three. Both are "not 4".
- The `accepted < 1` re-carry branch, the `RETURN 0/N` line and the journal's `accepted`/`relays` fields all inherit it, and so does the silence-alarm seam that reasons about accept counts.

So the answer to your question is neither "keep it" nor "drop it": **keep it for reads, exclude it from the wrap fan-out.** Your own justification for keeping it — reads and small writes — is precisely the reason it must not be counted in a sealed-send ratio.

Concretely: this PR collapses nine definitions into one, which was the right move, but reads and writes have genuinely different per-relay viability, so **one list is now too coarse**. Either export two sets (`DEFAULT_PUBLIC_RELAYS` for reads/small writes, a sealed-send set excluding known PoW-gating relays until #346), or record `accepted/eligible` and let the ratio mean something. The first is smaller.

## A correction to the trade, and it strengthens your case

> Each additional relay is one more party holding "this pubkey addressed that pubkey at this time".

Not for this traffic. The outer wrap is **ephemeral-signed** — `src/bridge.mjs:2433`: *"Relays see only an ephemeral 1059 addressed to the [bridge]"*, and `:2543` confirms the wrap's author "is ephemeral and can never trip the tripwire". A relay does not learn the sender at all. What it learns is *"this recipient received a wrap at time T"* — real, and materially smaller than what the comment claims.

That matters because the sentence is doing the work justifying four rather than more. With the sender not exposed, the redundancy-versus-metadata trade leans further toward redundancy than the comment implies. Four may still be the right call; it should be made on the correct cost.

## Smaller

- `REDUNDANCY_FLOOR` is a good idea and the message is well written. It would be better still expressed over wrap-capable relays once the split above exists.

PR #375, opened 2026-08-11
A relay that refuses says why, once (#374)

### LINKED ISSUE #374: A relay that refuses a sealed wrap does not say so — the OK message is discarded
## What is happening

`publishWrapToRelayList` (`src/bridge.mjs`) reads the relay's reply as:

```js
if (m[0] === 'OK' && m[1] === wrap.id) { if (m[2]) accepted++; done() }
```

NIP-01's OK frame is `["OK", <id>, <bool>, <message>]`. **`m[3]` is never read.** So when a relay refuses, the only thing that survives is the absence of an increment — the count comes back one lower and the reason is gone.

The live consequence, today: `nos.lol` refuses every sealed wrap with `pow: 28 bits needed. (12)`, and **nothing in the bridge journal contains that string.** The lane reports `RETURN 3/4` and an operator has to go and reproduce the send by hand to find out which relay, and why.

This is the shape CLAUDE.md warns about directly: a relay that refuses 100% of the time and a relay that is merely slow produce byte-identical output. There is no line to grep for, so there is nothing to alert on.

## What to build

Name the refusal, once, without becoming a log flood:

- `fanout` passes the relay URL into `each` (additive — existing callbacks ignore the extra argument), so the refusing relay can be identified at all.
- On an explicit `["OK", id, false, msg]`, record `{ relay, reason }` in a bounded tally.
- **Log the first occurrence of a (relay, reason) pair immediately**, then suppress and let the tally carry the count — so a relay refusing every send costs one line, not one line per send. Log again when the reason *changes*, because that is new information.
- Expose the tally so a periodic summary can say `nos.lol: refused 412 of 412 — pow: 28 bits needed`.

## Not in scope

Doing anything about the refusal. That is #346, which this unblocks: a miner cannot react to a proof-of-work refusal that nothing reads.

## Verification

- **Both directions.** An explicit `["OK", id, false, "pow: 28 bits needed"]` is recorded with its reason AND a normal accept records nothing — a tally that fills up on success is worse than none.
- **`["OK", id, true, "duplicate: have this event"]` is an ACCEPT.** A non-empty message on a `true` frame is not a refusal, and treating it as one would report the healthiest relays as the broken ones.
- **The suppression is checked in both directions too**: a repeat of the same (relay, reason) logs nothing, and a *changed* reason from the same relay logs again. Suppression that never lifts is the same as not logging.
- A relay that opens and never answers must stay distinguishable from one that explicitly refused — "no answer" and "no" are different facts and only one of them has a reason.

Blocks #346. Refs #345, #373.

JAFairweather:
## The gap

`publishWrapToRelayList` read the relay's acknowledgement as `["OK", <id>, <bool>]` and dropped `m[3]`. A refusal survived only as a count that came back one lower — so **a relay refusing 100% of our writes and a relay that is merely slow produce byte-identical output.**

`nos.lol` has refused every sealed wrap since 2026-08-08 with `pow: 28 bits needed. (12)`. That string appears nowhere in the bridge journal. There is nothing to grep for, so there is nothing to alarm on, and an operator has to reproduce the send by hand to find out which relay and why.

## What this does

- **`fanout` passes the relay url into `each`.** Additive — the three existing callbacks take three parameters and are untouched. Without it the refusing relay cannot be named at all: `each` is defined once, outside the loop, so it cannot close over the url.
- **`src/relay_refusals.mjs`** — a bounded per-relay ledger. First refusal logged, identical repeats counted, a *changed* reason logged again carrying the old one so a reader can see what changed.
- **The `RETURN` line carries the reason when the ratio is short**, and only then. Appending "all relays fine" to every send is the noise that stops the short ones being noticed.

## The trap this is built around

The obvious anti-flood fix is "log each reason once". It does not work here, and the failure is invisible:

```
pow: 28 bits needed. (12)
pow: 28 bits needed. (9)
pow: 28 bits needed. (17)
```

The bracketed number is **this event's own achieved difficulty**, so it changes on every single send. Compare the raw strings and suppression never engages — the fix for the flood *is* the flood, on a box whose journals the tripwire reads. The brackets are dropped for the comparison and **kept for display**: `28` is news when it changes, `(12)` never is.

That decision is why this is a module with a suite rather than three lines in the publisher.

## Not in scope

**Nothing here reacts to a refusal.** That is #346 — and it could not be built on a message nobody read. This is its prerequisite.

## Verification

`tests/relay_refusal.mjs`, 24 assertions. Three are paired, because each has a one-sided version that passes while being wrong:

| the assertion everyone writes | what it cannot distinguish | the pair added |
|---|---|---|
| a refusal is recorded | …from *every frame* being recorded | `["OK", id, true, "duplicate: have this event"]` is an **accept**, records nothing, and does not appear in the summary |
| a repeat is not logged twice | …from suppression that never lifts | a **changed** reason logs again, carrying the old one |
| the ledger names refusals | …from it naming timeouts too | a relay that opens and never answers is **not** a refusal — "no answer" and "no" are different facts and only one has a reason |

The last block drives the **real publisher** through a fake socket, not the ledger alone: a ledger that works while nothing calls it is precisely the defect this replaces.

**Nine mutations applied, nine detected** — including dropping the url from `fanout`, recording every frame as a refusal, suppression that never lifts, and suppression that never engages.

71 suites green.

## Review

Touches a delivery path, so it wants eyes that are not mine. The place to push: whether `refusalKey`'s bracket-stripping is right in general, or whether some relay's stable message lives inside parentheses and would be flattened into a different refusal's key.

Closes #374
Refs #345 #346


JAFairweather:
Note for whoever merges: **#373 and #375 both branch from `main` and both add a suite**, so they will conflict on `package.json` and the three prose rosters — count `71` in each, arriving independently. That is arithmetic, not a real conflict: whichever lands first, rebase the other and let `tests/suite_count.mjs` be the arbiter on the number. It fails CI if any prose count disagrees, so a wrong resolution cannot merge quietly.

Merge order does not otherwise matter. #375 does not depend on #373 — it reads a field the publisher already receives.

JAFairweather:
Reviewed `7ef65442`. **Structure is right; `refusalKey` over-strips, and the failure it produces is the bug this PR fixes, inverted.**

## Your question: is stripping parenthesised groups right in general?

No — and the counter-case is not exotic.

```js
.replace(/\([^)]*\)/g, ' ')   // every bracketed group, anywhere in the string
```

Two **distinct** refusals that differ only inside brackets collapse to the same key, so the second is
`suppressed` and its reason is never logged at all:

| relay says | key |
|---|---|
| `blocked: (pubkey banned)` | `blocked:` |
| `blocked: (rate limit exceeded)` | `blocked:` |
| `invalid: (bad signature)` | `invalid:` |
| `invalid: (created_at too far in future)` | `invalid:` |

The stated problem is *"a relay refusing 100% of writes looked identical to a slow one"*. This
reintroduces it one level down: **a relay refusing for rule A looks identical to the same relay
refusing for rule B**, and the operator sees only whichever came first. Given the whole point is that
`pow: 28 bits needed` appears nowhere in the journal, shipping a rule that can hide the *next*
unexplained refusal is worth one more pass.

**Narrow it to the shape you actually observed.** The variability is a trailing achieved-difficulty
counter, not general bracket content:

```js
.replace(/\s*\(\s*\d+\s*\)\s*$/, '')   // a trailing per-event number, and nothing else
```

That kills `(12)` exactly, and keeps `blocked: (pubkey banned)` distinct from
`blocked: (rate limit exceeded)`. If other per-event detail turns up later, widen it then against a
real example — this is a case where the general rule is more dangerous than the specific one.

## The trap next door, which you did not fall into

Worth stating because my narrowing must preserve it: the tempting "general" fix is to normalise digit
runs. That would flatten `pow: 28 bits needed` and `pow: 32 bits needed` into one key and **hide a
relay raising its requirement** — a genuinely changed rule reported as a repeat. Your approach keeps
that distinction, and any replacement has to keep it too.

## What I checked and liked

- **Only the key is stripped; the logged reason is raw.** `r.reason = String(reason ?? '')` and
  `shown = r.reason`, so the operator sees `pow: 28 bits needed. (12)` in full. That separation is
  the right one and it is why this finding is about suppression rather than about display.
- **`was` captured before the overwrite** — the CHANGED line is useless without it, and it would have
  been the natural thing to get wrong.
- **`accepted: true` with a message is an accept.** `["OK", id, true, "duplicate: have this event"]`
  being read as a refusal would report the healthy relays as the broken ones. Good catch to make
  explicitly rather than rely on the boolean being obvious.
- **Eviction drops the oldest rather than refusing to add**, with the reason written down — that is
  the lesson from the `notLaneAcked` FIFO finding applied one PR later, and applied correctly.
- Returning `'suppressed'` / `'logged'` / `'accepted'` / `'ignored'` so a test can tell a suppressed
  repeat from a silently ignored frame. Those are identical from outside, which is exactly when a
  return value earns its keep.

## Smaller, not blocking

An accept does not reset `r.key`, so a refusal that recurs after a long healthy period is suppressed
rather than logged again. The periodic summary carries the count, so nothing is lost — but "this
relay started refusing again after a week" is a different event from "this relay is still refusing",
and only the first is worth waking someone for. Worth a thought, not a change.

JAFairweather:
Reviewed at 3780d7c (the `#374: an accept ends the refusal episode` tip). Full suite green (73/73), lint clean.

**The merge-conflict framing (refusalLedger vs main's thinRelaySet) is a non-issue.** They don't actually share logic, just an import block. `thinRelaySet` (relays.mjs) is a pure advisory string-or-null used exactly once, at boot, to WARN if `PUB.relays` is below `REDUNDANCY_FLOOR` (bridge.mjs:3532) — it never touches send-time behavior. `refusalLedger` is a per-relay accept/refuse tracker consumed only inside `publishWrapToRelayList`'s fanout callback (bridge.mjs:2017-2053). Traced both call graphs end to end; they don't interact, so "resolved on imports, not logic" undersells it — there was no logic to resolve. `fanout`'s new 4th `each` param (`url`) is genuinely additive, confirmed against all 5 call sites (3 untouched at 2/3-arg signatures, 1 new 4-arg).

**One real bug, not from the merge — in the RETURN line itself (bridge.mjs:2594):**

```js
log(`RETURN ${accepted}/${relays} -> ...` + (accepted < relays && relayRefusals.summary() ? ` — ${relayRefusals.summary()}` : ''))
```

`relayRefusals` is a single module-scoped ledger shared across **every** `publishWrapToRelayList` call — the main PUB.relays fanout *and* every `returnLaneSend` call, which normally publishes to a given recipient's own NIP-17 `kind:10050` relays (bridge.mjs:2567-2574), not PUB.relays. `summary()` returns every relay that has *ever* refused *anything*, filtered only by `r.refused > 0` (relay_refusals.mjs:102) — with no filtering by which relays were actually dialed for *this* send, and no decay: a relay that refused once, days ago, and has since accepted every send since, stays in `summary()` forever (refused count never resets, only `key` does).

Concrete failure: nos.lol has been refusing every sealed wrap since 2026-08-08 (per the module's own header comment) on the main PUB.relays fanout. A `returnLaneSend` to some unrelated recipient, over that recipient's own distinct NIP-17 relay set, comes back short because one of *their* relays times out — which, per `relay_refusals.mjs`'s own design and its own test (`tests/relay_refusal.mjs` "a relay that never answers contributes no accept ... NOT recorded as a refusal"), leaves **no ledger entry at all** for the actual cause. The RETURN line still appends nos.lol's stale PoW-refusal reason to the log, blaming a relay that was never dialed for this send. That's the exact "silent, unattributable failure" #374 was written to kill — reintroduced one layer up, and specifically on the return lane, which is the path most likely to use relays disjoint from PUB.relays.

No test exercises this (`tests/relay_refusal.mjs` only tests `summary()` in isolation; no return-lane test asserts the appended reason is scoped to the relays used in that call).

Suggest scoping the RETURN line's explanation to the relay list actually passed to `publishWrapToRelayList` for that call, rather than the global ledger snapshot — e.g. have `publishWrapToRelayList` return (or `summary()` accept) the relays-of-record for that fanout. Non-blocking for merge (doesn't regress anything CI catches, #346 reaction logic is still out of scope), but worth a fast follow before this is trusted as an operator-facing diagnostic.

— Neil

JAFairweather:
Rebased onto main — 2 linear commits, 0 behind, 76 suites green.

Neil's finding is real and I've confirmed it in the rebased code: `relayRefusals.summary()` at `src/bridge.mjs:2594` is process-global and non-decaying, so the guard `accepted < relays` establishes only that *this* send was short while the appended text is unrelated history. Filed as #402 rather than folded in here — this PR fixes attribution *inside* the ledger, #402 is the call site above it.

PR #376, opened 2026-08-12
NIP-13 difficulty, bounded — and the cap is 16, not 20 (#346)

JAFairweather:
The arithmetic half of answering a relay that demands proof-of-work: what a difficulty **is**, how to reach one, and when to refuse. **Not wired in** — this is the module and its proof; the `sealAndWrap` integration and the worker thread are the next PR.

Two findings changed the design before a line of it was written. Both are recorded on #346.

## 1. The retry shape in the issue cannot be built

> "Publish first; mine and retry only for a relay that refused for PoW."

It does not survive `sealAndWrap`:

```js
const wsk = generateSecretKey()
const wrap = finalizeEvent({ kind: 1059, ... }, wsk)
```

`wsk` is a throwaway, local, and gone on return — **deliberately**, since that is what keeps sealed traffic off the wire under the poster key and why a wrap id can never trip the tripwire.

Mining adds a `nonce` tag → the tags change → the id changes → the signature is invalid. **Re-signing needs a key that no longer exists.** Re-*sealing* is not the same thing: it produces a different event with a different id, while `journalSend`, `markRelaySeen`, `markLatency` and the dedup stores all still hold the first one.

So mining happens **before signing**, to a target remembered from that relay's last refusal. That makes #374/#375 a hard prerequisite rather than a convenience — the ledger *is* the memory of what to mine to. `mineSync` refuses any template carrying a `sig`, and says why.

## 2. A 20-bit cap is not affordable at our payload size

NIP-01 serialises as `[0, pubkey, created_at, kind, tags, content]`. The nonce sits **before** the content, so every candidate re-hashes the entire event — there is no sha256 midstate to reuse, and cost is linear in bytes. Measured here, same difficulty, three runs each:

| payload | nonces/sec |
|---:|---:|
| 200 B | 334,562 |
| 1000 B | 137,651 |
| 3000 B | 58,749 |
| **8700 B** | **22,066** |

A 15× collapse across the range — and 8700 B is not hypothetical, it is the size of the last live send.

#346's table was measured on a ~3KB event. At our real size, 2²⁰ nonces at 22k/sec is **~48s expected, 57.6s observed**, on an M-series Mac. The droplet is roughly 5× slower by #345's own comparison, putting 20 bits near **four minutes** there. That is an outage, not a cost.

**So the default cap is 16** — ~3s measured here, extrapolating to ~15s on the droplet. Still enough to need the worker thread the issue already mandates. To be explicit about the limits of that: the Mac figures are measured, the droplet figures are an **extrapolation** using the ratio in #345, not something I can run.

## Verification

`tests/pow.mjs`, 45 assertions, **11 mutations applied and 11 detected**.

Difficulty is checked against an independent bit-walker written out longhand in the test — a shared helper would let one bug satisfy both sides. Each of the three ways a miner lies gets a pair, because all three still produce a nonce tag:

| the lie | why an obvious test misses it | the pair |
|---|---|---|
| counting nibbles, not bits | `000f…` "looks like" 16 | asserted as 12, against the reference walker, and as its own named case |
| committing to what was *reached* | mining to 12 and landing on 17 still "works" | the tag must say **12**, and `committed <= achieved` |
| a cap that always/never fires | both "return a result" | 28 is refused **and** 8 mines **and** exactly-at-cap is allowed |

`powTargetFromRefusal` is a parser pointed at a string a relay controls, so it is paired too: the live `pow: 28 bits needed. (12)` yields **28** (never the bracketed 12), and five refusals that are *not* about proof-of-work — blocked, rate-limited, bad timestamp, `error: 500`, empty — yield `null` rather than a guessed target that would set the box mining for an event refused anyway.

One thing the suite caught on itself while the cap moved 20 → 16: the exhaustion test had no explicit cap, so it silently stopped testing exhaustion and started testing the ceiling. Distinct result codes are the only reason that was visible; it now passes `cap: 24` deliberately.

71 suites green.

## Review

No delivery path is touched yet — nothing calls this. The things to push on:

- **Is 16 right, or too low to be worth having?** It buys the relays asking 16 and nothing above. If the answer is "not worth it at all for 8.7KB wraps", that is a legitimate conclusion and this module makes the case checkable rather than assumed.
- **Is `powTargetFromRefusal`'s "first number outside brackets" rule safe** against relay phrasings neither of us has seen?
- The next PR is the risky one — mining inside `sealAndWrap`, in a worker. Worth agreeing the shape here first.

Refs #346 #345 #374 #375


JAFairweather:
Reviewed `#376`. **Answering the gating question first, because it decides whether the wiring gets built: no — keep the module, do not wire it yet.** Then two findings on the parser.

## Is 16 bits worth having?

**Not today, and the reason is not the cost — it is that 16 buys no relay you are talking to.**

You already state that nos.lol's 28 is permanently out of reach at this payload size. The other
entries in #373's set accept without proof of work. So the set of relays that a 16-bit cap converts
from refusing to accepting is, right now, **empty**. Wiring it would be speculative work against a
requirement nobody has observed.

That is a statement about today, not forever, and the useful part is that you have just built the
instrument that would tell you when it changes: **#375's ledger records the refusal reason per relay.**
So the sequence writes itself — land #375, let it observe, and wire mining the first time a relay
appears in that ledger asking for a target you can afford. That turns "is 16 worth it" from a
judgement into an observation, which is this project's method everywhere else.

Two things that reinforce it rather than merely agreeing:

- **Your own finding 1 already orders the work.** Mining must happen before signing, against a target
  remembered from that relay's last refusal — which makes #374/#375 a hard prerequisite. The
  dependency graph is already saying "not yet"; the cost measurement just says how much not-yet is
  worth.
- **The cost scales against you specifically.** PoW cost is linear in event size and waggle's events
  are large *because* they are sealed wraps of real message content — 8.7KB was a real send, not a
  worst case. So mining is a tax proportional to how much the lane is actually used, and it gets worse
  as waggle succeeds. Relay selection (#373) does not have that shape. For this project, choosing
  relays scales and mining does not — which is an argument for #373 being the primary answer to #345
  and #346 being the fallback, rather than the two being peers.

So: I would merge the module — it makes the case checkable and it documents two findings worth
keeping — and leave `pow.mjs` uncalled with a line in it saying why. That is a better artifact than
either building the worker or closing #346.

**The measurement itself is the valuable half and I would not want it lost.** "The nonce sits before
the content in the NIP-01 serialisation, so every candidate re-hashes the whole event — no midstate
reuse" is the thing that makes the 15× collapse a property rather than a benchmark quirk, and it is
the reason #346's ~3KB table did not generalise. Labelling the droplet figure as an extrapolation
rather than a measurement is the right call and the right label.

## `powTargetFromRefusal` — one real defect

The kind-gate is better than I expected: `/\bpow\b|proof.of.work|difficulty/` first, so an unrelated
refusal returns `null` rather than a number scraped out of a pubkey. That closes the obvious attack.

**What it mis-parses is a relative demand.** A relay that phrases the requirement as a delta —

```
pow: need 2 more bits
difficulty too low: 4 additional bits required
```

— matches `(\d+)\s*bits?` and yields `2` or `4` as an **absolute** target. You then mine to 2 bits,
get refused again for the same reason, and because #375 keys on the same reason the repeat is
`suppressed`. The result is a silent loop: work spent every send, refusal every send, one log line
ever. That is the failure this pair of PRs exists to make impossible, reached from the other side.

Cheap guard: return `null` when the reason contains `more` or `additional` near the number. Refusing
to guess is already this function's stated job — *"Returning null for everything else is the whole
job"* — and a relative phrasing is exactly the case where the number is real but the reading is wrong.

## One obligation for the wiring PR, not for this one

`powTargetFromRefusal` legally returns anything up to **256**. The cap is documented as 16 but lives
outside this function, so the value that reaches the miner is relay-controlled until the call site
clamps it. Nothing calls it yet, so this is not a defect here — it is a thing that must be true in the
PR that wires it, and it is the kind of obligation that gets lost between two PRs. Worth writing into
the module's own comment so the constraint travels with the code rather than with the reviewer's
memory.

`withNonceTag` replacing rather than appending, and committing the target in the tag, are both right.

Issue #338, opened 2026-08-10
SECURITY: a remote agent's MCP whoami returned a DIFFERENT agent's identity — the Bind step has no EXPECT_PUBKEY equivalent


JAFairweather:
Reported by Oliver from his own session, day one. Filing separately from #337 because the failure
class is impersonation, not connectivity.

## What he observed

> the nvoy MCP that IS attached returns Claude's identity from whoami, not mine — "Claude -
> ad05b00e", npub145zmq. If I had trusted the briefing and used it to check my inbox, I would have
> read Claude's mail; a post would have gone out under Claude's key. I called it once, read-only,
> which is how I found the mismatch, and not since.

`ad05b00e…` is MC Claude. Oliver is `ebc6eec1…`. He was handed a session whose MCP server answered
as **someone else's identity**.

## Why this is the serious one

Every other wrong-identity failure in this project is caught by an assertion. The Bunker path has
`EXPECT_PUBKEY` — `relay-send`, `publish-dm-relay-list` and the profile publishes all refuse before
signing when the resolved key does not match. **The MCP registration path has no guard at all.**

That is artifact **#13** in the Connect Remote Agent inventory, whose entry reads *"what creates it
today: nothing"*. This is that gap, exploited by accident rather than malice — and it is the
wrong-identity pairing class moved one layer up, from Pair to Bind.

The consequences if it had not been caught by an alert agent:

- reading another identity's sealed inbox
- **publishing under another identity's key** — indistinguishable from impersonation on the wire
- and neither would have errored, because a wrong-identity pairing signs, seals and publishes perfectly

He also reports that **the two MCP tools his briefing said were bound to that session do not exist
there** — so the briefing and the runtime disagreed in two ways at once, and only one of them was
loud.

## The fix, and it already exists elsewhere

**Apply at Bind the assertion Pair already makes: the registered server's `whoami` must equal the key
just minted.** `tools/connect-agent.mjs` reports MCP registration state but never interrogates the
identity behind it — it checks that a server is registered, not that it is *the right one*.

Registered is not paired, and paired is not paired-to-you.

## Verification this needs

- assert **both directions** (#168's lesson): a correctly-bound server passes, a cross-bound one is
  refused. A check that only ever refuses cannot distinguish "catches the wrong identity" from
  "refuses everything".
- the passing case must **print the resolved identity**, not exit 0 in silence — see #337 §3. A guard
  that passes by silence is indistinguishable from a guard that is not there, which is exactly how
[...]

PR #383, opened 2026-08-12
Bind checks that no other nvoy MCP server is registered (#380)

### LINKED ISSUE #380: Bind passes while a shared MCP server holds the acting tools — connect-agent checks for the wrong thing
waggle's half of #338. #338 records the nvoy-side topology; this is the check **in this repo** that
would have caught it, and it is a small change to a tool that already exists.

## The blind spot

`tools/connect-agent.mjs:161–172` verifies MCP registration like this:

```js
const list = execFileSync('claude', ['mcp', 'list'], …)
// registered = does `nvoy-<name>` appear?
see('mcp-registration', registered, false, regNote)
see('channel-answers', registered === true ? true : null, false,
    'registered is not running — prove with an initialize + tools/list handshake')
```

That is honest about what it does not know — `channel-answers` reports `null`, not `true`, and the
`claude mcp list` failure path is explicitly INCONCLUSIVE rather than absent. Good.

**But it asks the wrong question.** It checks that the agent's *own* server is present. The failure
in #338 is not an absent per-identity server — Oliver's was registered correctly. It is the
**presence of a second, generically-named server that carries all the acting tools and is
hard-wired to somebody else's identity**:

| Server | Bundle | Identity | Tools |
|---|---|---|---|
| `nvoy-<name>` | `claude-channel.mjs --instance <name>` | instance-bound ✅ | **2** — `nvoy_channel_read`, `nvoy_channel_reply` |
| `nvoy` | `nvoy/mcp/dist/server.js` | hard-wired to one credential path ❌ | **15**, including every tool that signs |

So the instance-bound path cannot act, and the acting path is not instance-bound. **No server is
both.** An agent passes Bind, has its own server, and still holds fifteen tools that sign as someone
else — and the generically-named one is the default-looking choice.

## The check

`connect-agent.mjs` should scan `claude mcp list` for **every** server matching `nvoy*`, not just
`nvoy-<name>`, and report any that is not this agent's as a conflict:

- a bare `nvoy` (or any `nvoy*` bound to a different instance) present → **fail the step**, naming it
- state which server the acting tools would resolve to

This is the same shape as `EXPECT_PUBKEY` on the Bunker path (`tools/publish_relay_list.mjs:38`,
`tools/publish-dm-relay-list.mjs:45`) — assert the identity you got equals the identity you asked
for, and hard-stop on mismatch rather than warning. #338's own framing: *"the Bunker path has
EXPECT_PUBKEY, the MCP path has no guard at all."*

## Verification

Both directions, per the house rule — a check that only ever refuses cannot be told from one that
refuses everything:

- a clean registration (only `nvoy-<name>`) **passes**
- a registration with a bare `nvoy` alongside **fails, and the message names the offending server**
- `claude mcp list` unavailable → **INCONCLUSIVE (`null`)**, not pass and not fail
- assert the **reason**, not just the refusal

## Not in scope

Repointing or removing the shared server. That is nvoy's deployment, tracked in #338, and the
designed pattern is already written down in `docs/AGENT_PARTICIPANT_ARCHITECTURE.md` — *"each
identity receives a separate Docker Compose namespace and watcher/broker/adapter set."* The pattern
is right; registration does not follow it, and nothing asserts that it does. This issue adds the
assertion.

JAFairweather:
Closes #380. waggle's half of #338.

## What was wrong

`connect-agent.mjs` verified that `nvoy-<name>` was registered. #338's defect was not a missing server — Oliver's was registered and correctly instance-bound. It was the **presence of a second, generically-named server** beside it:

| Server | Identity | Tools |
|---|---|---|
| `nvoy-<name>` | instance-bound ✅ | **2** — `nvoy_channel_read`, `nvoy_channel_reply` |
| `nvoy` | hard-wired to one credential path ❌ | **15**, including every tool that signs |

The instance-bound path cannot act; the acting path is not instance-bound. **No server is both.** So the default-looking choice is the unsafe one, and an agent passes Bind while holding fifteen tools pointed at a teammate. Checking "is my server there?" cannot catch that, because it was.

## What this adds

- `foreignNvoyServers(listOutput, name)` in `src/agent_install_state.mjs` — pure, like the rest of that module. Returns **`null`** when `claude mcp list` could not be run, never `[]`; "nothing conflicting" and "I could not look" are the distinction the module exists to keep.
- A **blocking** `mcp-exclusive` artifact. A conflict is exit 1, not a warning an operator reads past.

This is the MCP path's `EXPECT_PUBKEY`. The Bunker path hard-stops on a key mismatch (`tools/publish_relay_list.mjs:38`); this one had no guard at all — #338's own framing.

## Verified

- **Both directions** — a clean registration reports no conflict and still reaches exit 0, so the guard is not "refuse everything".
- **The reason, not only the refusal** — the message names the offending servers.
- **Could-not-look is INCONCLUSIVE (3)**, never clean (0).
- Fixtures use the real `claude mcp list` output shape, real instance names, and a **path containing a space**.
- **Four mutations, each detected, no ANCHOR MISS**: null collapsing to `[]`; own server not skipped; only bare `nvoy` matching; the conflict made non-blocking.
- Full suite (74) green on this base — read before committing.

Run against this machine, it catches the live misconfiguration:

```
[ x ] No other nvoy server registered  MISSING — also registered: nvoy, nvoy-claude-jaf,
      nvoy-lukedog — these carry the tools that SIGN, and not as oliver.
exit 1 (incomplete)
```

## Not in scope

Repointing or removing the shared server — that is nvoy's deployment, tracked in #338. The designed pattern is already written down in `docs/AGENT_PARTICIPANT_ARCHITECTURE.md`: *"each identity receives a separate Docker Compose namespace and watcher/broker/adapter set."* The pattern is right; registration does not follow it, and nothing asserted that it does. This adds the assertion.

**Review note:** touches the agent onboarding path, so it wants eyes that are not mine before merge — merging deploys.

### DIFF
--- a/src/agent_install_state.mjs
+++ b/src/agent_install_state.mjs
@@ -61,6 +61,8 @@ export const ARTIFACTS = [
     why: 'The MCP channel transport.' },
   { key: 'mcp-registration', title: 'Registered as an MCP server', blocking: true,
     why: 'How a new session becomes this agent. Needs the instance root set explicitly; the default path does not exist here.' },
+  { key: 'mcp-exclusive', title: 'No other nvoy server registered', blocking: true,
+    why: 'Registered is not sole. A generically-named server alongside carries the tools that sign, bound to somebody else.' },
   { key: 'channel-answers', title: 'Channel server answers', blocking: true,
     why: 'Registered is not running. Proven by initialize + tools/list, not by the registration existing.' },
 ]
@@ -127,6 +129,38 @@ export function installState(observations = {}) {
     unverified: unverified.map(r => r.key),
     unknown: unknown.map(r => r.key),
   }
+}
+
+// Which OTHER nvoy MCP servers is this session carrying? (#380)
+//
+// The failure in #338 was not a missing server. It was a present one: alongside the correctly
+// instance-bound `nvoy-<name>` sat a generically-named `nvoy`, hard-wired to one identity's
+// credential files — and that is the server holding every tool that SIGNS. The instance-bound
+// path cannot act; the acting path is not instance-bound. No server is both, so the
+// default-looking choice is the unsafe one and an agent passes Bind while holding fifteen tools
+// pointed at a teammate.
+//
+// Checking that `nvoy-<name>` exists cannot catch that, because it was there. The question is
+// whether anything ELSE named nvoy is also there.
+//
+// Pure by design, like the rest of this module: it takes the text `claude mcp list` printed and
+// returns names. Passing `null` — the tool could not be run — must reach the UNKNOWN path, never
+// an empty array, because "nothing conflicting" and "I could not look" are the two things this
+// whole module exists to keep apart.
+export function foreignNvoyServers(listOutput, name) {
+  if (typeof listOutput !== 'string') return null
+  const mine = `nvoy-${String(name).toLowerCase()}`
+  const found = []
+  for (const line of listOutput.split('\n')) {
+    // `claude mcp list` prints `<server>: <command…> - <status>`. Take only the name.
+    const m = /^([A-Za-z0-9._-]+):/.exec(line.trim())
+    if (!m) continue
+    const server = m[1].toLowerCase()
+    if (server !== 'nvoy' && !server.startsWith('nvoy-')) continue
+    if (server === mine) continue
+    if (!found.includes(server)) found.push(server)
+  }
+  return found
 }
 
 // Render for a terminal. Kept here rather than in the CLI so the suite can assert that an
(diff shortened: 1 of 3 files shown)


JAFairweather:
Merge-order note: this PR and **#365** both touch `src/agent_install_state.mjs`, `tests/agent_install_state.mjs` and `tools/connect-agent.mjs` — the only three files this PR changes, so the overlap is total.

#365 is currently CONFLICTING and sits at the top of the `356 → 359 → 365` stack, so it is furthest from mergeable. This PR is 3 files, off `main`, and green.

Suggested order: **this one first**, then #365 rebases onto it. The reverse means rebasing a 24-file stacked branch to pick up a 3-file change.

Issue #331, opened 2026-08-10
Sources tab: ask outside authors to be carried, and show the six states honestly


JAFairweather:
Tab 2 of #330. Design: `docs/DESIGN_CONSOLE_IA.md`.

Give npubs of people outside the hive; waggle asks their permission to carry their public posts in.
**Most of the machinery already exists** — what is missing is the surface.

| Piece | Where | State |
|---|---|---|
| The ask | `src/bridge.mjs:927`, `src/nostr_egress.mjs:392` (`consent_request` template) | built |
| Prefilled grant | the ask carries a ready-to-sign 440 | built |
| Verification | `src/consent.mjs`, imported at `bridge.mjs:42` | built |
| Anti-spam | ask-once-ever, rate cap, never a muted/grandfathered/consented target | built, tested |
| Enforcement | `mirror_require_consent` | **ON in production** (read from the box 10 Aug) |
| Console surface | — | **none** |

## The name

Not "add remote content" — that frames it as the operator *taking*, and the operator cannot add
anyone's content. From `consent.mjs`: *"Every other grant in the estate is authored by the
AUTHORITY. A consent record is authored by the DATA SUBJECT."*

**Tab: Sources. Action: "Ask to carry".** Decided 10 Aug.

## Six states, not five

Taken from the bridge's own vocabulary rather than invented:

**Carrying · Asked — waiting on them · Held, no consent · Declined · Withdrawn · Grandfathered**

Two of these matter more than they look:

- **Held, not dropped.** The bridge logs `PUBLIC hold[no-consent]: … participant has not consented
  (default-closed, §8)`. Traced in production for a watched author: added to the watchlist 03 Aug,
  one post carried, enforcement engaged eight minutes later, every subsequent post held with the
  reason named and a consent ask sealed the same second. Use the bridge's word.
- **Grandfathered** — carried, no consent record, permanently exempt. The pre-consent crew who
  joined as members before the mechanism existed; `mirror_grandfathered` currently holds two.
  **A tab showing five states while reality has six lies in the worst direction**: it implies
  consent underpins a carry that is actually running on an exemption.

**Declined** and **Withdrawn** must also be first-class. Only the grantor may revoke their own
consent — the load-bearing check in `consent.mjs` — so a withdrawal happens *to* the operator,
unilaterally, and needs somewhere to appear rather than a row silently vanishing.

## ⚠ What this tab must not get wrong

[...]

PR #390, opened 2026-08-12
One honest consent word per author, so a Sources tab can exist (#389)

### LINKED ISSUE #389: The signed control state collapses four different consent realities into one word — a Sources tab built on it cannot be honest
Blocks #331 (Sources tab), and is a defect in the **signed, public** artifact regardless of whether
that tab is ever built.

#331 requires the tab to show six states, and says plainly why:

> A tab showing five states while reality has six **lies in the worst direction**: it implies
> consent underpins a carry that is actually running on an exemption.

The console cannot show them. `buildControlState()` (`src/bridge.mjs:2040`) projects one word per
watched author from a **four**-value vocabulary, closed at the signing schema
(`src/nostr_egress.mjs:170`): `pending | asked | active | revoked`.

## Proven, not inferred

Four authors in four genuinely different states, enforcement **on**, driven through the real
`buildControlState()`:

```
superseded-terms consenter   -> projected "pending"
grandfathered                -> projected "pending"
muted (approver rejected)    -> projected "pending"
never asked                  -> projected "pending"
```

One word, four realities — and one of them is being **carried**:

- **Grandfathered is carried.** `bridge.mjs:1490` — `if (!consentBinds && !PUB.mirrorGrandfathered.includes(author))` — so an author on the exemption list is never held. The record labels them identically to an author whose every post is being dropped. This is exactly the worst-direction lie #331 names, and it is in the signed artifact, not in some future UI.
- **Muted will never be asked.** `maybeAskConsent` returns false for a muted target, permanently. Showing them beside a never-asked author implies an ask is coming that will never be sent.
- **`pending` does not mean held.** It means held only when `mirror_require_consent` is on. Off, the same word means *carried, ungated*. The gate flag is published in the same record (`operations.gates.consent_required`), but `operations` is an additive/optional field, so the top-level meaning of `pending` depends on a field that may be absent.
- The superseded-terms case turned out to be handled correctly and is **not** a defect: `processConsentEvent` passes `expectedTosHash` to `verifyConsent`, so a consent bound to old terms never enters `mirrorConsent` at all. Recorded here because I expected it to project as `active` and it does not. `bridge.mjs:1488`'s second `rec.tosHash === ...` check is therefore belt-and-braces, not the gate.

## Proposed vocabulary

Additive — the four existing words keep their wire meaning, so records already published stay valid
and `console/config-operations.mjs` keeps reading:

| wire | means | derived from |
|---|---|---|
| `active` | carrying, on the subject's own consent | `mirrorConsent` |
| `grandfathered` | **carried**, no consent record, permanently exempt | `mirror_grandfathered` |
| `asked` | asked, waiting on them | `mirrorAsked` |
| `held` | no consent and the gate is on — every post dropped | `mirror_require_consent` |
| `pending` | no consent, gate off — carried, ungated | (the honest remainder) |
| `revoked` | the subject withdrew | `mirrorRevoked` |
| `muted` | the **approver** rejected them; will never be asked | `muted_authors` |

Precedence must mirror `routePublic`'s, not be invented: consent binds first, then grandfathering
(which beats a withdrawal, because `bridge.mjs:1490` carries a grandfathered author whether or not
they ever withdrew).

## One correction to hand back to #331

#331 lists **Declined** as a first-class state. The bridge cannot observe it. `docs/CONSENT.md` §6 is
explicit that **silence is a no** — there is no decline event, and asking twice is harassment, so
nothing ever arrives to distinguish "said no" from "has not answered". The state the bridge really
has is `muted`, which is the **approver's** rejection, not the subject's (`bridge.mjs:~1505`:
*"A muted author was explicitly rejected by the approver"*). Those are opposite parties and must not
share a label. #331's Declined should be renamed or dropped before the tab is built on it.

## Verification

`tests/control_state.mjs`. Assert both directions and the precedence, not just that the new words
appear: a grandfathered author who also withdrew must read `grandfathered`, because that is what the
bridge does to their posts. Include the negative control that the four legacy words still project
unchanged for the cases that already worked.

JAFairweather:
Closes #389. Unblocks #331 (Sources tab), and fixes a defect in the **signed, public** control
record whether or not that tab is ever built.

## The defect

`buildControlState()` published one word per watched author from a four-value vocabulary. Driven
through the real bridge with `mirror_require_consent` **on**:

```
superseded-terms consenter   -> projected "pending"
grandfathered                -> projected "pending"
muted (approver rejected)    -> projected "pending"
never asked                  -> projected "pending"
```

One of those four is being **carried**. The consent gate at `src/bridge.mjs` reads
`if (!consentBinds && !PUB.mirrorGrandfathered.includes(author))`, so a grandfathered author is
never held — and the record labelled them identically to an author whose every post was dropped.
#331 names that exact failure:

> A tab showing five states while reality has six **lies in the worst direction**: it implies
> consent underpins a carry that is actually running on an exemption.

## The fix

`src/consent_state.mjs` — pure, seven states, **precedence copied from `routePublic`'s gate rather
than chosen**. Two orderings look wrong until you check the gate, and both are asserted:

- **grandfathering beats a withdrawal** — the gate carries them either way, so `revoked` would be a
  false label. The withdrawal is not lost; it is just not what is happening to their posts.
- **`held` and `pending` are the same consent situation under different gates.** One word for both
  made the top-level meaning depend on `operations.gates.consent_required`, a field the schema
  treats as optional — so an omitted field silently flipped it.

The signing schema now **imports** the vocabulary instead of restating it beside the key. `console/`
keeps a deliberate second copy (nothing under `src/` is served to a browser — the `scope-hash.mjs`
pattern), and the suite asserts word-for-word agreement including order. That copy is not cosmetic:
`config-operations.mjs` rejects the **whole** signed state on an unknown word, so drift does not
mis-render one row — it renders "state unavailable" and the owner sees nothing.

## A non-defect, recorded

I expected a consent bound to superseded terms to project as `active` while being held. It does not:
`processConsentEvent` already passes `expectedTosHash` to `verifyConsent`, so such a record never
enters `mirrorConsent`. The second `rec.tosHash === …` check in the gate is belt-and-braces, not the
gate. Noted because I went looking for it and it was already right.

## One correction handed back to #331

**"Declined" is retired.** The bridge cannot observe it — `docs/CONSENT.md` §6 is that **silence is
a no**, there is no decline event, and asking twice is harassment, so nothing ever distinguishes
"said no" from "has not answered". The state that exists is `muted`, which is the **approver's**
rejection, not the author's. Opposite parties: labelling one as the other tells an owner the author
refused when in fact they did. `docs/DESIGN_CONSOLE_IA.md` is updated.

## Verification

Full suite green. **Seven mutations run, all detected** — but two only after a fix:

- **M3 was NOT detected first time.** My gate assertion read `!== 'pending'` against an author who
  was already `revoked`, so it was true whatever the gate did; the suite stayed green with the
  branch mutated flat to `pending`. It now asserts the word itself, on a clean author.
- The muted assertion **failed on its first run** against that same revoked author. That is the live
  projection proving it applies the precedence, rather than the pure helper being asserted against
  itself — so the check moved to a second author and now also proves the projection is per-author.

Both directions asserted throughout: the four legacy words still project unchanged (a negative
control against "renames everything and breaks every published record"), every one of the seven
survives signing, and a plausible-but-undefined word (`carrying`) is still refused.

**No new suite file**, so `package.json` is untouched and this does not conflict with the open
queue. Tests went into `tests/control_state.mjs`, which already owns this subject.

**Review note:** this touches the signed public artifact and the schema that sits beside the key, so
it is substantive — it needs eyes that are not mine before merge. The thing worth attacking is the
precedence table in `src/consent_state.mjs`: every branch claims to mirror what the routing gate does
to that author's posts, and a label that disagrees with the routing is worse than no label.


JAFairweather:
**Pushed a second commit — the first pass of this PR would have shipped a regression.** Worth reading
before review, because it changes what the PR is.

I found the vocabulary restated in **one** place (the signing schema) and fixed that. It was in
**three**. `console/routing.mjs` and `console/following.html` each carried their own
`['pending','asked','active','revoked']`, and both use it the same way:

```js
for (const f of s.follows)
  if (!/^[0-9a-f]{64}$/.test(f.pubkey) || !['pending','asked','active','revoked'].includes(f.consent))
    return null            // <- the WHOLE state, not the row
```

So the moment the bridge published `grandfathered` for a real author, those two pages would not have
mis-rendered a row. They would have shown **"No fresh valid signed state was found"** and nothing
else — the Following page and the Routing console both blank, on a change whose entire purpose was
making that page honest.

I only found it because the next task opened `following.html` for an unrelated reason.

## The fix is a property, not three edits

Fixing the three I found would have left a fourth to be found in production. The suite now asserts
that **no file under `console/`, `src/` or `tools/` restates the list**, and the failure names the
offending file:

```
FAIL - #389 no file restates the consent vocabulary — one missed copy blanks a whole page
     restated in: console/routing.mjs
```

Verified by mutation in both directions: restoring either hardcoded list fails the guard and names
the right file; with them removed it passes.

## Also in this commit

`following.html` renders `CONSENT_LABEL` instead of the raw wire word, and adds a separate
**"carried now" / "not carried"** mark — three of the seven states mean the author's posts are
crossing, and a reader should not have to know which three. Its explanatory panel goes from four
states to seven and says outright that grandfathered means **carried with no consent record**.

Full suite green.

JAFairweather:
Reviewed `d1148f52`. **The fix is right and incomplete: three of the seven labels do not describe routing at all, and the argument you used to split `held`/`pending` applies to each of them unchanged.**

## The invariant the module states

> Each branch below mirrors what the routing gate actually does to that author's posts, because a
> label that disagrees with the routing is worse than no label.

I checked that branch by branch against the gate at `src/bridge.mjs:1481-1494`. The routing outcome
for a watched author is a function of exactly **three** things:

```js
if (PUB.mirrorRequireConsent && consentGated lane) {
  const consentBinds = rec && (!expectedTos || rec.tosHash === expectedTos)
  if (!consentBinds && !PUB.mirrorGrandfathered.includes(author)) DROP
}
```

`mirrorRequireConsent`, `consentBinds`, `grandfathered`. **`revoked`, `muted` and `asked` are not
consulted anywhere in it.** I confirmed `PUB.muted` is read only by `maybeAskConsent` (`:941`, which
refuses to *ask*), and `mirrorAsked` and `mirrorRevoked` appear nowhere in the gate.

So:

| label | gate ON | gate OFF |
|---|---|---|
| `active` | carried | carried |
| `grandfathered` | carried | carried |
| **`revoked`** | **dropped** | **carried** |
| **`muted`** | **dropped** | **carried** |
| **`asked`** | **dropped** | **carried** |
| `held` | dropped | — |
| `pending` | — | carried |

Four branches mirror the routing. Three describe **consent history** and leave the routing outcome
undetermined — which is the defect this PR exists to remove, surviving in three of seven states.

And it fails for the same stated reason:

> One word for both made the top-level meaning depend on `operations.gates.consent_required`, a field
> the schema treats as optional — so an omitted field silently flipped the meaning.

That sentence is true, verbatim, of `revoked`, `muted` and `asked`. An owner reading `muted` cannot
tell whether that author's posts are being dropped or carried without reading the same optional field,
with the same silent flip. `revoked` is the sharpest: it reads as "this author is not being carried",
and with the gate off they are being carried, ungated.

**The precedence itself is correct** — I checked the two you flagged as looking wrong. `grandfathered`
beating `revoked` is right: `processConsentEvent` deletes from `mirrorConsent` on a valid 441
(`:873`), so a revoked author has no consent, and the gate still carries them on the exemption.
`muted` before `asked` is right for the reason given. Precedence is not the problem; the vocabulary
is one axis short.

**Suggested shape.** The state is genuinely two facts — *why this author is in this consent situation*
and *what is happening to their posts* — and `held`/`pending` is that second axis already, discovered
for one case and not generalised. Two fields (`consent: 'revoked'`, `carrying: false`) is cleaner than
doubling the vocabulary to `revoked-held` / `revoked-carried`, and it is safer given your second
commit: two of the three vocabulary sites reject the **whole** signed state on an unknown word, so
every word added is a blast radius.

## A question the two commits raise between them

The call site passes `consented: mirrorConsent.has(pubkey)` — presence. The gate uses
`consentBinds` — presence **and** terms-binding. Those are the same only if a superseded-terms record
can never be in the map, which `verifyConsent(..., expectedTosHash: PUB.mirrorExpectedTosHash)` at
ingest does appear to guarantee.

If that guarantee holds, the gate's own `rec.tosHash === PUB.mirrorExpectedTosHash` re-check at
`:1489` is dead code. If it does not hold, the label says `active` for an author every one of whose
posts is dropped — the worst direction, in the case your own header cites as motivating
(*"superseded-terms consenter -> pending"*). **They cannot both be live and agree.** Worth resolving
which, because the answer either removes a line or adds a parameter to `consentState`.

## The second commit is the better half

Finding the vocabulary restated in three places, two of which reject the entire signed state on an
unknown word, is the more valuable catch — a first grandfathered author blanking the Following page
and the Routing console is a much worse outcome than a vague label, and it would have shipped as a
consequence of the fix rather than of the bug. Exporting `CONSENT_STATES` as the single source is the
right remedy.

JAFairweather:
Dennis — attacked the precedence table in `src/consent_state.mjs`, cross-checked every branch against `routePublic` (`src/bridge.mjs:1018-1106`), not against the module's own comments.

**Found a real branch that disagrees with the gate: `muted`.** The precedence table's contract is explicit — "every branch mirrors what the routing gate actually does to that author's posts" — and `CONSENT_IS_CARRYING.muted = false` / `CONSENT_LABEL.muted = 'Rejected by you — never asked'` both assert this author's posts are being blocked. That's only true when `mirror_require_consent` is ON.

`PUB.muted` only ever blocks a post at one place — `bridge.mjs:1086` — and that line is gated `if (quarantine)`. Quarantine is only ever set for the `'reply to our note'` classification (an untrusted stranger); a **trusted, watched author** (`why = 'mirrored feed'`, the only population `consentState()` is ever called over — `follows: PUB.authors.map(...)`) never enters that branch, so line 1086 never runs for them. Mute, as coded, has zero effect on a watched author's mirrored-feed posts. The only thing that can hold them is the separate consent gate (`bridge.mjs:1060-1076`), which checks `consented`/`grandfathered`, never `muted`.

So take a watched author who is muted, not grandfathered, not consented, with `mirror_require_consent` OFF (the coded default — `bridge.mjs:1054`, "DEFAULT-OFF... the deploy changes nothing and the crew keeps mirroring"). The consent gate condition is `false` (whole block inert when the flag is off), so their posts are **fully carried** — identical to a `pending` author. But the projection reports `muted`, and the console tells the owner "Rejected by you — never asked" / not carrying. That's the same shape of lie #331 named as the worst direction, just on the branch this PR didn't retarget.

This isn't hypothetical — your own test constructs the exact overlap (`tests/control_state.mjs:592-595`, `PUB.authors.push(mutedOnly); PUB.muted.push(mutedOnly)`), and it's operator-reachable today: `mute` only ever fires on a quarantined stranger reply (`handleCommand`'s `mute` branch reads `st.author` from the staging map), and `waggle mirror <pubkey>` (`addWatchAuthor`, `bridge.mjs:1213-1224`) has no check against `PUB.muted` — mute a stranger who replied, then later `waggle mirror` them once they're trusted, and you have a muted, watched author. Your test asserts the label; it never asserts what the live gate does to that author's event.

One mitigating fact worth being precise about: `src/consent.mjs`'s new header says `mirror_require_consent` **is ON in production** right now. With the gate on, the consent-gate branch alone does correctly hold this author (unrelated to mute), so today's deployment doesn't show the lie live. But the module's own claim is unconditional ("copied from the routing gate," not "correct only while the gate is on"), the flag is a config toggle any operator (including a future self-hoster, which is the DEFAULT-OFF case) can flip, and nothing in the code or the schema prevents this state from existing with the gate off.

Suggest: either fold `muted` into the same `gated`-style branch (a muted, watched author with the gate off should report `pending`, exactly like an ordinary un-consented one — mute simply doesn't apply to them), or make `bridge.mjs:1086`'s mute check apply to `PUB.authors` regardless of quarantine so the label and the gate actually agree. Either way, add a test that constructs this exact overlap and asserts against `routePublic`'s real drop/carry decision, not just the projected word — the class of bug this PR exists to kill.

The rest holds up: grandfathered-beats-revoked and held/pending-by-gate are both correctly copied from the gate, verified by reading the same lines. The second commit's "restated in three places" fix is real and the property test (any file under `console/`, `src/`, `tools/` restating the list fails, names the offender) is the right shape.

Issue #330, opened 2026-08-10
Console IA: one board of ins and outs, not five pages of mechanism


JAFairweather:
Parent issue for the console reorganisation. Design: `docs/DESIGN_CONSOLE_IA.md`.

## Problem

The console is organised by **mechanism** — grants on one page, follows on another, agents on a
third. Five pages accreted one wave at a time and nobody went back to ask what the whole thing
looks like.

The evidence that this is presentation and not data: **four of the five pages read the same
event.** Following, Agents, Routing and Config all fetch kind 30078, each re-deriving the hive and
rendering one slice. That document already carries `follows[]` (with per-author consent),
`agents[]`, `operations` and `observed_at`.

## The reorganisation

Organise by **direction of travel**, which is how the architecture already thinks — waggle is four
lanes.

| Tab | Is | Issue |
|---|---|---|
| 1 | The board — what needs you, what's coming in, who's here | this issue |
| 2 | Sources — ask outside authors to be carried | (below) |
| 3 | Connect an agent | extends #309 |
| 4 | Config — settings with teeth | (below) |

Routing splits: **decisions** (moderate a quarantined note) to Tab 1, **policy** (the lanes) to
Tab 4. Following and Agents dissolve into the board. Access keeps the *act* of granting.
Setup stays outside the tabs.

## Tab 1 specifically

Order: **Needs you now** -> **Coming in** -> **Who's here**.

Leading with pending decisions is the one addition to the original sketch. Standing state is
reference material you scan; quarantined notes, unanswered consent requests and pending joins are
*work*, and they are the only time-sensitive thing in the console.

## Blocked by #321

The agents half renders **empty today** — `processGrantEvent` never writes agent rows, so
`agents[]` is genuinely `[]` while grants are live. This redesign does not hit that bug, it
**forces** it: you cannot build a board whose right half is agents while the roster is
disconnected from the thing that admits agents. #321 is the precondition.

## Do not flatten the two planes

`admit` and `task` are not peers and must not share a permissions column:

- `admit` — **this bridge enforces it**. `src/bridge.mjs:832` is the whole enforcement.
- `task` — **nothing here enforces it**. waggle never reads it; it is a signed letter of authority
  the receiving runtime may choose to honour.

Listing both as values of one field re-creates exactly the confusion the Connect page was built to
prevent, on the page the owner looks at most.

## Related

- #322 — the bridge key is re-asked on every page. One board is the structural fix.
- #317 — diagnostics as a private local plane.

[...]

Issue #321, opened 2026-08-09
Granting an agent a 440 does not create an agent row — Access and Agents are two disconnected registries


JAFairweather:
Reported from the live console: the Access page lists agents with grants in force, and the
Agents page says **"This hive has admitted no agents."** Both are telling the truth. They are
reading two different registries that nothing connects.

## Verified, not inferred

Cold read of the bridge's signed control state from all four relays (4/4 answered), newest
published `2026-08-09T13:42:24Z` — after the current build deployed:

- the state **does** carry the `agents` key, so the #309/#315/#316 projection is working
- its value is genuinely `[]`
- on the host, `data/agent-rows.json` does not exist, so `loadAgentRows()` returns `{}` on ENOENT

So the screen is correct and the projection is correct. The gap is upstream of both.

## The actual gap

`saveAgentRows` has exactly one caller: `handleAgentLifecycleCommand` (`src/bridge.mjs:2342`).
`processGrantEvent` — the path that consumes the NIP-DA 440 grants the Access page renders —
never touches agent rows.

That means **admitting an agent by signing a grant does not make it an agent as far as this hive
is concerned.** A row only appears if someone separately publishes a signed `agent_admit`
lifecycle command. Nobody has, so the roster is empty while the grants are live.

Two registries answering the same question — *who is an agent here?* — with no link and no
reconciliation. An owner who does the obvious thing (grant `task` to an agent's key on the
Access page) ends up with an agent that works and a roster that denies it exists.

## What this needs deciding

Not obvious which way it should go, so this wants a decision before code:

1. **Derive rows from grants.** The 440 set becomes the source of truth and rows are a
   projection of it. Simplest to reason about; loses the lifecycle states (`paused`, `revoked`,
   return-lane) that rows carry and grants do not.
2. **Have `processGrantEvent` seed a row** on first sighting of a task-family grant, leaving
   lifecycle commands to mutate it afterwards. Keeps both, adds a write to a path that currently
   only reads.
3. **Keep them separate and say so on screen.** The Agents page states plainly that it lists
   agents admitted through the lifecycle lane, and links to Access for grant-holders. Cheapest,
   and honest, but leaves the owner to hold two models in their head.

Whichever is chosen, the empty state on the Agents page should not read as a bare "no agents"
when grants exist — that is the sentence that made this look broken.

JAFairweather:
[...]

Issue #371, opened 2026-08-11
CLAUDE.md documents #141's rejected per-session-identity design as the designed onboarding


JAFairweather:
## `CLAUDE.md` describes #141's **rejected** alternative as the designed onboarding

`CLAUDE.md` is auto-loaded into every Claude Code session in this repo, and this paragraph sits in
the section headed *"Claims that must never drift"*:

> **You act as your OWN participant key, never as the bridge.** The designed onboarding (#141) is
> that a session mints an *ephemeral* key, requests a maintainer grant, acts, and burns it — no
> persistent key held. … (#141 is partly built — do not describe it as shipped.)

#141 says the opposite, under its own heading *"The maintainer's 'new key per session, burn it' idea
— and the nuance that saves it"*:

> The instinct (disposability, nothing long-lived to steal) is right, but applied to the **identity**
> it breaks three things: a fresh npub per session is not admitted … is not in `return_lane` … and
> has **no continuity** … That trades the whole first-class-participant model for disposability.
>
> **Burn the connection, not the identity.** … **one stable Claude identity**, key only ever in the
> maintainer's signer … **a fresh, revocable bunker client-key per session**.

So the mint-ephemeral-identity-and-burn-it shape is the alternative #141 **considered and rejected**,
and `CLAUDE.md` presents it as the design. Two errors in one paragraph:

1. **Wrong design.** The resolved shape is one stable identity in a NIP-46 bunker with a disposable
   *client credential* per session — the same pattern as #54 for the bridge key. What is disposable
   is the connection, not the key.
2. **Stale status.** "#141 is partly built" — #141 is **CLOSED** (2026-08-04).

## Why this is worth a fix rather than a note

It misled a review today. Reasoning from that paragraph, I wrote a custody matrix for #370 giving
`#141 mint-act-burn — key in the process` as the model for ephemeral runtimes, and contrasted it with
bunker custody for peripherals. Neil caught it by reading #141 itself. The correct framing is the one
he landed on: bunker custody is not a new bar for actuators, it is the pattern already established
for the bridge key (#54) and the session key (#141), and the real axis is the **client credential's
lifetime** — dies with the session, versus has to persist for a device's whole life somewhere you do
not control.

That is a better conclusion than the one I started with, and the only reason it took a correction is
that the auto-loaded doc says the rejected thing. Every future session reads the same paragraph.

## Suggested replacement

> **You act as your OWN participant key, never as the bridge.** The designed onboarding (#141) is a
[...]

PR #391, opened 2026-08-12
Publish the consent-ask budget, not just its cap (#331)

JAFairweather:
Part of #331. **Stacked on #390** — base is `fix/389-consent-state-vocabulary`, so review that first
and this retargets to `main` automatically when it merges.

## Why this and not the tab

#331 is a whole console tab, and the IA (`docs/DESIGN_CONSOLE_IA.md`) says Following *dissolves into
the board*, which is #330's job and is blocked on #321. So the layout half is not mine to decide
yet. What **is** independently broken is the data underneath it: one of #331's two "must not get
wrong" items could not be built at all.

> **The rate cap needs a visible budget.** A silent refusal at the cap is indistinguishable from a
> broken send.

The signed record published `ask_per_hour` and nothing else. The counter that would answer it lived
in a module-private variable in `bridge.mjs`. An owner at the cap saw asks stop and had no way to
tell that from a bug.

## What landed

**`askBudget()`** — `per_hour`, `used_this_window`, `remaining`, `window_resets_in`. It reports the
budget as it will be when **next consulted**, not as the counter happens to sit: an elapsed window is
spent, and saying "0 left" for a window that has already reset sends the owner hunting for the
opposite problem.

**The schema beside the key refuses a budget that does not add up.** Not belt-and-braces — a budget
where `used + remaining !== per_hour` tells the owner asks are available when the next one will be
refused, which is worse than publishing nothing. Additive like `agents` before it, so a bridge that
has not restarted keeps signing valid records; asserted in both directions.

**On the surface**, the rest of #331's list:

- **"waggle asks each writer once, ever" now sits above the ask button**, not below it. Asserted by
  *index*, not by presence — the position is the whole point. Someone who does not know reads a
  non-response as a failed send and goes looking for a retry that does not exist.
- The remaining budget renders on Config, and both surfaces say a refusal at the cap is a refusal.
- **`src/consent.mjs` no longer opens by claiming it is not wired into the bridge.** `bridge.mjs`
  imports it and `mirror_require_consent` is on in production, so that header described a live
  consent gate as a sketch. A stale comment on a live gate is how a reviewer talks themselves out of
  reading it.

## Verification

Full suite green. **Six mutations, all detected after a fix:**

- **N3 was NOT detected first time.** My fixture only ever had a never-opened window, where the
  elapsed branch is true either way — so it stayed green with the window reset mutated out. That is
  the direction that matters: reporting a spent window as still spent tells the owner they are capped
  when the next ask would go through. It now spends a real ask through `sendConsentRequest` first,
  then checks both the open window and the elapsed one.

Assert-the-reason throughout: the arithmetic refusal is matched on `does not add up`, and a missing
field on `unexpected fields`, so "refuses correctly" is distinguishable from "refuses everything".

**Review note:** touches the signed artifact and the schema beside the key — substantive, needs eyes
that are not mine. The thing worth attacking is `askBudget`'s elapsed-window branch: it is the only
place where the published number deliberately disagrees with the in-memory counter.


JAFairweather:
Reviewed `ee3cf26d`. **The elapsed-window branch is right, and publishing it as a countdown undoes it at the reader.**

## The branch you flagged is correct

> Reports the budget as it will be when NEXT consulted, not as the counter happens to sit: an elapsed
> window is spent, and saying "0 left" for a window that has already reset would send the owner
> hunting for the opposite problem.

Agreed, and it is the right call for the right reason — the record is read to decide whether asking is
possible, so it must answer about the next consultation rather than about the last one. Deliberately
disagreeing with the in-memory counter is correct here, and saying so in the comment is what makes it
reviewable.

## But `window_resets_in` is a countdown in a record that is republished on a timer

`askBudget()` is evaluated at signing time. The control state is republished every
`control_state_refresh_secs`, default **300**. So every field here is up to five minutes stale by
construction — and staleness does different things to different fields:

- `per_hour` — a cap; stale is harmless.
- `used_this_window` / `remaining` — a count; stale is *wrong but conservative*.
- `window_resets_in` — a **countdown**; stale is wrong in the one direction you set out to avoid.

Concretely: sign at T with 60s left in the window. At T+120 a reader sees `window_resets_in: 60` and
`remaining: 0` for a window that expired a minute ago. That is exactly *"0 left for a window that has
already reset"* — the failure the elapsed branch exists to prevent — reintroduced between the signer
and the reader rather than inside the function.

The elapsed branch fixes it for the process that computes it. It cannot fix it for the party that
reads it, because a countdown stops being true the moment it is written down.

**Publish the window start as an absolute timestamp instead** — `window_started_at`, or
`window_resets_at` — and let the reader subtract. An absolute instant does not go stale; a duration
does. That also lets a reader do exactly what your elapsed branch does: if `now >= window_resets_at`,
the window is spent and `remaining` is the full cap, regardless of what the record's `remaining` said
when it was signed. Today a reader has no way to make that inference, because the only thing that
would tell them the window rolled is the same stale countdown.

If you want to keep `window_resets_in` for a console that renders it directly, keep both — but the
absolute one is the field a correctness argument can be built on.

## Smaller

- `window_resets_in: 0` when no window is open, with the comment distinguishing it from "resets
  imminently", is a real distinction and it is the right one. It becomes cleaner still with an
  absolute field: `null` for no window, a timestamp otherwise, and no value that means two things.
- The optionality handling in `nostr_egress.mjs` — `exact(...)` branching on `consent_asks == null` so
  an older bridge's record stays valid — is the same additive discipline as `agents`, and it is the
  part that would have been easy to get wrong given #390's second commit is about exactly this class
  of rejection.
- Correcting `consent.mjs`'s header, which claimed the file was not live while it gated production
  traffic, is worth more than it looks. *"A stale comment on a live gate is how a reviewer talks
  themselves out of reading it"* — that is the right diagnosis and it is the same failure as #371.

JAFairweather:
Dennis — attacked `askBudget`'s elapsed-window branch (`bridge.mjs:916-927`) against the real counter in `askRateOk` (`bridge.mjs:929-934`). No finding — the deliberate disagreement is correct and it's tested for exactly the property that matters.

Traced the two functions side by side: both use the identical threshold (`now - askWindowStart >= 3600_000`), same `Date.now()` basis, no divergent clock source. `askBudget`'s elapsed branch predicts what `askRateOk` *will* do on its next real call, and since elapsed status can only go false→true as wall time advances against a fixed `askWindowStart` (never true→false without an intervening real ask, which itself updates `askWindowStart` and makes the new state correct again), there's no ordering where the prediction and the eventual real reset actually disagree.

The one place a snapshot function like this could quietly lie — reporting a stale `askWindowCount` after the window has technically elapsed but before any real ask has rolled it over — is exactly what `tests/control_state.mjs:685-696` exercises: spend one through the real path, read `askBudget()` mid-window (`used_this_window === 1`), then read it again with `now` pushed past the boundary (`used_this_window === 0`, `remaining === per_hour`) without ever calling `askRateOk()` again. That's the deliberate-disagreement branch, verified against the live counter, not just against itself.

Schema side also holds: `nostr_egress.mjs`'s `controlState()` rejects any `consent_asks` where `used + remaining !== per_hour` before signing (so a broken snapshot can't even be published), and `config-operations.mjs` enforces the same invariant on read plus the additive-field backward-compat case (a pre-#331 record with the field entirely absent still signs and reads). Both directions tested, both look correct.

Nothing to push back on here.

Issue #379, opened 2026-08-12
Triage: re-judge the open backlog against the four-registry participation model


JAFairweather:
A coherence pass over all 40 open issues, asked for after the native-roster path was proven live.
Several issues were filed against a premise that has since been falsified, and at least two of them
will send someone building the wrong thing if they are picked up as written.

Nothing here is a new build. It is close / re-scope / keep, with the reason.

---

## The architecture to judge against

Four **independent registries** have to agree before an outside agent is a full participant. This is
the thing that was not clear when most of these issues were written:

| # | Registry | What it buys | Who can write it | Who can remove it |
|---|---|---|---|---|
| 1 | 440 grant (`grantSet`, in memory) | the authority for everything else | community owner | owner, via 441 |
| 2 | `channel_members` row | `@Name` is addressable; the channel is readable | owner, or waggle when `public.seat_grantees` is on (#355/#356) | waggle's unseat (#356) |
| 3 | `relay_members` row | the key can NIP-42 auth to the relay **at all** | claimed by the key, from an invite (#357) | **only the key itself** (#366) |
| 4 | `users.display_name`, from a `kind:0` | the at-word resolves to something | **only the key itself** | **nobody** (#366) |

Rows 3 and 4 are the ones that were missing from the model. Row 3 is the wall — `enforce_relay_membership`
fires at AUTH time, before any channel is consulted — which is why a `channel_members` row alone is inert
(#344's live 2×2). Row 4 is why waggle cannot name an agent on its behalf.

Proven live, end to end: mint → relay invite → the agent published its own `kind:0` → seated in the
roster → a crew member typed `@LukeDog` → the post survived and reached the key.

---

## A. Superseded — stop building these

**#337 §1 (addressing) — close.** Its premise is written into the issue: *"`@handle` resolves against
the channel roster. Relay-lane agents are never on it."* That is now false. Registries 2, 3 and 4
together make `@Name` work natively, and it has been observed working. The `nostr:npub1…` NIP-21
convention, the exfiltration guardrail, the handle-table staleness argument — all of it was careful
engineering around a wall that now has a door. **Do not build it.** #336 (`--mention <hex>`) goes with it.

⚠ **#337 §2 and §3 are untouched and still exactly right**, and they are buried under a title that now
reads obsolete. They should be **split out as their own issues** before anyone closes #337:

- **§2 — kind:10050.** The return lane is still the only inbound path, it still silently no-ops without
[...]

Issue #366, opened 2026-08-11
Relay revocation is cooperative: an owner cannot remove a key from the relay


JAFairweather:
Settled against `block/buzz@origin/main`, 2026-08-11. Read via `git show origin/main:` — the local checkout was 488 commits behind, and this is exactly the case where that matters.

## Three facts

**1. A removal path exists.** NIP-43 leave request, `KIND_NIP43_LEAVE_REQUEST = 28936` (`buzz-core/src/kind.rs:404`), handled at `buzz-relay/src/handlers/ingest.rs:2321`. It requires `require_relay_membership` to be on, a NIP-70 `-` tag, and a `created_at` within ±120s of now. It refuses for the relay owner.

**2. It is self-service only.** The handler calls `remove_relay_member(tenant.community(), &sender_hex)` where `sender_hex` is `event.pubkey` — the signer removes *itself*. There is no HTTP route anywhere in `crates/buzz-relay/src/api/` that removes a relay member. The only third-party removal is `crates/buzz-admin/src/main.rs`, a box-side CLI holding database credentials.

So **neither waggle nor the console can evict a key.** Only the key itself, or an operator with `buzz-admin` on the host.

**3. Removal does not clear the name.** Nothing in the tree deletes or clears `users.display_name` — no `DELETE FROM users`, no `UPDATE users SET display_name`. The leave handler removes the row and publishes NIP-43 announcements; the `users` row is untouched. So removing membership stops the key **authenticating**; the at-word keeps resolving.

## The consequence

**Relay revocation is cooperative.** A key whose grant has been revoked, and which declines to sign its own `kind:28936`, keeps its `relay_members` row and keeps its name. The owner has no lever.

This is the same shape as the `kind:0` problem — only the key can act for itself — and it is arguably worse. That one prevents access being *granted*. This one prevents access being *taken away*.

Scope of what a revoked-but-unremoved key retains: it can still NIP-42 auth to the relay. #356's unseat removes it from the channel roster, so it cannot read the channel. It can still refresh its own `kind:0` and keep its name resolving. Bounded, but not nothing.

## Decision

**Descoped.** The goal's "revoking in the console removes it, and the name stops resolving" becomes **"revoking removes the grant and the agent loses the lane"** — which is ours, true today, and shippable. The second clause is not deliverable by the console on either count and will not be claimed.

## The exit path stays open, and cleanly

[...]

PR #393, opened 2026-08-12
Name the disagreements between the registries (#392)

### LINKED ISSUE #392: Nothing reconciles the registries: four answers to "who is here?" and no way to see them disagree
Generalises #321, and builds the debt view #366 left open. Both are the same defect, found twice.

## The shape

Participation in this hive is spread across four registries (#379):

| # | Registry | Who writes | Who can remove |
|---|---|---|---|
| 1 | live 440 grant (`grantSet`) | owner | owner (441) |
| 2 | `channel_members` row | owner, or waggle when seating is on | waggle's unseat |
| 3 | `relay_members` row | **the key**, from an invite | **only the key** (#366) |
| 4 | `users.display_name` from a self-signed `kind:0` | **only the key** | **nobody** (#366) |

Plus waggle's own `agent-rows.json` lifecycle rows.

**Nothing compares any pair of them.** That is not a hypothetical:

- **#321** — the Access page lists agents with grants in force while the Agents page says *"This hive
  has admitted no agents."* Both are telling the truth about a different registry. Verified against a
  cold read of the signed state: `agents` is genuinely `[]` while grants are live.
- **#366** — a key whose grant was revoked keeps its `relay_members` row unless it removes itself.
  The issue closes by naming the fix: read the relay's own `KIND_NIP43_MEMBERSHIP_LIST` (13534) and
  *"diff it against live grants, and show exactly which keys hold a relay row with nothing behind
  it — no risk of our records having drifted from the relay's."*

Two issues, one missing capability: **nothing computes the disagreement.**

## What this asks for

A pure module that takes the registries as data and returns the disagreements. Not a fix for #321 —
deliberately **not** picking between its three options — but the thing that makes any of them
checkable, and makes its option 3 ("keep them separate and say so on screen") honest rather than an
excuse.

The disagreements worth naming, each of which is a real state someone has hit:

- **grant, no lifecycle row** — #321 exactly. The agent works and the roster denies it exists.
- **lifecycle row, no grant** — a row that outlived its authority; the screen shows an agent that
  cannot act.
- **relay row, no grant** — #366's cooperative-revocation debt. The key can still authenticate.
  Enumerable only by reading the relay's 13534, never from our own records.
- **grant, no relay row** — the invite was never claimed. Looks admitted, cannot authenticate. This
  is the failure mode #357 exists to fix and the one most likely to be misread as "the bridge is
  broken".

## Why it must report rather than resolve

Every one of these has a different *right* answer, and three of them are not ours to enforce: only
the key can create #3 and #4, and only the key can remove them. A reconciler that silently "fixed"
what it could would produce a screen that agrees with itself and disagrees with the relay — which is
the exact failure #366 warns about when it insists on reading the relay's own list rather than ours.

So: name the disagreement, name which side is authoritative, and say who can act. Nothing else.

## Verification

Both directions, and the reason. A reconciler that reports everything as a disagreement and one that
reports nothing fail an `ok`-only assertion identically — so pair every "this is flagged" assertion
with a fully-consistent fixture that must come back clean, and assert that each disagreement names
the key and the registry rather than merely being non-empty.

Fixtures must include a key in **more than one** disagreement at once — that is the case a
first-match-wins implementation gets wrong, and it is the normal state for a revoked agent.

JAFairweather:
Closes #392. Generalises **#321** and builds the debt view **#366** left open — they are the same
defect found twice.

## The defect

Participation here is spread across four registries (#379), and nothing compares any pair of them:

- **#321** — the Access page lists agents with grants in force while the Agents page says *"This hive
  has admitted no agents."* Both are telling the truth about a different registry. Verified there
  against a cold read: `agents` is genuinely `[]` while the grants are live.
- **#366** — a key whose grant was revoked keeps its `relay_members` row unless it removes itself,
  and that issue closes by naming the fix: read the relay's own membership list and *"diff it against
  live grants … no risk of our records having drifted from the relay's."*

Two issues, one missing capability: **nothing computes the disagreement.**

## It reports, it does not resolve

Not timidity. Three of the four registries are **not ours to write** — only the key can create or
remove its relay row and its name (#366). A reconciler that silently fixed what it could would
produce a screen that agrees with itself and disagrees with the relay, which is the exact failure
#366 avoids by insisting on the relay's own list rather than ours.

So every finding names the key, the disagreement, **which side is authoritative**, and **who can
act** — the last being what stops a screen offering a button nobody can press:

| finding | authority | who can fix it |
|---|---|---|
| has a grant, missing from the roster (#321) | grant | owner |
| on the roster, no live grant | grant | owner |
| can still authenticate, no live grant (#366) | grant | **the key itself** |
| has a grant, never claimed its invite (#357) | **relay** | the key itself |
| name still resolves, no live grant (#366) | grant | **nobody — filed with Block** |

**It deliberately does not pick between #321's three options.** It makes all three checkable, and
makes option 3 — *"keep them separate and say so on screen"* — honest rather than an excuse.

## Being unable to read is not being fine

An unread relay list reports as unread rather than as consistent. Both directions are asserted,
because this is the pair that is easy to collapse: an **empty** relay list is a real observation and
must behave differently from an **absent** one — with an empty list, a granted key is correctly
flagged as never having claimed its invite.

## Verification

New suite `tests/registry_reconcile.mjs`, 23 checks. **Eight mutations, all detected.**

The one worth naming is **R1 — first-match-wins per key.** A revoked agent is normally in *three*
disagreements at once: no grant, but still a roster row, still a relay row, and still a resolving
name. Stopping at the first hides two, and it is the most common real state, so there is a fixture
for exactly that.

The **negative control runs first in the file**: a fully consistent hive must report nothing, with
every counter zero rather than merely an empty array. A reconciler that flags everything and one that
flags nothing fail an `ok`-only assertion identically.

Findings are asserted by **id and by rendered text**, and rendering is a separate function from the
computation — a rendering bug and a reconciliation bug are different failures, and a test that only
reads the string cannot tell them apart.

## Merge note

This **does** add a suite file (71 → 72), so it touches `package.json` and the rosters and will
conflict with any other suite-adding PR in the queue. #388 exists precisely for that:
`node tools/merge-suites.mjs` resolves it without dropping a suite. Full suite green at 72.

**Review note:** pure module, no delivery path and no key — docs-and-tests adjacent rather than
substantive. The thing worth attacking is the FINDINGS table: each row claims an authority and an
actor, and if either is wrong the screen will confidently tell an owner to do something they cannot.


JAFairweather:
Reviewed `aaf3c42b`. **One `actor` is wrong in the way you named — it tells the owner someone else can fix it.**

Making `FINDINGS` a table with a mandatory `authority` and `actor` per entry, rather than a chain of
string comparisons, is the right structure: a new finding cannot be added without answering both
questions, and that is precisely the discipline that stops a reconciliation screen from being
confidently unactionable.

## The finding: `actor: 'the key itself'` on the unclaimed-invite row

> `authority: 'grant', actor: 'the key itself'`
> *"Looks admitted and cannot authenticate. Reads as a broken bridge; it is an unclaimed invite."*

That is true only while an **unexpired** invite exists. It usually will not.

`console/relay-invite.mjs` mints with `ttlSecs = 3600`, and Buzz's claim path returns a distinct
`ClaimOutcome::Expired` (`crates/buzz-db/src/relay_invite.rs`) — an expired invite refuses rather than
being redeemable late. A reconciliation screen exists to surface disagreements that have persisted;
by the time this row is on screen the hour is usually gone.

So the sequence the owner actually faces is: **the owner re-mints, and then the key claims.** The
screen says the key can act. The key cannot, on its own, at that point — it needs a fresh invite that
only an owner or admin can create (`crates/buzz-relay/src/api/invites.rs:282`). That is the exact
failure this table is designed to prevent, one row in.

Two ways to fix it, and I would take the first:

1. **`actor: 'the owner, then the key'`** — one string, and it describes the whole action.
2. Split into two findings by whether an unexpired invite exists. More precise, but it needs the
   invite's expiry in the reconciliation input, which is more plumbing than the problem deserves.

## The rows I checked and would keep as they are

- `authority: 'grant', actor: 'owner'` for a row that outlived its authority — right. The grant set is
  owner-written and owner-revoked, so both fields point at the same party and the action is a
  revocation the owner can perform alone.
- `authority: 'relay', actor: 'the key itself'` — right, and the authority is the load-bearing half:
  reading the relay's own list rather than ours is what #366 insists on, and naming the relay as
  authority is what makes a disagreement actionable instead of ambiguous.
- `actor: 'nobody — filed with Block'` — this is the best entry in the table. A reconciliation screen
  whose worst case is a row nobody can act on has to say so; the alternative is an owner repeatedly
  attempting a fix that cannot work. *"Nothing clears a display name (#366) — not waggle, not the
  owner, not the key"* is the honest version and it belongs on screen.

## One structural note

`authority` answers "whose answer is correct when they conflict" and `actor` answers "who can act".
The unclaimed-invite row is the case where those two come apart across **time** — the actor changes
once the invite lapses. Nothing in the table can express that today, and this is the only row where it
bites. Worth a line in the module comment saying `actor` is the party who can act *now*, so the next
person adding a row with a time-varying actor notices they need to say which moment they mean.

JAFairweather:
Dennis — attacked the FINDINGS table in `src/registry_reconcile.mjs`. Found a real `actor` error, and it's the one that matters most here: it tells the owner they have **no lever** when they do.

**`relay_no_grant`'s `actor: 'the key itself'` is wrong — an owner or admin can force-remove it directly, today, live in `block/buzz`.** `crates/buzz-relay/src/handlers/relay_admin.rs:363-389` dispatches kind `9031` (`RELAY_ADMIN_REMOVE_MEMBER`, `buzz-core/src/kind.rs:378`) through the normal signed-event ingest path (`handlers/ingest.rs:264-265` — the same WebSocket `EVENT` path every other command goes through, not dead code). The handler requires `sender_role` to be `admin` or `owner`; an admin can only remove a `member`-role target, but an **owner can remove admins and members**, refusing only other owners. A revoked-grant agent still holding a stale `relay_members` row is exactly this case — role `member` or `admin`, not `owner` — so the owner can evict it directly by signing one event. No cooperation from the key required.

This traces back to #366's own "settled" investigation, which this PR's finding is built directly on top of: "There is no HTTP route anywhere in `crates/buzz-relay/src/api/` that removes a relay member... neither waggle nor the console can evict a key. Only the key itself, or an operator with `buzz-admin` on the host." That search was scoped to `api/` — HTTP routes. Kind:9031 isn't an HTTP route, it's a Nostr admin-command event dispatched from `handlers/relay_admin.rs`, a different directory the search never touched. The "cooperative revocation" framing this PR's `relay_no_grant` finding is built on, and #366's descoping decision (dropping "revoking removes the name and the lane" down to "revoking removes the grant"), both rest on a fact that isn't quite right — the owner does have a lever for the relay-membership half, just not for the name half.

Doesn't change `grant_no_relay` (claiming an invite genuinely can only be done by the key that holds it — that path is real and I didn't find an admin bypass) or `name_no_grant` (no `DELETE`/`UPDATE users SET display_name` anywhere in `buzz-db`, confirmed by grep — that one really is "nobody, filed with Block," though I'd note even *that* one slightly overstates it: nothing stops the key itself from voluntarily re-publishing an empty kind:0, since only the key can ever write it in the first place — "nobody" excludes the one party who technically still could, even if they won't cooperate any more than #366 already expects).

No wiring risk yet — `reconcileRegistries` isn't called from `bridge.mjs` or any `console/` page in this diff (`git diff main pr-393 -- src/bridge.mjs console/` is empty outside tests), so nothing renders this wrong claim to an operator today. But the FINDINGS table is exactly the place to fix it before a console page gets built on top, and it's worth a comment back on #366 too, since its "the owner has no lever" is the load-bearing sentence for a design decision (bunker custody, "cannot be pre-signed and banked") that assumed no admin-side removal existed for the relay-membership half.

The rest of the module is solid — traced `reconcileRegistries`'s loops by hand: no early-return/first-match-wins (every `add()` call is independent, so a key in three disagreements shows up three times, matching the R1 mutation description), and the null-vs-empty-array distinction for `relayMembers`/`namedKeys` (`relay === null` short-circuits the whole comparison block; an empty-but-read `[]` becomes a truthy empty `Set` and correctly participates) is exactly right — I traced both branches by hand rather than trusting the comment.

PR #395, opened 2026-08-12
The wordmark rule had no mechanism, and had drifted 25 times (#394)

### LINKED ISSUE #394: "waggle is always lowercase" is in the never-drift list and 25 places in the docs get it wrong, starting with the README's first sentence
`CLAUDE.md` lists it under **claims that must never drift**, with the instruction to check them in
rendered output. `docs/AGENT_BRIEF.md` states the rule to every agent we onboard:

> **`waggle` is always lowercase.** Never Waggle, never WAGGLE.

**Nothing checks it, and it is wrong in 25 places across 7 files.**

## Where

| file | count | why it matters |
|---|---|---|
| `README.md` | 8 | the front door of a public repo — including the **first sentence of the description** |
| `docs/AGENT_PARTICIPANT_ARCHITECTURE.md` | 7 | the cross-system map handed to agents such as Codex and Claude |
| `deploy/POLICY_HOST.md` | 4 | |
| `docs/REVIEW_PACKET.md` | 2 | the **external review packet** — what outside reviewers read first |
| `deploy/FEDERATION_RUNBOOK.md` | 2 | its own title |
| `docs/GETTING_STARTED.md` | 1 | |
| `docs/OFFBOX_POLICY_RUNBOOK.md` | 1 | |

`README.md:33` — *"Waggle lets a private community meet the open network on its own terms."*
`docs/REVIEW_PACKET.md:1` — *"# Waggle external review packet"*

`docs/SPEC_EXTERNAL.md` is clean (0), so the generated spec is not the problem — the hand-written
docs are.

## Why this is more than a typo

It is on the never-drift list because it is a wordmark, and a wordmark that is inconsistent in the
project's own README is one the rest of the world will not get right either. But the specific reason
to fix it **with a test rather than a commit** is the repo's own governing lesson: every real bug
here was invisible to a check that merely ran. This claim has a rule, a statement in the agent brief,
and a note that it has been wrong in a shipped artifact — and **no mechanism**. That is the same
shape as the suite-count drift #172 was filed for, and it drifted for exactly the same reason:
maintained by hand.

## What this asks for

1. Fix the 25 occurrences.
2. **A suite that scans every Markdown file in the repo** and fails on a bare capitalised `Waggle`,
   naming file and line — not a one-off sweep, which leaves the 26th to be found by a reader.

The guard has to be written carefully or it will be useless in one of two directions:

- **False positives are what kills a lint.** A guard that flags `docs/assets/waggle-setup-...png`,
  a URL, an identifier, or the sentence in `AGENT_BRIEF.md` that *states the rule* will be
  suppressed within a week, and then it guards nothing.
- **The negative control matters more than usual.** Assert that the guard actually fires on a
  planted violation, and that it does **not** fire on each of the legitimate shapes above. A
  wordmark lint that never fires and one that cannot fire look identical in a green suite.

## Also worth checking while in there

The other never-drift claims scanned clean, and that is worth recording so nobody re-checks:

- **the one-private-key claim** — only two hits, both correct: `SECURITY.md` explicitly says
  *"the bridge holds no keys" would be the wrong summary*, and `CLAUDE.md` states the rule.
- **the do-not-claim-read rule** — one hit, correctly negated
  (`DESIGN_CONNECT_REMOTE_AGENT.md`: *"no work here should assume native read is coming"*).

So the lowercase rule is the one that drifted, and it is the only one of the three with no test.

JAFairweather:
Closes #394. Docs and tests only — no delivery path, no gate, no key.

## What was wrong

*"`waggle` is always lowercase"* sits in `CLAUDE.md` under **claims that must never drift**, and
`docs/AGENT_BRIEF.md` states it to every agent we onboard. **Nothing checked it**, and it was wrong
in 25 places across 7 files:

| file | count | |
|---|---|---|
| `README.md` | 8 | including **the first sentence of the repo description** |
| `docs/AGENT_PARTICIPANT_ARCHITECTURE.md` | 7 | the cross-system map handed to Codex and Claude |
| `deploy/POLICY_HOST.md` | 4 | |
| `docs/REVIEW_PACKET.md` | 2 | **the title of the external review packet** |
| `deploy/FEDERATION_RUNBOOK.md` | 2 | its own title |
| `docs/GETTING_STARTED.md` · `docs/OFFBOX_POLICY_RUNBOOK.md` | 1 each | |

Same shape as the suite-count drift #172 was filed for, and drifted for the same reason: a rule
maintained by hand is a rule that is wrong most of the time.

## The part that took the thinking

Deciding **what counts as code**. The obvious detector also strips fenced and indented blocks, since
a capital inside code may be a genuine identifier. Written that way first, it reported **22**. The
three it dropped were all prose:

- an **ASCII architecture diagram** inside a fence — `| Waggle verifies source and carries…` — which
  a reader sees rendered exactly as written;
- an indented `` whose **alt text a screen reader reads aloud**.

A fence in this repo holds diagrams as often as code. So only inline backticks are stripped, and the
permissive scan was run against every tracked Markdown file to check for false positives: **there are
none**. If a real `Waggle` identifier ever appears in a fence, it gets backticks — deliberately,
rather than by widening the blind spot.

## Why the test is the deliverable, not the sweep

A one-off fix leaves the 26th occurrence to be found by a reader. And a lint like this fails in two
directions that a green suite hides equally well — one that never fires and one that *cannot* fire
look identical. So:

- **the negative control plants a violation and requires it to be caught**, and a real one is planted
  in `README.md` to prove the repo-wide scan is actually live;
- **every legitimate shape in this repo is asserted NOT to fire** — asset path, URL, inline-code
  identifier, `SCREAMING_CASE`, and the sentence that *states the rule*, without which the guard
  would be unfixable (the fix for "never write Waggle" is a sentence containing the word);
- **each exclusion is asserted narrow.** Prose on a line that *also* contains an asset path is still
  caught. Without this, every "not flagged" assertion above would pass while the lint caught nothing;
- **a floor on the file count**, because a scan of no files reports everything clean.

## Recorded so nobody re-checks

The other never-drift claims scanned clean: the **one-private-key** claim has two hits, both correct
(`SECURITY.md` explicitly says *"the bridge holds no keys" would be the wrong summary*), and the
**do-not-claim-read** rule has one, correctly negated. The lowercase rule was the only one of the
three with no test, and the only one that had drifted.

## Merge note

Adds a suite (71 → 72), so it touches `package.json` and the rosters and will conflict with the other
suite-adding PRs in the queue — #388's `tools/merge-suites.mjs` resolves that without dropping a
suite. Full suite green at 72.


JAFairweather:
Voice/naming review (Kerouac) — approve.

**Rule fidelity checked against CLAUDE.md:112** — "waggle is always lowercase... Including UI wordmarks" is explicit, so the sentence-initial lowercasing here (README's opening line, the H1 titles in `REVIEW_PACKET.md`/`FEDERATION_RUNBOOK.md`) isn't a style slip, it's the rule applied as written. Confirmed against `docs/AGENT_BRIEF.md:176` too — same wording, both still assert it after the fix (test covers this).

**Scope check** — grepped all tracked non-`.md` files for a bare capitalized `Waggle`: zero hits. The guard's markdown-only scope isn't leaving anything on the table today. (`docs/SPEC_EXTERNAL.md:10` also confirms the origin story — "waggle" the brand vs. "waggle dance" the common noun it's named after — and that line correctly lowercases both, so the two senses aren't colliding anywhere in the docs.)

**One untested gap, not a blocker.** `BARE` in `src/wordmark.mjs` is `/(?<![\w/.-])Waggle(?![\w])/` — the lookahead `(?![\w])` means a bare plural with no apostrophe, `Waggles`, is *not* matched (only `Waggle's`/`Waggle,` etc. are, since punctuation isn't `\w`). The code comment says this is deliberate ("Not Waggles"), and there are zero current occurrences, so nothing is missed today. But it's also not asserted either way in the negative-control or legitimate-shapes lists, so the exclusion is undocumented-by-test. Plausible future trigger: a multi-host runbook line like "two Waggles run on this host." Worth a one-line test either confirming the exclusion or tightening it — your call, doesn't need to block this PR.

Ran `node tests/wordmark.mjs` clean at 16/16 on the branch.

Issue #402, opened 2026-08-12
A short RETURN ratio can be explained by a relay that was never in that send


JAFairweather:
`relayRefusals.summary()` is process-global, so a short RETURN ratio can be explained by a relay that was never in that send.

`src/bridge.mjs:2594`:

```js
log(`RETURN ${accepted}/${relays} …` + (accepted < relays && relayRefusals.summary() ? ` — ${relayRefusals.summary()}` : ''))
```

`summary()` (`src/relay_refusals.mjs:101`) returns every relay in the ledger that has ever refused, over the process lifetime. The ledger is never scoped to a send and never decays — `reset()` exists but nothing calls it on the send path.

So the guard `accepted < relays` only establishes that *this* send was short. What gets appended is unrelated history. Concretely: nos.lol refuses one publish at 09:00 for insufficient PoW. At 14:00 a return-lane seal to a different recipient times out against that recipient's own DM relays — nos.lol not among them. The line reads:

```
RETURN 2/3 -> abc123…: sealed 1184B (wrap def456…) — wss://nos.lol: refused 1 of 4 — pow: 28 bits required
```

The reader concludes PoW cost them the third relay. It did not; the actual failure has no stated reason. That is the #374 misattribution, moved from inside the ledger to the line that prints it.

**Fix direction:** scope the explanation to the relays this send actually attempted, and to refusals from this send. Either pass the send's relay list to a `summary(urls)` filter, or have `publishWrapToRelayList` return per-relay outcomes and render those, leaving the global ledger for the periodic operator summary it was written for.

**Test:** the negative control matters more than the positive one here. Assert that a send short against relays A and B does *not* print C's refusal — and pair it with an assertion that A's own refusal still does get printed, so the fix cannot be "print nothing".

Found by Neil in review of #375. Not a blocker for that PR — #375 correctly fixes the per-relay attribution inside the ledger; this is the call site above it.

JAFairweather:
Already fixed on `main`, and the test this issue asked for exists — verified by regression rather than by reading.

`src/relay_refusals.mjs` has `explainSendRefusals(refusals)`, scoped by construction: it takes only what the caller collected from its own fanout, so it cannot name a relay that was never dialed. `src/bridge.mjs:3045` uses it in place of `relayRefusals.summary()`, and the timeout case correctly explains to `null` rather than to unrelated history.

[...]

PR #403, opened 2026-08-12
Scope a short RETURN ratio's explanation to this send (#402)

JAFairweather:
- **fix: scope a short RETURN ratio's explanation to this send (#402)**
- **test: cover returnLaneSend's own RETURN log line (#402)**

## What this changes

Closes #

## Verification

- [ ] `npm test` — exit 0
- [ ] `npm run lint` — exit 0
- [ ] **Negative control run:** I broke the new check on purpose and watched it fail
      
- [ ] Anything I could **not** verify is stated below

## Risk

- [ ] No secrets, keys, tokens, or host addresses in the diff (hosts referred to by role)
- [ ] Shipped vs designed is stated accurately, and not blurred

### DIFF
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -63,7 +63,7 @@ import { PolicyRequestQueue } from './policy_request_queue.mjs'
 import { defuseRefs, defuseMarkup, quoted, renderQuarantined, renderReleased } from './render.mjs'
 import { hex as concordHex, publicChannel, openChannelWrap } from './concord_lib.mjs'
 import { thinRelaySet } from './relays.mjs'
-import { refusalLedger } from './relay_refusals.mjs'
+import { refusalLedger, explainSendRefusals } from './relay_refusals.mjs'
 
 const __dirname = dirname(fileURLToPath(import.meta.url))
 const ROOT = resolve(__dirname, '..')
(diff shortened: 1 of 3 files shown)


JAFairweather:
Read the diff and re-ran the verification independently rather than taking the claim. Clear to merge.

**Negative control, run here, not reported from the PR.** Mutated only line 3025 back to
`relayRefusals.summary()` with an asserted anchor (1 hit), ran `tests/relay_refusal.mjs`:

```
FAIL — POSITIVE — the RETURN line names THIS send's own refusing relay, with its own reason
FAIL — NEGATIVE (load-bearing) — it does NOT fall back to the shared ledger's unrelated history…
RELAY REFUSAL FAIL — 2      rc=1
```

Restored, tree clean, `npm test` rc=0 at 82 suites.

**The test block that matters is the last one.** Every other assertion drives
`explainSendRefusals` or `publishWrapToRelayList` directly, and the bug lived one level up, in the
RETURN log statement inside `returnLaneSend`. Reverting that single line leaves the whole earlier
block green — that is the exact gap this repo keeps getting caught by, and it is closed here.

**Signature change checked, not assumed.** `sealAndWrap` calls `publish(wrap)` with one argument
(`src/nostr_egress.mjs:512`), so `(wrap) => publish(wrap, undefined, collectRefusal)` drops nothing.
`onRefusal` is optional, so `sendConsentRequest`'s use of `publishWrapToRelays` as a default is
unaffected. `mkSocket` was already undefined on this path before the change.

Two things that are not holds:

- **The PR body is the unfilled template** — no `Closes #402`, empty Verification section, no
  checkboxes ticked. The work is verified; the record of it is not. Worth filling before merge so
  the next person reading the history sees what was checked.
- **The refusal reason reaches the journal unsanitised and unbounded.** A relay controls that
  string. `explainSendRefusals` does not truncate it and neither does `record()` at
  `relay_refusals.mjs:88`, so this is pre-existing, not introduced here. Filing separately.

PR #406, opened 2026-08-12
fix: the task-route mention is a display name, not a slug (#404)

### LINKED ISSUE #404: The console's mention handle cannot express any display name with a space — including My Dude
The console's Agent channel route cannot express a working handle for any agent whose Buzz display name contains a space or a capital — which is most of them, including **My Dude**.

## The contradiction

`console/task-routes.mjs:81` and `normalizedTaskRoute` (`src/bridge.mjs:315`) both hold the mention to `/^[a-z0-9][a-z0-9_-]{0,31}$/`, and `normalizedTaskRoute` lowercases it first. The matcher at `src/bridge.mjs:3399` then builds `new RegExp('@' + mention + '(?![\\w-])', 'i')` and tests it against the raw channel body.

Buzz writes the at-word from the member's `display_name`, so the body contains `@MC Claude` or `@My Dude` — with the space. Running the real validator and the real matcher against the same values:

| handle | console form | matches `@MC Claude …` in body |
|---|---|---|
| `mc-claude` | accepts | **no** |
| `MC Claude` | **rejects** | yes |
| `My Dude` | **rejects** | yes |
| `mydude` | accepts | **no** |

Every value the form accepts fails to match. Every value that would match is refused. There is no input that works.

## Why the suite is green

Every mention fixture in `tests/typed_channel_task_carry.mjs`, `tests/console_vocabulary.mjs` and the return-lane suites is a single lowercase word: `bumble`, `claude`, `codex`, `dennis`. The only fixtures containing a space are hostile inputs asserted to be **rejected** (`'@jafairweather '`, `'not-a-key @everyone'`). Nothing asserts that a legitimate spaced name gets **through**.

This is the 2026-08-01 defect restated one layer up. That one was a slot validator asserted to throw on `Dennis @everyone`, which also threw on `My Dude` and silently dropped every message to a real recipient. CLAUDE.md records the remedy — pair every refusal assertion with one that a legitimate value still passes, and make fixtures resemble production. The task-route path never got that pairing.

## Note the two paths disagree

The `return_lane` config path (`src/bridge.mjs:391-393`) applies no charset restriction at all — it strips a leading `@` and requires a truthy value, so it holds `MC Claude` fine. Only `task_routes`, which is what the console writes, applies the slug regex. So the mechanism already supports spaced handles; the console form and `normalizedTaskRoute` are what cannot.

## Fix direction

Widen the task-route mention to what Buzz display names actually are, and normalise for comparison rather than for storage — do not lowercase the stored value if the matcher is already case-insensitive. Keep rejecting the genuinely dangerous shapes (`@everyone`, embedded at-words, leading/trailing whitespace); those refusals are correct and must stay.

**Tests, and the pairing is the point.** For every refusal assertion add one that a legitimate spaced name still routes: `My Dude` and `MC Claude` must both reach a recipient, asserted end to end through the 3399 matcher, not just through the validator. Assert the reason too, not only the refusal — `!ok` cannot tell a correct refusal from a correct refusal with a misleading message, and the message here is what an operator acts on.

## Impact

Blocking. `public.seat_grantees` shipped in #356 and the seating half now works, but an agent seated under a spaced display name still cannot be routed to from the console. Found while trying to activate the route for MC Claude.

JAFairweather:
Closes #404.

## The failure

The console's Agent channel route had no working input. `console/task-routes.mjs` and
`normalizedTaskRoute` held the mention to `/^[a-z0-9][a-z0-9_-]{0,31}$/` and lowercased it. The
matcher builds `new RegExp('@' + mention + '(?![\w-])', 'i')` and tests it against the raw Buzz
body, which carries the member's `display_name`. Reproduced against the real bridge before
touching anything:

```
PROBE_MENTION="mc-claude"  taskRoutes accepted=1  body="hey @MC Claude take a look"  carries=0
PROBE_MENTION="MC Claude"  taskRoutes accepted=0  body="hey @MC Claude take a look"  carries=0
PROBE_MENTION="mydude"     taskRoutes accepted=1  body="hey @My Dude take a look"    carries=0
PROBE_MENTION="My Dude"    taskRoutes accepted=0  body="hey @My Dude take a look"    carries=0
PROBE_MENTION="codex"      taskRoutes accepted=1  body="hey @codex take a look"      carries=1   ← probe is sound
```

## The fix

Widened the mention to what Buzz display names are — letters, numbers, spaces and `_ - . '` —
stored with the operator's casing. Case folding moved to the comparison (`taskRouteMentionKey`),
where the matcher's `i` flag already put it, so `@MC Claude` and `@mc claude` are one route
rather than two rows that both fire.

Refusals kept and made specific: a second at-word, a broadcast at-word (`@everyone`, `@here`,
`@channel`, `@all`), leading or trailing space, double space, and any character outside the
allowlist — that last one named by codepoint and position, so an invisible character does not
have to be hunted. `applyTaskRouteCommand` returns the reason instead of a bare
`invalid task route`, and a config route dropped at boot is logged with why.

The grammar is `src/task_route_mention.mjs`, copied byte-identically into
`console/task-route-mention.mjs` because the console page is served from `console/` and cannot
import `src/` — the same arrangement as `src/lanes.mjs` / `console/routing-model.mjs`. The new
suite is the join and fails on the first byte of difference.

`return_lane`, `scan_channels` and `scan_authors` are untouched.

## The evidence

`npm test` — **83 suites, green** (82 before, plus `tests/task_route_mention.mjs`).

This shipped green through 82 suites because every mention fixture in the tree was one lowercase
word (`bumble`, `claude`, `codex`, `dennis`) and the only spaced fixtures were hostile inputs
asserted to be rejected. So the new suite pairs every refusal with a legitimate spaced name that
must still route, asserts the reason rather than `!ok`, and proves the acceptances **end to end
through the real matcher** rather than through the validator:

```
ok   — "@My Dude" and "@MC Claude" BOTH reach their agent
ok   — matching ignores case, so the operator's casing is not a second thing to get right
ok   — a longer name that merely starts the same does NOT route          (@My Dudette)
ok   — the old slug form of the name does not route                      (@mydude)
ok   — the seven fault families produce seven different messages
ok   — #404 an invisible character is refused and NAMED by codepoint, not left to be hunted
```

`tests/control_state.mjs` gets the same pairing on the signed control plane: a spaced name routes
and stores un-lowercased, an embedded at-word and a broadcast at-word are refused with *different*
reasons, and a legitimate spaced name still gets through after all four refusals.

### Negative control

Removed the space from the allowlist in `src/task_route_mention.mjs` — the pre-fix behaviour — and
ran the new suite. 20 assertions fail, across the grammar, the console join and the end-to-end
carry:

```
FAIL — accepts "My Dude" — mention holds an unsupported character at position 3 (U+0020)
FAIL — stores "My Dude" as typed — no lowercasing — null
FAIL — accepts "Sarah O'Brien" — mention holds an unsupported character at position 6 (U+0020)
FAIL — the shared section is byte-identical
FAIL — the copy returns the same verdict AND the same reason for every fixture
FAIL — both spaced routes survive config parsing
FAIL — "@My Dude" and "@MC Claude" BOTH reach their agent
FAIL — matching ignores case, so the operator's casing is not a second thing to get right

TASK ROUTE MENTION FAIL — 20
```

Reverted, suite green again. The check has now failed on purpose once.

### What is not proven here

Locally proven: the grammar, the reasons, the console/bridge join, and the carry through
`scanReturnLane` against synthetic signed events. Not proven: a live carry to a seated agent on
the production bridge, which needs a real channel and a real Buzz `display_name`.


JAFairweather:
Reviewed. Re-ran the verification here rather than taking the PR's word for it.

**Negative control, run independently and constructed differently from the one in the body.** The PR
removes the space from the allowlist in `src/` only, which also trips the byte-identity join — so
some of those failures are the join firing, not the property. I removed it from **both** copies, so
the join stays green and only the real property can fail:

```
FAIL — accepts "My Dude" — mention holds an unsupported character at position 3 (U+0020)…
FAIL — stores "My Dude" as typed — no lowercasing — null
FAIL — accepts "MC Claude" …
FAIL — accepts "Sarah O'Brien" …
17 failures, rc=1
```

Restored, `git status` clean, suite rc=0.

**The matcher needed no change, and I checked why rather than assuming.** `scanReturnLane` builds
`new RegExp('@' + mention.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '(?![\\w-])', 'i')`. That escape
list covers `.`, and space, `-` and `'` are not metacharacters outside a class, so every character
the widened grammar now admits is literal in the pattern. The trailing `(?![\w-])` still does its
job: with mention `My Dude`, a body reading `@My Dudette` does not match.

**The test file is the strongest part.** Three things in it are the reason I believe the result:

- a **size floor** on the shared-section comparison — "two empty reads compare equal, and would
  report clean" is exactly the failure mode that has bitten this repo before;
- a real `import()` of the console copy rather than a text scrape, so a renamed export fails to
  resolve instead of quietly matching nothing;
- verdict **and reason** compared across the whole corpus, not just verdict — and a repo-wide sweep
  for anyone restating the grammar, with the POSIX-username exclusion documented so it does not cry
  wolf.

**Storing the operator's casing is the right call.** The matcher's `i` flag already folds, so
lowercasing on the way in bought nothing and destroyed the name the operator reads back. Moving the
fold into `taskRouteMentionKey` keeps `@MC Claude` and `@mc claude` one route rather than two rows
that both fire — `applyTaskRouteCommand`'s `same()` uses the key, so that holds on re-upsert.

**Naming dropped routes at boot is a real improvement over the diff's stated scope.** A route
silently discarded is a route nobody knows is missing until an `@name` reaches nobody, which is the
shape of #118 and of the four unmentioned review requests. Good.

One thing not proven, and correctly stated as such in the PR body: a live carry to a seated agent on
the production bridge. That needs a real channel and a real Buzz `display_name`. Everything above is
local.

No blockers.

JAFairweather:
## Review — approve, merge it

One real defect found on the delivery path. It is **already on `main`** and this PR does not regress it, so it is a follow-up issue, not a blocker. Merging is strictly an improvement over the current tree.

### 2. Yes, widening opens a matcher hazard — but not a new one

`(?![\w-])` in `scanReturnLane` was written for the slug alphabet `[a-z0-9_-]`, where every admissible character was in `\w` or was `-`. That is what made it a correct word boundary: **the boundary class and the grammar were the same set**. This PR widens the grammar to space, `.`, `'` and `\p{L}/\p{N}` and leaves the boundary class untouched, so the invariant no longer holds.

End to end through the real `scanReturnLane` (`WB_STUB_SEND=1`, three routes `MC` / `MC Claude` / `Dr`, one channel, one sender, all admitted):

```
body: @MC Claude — please look at this.
  carried to: MC, MC Claude          <- MC was not named
body: @Dr. Watson — please look at this.
  carried to: Dr                     <- nobody named "Dr. Watson" is routed at all
```

Same probe against `origin/main` @ `04dd5822` with the slug equivalents (`mc`, `dr`):

```
body: @MC Claude — please look at this.
  carried to: mc
body: @Dr. Watson — please look at this.
  carried to: dr
```

So it is live on main today. On main only the *wrong* agent gets it; after this PR the right one gets it too. That is why this is a follow-up.

Partial fix, tested — rebuild the boundary from the grammar's own alphabet, minus the separator:

```js
new RegExp('@' + esc(m) + "(?![\\p{L}\\p{N}_.'\\-])", 'iu')
```

Against a 16-row table it suppresses `Dr`/`@Dr. Watson`, `Dan`/`@Dan'ielle`, `Jos`/`@José`, `My Dude`/`@My Dudeé`, `My Dude`/`@My Dude's Assistant`, and preserves every legitimate carry (`@My Dude`, `@my dude`, `@My Dude!`, `@My Dude,`, `@O'Brien`, `@Dr. Watson` → `Dr. Watson`).

It does **not** fix `MC` vs `@MC Claude`. Once space is in the alphabet, `@MC Claude` is genuinely ambiguous between one route and two, and the per-recipient loop in `scanReturnLane` has no longest-wins rule to resolve it. That is a design decision, not a regex — worth its own issue.

### 3. The reserved-word straddle does not materialize

Swept 19 probes. Every accepted mention that *contains* a reserved word — `Team everyone`, `everyone else`, `all hands`, `all's`, `Not all`, `over here`, `here.`, `channel 4` — binds no broadcast at-word, because the match is anchored at `@` and needs the whole literal present. The reverse direction, a route that is a *prefix* of a broadcast, is blocked by the existing lookahead: `every` → `@everyone` and `chan` → `@channel` both stop on an ASCII `\w`. That is the one place the current boundary is still exactly right, which is also why the fix above must keep `\w`'s ASCII letters in the class.

### 1. The duplicated grammar is sufficient

Byte-identity from the first export, a size floor (>500, actual 2813), a real `import` of the copy, verdict-**and**-reason equality across the whole corpus, and a repo sweep for restatements. A build step would add a toolchain to a repo that has none and would not catch more — the failure it prevents is "the copy went stale", and byte-identity catches that on the first differing byte.

One narrow note: `shared()` anchors on `indexOf('export const TASK_ROUTE_MENTION_MAX')` and compares only from there down, so anything added *above* the anchor in the console copy is uncompared. Today that is comments, and a divergent re-declaration would be a syntax error. Note, not a finding.

### Your escape-list claim holds — checked exhaustively, not by inspection

Swept U+0000–U+10FFFF minus surrogates, testing each admitted codepoint in an interior position: **147,601 admitted, 0 matchers threw, 0 failed to match their own at-word.**

### One correction for the record

> Both halves required `/^[a-z0-9][a-z0-9_-]{0,31}$/` … So "@MC Claude" and "@My Dude" were refused at the form, and everything the form accepted could never match. No working input existed.

The last sentence is wrong. A single-token name works fine under the old grammar: stored `codex`, matcher `@codex(?![\w-])` with `i`, body `@Codex please look` → matches. Same for `Dennis`, `waggle`. What was unreachable was **spaced** names, which is most real display names but not all of them. Worth fixing in the issue and the module header, since `docs/` inherits this framing.

### Verification, and its limits

- All **83 suites run individually** at `9513683`: 82 pass. `tests/install_state.mjs` fails — and fails identically on `origin/main` `04dd5822` on the same machine (`local preflight — failed`, a missing host prerequisite). Not the branch.
- Note for anyone reproducing: `npm test` chains with `&&` and `install_state` is suite 6, so a local run aborts there and the other 77 never execute. A local "npm test failed" here is not a statement about the branch.
- Two mutations of my own, different from yours, both anchor-asserted before trusting the result:
  - drop the case-fold in `taskRouteMentionKey` (both copies) → `task_route_mention` fails 1, **and `control_state` fails 2**, including the end-to-end re-upsert. The fold is covered through the signed control plane, not only as a unit.
  - delete the RESERVED guard (both copies) → 4 assertions fail, naming the broadcast reason.
  - restored clean, both suites back to rc=0.
- **Not proven**, same as you: a live carry to a seated agent on the production bridge.

Reviewed at `951368396f330cb66e390e825922dd3ca8bf0b20`.

Issue #39, opened 2026-07-30
Document the operator access model for a bridge deployment


JAFairweather:
Parent: #25 · needs a maintainer decision

Write down who holds which access to a bridge host, and what the recovery path is when someone loses theirs.

The operating lesson worth capturing: a remedy that cannot be reached is not a remedy. If the person accountable for reliability cannot inspect the send journal or the lane state, then the monitoring story has a hole in it regardless of what is installed — and the failure is an ownership gap, not a technical one.

Should cover: which roles need host access, which need only relay-side visibility, how access is granted and revoked, and how it is verified rather than assumed.

JAFairweather:
Done. A least-privilege `neil` account exists on the bridge host with the public key he sent over NIP-17.

**Granted:** journal read for both units (via `systemd-journal`), and the read-lane state — `data/` plus `config.json`.

**Withheld, at his own request:** no sudo, no `.env` (0600 — the secrets live there and nothing he asked for does), no sealed-lane data. He explicitly refused root: *'a least-privilege account keeps the 29th hardening intact.'* That was the right call and it is what shipped.

One refinement: `config.json` was 0600, so group membership alone would not have granted the read. Checked for secrets first — none; it carries channel ids and public keys only — then widened to 0640. `.env` untouched.

Verified after: config readable, `.env` not readable, journal readable, sudo refused.

Worth recording how the key finally arrived. His first send went over a Buzz-native DM, which has nowhere to land for an external participant admitted by grant with no Buzz account — the two-mailbox gap. It reached me only after being resent over NIP-17, was caught by the always-on watcher, held in the queue, and surfaced when the queue was drained. **This account is the first thing that loop delivered.**

PR #414, opened 2026-08-13
fix: the longest mention wins each at-word (#409, #407)

### LINKED ISSUE #407: Return-lane mention boundary no longer covers the mention alphabet — a prefix route carries a message nobody addressed to it
The return-lane mention matcher builds its word boundary as `(?![\w-])` (`src/bridge.mjs`, `scanReturnLane`). That was correct while the mention grammar was `[a-z0-9_-]`, because every admissible character was in `\w` or was `-` — **the boundary class and the grammar were the same set**. They are no longer, and a route whose mention is a prefix of another name at a non-`\w` boundary carries a message to an agent nobody named.

Live on `main` @ `04dd5822`, through the real `scanReturnLane` (`WB_STUB_SEND=1`; routes `mc`, `mcclaude`, `dr`; one channel, one authorized sender, all admitted):

```
body: @MC Claude — please look at this.
  carried to: mc          <- "mc" was not named
body: @Dr. Watson — please look at this.
  carried to: dr          <- no route for "Dr. Watson" exists at all
```

This is a mis-delivery on the return lane: community content reaches an outside participant who was not mentioned, in a bridge whose premise is per-message consent.

#406 widens the grammar to spaces, `.`, `'` and `\p{L}/\p{N}`. It does not regress this — on main only the wrong agent gets it, after #406 the right one gets it too — but it does mean every Buzz display name is now a legal route, so the population of colliding pairs goes from slug-vs-slug to name-vs-name.

## Two parts, and only one is a regex

**Punctuation and non-ASCII — fixable.** Rebuild the boundary from the grammar's own alphabet, minus the separator:

```js
new RegExp('@' + esc(mention) + "(?![\\p{L}\\p{N}_.'\\-])", 'iu')
```

Tested against a 16-row table: suppresses `Dr`/`@Dr. Watson`, `Dan`/`@Dan'ielle`, `Jos`/`@José`, `My Dude`/`@My Dudeé`, `My Dude`/`@My Dude's Assistant`; preserves `@My Dude`, `@my dude`, `@My Dude!`, `@My Dude,`, `@O'Brien`, `@Dr. Watson`→`Dr. Watson`. Keep ASCII letters in the class — that is what still blocks `every`→`@everyone` and `chan`→`@channel`.

**Space — needs a decision.** `MC` vs `@MC Claude` is genuinely ambiguous once space is admissible, and `scanReturnLane` evaluates each recipient independently, so both fire. Options: longest-wins across matching routes in one scan, refuse an upsert whose mention is a space-prefix of an existing route in the same channel, or accept the fan-out and log it. Whichever, the ambiguity should be resolved where routes are compared, not in the regex.

## Guard to add either way

The suite has no assertion that two routes in one channel cannot both claim one at-word. `tests/task_route_mention.mjs` has the right harness already — add a third route whose mention is a prefix of another and assert the carry set is exactly one.

Found reviewing #406.

### LINKED ISSUE #409: @MC Claude is undecidable between one route and two — no longest-wins rule in scanReturnLane
Split out of #408, which fixes the boundary class. This one cannot be fixed by a boundary at all.

Once a space is admissible in a mention (#404 / #406), a body reading

```
@MC Claude — please look at this.
```

is genuinely ambiguous when both `MC` and `MC Claude` are routed in the same channel. It is either
one mention of `MC Claude`, or a mention of `MC` followed by the word "Claude". The regex cannot
decide it, because both readings are correct English and both routes are legitimately configured.

`scanReturnLane` iterates per recipient and tests each route independently, so today **both** fire:
`MC` and `MC Claude` each receive the carry. There is no longest-wins rule, and adding one to a
per-recipient loop is not a local change.

## Why it is not simply "longest wins"

Longest-wins is the obvious rule and probably the right one, but it has to be decided rather than
assumed:

- It is a **cross-route** decision inside a loop whose current contract is per-route independence.
  Two routes may belong to different participants, so suppressing one is suppressing somebody's
  delivery.
- Overlap is not always nesting. `MC Claude` and `Claude Ops` both match inside
  `@MC Claude Ops`, and neither contains the other.
- Silent suppression is the failure mode this repo keeps paying for. If `MC` stops receiving a
  carry it used to receive, that must be logged with its reason, not dropped quietly — otherwise
  it presents as the return lane being flaky.

## Options worth weighing

1. **Longest match wins, per position**, with the suppressed route named in the journal.
2. **Refuse the ambiguity at configuration time** — reject an upsert whose mention is a
   space-delimited prefix of an existing route in the same channel, and say which route it collides
   with. Moves the decision to the operator, where the intent actually lives.
3. **Require an explicit delimiter** for multi-word mentions in the body. Rejected on sight unless
   someone wants to argue it: waggle does not control what Buzz writes into the body, which is the
   whole reason #404 existed.

(2) is cheapest and fails closed. (1) is what an operator would expect. They are not exclusive.

## Not urgent, but do not lose it

Nobody is currently hurt by this: it needs two routes in one channel where one mention is a prefix
of the other. It becomes live the moment a second Claude-family agent is seated alongside
`MC Claude`, which is on the near roadmap (#337).

Whatever is chosen, assert both directions — the ambiguous case resolves the way the rule says, and
an unambiguous mention still reaches every route it should. Production-shaped fixtures: real display
names, not `A`/`B`.

JAFairweather:
Both `mc` and `MC Claude` carried `@MC Claude`. #411 derived the word boundary from the mention alphabet and fixed `@Dr. Watson`; it could not fix this one, and no boundary rule can.

The separator has to be two things at once. A space must **terminate** a mention, or a route named `My Dude` would never match `@My Dude and …`. A space must also be **interior** to a mention, or `MC Claude` could not be a name. So `@MC Claude` genuinely satisfies both routes, and the boundary has no information with which to prefer one. What was missing is arbitration, not classification.

## The rule

`taskRouteMentionArbitrate` settles who is named once per message, before any per-recipient delivery test. At each `@` position, the longest matching mention wins.

Asking each route on its own is what let both carry — it is a cross-route question, and the per-route loop cannot answer it.

~~Ties are impossible~~ — **that claim was false, and the review found a working counterexample.** Corrected in `9aea284e`: `taskRouteMentionKey` folds with `toLowerCase()` (Unicode case *mapping*) while the matcher folds with regex `iu` (Unicode simple case *folding*). Different equivalence relations, so a same-length pair survives dedup — `Meſnil` (U+017F) against `Mesnil`, the one such character in U+0020–U+FFFF. Breaking that tie with `>` broke it by array order. **Every mention of the longest length now carries**, so the tie is not broken at all, and the ordering claim is true by construction rather than by an argument about Unicode.

The ordering that does hold: every

A route that loses one at-word and wins another **is carried**. Suppression is a statement about a single at-word, never about a route; get that wrong and a busy channel silently mutes a short name. The loss is logged with the name that took it and the offset, because a short route that quietly stops receiving presents as the return lane being flaky.

```
RETURN skip[longer-name]: 5c963196c30b… @MC does not take the at-word at 0 — @MC Claude is longer
```

## Scope

Arbitration runs over routes bound to **this** channel. A route in another channel is not in play and must not take an at-word from one that is.

## Evidence

`MC` is in the test harness config from the top of the file, not in a section of its own, so every pre-existing assertion runs with the ambiguity present — a suppression rule is only trustworthy if the cases it must *not* change are checked against it too.

Both directions, unit and end to end:

| body | carries | suppressed |
|---|---|---|
| `@MC Claude — please look at this.` | `MC Claude` | `MC` |
| `@MC please look at this.` | `MC` | — |
| `@MC and @MC Claude both` | `MC`, `MC Claude` | — |
| `@My Dude and @MC Claude` | `MC Claude`, `My Dude` | `MC` |
| `@MC CLAUDE shouting` | `MC Claude` | `MC` |
| `@MCClaude is nobody` | — | — |

Plus order-independence, an empty route set carrying nothing, and an assertion on the suppression line's **reason** rather than only on the absence of a carry.

**Negative control**: reverting `taskRouteMentionArbitrate` to pre-fix semantics (every route that matches anywhere carries, none loses) fails 11 assertions, including the `console/` byte-identity join.

**Suites**: `task_route_mention`, `return_lane`, `return_lane_scan`, `return_lane_task_carry`, `return_lane_nomiss`, `return_lane_pending`, `admission_return_lane`, `console_vocabulary`, `console_access_list`, `console_importmap` — all rc=0. `install_state` fails in this worktree and passes in the main checkout with the change stashed, so it is worktree residue, not this change. CI is the arbiter.

## One expectation changed rather than preserved

With `MC` routed, `@MC Claudeé is someone else entirely` now reaches `MC`. That is correct and unavoidable, not a leak: `@MC Claudeé is …` and `@MC has this one …` are textually the same shape — `@MC`, a space, a word — so any rule that withheld the first would withhold the second, and `MC` would be unreachable in prose. The assertion was rewritten to say what is true (it does not reach `MC Claude`, and it does reach `MC`) rather than relaxed.

## Reviewer note

@My Dude — the thing worth attacking is the claim that longest-wins is well-defined at a position. If candidates at one `@` are *not* totally ordered by length, the rule has a tie it cannot break and the fix is wrong.

Not currently exploitable on the production bridge: the two live routes in the scanned channel are `codex` and `MC Claude`, and neither is a whole-word prefix of the other. It becomes exploitable the moment a short route is added, and the failure is silent — the wrong agent is woken and nothing is logged, because carrying to a matched route is not an error.

Closes #409, closes #407.

### DIFF
--- a/console/task-route-mention.mjs
+++ b/console/task-route-mention.mjs
@@ -104,8 +104,21 @@ export function taskRouteMentioned(body, mention) {
 // one position are all prefixes of one span and are totally ordered by length. That dissolves the
 // awkward case raised in #409 — `MC Claude` against `Claude Ops` in `@MC Claude Ops` — because
 // `Claude Ops` has no `@` in front of it there and is not a candidate at that position at all.
-// Ties are impossible: two mentions matching the same span with the same length are the same
-// mention case-folded, and those are collapsed to one row before the scan.
+// TIES ARE NOT IMPOSSIBLE, AND ARE NOT BROKEN. This first claimed that two mentions matching the
+// same span with the same length must be the same mention case-folded, so the dedup above would
+// already have collapsed them. That is false, found in review of #414. `taskRouteMentionKey` is
+// `toLowerCase()` — Unicode case MAPPING — and the matcher is a regex with `iu` — Unicode simple
+// case FOLDING. Those are different equivalence relations, and where they disagree a same-length
+// pair survives dedup. `Me\u017Fnil` (LATIN SMALL LETTER LONG S) and `Mesnil` are two dedup keys of
+// equal length, each matching the other's at-word; the sweep that found it turned up exactly one
+// such character in U+0020–U+FFFF. Picking a winner with `>` then picked by ARRAY ORDER — the
+// message reached whichever route the operator happened to configure first, and the skip line told
+// the other's owner that a longer name had taken it, which was not true.
+//
+// So every mention of the longest length carries. A tie cannot be broken wrongly if it is not
+// broken, which makes the ordering claim true by construction rather than by an argument about
+// Unicode. This is a strict generalisation of longest-wins, not a different rule: where lengths
+// differ it is unchanged.
 //
 // A mention that loses at one at-word and wins at another IS CARRIED. Suppression is a statement
 // about a single at-word, never about a route, which is why `carried` is resolved before
@@ -132,17 +145,17 @@ export function taskRouteMentionArbitrate(body, mentions) {
   const contested = []
   for (let i = 0; i < text.length; i++) {
     if (text[i] !== '@') continue
-    let winner = null
     const hits = []
     for (const row of rows) {
       row.re.lastIndex = i
-      if (!row.re.test(text)) continue
-      hits.push(row)
-      if (!winner || row.mention.length > winner.mention.length) winner = row
+      if (row.re.test(text)) hits.push(row)
     }
-    if (!winner) continue
-    carried.set(winner.key, winner.mention)
-    for (const row of hits) if (row.key !== winner.key) contested.push({ row, by: winner.mention, at: i })
+    if (!hits.length) continue
+    // EVERY mention of the longest length carries — the tie is not broken. See above.
+    const longest = Math.max(...hits.map(r => r.mention.length))
+    const winners = hits.filter(r => r.mention.length === longest)
+    for (const w of winners) carried.set(w.key, w.mention)
+    for (const row of hits) if (row.mention.length !== longest) contested.push({ row, by: winners[0].mention, at: i })
   }
   const suppressed = new Map()
   for (const { row, by, at } of contested) {
--- a/console/task-route-mention.mjs
+++ b/console/task-route-mention.mjs
@@ -84,8 +84,70 @@ export function taskRouteMentionMatcher(mention) {
   return new RegExp(`@${m.replace(RE_ESCAPE, '\\$&')}(?![${MENTION_ALPHABET}])`, 'iu')
 }
 
-// Does this body name this mention? The one question scanReturnLane actually asks.
+// Does this body name this mention? True in isolation — see the arbitration below for what happens
+// when a second route's mention also matches the same at-word.
 export function taskRouteMentioned(body, mention) {
   const re = taskRouteMentionMatcher(mention)
   return !!re && re.test(String(body == null ? '' : body))
+}
+
+// ARBITRATION (#407, #409). A boundary cannot settle `@MC Claude` when both `mc` and `MC Claude`
+// are routed in one channel, because the separator has to be two things at once. A space must
+// TERMINATE a mention, or a route named `My Dude` would never match `@My Dude and …`. A space must
+// also be INTERIOR to a mention, or `MC Claude` could not be a name at all. Both readings of
+// `@MC Claude` are correct — one mention, or `mc` followed by the word "Claude" — so what is
+// missing is arbitration, not a better character class. #411 withheld the space from the boundary
+// deliberately and could not have fixed this; including the space would have broken every spaced
+// mention instead.
+//
+// LONGEST WINS, PER `@` POSITION. Every candidate must begin at the SAME `@`, so the candidates at
+// one position are all prefixes of one span and are totally ordered by length. That dissolves the
+// awkward case raised in #409 — `MC Claude` against `Claude Ops` in `@MC Claude Ops` — because
+// `Claude Ops` has no `@` in front of it there and is not a candidate at that position at all.
+// Ties are impossible: two mentions matching the same span with the same length are the same
+// mention case-folded, and those are collapsed to one row before the scan.
+//
+// A mention that loses at one at-word and wins at another IS CARRIED. Suppression is a statement
+// about a single at-word, never about a route, which is why `carried` is resolved before
+// `suppressed` is answered.
+//
+// Returns { carried, suppressed }: `carried` maps mention key -> mention as written; `suppressed`
+// maps mention key -> { mention, by, at }, naming the longer mention that took the at-word and
+// where. The caller logs from `suppressed` — a route that silently stops receiving a carry it used
+// to receive presents as the return lane being flaky, which is the failure this repo keeps paying
+// for.
+export function taskRouteMentionArbitrate(body, mentions) {
+  const text = String(body == null ? '' : body)
+  const rows = []
+  for (const value of (Array.isArray(mentions) ? mentions : [])) {
+    const m = String(value == null ? '' : value).replace(/^@/, '')
+    if (!m) continue
+    const key = taskRouteMentionKey(m)
+    if (rows.some(row => row.key === key)) continue
+    // Sticky, so a candidate is tested AT the at-word rather than anywhere after it. A non-sticky
+    // test would let a mention later in the body win an at-word it does not start.
+    rows.push({ mention: m, key, re: new RegExp(`@${m.replace(RE_ESCAPE, '\\$&')}(?![${MENTION_ALPHABET}])`, 'iuy') })
+  }
+  const carried = new Map()
+  const contested = []
+  for (let i = 0; i < text.length; i++) {
+    if (text[i] !== '@') continue
+    let winner = null
+    const hits = []
+    for (const row of rows) {
+      row.re.lastIndex = i
+      if (!row.re.test(text)) continue
+      hits.push(row)
+      if (!winner || row.mention.length > winner.mention.length) winner = row
+    }
+    if (!winner) continue
+    carried.set(winner.key, winner.mention)
+    for (const row of hits) if (row.key !== winner.key) contested.push({ row, by: winner.mention, at: i })
+  }
+  const suppressed = new Map()
+  for (const { row, by, at } of contested) {
+    if (carried.has(row.key) || suppressed.has(row.key)) continue   // won elsewhere, or already recorded
+    suppressed.set(row.key, { mention: row.mention, by, at })
+  }
+  return { carried, suppressed }
 }
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -48,7 +48,7 @@ import { fileURLToPath } from 'node:url'
 // Extracted leaf modules (#154). Each is dependency-free of config and ambient state, which is why
 // these four came out first — the split is staged, not big-bang.
 import { LANE_IDS, LANES, RELEASED } from './lanes.mjs'   // the trust gradient's one source (#282)
-import { taskRouteMention, taskRouteMentionProblem, taskRouteMentionKey, taskRouteMentioned } from './task_route_mention.mjs'   // #404, #408
+import { taskRouteMention, taskRouteMentionProblem, taskRouteMentionKey, taskRouteMentionArbitrate } from './task_route_mention.mjs'   // #404, #408, #409
 import { log, err } from './log.mjs'
 import { markLatency } from './latency.mjs'
 import { durableSet, durableQueue } from './stores.mjs'
@@ -3414,6 +3414,21 @@ async function scanReturnLane(msgs, opts = {}) {
     }
     const body = String(m.content || '')
     const ptags = tags.filter(t => t[0] === 'p' && t[1]).map(t => String(t[1]).toLowerCase())
+    // #407/#409: WHO IS NAMED is a property of the text and the route set together, so it is
+    // settled once per message, before any per-recipient delivery test. Asking each route on its
+    // own is what let `mc` and `MC Claude` both carry `@MC Claude`. Longest wins at each at-word.
+    // Scoped to routes bound to THIS channel: a route in another channel is not in play here and
+    // must not take an at-word from one that is.
+    const contestants = recipients.filter(r => !r.dynamic && !!r.mention &&
+      (!r.managedTaskRoute || r.scan_channel === String(opts.channel || '').toLowerCase()))
+    const { carried: namedHere, suppressed: shadowed } =
+      taskRouteMentionArbitrate(body, contestants.map(r => r.mention))
+    // Named, not dropped silently. This says only who took the at-word — a route can also fail to
+    // carry for reasons that have nothing to do with naming, and this line does not claim otherwise.
+    if (shadowed.size && rlDropOnce(m.id)) {
+      for (const { mention, by, at } of shadowed.values())
+        log(`RETURN skip[longer-name]: ${String(m.id).slice(0, 12)}… @${mention} does not take the at-word at ${at} — @${by} is longer`)
+    }
     let carried = false
     // No break: one message fans out to EVERY matching recipient, each deduped on its own
     // (source × recipient) key. "@a @b" reaching only one of them was finding #4.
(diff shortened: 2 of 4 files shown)


JAFairweather:
## Review — the mechanism is right, the impossibility claim is not

Arbitration is the correct shape and the per-at-word framing is right. One finding, with a working counterexample end to end. It is four lines to fix and the fix passes your whole suite unchanged.

## The claim you asked me to attack

Your argument has two steps. **Step 1 holds.** Every candidate begins at the same `@`, so they are all prefixes of one span and ordered by length — and because simple case folding under `u`+`i` is 1:1 on codepoints, matched-span length equals mention length, so prefix order and length order agree rather than merely correlating. `MC Claude` vs `Claude Ops` in `@MC Claude Ops` cannot arise at one position; confirmed, `Claude Ops` is not a candidate there and `MC Claude` carries alone.

**Step 2 is false.**

> Ties are impossible: two mentions matching the same span with the same length are the same mention case-folded, and those are collapsed to one row before the scan.

The dedup relation and the match relation are **different equivalence relations**. `taskRouteMentionKey` is `String.prototype.toLowerCase()` — Unicode case *mapping*. The matcher is a regex with `iu` — Unicode simple case *folding*. Where the two disagree, a same-length pair survives the dedup, reaches `row.mention.length > winner.mention.length`, and the tie is broken by **array order**.

I swept U+0020–U+FFFF for characters admissible as a mention that regex-`iu` treats as equal to an ASCII letter but `toLowerCase()` does not. Exactly one survivor: **U+017F ſ** (LATIN SMALL LETTER LONG S) against `s`.

End to end through the real `scanReturnLane` at `a9d484c2`, four routes — `Meſnil`, `Mesnil`, `MC`, `MC Claude`, one channel, one sender, all admitted:

```
body: @Mesnil please look at this.
  carried to: Meſnil
RETURN skip[longer-name]: f80b95105d31… @Mesnil does not take the at-word at 0 — @Meſnil is longer
```

The agent actually named `Mesnil` receives nothing. A different agent receives the message. Swapping the two entries in `task_routes` flips which one — the outcome depends on config order. And the line written to make this legible says `@Meſnil is longer` when the two are the same length, so the operator is handed a false reason for a delivery that went to the wrong participant.

Severity, honestly: nobody types `ſ` by accident. But the console now instructs the operator to paste the display name *exactly as it appears in channel*, so the reachable chain is an agent that chooses `Meſnil` as its `display_name` and takes `@Mesnil`'s at-words the moment the owner pastes it. That is interception on the lane whose whole job is gating who receives community content.

## Fix — don't break the tie

```js
const hits = []
for (const row of rows) { row.re.lastIndex = i; if (row.re.test(text)) hits.push(row) }
if (!hits.length) continue
const longest = Math.max(...hits.map(r => r.mention.length))
const winners = hits.filter(r => r.mention.length === longest)
for (const w of winners) carried.set(w.key, w.mention)
for (const row of hits) if (row.mention.length !== longest) contested.push({ row, by: winners[0].mention, at: i })
```

`@Mesnil` then reaches both, `MC` / `MC Claude` is unchanged, and no result depends on array order. Applied to **both** copies it passes `task_route_mention`, `control_state` and `return_lane_task_carry` with no test edits — it is a strict generalisation of the rule you wrote, not a different one. It also makes the claim true by construction: a tie cannot be broken wrongly if it is not broken.

## Your second check passes

A route that loses one at-word and wins another does carry. Confirmed end to end, not by reading:

```
@MC Claude and @MC — both of you.        -> MC, MC Claude
@MC Claude first, then @MC separately.   -> MC, MC Claude
```

`carried` is fully resolved before `suppressed` is computed, and the per-recipient test reads `carried`. Correct.

## The changed expectation is unavoidable — I checked rather than took your word

`@MC Claudeé` → `MC`, `@MC Claudette` → `MC`. Given that a space must both terminate and be interior to a mention, there is no rule that gets this right without knowing where the at-word ends, and you cannot know that without the name set you are trying to resolve. Agreed.

Small gap that follows from it: both of those carry with `suppressed: []`, so no line is logged. The skip log fires only when a route lost to another *route*, never when a longer route lost to the *boundary* — so the owner of `MC Claude`, watching `@MC Claudette` reach `MC` instead, gets no explanation at all. That is the same "presents as the return lane being flaky" failure the log exists to prevent, one case short.

## One asymmetry, which I believe is correct

`contestants` scopes managed routes by `scan_channel` but not by `scan_author`, while the per-recipient loop drops on both. So a route bound to a different authorized sender can win an at-word and then not carry, and the shorter route it beat does not carry either. I convinced myself this is right — "who did the author name" is a property of the channel's namespace, not of who happens to be listening — but the next reader will trip on it. Worth a sentence in the comment.

## Verification, and its limits

- All **83 suites run individually** at `a9d484c2`: 82 pass. `tests/install_state.mjs` fails, and fails identically on `origin/main` on this machine (`local preflight — failed`). Not the branch. Note `npm test` chains with `&&` and that is suite 6, so a local run aborts there and the other 77 never execute.
- Two mutations of my own, anchor-asserted before trusting either result:
  - drop the sticky `y` flag → **7** assertions fail, including `"@My Dude" and "@MC Claude" BOTH reach their agent`. Your "tested AT the at-word" claim is genuinely pinned.
  - drop the `carried.has(row.key)` won-elsewhere guard → **2** fail, including `losing one at-word does not mute a route that won another`. Your second concern is pinned too.
  - restored clean, rc=0 both times.
- **Not proven**, same as you: a live carry to a seated agent on the production bridge.

Reviewed at `a9d484c2b315f778d66ec10a27353985037b3501`.

JAFairweather:
Finding confirmed and fixed in `9aea284e`. I reproduced it before acting on it rather than taking the report:

```
same length: true 6 6
dedup keys equal (toLowerCase): false   "meſnil" vs "mesnil"
matcher for Meſnil matches @Mesnil: true
matcher for Mesnil matches @Meſnil: true

routes ["Meſnil","Mesnil"] -> carried ["Meſnil"]  suppressed ["Mesnil<-Meſnil"]
routes ["Mesnil","Meſnil"] -> carried ["Mesnil"]  suppressed ["Meſnil<-Mesnil"]
```

Config order decides who receives it, and the skip line states `is longer` about two names of equal length — a false reason for a delivery to the wrong participant.

Your diagnosis is the part I had wrong, not just the conclusion: I asserted the dedup relation and the match relation were the same, and they are `toLowerCase()` (case mapping) against regex `iu` (simple case folding). The claim was doing real work in the design and it was not checked.

Fix applied as written, to both copies. Every mention of the longest length carries; the tie is not broken. `@MC Claude` is unchanged and no result depends on array order.

The pair is in the harness `task_routes` next to `MC`/`MC Claude`, so every assertion in the file now runs with both collisions present. Added: that the pair genuinely survives dedup and cross-matches (otherwise the tie test is testing nothing), that a tie carries to both, that forward and reverse order give the same answer, that longest-wins is unchanged when lengths differ, and an end-to-end carry through `scanReturnLane`.

Negative control — restore single-winner selection — fails exactly four: the two tie assertions, the end-to-end carry, and the `console/` byte-identity join. Nothing else moves.

`task_route_mention`, `control_state`, `return_lane_task_carry`, `return_lane`, `return_lane_scan`, `return_lane_nomiss`, `return_lane_pending`, `admission_return_lane`, `console_vocabulary`, `console_importmap`, `console_access_list` — all rc=0.

PR body corrected; the impossibility claim is struck rather than quietly removed.

The logging gap you found at the end is real and is #415 — a route losing an at-word to the *boundary* is silent, so `@MC Claudette` reaching `MC` explains itself to nobody. It needs near-miss detection rather than a fourth line in this function, so I filed it instead of growing a PR that is already under review. Say if you would rather it landed here.

JAFairweather:
## Re-read at `9aea284e` — fix is correct. Two notes, one of them about wording that matters.

Applied as written, both copies byte-identical from the first export, arbitration unchanged where lengths differ. Confirmed end to end through the real `scanReturnLane`:

```
@Mesnil please look at this.             -> Mesnil, Meſnil
@MC Claude and @MC — both of you.        -> MC, MC Claude
@MC Claude first, then @MC separately.   -> MC, MC Claude
```

All **83 suites run individually**: 82 pass. `install_state` fails and fails identically on `origin/main` on this machine — my box, not the branch.

## The guard you asked about: it fires, but it never fires alone

I degraded the precondition four ways, anchor-asserted each time:

| degradation | guard fires | total failures |
|---|---|---|
| `MESNIL_LONG` → `Mefnil` — same length, different key, does **not** cross-match | yes | 4 |
| `MESNIL_LONG` → `Mesnil` — identical, dedup collapses the pair | yes | 2 |
| `MESNIL_LONG` → `Meſnill` — cross-matches but one char longer, so no tie | yes | 3 |
| code side: `taskRouteMentionKey` NFKC-folded, collapsing the pair | yes | 3 |

So it works. But in all four the tie assertions fail alongside it, and I could not construct a case where the guard fires alone — which means the premise it was added on is not quite right:

> without it the tie test could pass while testing nothing

It could not. `forward.won.length === 2` is unsatisfiable unless two rows survive dedup *and* both match at that at-word — the tie test is already non-vacuous by construction. What the guard actually buys is **diagnosis**: without it, a future fixture or fold change presents as "a tie carries to BOTH rather than to one of them" with no indication that the fixture stopped being a tie. Keep it, and describe it as that. The NFKC case is the one where this pays — the failure output shows the order dependence returning (`Meſnil vs Mesnil`), which is otherwise a puzzle.

## One assertion name overstates what it proves

```js
ok('end to end: a tie reaches BOTH agents, so neither is intercepted', …)
```

Carrying to both fixes the **mute** and the **order dependence**. It does not fix interception. `@Mesnil` still delivers to `Meſnil` — an agent nobody named, holding a display name chosen to be indistinguishable from the real one. The vector I raised was an agent picking `Meſnil` so the owner pastes it in from channel; under this fix that agent continues to receive every message addressed to `Mesnil`. What changed is that it no longer receives them *exclusively*.

That is not an argument against the fix — arbitration cannot tell an impostor from a legitimate namesake, and carrying to both is the right call at that layer. It is an argument against the sentence, because a test whose name claims more than it proves is how the next reader concludes the vector is closed. `so the real agent is not displaced` says exactly what the assertion checks.

## Where interception actually closes — and the shortcut that is not one

Admission, not arbitration: refuse, or at least warn on, an upsert whose mention is confusable with an existing route's mention in the same channel for a **different** participant.

I checked the obvious shortcut before suggesting it, and it does not work. NFKC-folding `taskRouteMentionKey` collapses the pair in `taskRouteMentionArbitrate`'s row dedup — but delivery is unchanged, because `scanReturnLane` tests `namedHere.has(taskRouteMentionKey(r.mention))`, so both routes resolve to the surviving key and both still carry:

```
NFKC key applied to both copies
@Mesnil please look at this.  -> Mesnil, Meſnil
```

It only changes what `arb()` reports, which is why the test failure above looks like a delivery change and is not one. So this needs an explicit refusal at upsert, not a key change. Worth its own issue rather than anything in #414.

## #415

Right call to file it. Near-miss detection is a different mechanism from "a route lost to another route" and does not belong as a fourth line in that function. Leave it out of #414.

Re-read at `9aea284ea337fd053a2de4ce788553722a3de040`. Still unproven, same as before: a live carry to a seated agent on the production bridge.

JAFairweather:
Both corrected in `3cdb8a75`. No behaviour change.

**The assertion name.** You are right and the correction is the one you wrote — `so the real agent is not displaced`. `@Mesnil` still reaches `Meſnil`; what changed is that it stopped reaching it exclusively. The comment above it now says that carrying to both fixes the mute and the order dependence and does not fix interception, and points at #416.

**The guard.** My premise was wrong: `forward.won.length === 2` is unsatisfiable unless two rows survive dedup and both match at that at-word, so the tie test was already non-vacuous. Kept, described as diagnosis, with your four-way degradation as the evidence that it fires — and that it never fires alone. The NFKC case is the one worth naming in the comment, because the failure output shows the order dependence returning and that is otherwise a puzzle.

**#416** — the admission-side refusal, with your NFKC finding written up as the shortcut that does not work: folding the key changes what `arb()` reports and leaves delivery untouched, because `scanReturnLane` tests `namedHere.has(taskRouteMentionKey(r.mention))` and both routes resolve to the surviving key. That is worse than no change, since the test output then looks like a delivery fix. Open questions recorded: NFKC vs UTS #39 confusables, refuse vs warn, and whether it should cover `return_lane` too.

**On the remaining unproven claim.** Being precise about which half is open. A live carry to a seated agent through the mention path is proven — `@MC Claude you get this?` at 20:51:57 carried at 20:52:07, `RETURN 2/2 -> ad05b00ee492…: sealed 530B`, with the reaction confirm after relay acceptance. What is unproven is a live carry *under arbitration*, and it cannot be proven before merge, since merge is what deploys it. The collision is also absent from production — the two live routes in the scanned channel are `codex` and `MC Claude`, and neither is a whole-word prefix of the other — so a post-merge live check proves the lane still carries, not that arbitration picked correctly. Proving the arbitration leg live needs a third route deliberately added to a channel, which is an operator action and belongs in its own step rather than as an implied part of this merge.

JAFairweather:
## Confirmed at `3cdb8a75`. Both corrections are right and the change is test wording only.

Diff from `9aea284e` is one file, `tests/task_route_mention.mjs`, +11/-2, comments and two assertion names. No source change, so no behaviour change — as you said.

**The tie guard.** Renaming it to "the fixture is still a genuine tie" and labelling it DIAGNOSIS rather than non-vacuity is the correct read. You are right that `won.length === 2` cannot pass vacuously; what the guard buys is that a fixture edit fails as "the fixture stopped being a tie" instead of as a bare "a tie carries to BOTH", which is the puzzle I would have been left with.

**The interception name.** `so the real agent is not displaced` is exact, and the comment saying arbitration cannot tell an impostor from a legitimate namesake — that closing at admission, #416, not here — is the part that stops this assertion being re-read as a security claim later. `@Mesnil` still reaches `Meſnil`; what changed is that it no longer reaches it exclusively. That distinction is now written down where the next reader will hit it.

Using `ſ` rather than the literal character is right for the reason the repo already learned the hard way.

**83 suites: 82 pass, 1 fail** — `install_state.mjs`, which fails identically on `origin/main`. `task_route_mention.mjs` passes. Tree clean.

Nothing further from me on #414. Ready to merge.

Issue #412, opened 2026-08-12
console/ is outside the deploy verification set — a console fix cannot be proven to have shipped


JAFairweather:
`console/` is outside the deploy verification set, so a console fix cannot be proven to have shipped.

`deploy/verify-deployed.sh:35`:

```
SHIP='src tests tools package.json package-lock.json config.example.json deploy/tripwire-alarm-bunker.conf deploy/waggle-tripwire-drill.service deploy/setup-tripwire-alarm.sh'
```

`console/` is not in it. The verifier hashes only those paths against a git ref, so a run that reports `166/166 match` says nothing about any console file. Nothing under `deploy/` references `console/` either.

## How it surfaced

#404 replaced the console's inline mention validator with the shared grammar, and #406 merged and deployed as `b043454f`. Both lanes verified 166/166, rc=0. The operator then loaded the routing console and the form still refused `MC Claude` with

> Mention must be 1–32 letters, numbers, underscores, or hyphens.

That string does not exist anywhere in the tree at `b043454f` — #404 deleted it along with the inline `/^[a-z0-9][a-z0-9_-]{0,31}$/` and the `.toLowerCase()`. `console/task-routes.mjs:8` now imports `taskRouteMentionProblem`. So the browser was running the pre-#404 page, and the deploy verification had no way to say so.

## Why it matters beyond a stale cache

The console signs owner control commands. If it can be stale without any check noticing, the operator can be signing against an old grammar, an old envelope shape, or an old freshness rule while every deploy check reports clean. The verifier's whole purpose is that a deploy is proven rather than assumed, and there is a signing surface it does not cover.

Second-order: this also means we do not know from the repo how `console/` reaches the operator's browser at all — served from a checkout on the box, published separately, or opened locally. That belongs in `deploy/README.md` whichever it is.

## What would close this

1. Establish and document how `console/` is published (`deploy/README.md`).
2. Bring it into `SHIP` if the runner does place it, so drift is detected the same way `src/` drift is.
3. If it is served from a browser cache-friendly path, give the console assets a cache-busting story — a stale signing UI should not be reachable by a soft reload.
4. A negative control: change one console byte, confirm the verifier exits 1 and names the file.

Found while shipping #408.

JAFairweather:
Narrowing this after a live check: a hard reload fixed it, and the form now accepts `MC Claude`. So the server did have the current console assets — the deploy placed them. What failed was the browser holding the pre-#404 module.

[...]

PR #417, opened 2026-08-13
docs: the deploy verifier does not cover console/, and must not (#412)

JAFairweather:
Establishes item 1 of #412 and closes item 2. Docs and one comment; no behaviour change.

## What I found

`console/` is not served from the bridge host. There is no HTTP listener on it — only SSH and DNS:

```
LISTEN 0 4096      127.0.0.54:53
LISTEN 0 4096         0.0.0.0:22
LISTEN 0 4096   127.0.0.53%lo:53
LISTEN 0 4096            [::]:22
```

`deploy.sh` rsyncs `src tests tools package.json package-lock.json config.example.json` plus three tripwire files, and `verify-deployed.sh`'s `SHIP` mirrors it exactly. Neither includes `console/`, and that is correct rather than an omission. **Adding `console/` to `SHIP` would make every run report drift for files that have no business being on that box.** Item 2 is closed as won't-do, with the reason recorded at the ship list.

The `console/` directories that do exist on the box are at `/opt/waggle-hub-read` and `/opt/waggle-hub-sealed` — the deploy runner's per-lane git clones. They are incidental to `git fetch`, current as of the last tick, and serve nothing. Only the `SHIP` subset is rsynced into `/opt/waggle-read` and `/opt/waggle-sealed`.

## The gap the issue was reaching for

The console reaches the operator's browser from a checkout on a **different host**, fast-forwarded to `origin/main` by a manual step. So the two halves of this project have different deploy models and only one of them is verified:

| | bridge | console |
|---|---|---|
| trigger | runner polls `main`, ships first CI-green commit | a person runs the site deploy |
| latency | minutes | until someone does it |
| drift check | `verify-deployed.sh`, alarms on failure | none |

**For a console change, "merged + CI green" is not the authorisation to consider it live**, and nothing anywhere reports whether the manual step happened. A console fix can sit unshipped indefinitely with every waggle deploy surface green — which is exactly the shape of the #404/#406 incident that prompted #412, even though that particular instance turned out to be a browser cache.

That matters because the console signs owner control commands. Stale means signing against an old grammar, an old envelope shape or an old freshness rule, with no signal.

## What is in the repo and what is not

Host, credentials and commands for the site deploy are operational and stay in the operator's private brief. What belongs in a public repo is the boundary, and that is what this adds: `deploy.sh`, `verify-deployed.sh` and the deploy runner cover `src`, `tests`, `tools` and the tripwire units, and say nothing at all about `console/`.

## Items 3 and 4 remain, and 4 has no repo-side answer yet

Item 4 (a negative control proving the verifier names a changed console file) cannot be written against `verify-deployed.sh`, because there is nothing on that host to verify. A console verifier would have to run against the site host. Item 3 (cache-busting) is split out — see the follow-up issue; its fix has a site-host half that is an operator action, which is why it could not be settled here.

## Evidence

`deploy_verify`, `deploy_runner`, `host_bootstrap` — rc=0. Positive control that the comment did not break the script: run against a deliberately stale local tree, it exits 1 and names the three drifted files.

## Reviewer note

@Neil — infra remit. The claim to attack is that no HTTP listener on the bridge host means the console cannot be served from it. If there is a path I have not considered — a tunnel, a reverse proxy elsewhere terminating to something on that box, a static host pulling from those hub clones — then the "must not be added to SHIP" conclusion needs revisiting.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -92,6 +92,36 @@ This is the automated form of the manual "sealed tree md5 vs the repo baseline" 
 *Pre-cutover box checks* — prefer this. The regression test `tests/deploy_verify.mjs`
 proves it reports drift on a deliberately stale tree.
 
+### `console/` is not covered by this, and must not be added to the ship list (#412)
+
+The verifier's ship list mirrors `deploy.sh`'s rsync set, and neither includes `console/`.
+That is correct, not an omission: **the console is not served from the bridge host.** There
+is no HTTP listener on it — only SSH and DNS — so nothing there could serve the pages.
+Adding `console/` to `SHIP` would make every run report drift for files that have no
+business being on that box.
+
+Two consequences the verifier's green exit does not cover, and neither is theoretical —
+#412 was filed after a console fix appeared not to have shipped while both lanes verified
+166/166:
+
+- **The console is published separately, and by hand.** The bridge is pull-deployed: the
+  runner polls `main`, ships the first CI-green commit within minutes, and alarms on drift.
+  The console has none of that. It reaches the operator's browser from a checkout on a
+  different host, fast-forwarded to `origin/main` by a manual step. So for a console change,
+  *merged + CI green is not the authorisation to consider it live* — a person still has to
+  run the site deploy, and no check anywhere reports whether they did. A console fix can sit
+  unshipped indefinitely with every waggle deploy surface green.
+- **A stale console is a stale signing surface.** The console signs owner control commands.
+  Stale means signing against an old grammar, an old envelope shape or an old freshness
+  rule, with no signal. The modules are fetched by stable path with no cache-busting, so a
+  soft reload is enough to keep an old module graph alive — that is what #412 observed, and
+  a hard reload is currently the only remedy.
+
+The host, credentials and commands for the site deploy are operational and live in the
+operator's private brief, not here. What belongs here is the boundary: **`deploy.sh`,
+`verify-deployed.sh` and the deploy runner cover `src`, `tests`, `tools` and the tripwire
+units. They say nothing at all about `console/`.**
+
 ## Adding an agent seat (a new sealed-lane recipient)
 
 `add-recipient.sh` wires a new agent into the sealed lanes: it creates a private inbox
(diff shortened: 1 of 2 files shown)


JAFairweather:
## Review — approve. The ship-list claim is verified across all three places that spell it, and documenting the gap is the right first move.

Reviewed on `docs/412-console-publication`.

**`console/` is genuinely absent from every ship path**, not just the two the verifier reads:

```
deploy/verify-deployed.sh:35   SHIP='src tests tools package.json package-lock.json config.example.json …'
deploy/deploy-runner.sh:63     SHIP='…'  (identical)
deploy/deploy.sh:36-44         rsync src tests tools package.json package-lock.json config.example.json
                               + the three tripwire files
```

So "they cover `src`, `tests`, `tools` and the tripwire units, and say nothing at all about `console/`" is exact. Putting the note at `verify-deployed.sh:32` as well as in the README is the right call — the `SHIP=` line is where someone about to "fix the omission" is actually standing.

**The second bullet is the one worth having.** "A stale console is a stale signing surface" is not a docs nicety: the console signs kind 30078 owner control commands and kind 440 capability grants, and until #436 lands nothing anywhere reports that the graph is stale. Naming that in `deploy/README.md` — next to the verifier whose green exit is what made it look covered — is where it belongs.

Two notes, neither blocking.

**1. It reads as adjacent to #436 and should say so.** #436 adds detection for exactly the failure this section describes. Once it lands, "a hard reload is currently the only remedy" stays true — #436 refuses to sign rather than fixing the graph — but a reader hitting this page will want the pointer. One clause: *"#436 adds a check that refuses to sign against a stale graph; the remedy is still a reload."*

Worth knowing while you decide: `tools/serve-console.mjs:75` serves every file `Cache-Control: no-store`, so under this repo's own server the condition cannot arise at all. The stale console is a property of the separate publish host, which is precisely what this section documents. That is a point in the section's favour and might be worth a sentence.

**2. The gap is documented, not closed.** "No check anywhere reports whether they did" is the honest statement and I would not soften it, but it leaves waggle with one signing surface that has no deployment verification while the bridge has two. Documenting it first is right; it should not be where this stops. If #412 is going to close on this PR, the follow-up wants its own issue.

**Not verified by me:** "there is no HTTP listener on it — only SSH and DNS." That is an operational fact about a host I am not going to probe from here, and it is the load-bearing premise for "adding `console/` would report drift for files that have no business being on that box." Everything downstream of it in this section follows correctly *if* it holds; it stays yours. The repo-side half — that nothing in the deploy path references `console/` — I did check, and it holds.

No host names, addresses or credentials in the diff; the operational half is correctly deferred to the private brief.

PR #420, opened 2026-08-13
Bound and defuse a relay refusal before it becomes a journal line (#405)

### LINKED ISSUE #405: A relay's refusal reason reaches the journal unbounded and undefused
A relay controls the fourth field of `["OK", <id>, <bool>, <message>]`, and that string now reaches
the bridge journal twice — once from the ledger, once on the RETURN line — with no length bound and
no sanitisation.

`src/relay_refusals.mjs:88`

```js
const shown = r.reason || '(no reason given)'
log(was === null ? `RELAY REFUSED ${url}: ${shown} — …` : …)
```

and `explainSendRefusals` (added in #403) does the same:

```js
byRelay.set(relay, String(r?.reason ?? '') || '(no reason given)')
```

Neither truncates. A relay returning a megabyte, or a newline, or an ANSI escape, puts it straight
into the journal on the box.

## Why it matters here specifically

The tripwire reads those journals. A refusal message carrying newlines can synthesise what look like
additional journal lines, which is the same defusing problem `render.mjs` already solves for note
bodies on the in-door — an outsider's text rendered somewhere an operator reads and acts on. The
in-door has `defuseRefs`/`defuseMarkup`; this path has nothing.

The relay set is not fully trusted input either: on the return lane it comes from the recipient's own
`kind:10050` list, so the recipient chooses which relays the bridge dials and therefore who gets to
write into this string.

## Not a regression

Pre-existing since #374. #403 did not introduce it and is not blocked on it — flagged there during
review and split out so it does not hold up that merge.

## What a fix asserts

Bound and defuse the reason at the point it becomes a log line, in both `record()` and
`explainSendRefusals`, so there is one rule rather than two. Then, per the house discipline:

- Pair the refusal assertion with the positive one — a long/hostile reason is truncated and made
  inert, **and** an ordinary reason like `pow: 28 bits needed. (12)` still arrives intact and
  readable. A guard the operator cannot read through is a guard that hides the fault.
- Assert the reason, not only that something was trimmed.
- Fixtures must resemble production: real multi-line and escape-bearing input, not `'x'.repeat(n)`
  alone. Put any invisible character in as `\uXXXX`, never literally in the source.
- Run the negative control.

Note `refusalKey()` already strips bracketed detail for the comparison, so a fix must not change what
counts as "the same refusal" — suppression behaviour is tested and should stay put.

JAFairweather:
Closes #405.

#374 put the relay's own refusal text into the journal, because a refusal that survived only as a count coming back one lower was not actionable. Nothing bounded or sanitised that text, and the relay chooses it.

## The two faults

**A newline synthesises a journal line.** The tripwire reads these journals. A reason carrying a break followed by `RELAY[buzz] ok -> ...` writes a sentence in the bridge's mouth. An ESC sequence rewrites what the operator sees more directly.

**Length.** A megabyte reason was a megabyte journal line.

## The fix

`defuseJournalText` collapses C0, DEL and C1 to a single space and caps at `REFUSAL_TEXT_MAX = 200`. Whole ranges rather than the characters anyone has abused so far — naming what may pass is what stays correct when someone finds a new one.

Wired in at the point of entry, so `record()`, `summary()` and `rows()` are safe by construction rather than by each of them remembering.

**Beyond the issue's literal ask, the relay URL is defused too.** On the return lane the relay set comes from the recipient's own `kind:10050`, so the URL is chosen by the same party as the reason. Defusing one while interpolating the other into the same line fixes half an injection.

**`refusalKey` still receives the raw reason.** Defusing before comparison would change what counts as the same refusal, and the suppression behaviour is asserted unchanged rather than assumed.

## Readability is the constraint, not an afterthought

`pow: 28 bits needed. (12)` survives byte for byte. A guard that mangles the ordinary case hides the exact fault #374 exists to reveal. Control characters collapse to a space rather than being stripped or escaped into noise, and truncation **says** it truncated and by how much — a reason 50 000 characters long is itself the news, and silently cutting it to 200 reports a hostile relay as a chatty one.

## Evidence

83 suites, exit 0. 62 assertions in `relay_refusal`, 27 of them new.

**Negative control.** With `defuseJournalText` reduced to a pass-through:

```
11 FAIL, 51 still pass

FAIL - a newline cannot synthesise a second journal line
FAIL - an ESC is removed so a refusal cannot rewrite the operator terminal
FAIL - NUL, DEL and C1 go the same way as CR and LF
FAIL - a megabyte reason does not become a megabyte journal line
FAIL - the line SAYS it truncated and by how much
FAIL - the ledger line carries no line break either
FAIL - a relay URL cannot forge a line either
FAIL - summary() defuses the relay it prints, not just the reason
FAIL - explainSendRefusals follows the SAME rule rather than a second one

ok   - an ordinary refusal survives byte for byte
ok   - so does a long-but-plausible one, unchanged and untruncated
ok   - the text is still THERE, flattened rather than censored
ok   - defusing did not change what counts as the same refusal
ok   - an ordinary reason still comes through it untouched
```

Every readability assertion survives the neutering. That is what separates "refuses the dangerous thing" from "refuses everything" — the failure mode that shipped a live outage here on 2026-08-01.

Fixtures are shaped like production, not `'x'.repeat(n)`: the forgery is a real-looking timestamped `RELAY[buzz] ok ->` line. Every invisible character is written as a `\uXXXX` escape; the test source contains no literal control character, verified by grep. The generator that wrote it builds each escape from `chr(92)` rather than typing it, after a first attempt put a real NUL in the file and `node --check` caught it.

## Review

@My Dude — two things worth attacking.

**The character class.** It is a range whose endpoints are both invisible characters. That is exactly the shape this repo has been burned by before: a class nobody can read is a class nobody can check. Worth confirming the ranges are the ones intended and that nothing printable falls inside them.

**`refusalKey` sees the raw reason while everything else sees the defused one.** Corrected after review: the sentence that stood here claimed two reasons differing only in control characters are one refusal and the second is suppressed. They are not. JS `\s` does not cover NUL or ESC, so the control character survives the squeeze and varies the key while every one of those lines renders identically — both log, and the second prints CHANGED where nothing visible changed. The decision to leave `refusalKey` on the raw string still stands, because defusing before the comparison does not fix it either. `src/relay_refusals.mjs` now records what actually happens; #422 carries the root, which is that line COUNT is unbounded whether or not any control character is involved.

### DIFF
--- a/src/relay_refusals.mjs
+++ b/src/relay_refusals.mjs
@@ -29,6 +29,42 @@
 // dropped for the comparison and KEPT for display, because `28` is news when it changes and `(12)`
 // never is.
 
+/// THE JOURNAL IS A SURFACE AN OPERATOR READS AND ACTS ON (#405), so text that arrived from outside
+/// is bounded and defused before it becomes a line — the same reason `render.mjs` defuses a note
+/// body on the in-door. Two things a relay controls can hurt here, neither needing it to be hostile
+/// on purpose:
+///
+///   - A NEWLINE synthesises what looks like another journal line. The tripwire reads these
+///     journals, so a reason carrying a line break followed by `RELAY[buzz] ok -> ...` puts a
+///     sentence in the bridge's mouth. An ESC sequence rewrites what the operator sees more
+///     directly still.
+///   - LENGTH. Nothing truncated this: a megabyte reason was a megabyte journal line.
+///
+/// THE RELAY URL IS DEFUSED TOO, not only the reason. On the return lane the relay set comes from
+/// the recipient's own `kind:10050`, so the URL is chosen by the same party as the reason — and
+/// defusing one while interpolating the other into the same line fixes half an injection.
+///
+/// READABILITY IS THE CONSTRAINT, not an afterthought. `pow: 28 bits needed. (12)` has to survive
+/// byte for byte, because the whole point of #374 was that this string reaches the operator; a
+/// guard that mangles the ordinary case hides the fault it exists to reveal. So control characters
+/// collapse to a space rather than being stripped or escaped into noise, and truncation SAYS that
+/// it truncated and by how much — a reason 40 000 characters long is itself the news, and silently
+/// cutting it to 200 reports a hostile relay as a chatty one.
+export const REFUSAL_TEXT_MAX = 200
+
+export function defuseJournalText(value, max = REFUSAL_TEXT_MAX) {
+  const raw = String(value ?? '')
+  // C0 (U+0000-U+001F: CR, LF, TAB, ESC, NUL), DEL (U+007F), and C1 (U+0080-U+009F, which some
+  // terminals still act on). Whole ranges rather than the characters anyone has abused so far —
+  // naming what may pass is what stays correct when someone finds a new one.
+  const flat = raw
+    .replace(/[\u0000-\u001F\u007F-\u009F]+/g, ' ')
+    .replace(/ {2,}/g, ' ')
+    .trim()
+  if (flat.length <= max) return flat
+  return `${flat.slice(0, max)}... (truncated, ${raw.length} chars)`
+}
+
 /// Everything outside brackets, lowercased and squeezed: the part of a refusal that is about the
 /// relay's rule rather than about this particular event.
 export function refusalKey(reason) {
@@ -85,11 +121,16 @@ export function refusalLedger({ cap = 64, log = () => {} } = {}) {
       if (r.key === key) return 'suppressed'
       const was = r.reason                 // captured BEFORE the overwrite; the line is useless without it
       r.key = key
-      r.reason = String(reason ?? '')
+      // Defused on the way IN, so every consumer of the row — this line, `summary()`, `rows()` —
+      // is safe by construction rather than by each of them remembering (#405). The KEY above is
+      // computed from the raw reason and is deliberately untouched: defusing before comparison
+      // would change what counts as the same refusal, and that behaviour is tested.
+      r.reason = defuseJournalText(reason)
       const shown = r.reason || '(no reason given)'
+      const shownRelay = defuseJournalText(url)
       log(was === null
-        ? `RELAY REFUSED ${url}: ${shown} — first time; further identical refusals are counted, not logged (#374)`
-        : `RELAY REFUSAL CHANGED ${url}: ${shown} — was "${was || '(no reason given)'}"; ${r.refused} refusal(s) from this relay so far`)
+        ? `RELAY REFUSED ${shownRelay}: ${shown} — first time; further identical refusals are counted, not logged (#374)`
+        : `RELAY REFUSAL CHANGED ${shownRelay}: ${shown} — was "${was || '(no reason given)'}"; ${r.refused} refusal(s) from this relay so far`)
       return 'logged'
     },
 
@@ -101,7 +142,7 @@ export function refusalLedger({ cap = 64, log = () => {} } = {}) {
     summary() {
       const bad = [...rows.values()].filter(r => r.refused > 0)
       if (!bad.length) return null
-      return bad.map(r => `${r.relay}: refused ${r.refused} of ${r.refused + r.accepted} — ${r.reason || '(no reason given)'}`).join(' · ')
+      return bad.map(r => `${defuseJournalText(r.relay)}: refused ${r.refused} of ${r.refused + r.accepted} — ${r.reason || '(no reason given)'}`).join(' · ')
     },
 
     size() { return rows.size },
@@ -125,8 +166,8 @@ export function explainSendRefusals(refusals) {
   for (const r of refusals) {
     const relay = String(r?.relay || '')
     if (!relay || byRelay.has(relay)) continue
-    byRelay.set(relay, String(r?.reason ?? '') || '(no reason given)')
+    byRelay.set(relay, defuseJournalText(r?.reason) || '(no reason given)')
   }
   if (!byRelay.size) return null
-  return [...byRelay].map(([relay, reason]) => `${relay}: ${reason}`).join(' · ')
+  return [...byRelay].map(([relay, reason]) => `${defuseJournalText(relay)}: ${reason}`).join(' · ')
 }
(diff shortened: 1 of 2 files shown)


JAFairweather:
## Review — approve. Point 1 confirms clean; point 2's conclusion holds but its stated reason is inverted.

All **83 suites run individually** at `3005fd8e`: 82 pass. `install_state` fails and fails identically on `origin/main` on this machine — my box, not the branch. (`npm test` chains with `&&` and that is suite 6, so a local run aborts there and the other 77 never execute.)

## 1. The class is exactly what it says — checked exhaustively, not read

I copied the class verbatim out of `src/relay_refusals.mjs` programmatically and swept U+0000–U+10FFFF with it:

```
matched count : 65
matched ranges: U+0000-U+001F, U+007F-U+009F
caught but NOT category Cc (want 0): 0
category Cc NOT caught (want 0):     0
```

The ranges are the intended ones, **nothing printable falls inside**, and nothing in the named category falls outside either — exactly the 65 Unicode Cc codepoints, no more and no less.

Worth saying, since you raised the invisible-endpoints shape: it does not apply here, and the reason is that the class is written with `\uXXXX` escapes. That is what let me lift it out of the file and sweep it mechanically. The trap this repo was burned by was a class with *literal* invisibles; this is the same rule working.

**What survives, and whether it matters.** U+2028, U+2029, U+202D/E, U+2066–9, U+200B, U+00AD, U+00A0, U+FEFF. I checked the consequence rather than assuming one: `tools/tripwire.mjs` splits on `'\n'` only (lines 149 and 235), so **none of these synthesises a journal line for the alarm.** They are terminal-rendering only. Of those the one with teeth is **U+202E RIGHT-TO-LEFT OVERRIDE**, which reorders what the operator reads with no C0/C1 character anywhere in the string — squarely inside "text an operator reads and acts on". Narrow, real, and an issue rather than a reason to hold this.

## 2. The judgement call is fine. The sentence justifying it is false.

> refusalKey sees the raw reason … two reasons differing only in control characters are one refusal and the second is suppressed

They are not one refusal, and the second is not suppressed. `refusalKey` squeezes `\s+`, and JS `\s` does **not** include NUL, ESC, or most of C0 — so a non-whitespace control character survives into the key and *varies* it, while `defuseJournalText` maps it to a space so every line renders identically. Three reasons, differing only by one NUL and one ESC:

```
refusalKey, computed from RAW:
  A: "pow: 28 bits needed"
  B: "pow: 28<NUL> bits needed"
  C: "pow:<ESC> 28 bits needed"

all three display identically?  true
all three share one refusal key? false
```

Six sends of those three reasons produce six lines, `logged` every time, five of them:

```
RELAY REFUSAL CHANGED wss://nos.lol: pow: 28 bits needed. — was "pow: 28 bits needed."; …
```

A *changed* line where nothing changed — and after this PR the operator cannot see why, because the only difference has been normalised away. The behaviour is the exact inverse of what the new comment asserts, and that comment is what the next reader will trust.

## The finding underneath it: line length is bounded, line COUNT is not

I proposed collapsing C0/DEL/C1 inside `refusalKey` and tested it before recommending it. **It does not work**, and it is worth saying why rather than shipping it: it fixes the six-send case (1 line, 5 suppressed), but a NUL *walked along* the string still yields 401 lines from 500 sends, because inserting a space mid-word is a genuinely different string after the squeeze.

Then the test that reframes it — no control characters at all, just a relay varying its own wording:

```
500 sends, visible text varied, zero control characters -> 500 journal lines
ledger rows (bounded by relay count, as designed): 1
```

So the control-character angle is a special case, not the root. **`defuseJournalText` bounds the length of one line; nothing bounds how many lines a relay can cause.** 500 sends becomes 500 lines of up to ~260 characters each, on a box whose journals the tripwire reads. The module header says the anti-flood rule is the entire point — "log the first occurrence, suppress repeats, and log again when the reason is genuinely DIFFERENT" — and *genuinely different* is the relay's choice.

This is on `main` today (`0efe7cd1`, 500 both ways), so **#420 does not introduce it** and it is not a merge blocker. Two things follow:

1. Leaving `refusalKey` on the raw string is the right call — changing it closes nothing. Keep the decision, drop the justification, because as written it claims a suppression that does not happen.
2. The real bound is a per-relay cap on logged lines in a window, not a better key. Different mechanism, own issue — I will file it unless you would rather.

## Your negative control, and four of mine

Your 11-fail / 51-pass split is the right shape and I did not re-run it. Four different mutations instead, each anchor-asserted:

| mutation | failures | named assertion |
|---|---|---|
| narrow the class to only LF and CR — the two anyone has abused | 3 | `an ESC is removed so a refusal cannot rewrite the operator terminal` |
| truncate silently, dropping the "and by how much" suffix | 1 | `…the line SAYS it truncated and by how much` |
| raise `REFUSAL_TEXT_MAX` to 100000 | 1 | same assertion |
| defuse the reason but not the relay URL | 1 | `a relay URL cannot forge a line either` |

All four caught, restored clean. The half-injection case having its own assertion is the one I most expected to be missing, and it is there.

**Not proven**, as always: a live refusal from a real relay on the production bridge.

Reviewed at `3005fd8efee0fa228626c52db6d10e3abfd2b596`.

JAFairweather:
Point 2 is right and I have corrected it at the source. `1b13114a`.

I had the mechanism backwards: `\s` does not cover NUL or ESC, so the control character survives the squeeze and **varies** the key while `defuseJournalText` renders every one of those lines identically. Two refusals that both log, the second printing CHANGED where nothing visible changed. The comment in `src/relay_refusals.mjs` now says that, including the `401 lines from 500 sends` result that rules out defusing before the comparison — the decision stands, the justification was wrong, and it was the sentence the next reader would have trusted. The PR body is corrected too.

Both findings are filed, so neither rides on this PR:

- **#422** — line COUNT is unbounded whether or not a control character is involved. Your 500-sends-500-lines measurement is the issue's evidence, marked as pre-existing on `0efe7cd1`.
- **#423** — the bidi and zero-width survivors, with your sweep result and the tripwire finding that they cannot synthesise a line for the alarm. U+202E named as the one with teeth.

Your sweep is better evidence than my assertion and I have said so in #423 rather than restating it here. The four mutations are the part I would not have run — the half-injection case having its own assertion was the one I most expected to be missing too.

One correction to the environment note: `install_state` fails on a checkout with no `node_modules` and no local `config.json` — the preflight step reports `dependencies not installed`, and the suite reads that as a blocking problem. `npm ci` plus a local `config.json` in the worktree took me from that failure to 83 suites exit 0. Worth knowing before it gets read as a branch fault twice.

PR #421, opened 2026-08-13
fix: refuse a confusable mention at admission, not at arbitration (#416)

### LINKED ISSUE #416: A confusable mention can be admitted as a route: interception closes at upsert, not at arbitration
Split from the #414 re-read.

#414 makes `@Mesnil` carry to both `Mesnil` and `Meſnil` (U+017F). That fixes the mute and the order dependence — the legitimate agent is no longer displaced. It does not close interception: an agent that chooses `Meſnil` as its `display_name` still receives every message addressed to `Mesnil`. What changed is that it no longer receives them exclusively.

That is the right call at the arbitration layer. Arbitration cannot tell an impostor from a legitimate namesake, and refusing a carry on suspicion would mute real agents. The check belongs at **admission**: refuse, or at minimum warn on, a task-route upsert whose mention is confusable with an existing route in the same channel bound to a **different** participant.

The reachable chain is short, which is why this is worth doing rather than noting. The console instructs the operator to paste the display name exactly as it appears in channel, so the operator does not need to be fooled by the glyph — they only need to copy it.

## The shortcut does not work — checked, not assumed

NFKC-folding `taskRouteMentionKey` collapses the pair in the row dedup inside `taskRouteMentionArbitrate`, so `arb()` reports one route. **Delivery is unchanged.** `scanReturnLane` tests `namedHere.has(taskRouteMentionKey(r.mention))`, so both routes resolve to the surviving key and both still carry:

```
NFKC key applied to both copies
@Mesnil please look at this.  -> Mesnil, Meſnil
```

So a key change alters what the arbitration function reports without altering who receives the message — worse than no change, because it makes the test output look like a delivery fix. This needs an explicit refusal at upsert.

## Open questions for whoever takes it

- Which confusability relation. NFKC is the obvious candidate; the Unicode confusables set (UTS #39) is wider and catches Cyrillic/Latin homoglyphs that NFKC does not.
- Refuse or warn. Refusing is safer; warning avoids blocking a legitimate namesake in a community that has two people with similar names. The console error line is the product either way — see `taskRouteMentionProblem`, where the reason is the thing the operator acts on.
- Scope. Same channel, different participant, is the minimum. Whether it should also cover `return_lane` entries is a separate question.

JAFairweather:
Closes #416.

#414 made `@Mesnil` carry to a long-s twin as well. That fixed the mute and the order dependence, and left interception open: the twin still receives everything addressed to the real name, just no longer exclusively.

Arbitration cannot close it. It cannot tell an impostor from a legitimate namesake, and refusing a carry on suspicion mutes real agents. So the refusal is at **upsert**, in `applyTaskRouteCommand`, before a lookalike becomes a route at all.

## Two guards, two layers

**Script mixing, in the grammar.** A mention that mixes Latin with Cyrillic or Greek letters is refused. No normalisation catches that pair, because they are genuinely different letters, and nobody types it by accident. Only those three scripts, and only in combination with each other — a name in Japanese, Arabic or Hebrew is a name, and refusing it would be this guard mistaking "foreign" for "hostile". That direction is asserted, not assumed: wholly-Cyrillic, wholly-Greek and Katakana names are in the ACCEPT table.

**A confusability skeleton, at admission.** `taskRouteMentionConflict` refuses an upsert whose NFKD-strip-marks-NFKC skeleton matches an existing route in the same channel bound to a **different** participant. That folds the long s, the fi ligature, fullwidth text and stripped accents.

## The two judgement calls

**The skeleton is deliberately not `taskRouteMentionKey`.** #416 measured what happens if you widen the comparison key instead: the row dedup inside `taskRouteMentionArbitrate` collapses the pair, `arb()` reports one route — and `scanReturnLane` still carries to both, because it tests membership on the same folded key. Delivery unchanged, test output reading like a delivery fix. The key answers "is this the same route" and must stay exactly as tight as the matcher's `i` flag. The skeleton answers "would the operator see these as the same name", and nothing routes on it.

**An identical mention for a second agent is still allowed.** The operator typed the same name twice; that is a fan-out they can see. A lookalike is the thing they cannot see, and that is the line here. Refusing identical mentions too would have broken a config nobody complained about.

## What this does not close

A name written **wholly** in another script that happens to look Latin. No normalisation folds a Cyrillic `s` onto a Latin `s`, and the script rule does not fire because nothing is mixed. Closing it needs the UTS #39 confusables table. That gap is asserted in the suite rather than described, so it cannot quietly be assumed shut — the pair of assertions is "the skeleton does NOT fold this" plus "the grammar is what refuses the mixed form".

Manual `return_lane` entries are not consulted. They are unmanaged config, edited by hand, and #416 leaves them as a separate question.

The console form is not changed. It signs a command the bridge validates, so a pre-flight check there would be a convenience, not the enforcement point; the bridge's reason is what reaches the operator either way.

## Evidence

83 suites, exit 0. `task_route_mention` 170 assertions (33 new), `control_state` 8 new.

**Negative control.** With the conflict check stubbed to `null` and the script rule short-circuited to `false`:

```
task_route_mention   151 ok, 19 FAIL      (was 170 ok, 0 FAIL)
control_state          5 FAIL
```

Every failure is a refusal assertion, the console-copy join, or the config-cleanup assertion that cascades from routes no longer being refused. **Not one acceptance assertion moved** — the ACCEPT table, the boundary table and the end-to-end carry all stayed green under the neutered guard. That is what separates "refuses the dangerous thing" from "refuses everything", which is the shape that shipped a live outage here on 2026-08-01.

Every confusable fixture is written as a `\uXXXX` escape, and for a reason that is not the usual one: these characters are perfectly visible, and that is the problem. The whole attack is that the two spellings are one picture, so a literal would be a fixture no reviewer could check by reading it. Verified by grep that no fixture line carries a non-ASCII literal.

## Review

@My Dude — two places to push.

**The script rule is in the grammar, so it applies to every mention, not only to one that collides with an existing route.** A legitimate name that mixes Latin and Greek — a transliteration, a stylised handle — is now refused outright. I think that is right, because the cost of a false refusal is an error message the operator reads and works around, while the cost of a false accept is silent misdelivery. It is still a widening of what the form rejects, and it is worth a second opinion.

**The identical-mention carve-out.** Two agents may still share an exact name in one channel, and both will carry. I left it deliberately, argued above. If you read that as the same defect wearing a different hat, say so — it is a one-line change to close, and a config change to un-break afterwards.

### DIFF
--- a/console/task-route-mention.mjs
+++ b/console/task-route-mention.mjs
@@ -34,18 +34,18 @@ const STARTS_WELL = /^[\p{L}\p{N}]/u
 
 const codepoint = ch => `U+${ch.codePointAt(0).toString(16).toUpperCase().padStart(4, '0')}`
 
-// SCRIPT MIXING (#416). Latin, Cyrillic and Greek share glyphs: `а` U+0430 and `a` U+0061 are one
-// shape in every font an operator will ever read the console in. A name written half in one and
-// half in another is not a name, it is a disguise — nobody types `Меsnil` by accident, and no
-// normalisation catches it, because the two letters are genuinely different letters.
+// NO SCRIPT-MIXING RULE HERE, and that is a decision, not an omission (#416, #426).
 //
-// Refused here rather than at the conflict check below because it does not need a second route to
-// compare against: a mention that lies about which letters it contains is unusable on its own.
-// Only these three are named, and only in combination with each other — a name in Japanese, Arabic
-// or Hebrew is a name, and refusing it would be this guard mistaking "foreign" for "hostile".
-const HAS_LATIN = /\p{Script=Latin}/u
-const HAS_CYRILLIC = /\p{Script=Cyrillic}/u
-const HAS_GREEK = /\p{Script=Greek}/u
+// One was written and removed under review. Refusing a mention that mixes Latin with Cyrillic or
+// Greek sounds like the obvious hardening and prevents nothing: cross-script never folds in the
+// matcher, so `@Max` in a body cannot reach a route spelled `Мах` — the only spelling that DOES
+// intercept is one the skeleton below already refuses. What the rule caught instead was
+// `Nikos Παπάς`, `Σigma bot` and `Дenis`, and it caught them in the function that re-validates
+// deployed config, so a name that routed yesterday would have stopped routing on merge.
+//
+// It cannot be moved to the conflict check either: that compares skeletons, and no normalisation
+// folds Cyrillic onto Latin, so the rule would never fire there. The remaining case — a human
+// pasting `@Мах` at a route named `Max` — is a compose-time hazard, not interception, and is #426.
 
 // null when the value is a usable mention, otherwise the reason — and the reason is the product.
 // `!ok` cannot tell a correct refusal from a correct refusal with a misleading message, and this
@@ -64,9 +64,6 @@ export function taskRouteMentionProblem(value) {
   if (!STARTS_WELL.test(raw)) return 'mention must start with a letter or number'
   if (raw.includes('  ')) return 'mention must separate words with a single space'
   if (RESERVED.has(raw.toLowerCase())) return `@${raw} is a broadcast at-word, not an agent`
-  if (HAS_LATIN.test(raw) && (HAS_CYRILLIC.test(raw) || HAS_GREEK.test(raw))) {
-    return `@${raw} mixes Latin with Cyrillic or Greek letters — those scripts share glyphs, so the name is not the one it appears to be`
-  }
   return null
 }
 
@@ -93,9 +90,13 @@ export const taskRouteMentionKey = value => String(value == null ? '' : value).t
 //
 // NFKD, drop the combining marks, back to NFKC: that folds `Meſnil` (U+017F long s), `ﬁnn` (the fi
 // ligature), fullwidth `ＭＣ` and `Mésnil` onto their plain forms. It does NOT fold a Cyrillic `ѕ`
-// onto a Latin `s` — no normalisation does, they are different letters. That case is refused one
-// layer up by the script-mixing rule, and the residual gap is a name written WHOLLY in another
-// script that happens to look Latin. Closing that needs the UTS #39 confusables table.
+// onto a Latin `s` — no normalisation does, they are different letters.
+//
+// That is complete against interception, and the completeness was measured rather than argued: a
+// sweep of U+0020–U+10FFFF found ZERO characters that the matcher folds onto an ASCII letter but
+// this skeleton does not fold the same way (#421 review). A lookalike the matcher can be fooled by
+// is a lookalike this refuses. What is left over is a name written in another script that an
+// operator might mis-type — a compose-time hazard, not a delivery one, tracked as #426.
 export function taskRouteMentionSkeleton(value) {
   return String(value == null ? '' : value)
     .replace(/^@/, '')
@@ -115,10 +116,13 @@ export function taskRouteMentionSkeleton(value) {
 // the interception in #416:
 //
 //   - SAME participant, lookalike spelling — that is the agent renaming itself. Not interception.
-//   - IDENTICAL mention, different participant — deliberately still allowed. The operator typed the
-//     same name twice; that is a fan-out they can see, and refusing it would break a config nobody
-//     complained about. A LOOKALIKE is the thing an operator cannot see, which is the whole reason
-//     the check exists. Refusing only what the eye cannot separate is the line here.
+//   - IDENTICAL mention, different participant — deliberately still allowed, and it IS interception:
+//     a second `Max` takes a share of every `@Max`. It stays allowed because of who authored the
+//     name, not because anyone can see it. Two identical rows in a rendered route list are no more
+//     distinguishable than `Max` and `Мах`. The difference is that upsert is approver-signed and
+//     participant-admitted, so a duplicate requires an approver to type a name they already know —
+//     which is exactly the knowledge a lookalike denies them. Authorship at typing time is the line,
+//     not visibility at reading time (#421 review).
 //
 // Manual `return_lane` entries are not consulted: they are unmanaged config, edited by hand and by
 // definition not going through this path (#416 leaves that as a separate question).
--- a/console/task-route-mention.mjs
+++ b/console/task-route-mention.mjs
@@ -34,6 +34,19 @@ const STARTS_WELL = /^[\p{L}\p{N}]/u
 
 const codepoint = ch => `U+${ch.codePointAt(0).toString(16).toUpperCase().padStart(4, '0')}`
 
+// SCRIPT MIXING (#416). Latin, Cyrillic and Greek share glyphs: `а` U+0430 and `a` U+0061 are one
+// shape in every font an operator will ever read the console in. A name written half in one and
+// half in another is not a name, it is a disguise — nobody types `Меsnil` by accident, and no
+// normalisation catches it, because the two letters are genuinely different letters.
+//
+// Refused here rather than at the conflict check below because it does not need a second route to
+// compare against: a mention that lies about which letters it contains is unusable on its own.
+// Only these three are named, and only in combination with each other — a name in Japanese, Arabic
+// or Hebrew is a name, and refusing it would be this guard mistaking "foreign" for "hostile".
+const HAS_LATIN = /\p{Script=Latin}/u
+const HAS_CYRILLIC = /\p{Script=Cyrillic}/u
+const HAS_GREEK = /\p{Script=Greek}/u
+
 // null when the value is a usable mention, otherwise the reason — and the reason is the product.
 // `!ok` cannot tell a correct refusal from a correct refusal with a misleading message, and this
 // message is the entire content of the console's error line. An invisible character in particular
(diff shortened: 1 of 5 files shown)


JAFairweather:
## Review — the skeleton guard is right and provably complete. Move the script rule before merging.

All **83 suites run individually** at `ec8416bb`: 82 pass. `install_state` fails identically on `origin/main` on this machine — my box, not the branch.

## The skeleton guard closes #416 exhaustively, which is more than the PR claims

I swept every codepoint U+0020–U+10FFFF for a character that the **matcher** folds onto an ASCII letter but the **skeleton** does not fold the same way. Each one would be an interception this guard misses.

```
characters the matcher folds onto an ASCII letter but the skeleton does not: 0
```

There is no such character. Restricted to mentions the grammar admits, `taskRouteMentionConflict` catches every lookalike the matcher can actually be fooled by. Control, through the real predicate:

```
admission of "Meſnil" against a routed "Mesnil" : REFUSED
had it been admitted, "@Mesnil" would reach     : real Mesnil + IMPOSTOR
```

That is #416, closed. The two-questions split in the comment at line 100 is the right call and the sweep is the evidence for it.

## 1. The script rule refuses every mixed-script mention, and prevents zero interceptions

Cross-script never folds in the matcher — that is what the sweep says, and directly:

```
body "@Max"     route "Мах"     -> no match
body "@Mesnil"  route "Меsnil"  -> no match
body "@Mesnil"  route "Meſnil"  -> INTERCEPTS   (and is refused at admission)
```

The only pair that intercepts is the one the skeleton already catches. The script rule buys nothing at delivery. What it costs:

```
"Nikos Παπάς"   REFUSED        "René Descartes"  accepted
"Σigma bot"     REFUSED        "トヨタ bot"        accepted
"pi π calculus" REFUSED        "אבג bot"          accepted
"Дenis"         REFUSED        "مرحبا bot"        accepted
```

Greek and Cyrillic personal names are names. The comment defending the Japanese/Arabic/Hebrew exemption says refusing them would be the guard "mistaking foreign for hostile" — that reasoning does not stop at three scripts, and `Nikos Παπάς` is the case it does not survive.

**And unlike the conflict check, the script rule is retroactive.** `bridge.mjs:328` re-validates every configured route through `taskRouteMentionProblem` at boot, and merging deploys and restarts the lane:

```
a route already in public.task_routes, at the next restart:
  @MC Claude      still routes
  @Nikos Παπάς    DROPPED by this PR
  @Σigma bot      DROPPED by this PR
  @Дenis          DROPPED by this PR
  @My Dude        still routes
  @Mésnil         still routes
```

Named at boot with its reason, so not silent — line 326 earns itself here. But a name that routed yesterday stops routing on merge, with no migration step. The conflict check cannot do this, because it only runs on upsert. Putting a name-refusing rule in the function that re-validates deployed config is what makes it retroactive.

## 2. Identical names — the decision is right, the justification is not

It is real interception, not a benign fan-out. Checked, not assumed:

```
admission of a 2nd "Max" for a different participant : ADMITTED
"@Max ship it" reaches                                : real Max + IMPOSTOR
```

Keep allowing it anyway. The reason that holds is the one not in the comment: upsert is approver-signed (`2902`) and participant-admitted (`2915`), so a second `Max` requires an approver to type a name they authored. That is exactly the knowledge they cannot have about a lookalike, and it is a sharper line than the one drawn.

Drop "a fan-out they can see." In a rendered route list two identical `Max` rows are exactly as indistinguishable as `Max` and `Мах` — more so, if anything. What separates the cases is authorship at typing time, not visibility at reading time, and the next reader will act on the version that says the operator can see it.

## The residual is narrower than the comment says — UTS #39 is not needed for it

`Мах` (U+041C U+0430 U+0445) clears the grammar and the skeleton and is admitted alongside `Max`. But it cannot take a message addressed to `Max`:

```
"@Max ship it"           -> real Max only
"@Мах ship it" (pasted)  -> IMPOSTOR only
```

It only catches a human who pastes the Cyrillic form. That is a compose-time hazard, the same class the script rule addresses — not the interception #416 was filed for. Worth an issue, not the confusables table.

## Recommendation

Move the script rule out of `taskRouteMentionProblem` and into `taskRouteMentionConflict`: refuse a mixed-script name only when it collides with an existing route in that channel. That removes the false positives, makes the rule non-retroactive, and treats `Дenis` and `Мах` the same way — each admitted alone, each refused against a lookalike. The skeleton guard needs no change.

## Mutations

Five, each anchor-asserted, both copies mutated, both files restored byte-identical. All five caught:

| mutation | task_route_mention | control_state |
|---|---|---|
| script rule never fires | 14 FAIL | 2 FAIL |
| conflict check always returns null | 4 FAIL | 4 FAIL |
| skeleton reduced to a bare lowercase | 6 FAIL | 4 FAIL |
| conflict ignores the participant scope | 2 FAIL | 1 FAIL |
| conflict ignores the channel scope | 2 FAIL | 1 FAIL |

The last two are the ones I expected to be missing, because they are the false-positive direction — an agent renaming itself, and a lookalike in a different channel. Both caught. The parity join is the strongest part of the suite: byte-identity, a size floor, a real import, and same-verdict-**and**-same-reason across the corpus. (`the copy exports the same surface` still lists only the four original exports; byte-identity covers the three new ones, so that is cosmetic.)

**Not proven:** whether any deployed route mixes scripts. I cannot read production config — before merging, that is one grep of `public.task_routes`, and it decides whether the retroactive drop is theoretical or live. No live carry either, as always.

Note for sequencing: #414 is still open and edits this same module, and the comment at `2916` cites its behaviour. Whichever lands second wants a re-read, not just a rebase.

Reviewed at `ec8416bbfb14a9044352b066aff9944c6ab7c024`, against `origin/main` `0efe7cd1`.

JAFairweather:
Script rule removed, both comments corrected, residual filed as #426. Pushed as `2db206c4`.

## The script rule

You are right and the evidence is yours: it prevents zero interceptions and refuses real names, in
the function that re-validates deployed config. That last part is the one I did not see — I put a
name-refusing rule in `taskRouteMentionProblem` without noticing it runs at boot against rows that
already route.

**One correction to the recommendation, and it changes the fix.** It cannot move into
`taskRouteMentionConflict`. That check compares skeletons, and the skeleton does not fold Cyrillic
onto Latin:

```
"Max"    vs "Мах"    same skeleton? false
"Denis"  vs "Дenis"  same skeleton? false
"Mesnil" vs "Meſnil" same skeleton? true
```

A mixed-script name therefore never collides there, so the moved rule would never fire — a guard
that reads like protection and is dead. Making it fire needs a Cyrillic-to-Latin homoglyph map,
which is the confusables table you argued is not needed here, and I agree it is not.

So: deleted, not moved. `Дenis` and `Мах` are treated the same way — both admitted, neither able to
intercept — which is the outcome you asked for, reached by removing the rule rather than relocating
it.

## What replaces the four refusal fixtures

They now assert the property that makes admitting them safe, both directions:

```
a cross-script lookalike is a legal NAME: "Meſnil" / "Аpple" / "Dennisа" / "Μy Dude"
  …the skeleton does NOT fold it
  …it is admitted beside its twin rather than refused
  …a body addressed to the Latin name does NOT reach it
  …while the Latin name itself still carries
```

That last pair is there because a matcher reaching *neither* would satisfy a one-sided assertion.

## The identical-name comment

Rewritten to your version. It says it is real interception, drops "a fan-out they can see", and
gives the reason that holds: upsert is approver-signed and participant-admitted, so a duplicate
requires an approver to type a name they already know — the knowledge a lookalike denies them.
Authorship at typing time, not visibility at reading time.

## Evidence

83 suites, exit 0. Negative control on what remains — `taskRouteMentionConflict` forced to return
`null`, both copies, restored byte-identical:

```
task_route_mention   3 FAIL   the long-s refusal, and both halves of its reason
control_state        4 FAIL   the fullwidth refusal, its reason, "nothing was written",
                              and the config-returns-clean assertion
```

**Stated rather than counted:** the four new TWINS assertions stay green under that mutation,
because three of the five per-twin checks assert an admitted route and an empty conflict, which is
what a dead guard produces. They are worth having — they are what would catch a guard that refuses
everything — but they prove nothing about this mutation.

## Your open question

> before merging, that is one grep of `public.task_routes`

Moot now. Nothing is refused retroactively, so no deployed route can stop routing on this merge.

@Dennis — #426 has the three options for the compose-time residual and a note that they all act on
the approver while the hazard belongs to the person composing the message. Your call whether that is
worth building at all.

JAFairweather:
**Re-read at `2db206c4` (My Dude). Approve.** The script rule is gone, the residual is filed as #426, and the removal's justification holds when exercised rather than read.

Removing a guard needs the stronger proof, so I checked the load-bearing claim — that cross-script never folds, so nothing was opened by dropping it:

```
key(Max)      = "max"      skel(Max)     = "max"
key(Cyr Мах)  = "мах"      skel(Cyr Мах) = "мах"
keys equal?    false       skels equal?   false
```

Neither the matcher key nor the skeleton folds Cyrillic onto Latin, so `@Max` in a body cannot reach a route spelled `Мах`. No interception path exists for the rule to have been closing. And the three names it was actually refusing now pass:

```
prob(Nikos Παπάς) = null
prob(Σigma bot)   = null
prob(Дenis)       = null
```

That is the defect I raised: a rule in the function that re-validates deployed config, refusing names that route today. Removing it is the right call over relocating it, and the comment explaining *why there is no rule here* is worth more than the rule was — it stops the next reader re-adding it.

Also confirmed the two copies did not drift: `src/task_route_mention.mjs` and `console/task-route-mention.mjs` differ only in their header comments, byte-identical from the first `export` down, which is what `tests/task_route_mention.mjs` asserts. #426 is open and states the compose-time hazard accurately.

Nothing further from me.

PR #425, opened 2026-08-13
fix: report a route that lost its at-word to the WORD, not only to another route (#415)

### LINKED ISSUE #415: A route that loses an at-word to the boundary is never logged — only losses to another route are
Found in review of #414.

`RETURN skip[longer-name]` fires only when a route lost an at-word to another **route**. A route that loses to the **boundary** produces nothing.

`@MC Claudette` and `@MC Claudeé` both carry to `MC` with `suppressed: []`. The owner of `MC Claude`, watching messages that visibly begin `@MC Claude…` reach a different agent, gets no explanation anywhere. That is the same "presents as the return lane being flaky" failure the skip log exists to prevent, one case short of covered.

The carry is correct — #414 argues at length that `@MC Claudeé is …` and `@MC has this one …` are the same shape and no rule can withhold one without withholding the other. This is about the silence, not the routing.

Shape of a fix: at each at-word, a route whose mention is a prefix of the span but whose boundary lookahead failed is a *near miss*, and near misses are the thing worth logging. That is more machinery than the tie fix needed, which is why it is here and not in #414.

Worth checking while doing it: whether the near-miss log wants a rate limit. `rlDropOnce` is per message id, so a channel where one agent is habitually mistyped would log once per message rather than once per pattern.

JAFairweather:
Closes #415. **Stacked on #414** (`fix/409-longest-mention-wins`) — it extends the arbitration that PR introduces, so merge that first or this diff will not read correctly.

`RETURN skip[longer-name]` fires only when a route lost an at-word to another **route**. `@MC Claudette` carries to `MC` and not to `MC Claude`: no route took it, so nothing was contested and nothing was logged. The owner of `MC Claude` watched messages that visibly begin with their name reach somebody else, with no line anywhere saying why — the same "presents as the return lane being flaky" failure the suppression log exists to prevent, one case short of covered.

## The fix

A route whose mention is a prefix of the at-word but whose boundary lookahead failed is a **near miss**, reported under its own heading with the at-word as written:

```
RETURN skip[longer-word]: 9f2c… @MC Claude does not take the at-word at 0 — @MC Claudette continues past it
```

The at-word is the whole explanation. Seeing `@MC Claudette` next to the route `MC Claude` ends the question, which `skip[longer-name]` could never do because there was no longer name to name.

**Two headings, not one.** Losing to another route is a naming collision the operator resolves. Losing to the word is usually somebody typing a name that is not anyone's. Collapsing them into one line would report a name clash that does not exist.

**Precedence: carried > suppressed > near miss.** A route that carried anywhere in the body is not reported at all; a route already named in `suppressed` is not reported again under a second heading. One message yields at most one line per route.

## Bounding, because this is outside-controlled text on its way to a journal

Same lesson as #405. The reported word is capped at 64 characters with a visible ellipsis — a 4 000-character word would otherwise be a 4 000-character journal line chosen by whoever wrote the message. It cannot carry a control character by construction rather than by sanitising afterwards: the prefix is the configured mention, which the grammar has already refused control characters in, and the continuation is drawn from `MENTION_ALPHABET`, which is letters, numbers and `_ . ' -`.

Not rate limited beyond that, deliberately. Both maps are keyed by mention, and the caller's `rlDropOnce` is per message id, so the exposure is the one `suppressed` already has, not a new one. #415 asked the question; that is the answer.

## Evidence

83 suites, exit 0. `task_route_mention` 181 assertions, 13 new.

**Negative control**, with `nearMissed` forced to return empty:

```
174 ok, 7 FAIL   (was 181 ok, 0 FAIL)

FAIL — a route whose name is a prefix of a longer WORD is reported
FAIL —   …and the at-word as written is IN the line
FAIL —   …a possessive counts too — a different name, not this one
FAIL — a body with one of each reports one of each, under its own heading
FAIL — a monstrous word does not become a monstrous log line
FAIL —   …and it SAYS it was cut
FAIL — the shared section is byte-identical      (src neutered, console copy not)
```

**Stated plainly rather than counted as proof:** the precedence assertions and the "not a near miss for anyone" assertions all stay green under the neutered guard, because they assert an empty set and an empty set is what a dead guard produces. They are still worth having — they are what would catch a guard that fires on every at-word — but they prove nothing about this mutation and I am not claiming they do. The six that moved are the ones carrying the weight.

One fixture was changed after the first negative-control run: `long.near[0].endsWith(…)` threw on `undefined` under the neutered guard and took every later assertion down with it, so the run reported 161 ok and looked smaller rather than failing louder. It now reads through `String(… ?? '')` and fails as an assertion. A suite that dies mid-way tells you less than one that fails.

## Review

@Dennis — one thing to attack, and it is a judgement call about noise rather than correctness.

**Every at-word that merely starts with a route's name now produces a line.** `@MCX`, `@Dennisson`, `@MC Claude's` — all near misses. In a channel where an agent's name is a common word or a common prefix, that is a line per message, forever, and the operator learns to skim past exactly the heading that is supposed to tell them why a delivery went missing. I think it is right at this volume, and the per-message keying keeps it to one line per route, but the failure mode of a too-chatty log is the same as a silent one and it is worth a second opinion on whether this wants a threshold.

### DIFF
--- a/console/task-route-mention.mjs
+++ b/console/task-route-mention.mjs
@@ -201,11 +201,55 @@ export function taskRouteMentioned(body, mention) {
 // about a single at-word, never about a route, which is why `carried` is resolved before
 // `suppressed` is answered.
 //
-// Returns { carried, suppressed }: `carried` maps mention key -> mention as written; `suppressed`
-// maps mention key -> { mention, by, at }, naming the longer mention that took the at-word and
-// where. The caller logs from `suppressed` — a route that silently stops receiving a carry it used
-// to receive presents as the return lane being flaky, which is the failure this repo keeps paying
-// for.
+// The at-word as written, for a near-miss line: the `@`, the route's own name, and the run of
+// mention characters that CONTINUED past it. The continuation stops at the first character outside
+// the alphabet, so it takes `@MC Claudette` out of `@MC Claudette please look` and not the rest of
+// the sentence. A space deliberately ends it: a space may be interior to a name, but a name is what
+// we are trying to identify here and swallowing them would print the whole message.
+//
+// SAFE FOR A JOURNAL BY CONSTRUCTION, not by sanitising afterwards. The prefix is the configured
+// mention, which the grammar has already refused control characters in, and the continuation is
+// drawn from `MENTION_ALPHABET` — letters, numbers and `_ . ' -` — so no newline or escape can
+// enter this string from the channel body. The cap is what bounds it: a 4 000-character word is
+// otherwise a 4 000-character journal line chosen by whoever wrote the message (cf. #405).
+const NEAR_MISS_WORD_MAX = 64
+function atWordAt(text, at, mentionLength) {
+  const tail = new RegExp(`[${MENTION_ALPHABET}]+`, 'uy')
+  tail.lastIndex = at + 1 + mentionLength
+  const run = tail.exec(text)
+  const word = `@${text.slice(at + 1, at + 1 + mentionLength)}${run ? run[0] : ''}`
+  return word.length <= NEAR_MISS_WORD_MAX ? word : `${word.slice(0, NEAR_MISS_WORD_MAX)}…`
+}
+
+// LOSING TO THE BOUNDARY IS ALSO A LOSS (#415). `suppressed` answers "which other ROUTE took this
+// at-word", and that leaves the more common case unreported: `@MC Claudette` carries to `MC` and
+// not to `MC Claude`, because the boundary correctly says the at-word did not end after `Claude`.
+// No route took it from `MC Claude` — the word itself did — so nothing was contested and nothing
+// was logged. The owner of `MC Claude` watched messages that visibly begin with their name reach
+// somebody else, with no line anywhere saying why.
+//
+// That is the same "presents as the return lane being flaky" failure the suppression log exists to
+// prevent, one case short of covered. So a route whose mention is a PREFIX of the at-word but whose
+// boundary lookahead failed is reported as a NEAR MISS, with the at-word as written, because the
+// at-word is the whole explanation: seeing `@MC Claudette` next to the route `MC Claude` ends the
+// question immediately.
+//
+// PRECEDENCE, and it matters more than it looks. carried > suppressed > near miss. A route that
+// carried anywhere in the body is not reported at all; a route already named in `suppressed` is not
+// reported twice under a second heading. Two lines about one route in one message is how a log
+// starts being skimmed instead of read.
+//
+// Returns { carried, suppressed, nearMissed }: `carried` maps mention key -> mention as written;
+// `suppressed` maps key -> { mention, by, at }, naming the longer mention that took the at-word and
+// where; `nearMissed` maps key -> { mention, word, at }, naming the at-word that continued past the
+// route's name. The caller logs from the latter two — a route that silently stops receiving a carry
+// it used to receive presents as the return lane being flaky, which is the failure this repo keeps
+// paying for.
+//
+// NOT RATE LIMITED HERE, deliberately. Both maps are keyed by mention, so one message yields at most
+// one line per route no matter how many at-words it holds. Across messages the caller's `rlDropOnce`
+// is per message id, so a channel where an agent's name is a common prefix logs once per message —
+// the same exposure `suppressed` already has, not a new one.
 export function taskRouteMentionArbitrate(body, mentions) {
   const text = String(body == null ? '' : body)
   const rows = []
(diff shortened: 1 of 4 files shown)


JAFairweather:
**On the threshold question — no, and I'd resist adding one. The dedup key is what wants fixing, not the frequency.**

Read the diff against #414's tip, ran `tests/task_route_mention.mjs` clean (181/181, matches the PR's count), and traced `log()` to `src/log.mjs` — it's `console.log` to the box's journal, not a Buzz post. That matters for the noise question: nobody in a channel sees `RETURN skip[longer-word]` lines, only whoever greps the journal or runs the tripwire. So "the operator learns to skim past it" is a real cost (signal density in a log a human reads on-demand), but it's not the same severity as spamming a live chat — worth naming because it changes what kind of fix is proportionate.

**A frequency/count threshold is the wrong lever here, because it would suppress exactly the case this PR exists to stop suppressing.** The bug being fixed is: a route's owner watches messages that visibly start with their name land elsewhere, silently, forever. If the fix is "log the first N occurrences, then go quiet," a persistently-mistyped or persistently-common-prefix name reproduces the *identical* failure mode one layer down — informative for a day, then silent again, and the owner has no way to tell "fixed" from "rate-limited" from the log alone.

**What actually drives the volume you're worried about is repetition of the same confusion, not the existence of a near miss.** `@MCX` near-missing on `MC Claude` once is signal. The same route near-missing on the same literal at-word 40 times because one chatty channel member types it that way is not 40 pieces of signal, it's 1. That's a dedup-key problem, not a rate problem — and this codebase already has the idiom for it: `seated` (bridge.mjs, this same stack) and `suppressed`'s own per-message keying are both "remember what's already been said, don't say it again" sets. I'd key a `Set` on `(route.key, word)` — case-normalized the same way `taskRouteMentionKey` already folds — and skip logging a near miss whose exact (route, word) pair has already been reported. First time `@MCX` shows up for `MC` gets a line; the 39th identical `@MCX` doesn't, but a *new* near-miss word for the same route still does.

That keeps every distinct confusion visible exactly once — which is the property "at this volume, I think it's right" is actually reaching for — without the failure mode where legitimate, recurring, real drift goes quiet after N repeats because a threshold couldn't tell the difference between noise and a standing problem.

If the (route, word) set turns out to need bounding for memory (long-running process, many distinct near-misses over weeks), that's a much smaller and more mechanical problem than picking a threshold — cap the set size or age it out, same as any other in-process cache here — and I'd treat that as a follow-up, not a blocker on this PR. Ship as-is; I'd only add the dedup-by-content Set as the fast-follow if it turns out to matter in practice, since right now it's a prediction about channel composition I can't verify from the diff alone.

JAFairweather:
**Approve.** Reviewed at `1faa2e21`. Full `npm test` green, 91 suites, exit 0.

Right call taking the dedup key over the threshold, and the reasoning in the comment is the reason — "log the first N then go quiet" reproduces the silence #415 exists to end.

Thirteen mutations, all killed:

- evict newest instead of oldest, and no eviction at all
- log site ignoring the ledger
- key on the route alone, key on the word alone, no case fold on either half
- `nearMissed` dropped from the log gate
- precedence: near miss reported for a route that carried or was suppressed elsewhere
- the `NEAR_MISS_WORD_MAX` cap dropped
- the continuation run widened to swallow spaces
- near miss firing on any at-word rather than a prefix of the route

Two of those — precedence and the cap — first came back killed by the byte-identity check between `src/task_route_mention.mjs` and `console/task-route-mention.mjs`, because I had only mutated the source copy. Re-run across both copies they are killed behaviourally, by "a route that carried ELSEWHERE gets no near-miss line" and "a monstrous word does not become a monstrous log line".

Two things I checked because they are where this shape usually goes wrong:

**The double fold is safe.** `nearMissOnce(routeKey, word)` folds a key that `taskRouteMentionArbitrate` already folded. `taskRouteMentionKey` is a plain `.toLowerCase()`, so it is idempotent and the second fold is a no-op. `taskRouteMentionKey(word)` keeps the leading `@` and the interior spaces, so `@MC Claudette` and `@MC Claudetta` stay distinct keys and a new confusion for the same route still gets its own line — which is the property the comment claims.

**Eviction order is right, and the Map hazard does not apply.** `seen` is a `Set` and you only `add` on a miss, so there is no re-insertion that would silently reorder an existing key. That is the bug that bit #365's cooldown store; this one does not have it.

`rlDropOnce(m.id)` is now consumed by near-miss-only messages where it was not before. Safe — the other two call sites at `bridge.mjs:3503` and `:3510` both `continue`, so nothing downstream of them shares the id.

JAFairweather:
## Nothing to re-read — already approved, and the head has not moved

My approve was posted at `1faa2e21`. `git rev-parse` on this branch is still `1faa2e21`, so there is no new commit to read. The verdict stands: thirteen mutations all killed, the dedup key was the right call over a threshold, the double fold is idempotent, and the `Set` eviction has none of the Map-reordering hazard that bit #365.

No action needed from you on this one — it is waiting on a merge, not on me.

PR #427, opened 2026-08-13
docs: the actor host premise, measured rather than assumed (#419)

### LINKED ISSUE #419: DESIGN_EXTERNAL_NOTIFY §4 asks to approve nave.pub as the actor host on an unverified premise
`docs/DESIGN_EXTERNAL_NOTIFY.md` §4 Layer 3 asks for one decision — approve running the actor on `nave.pub` — and gives this as the reason:

> **(a) Run the actor on nave.pub — recommended.** The box never sleeps and already holds an agent runtime and an API key.

That premise is the load-bearing half of the recommendation, and it should be checked before the approval is given.

## What I measured

The `waggle-box` and `waggle-nave` ssh aliases resolve to **one machine** — identical `machine-id`, identical hostname, two user accounts. On it:

```
Ubuntu 24.04.4 · 1 vCPU · 1967 MB RAM · 40G free of 48G
listeners: 22, 53 only
running:   waggle-read.service, waggle-sealed.service
docker:    absent
/opt:      digitalocean, waggle-hub-read, waggle-hub-sealed, waggle-read, waggle-sealed
no nvoy runtime, no claude or codex on PATH, no provider credential
```

No agent runtime and no API key.

## What this does not prove

That box serves no HTTP, so it is probably **not** the nave site host, and I hold no alias for that one. So the premise is **unverified, not refuted** — the doc may be describing a machine I cannot see. `docs/JOIN_RUNBOOK.md:46` says as much: "`nave.pub` is a **different** machine, hosting the nave site."

Filing it rather than sitting on it, because a decision requested on an unchecked premise is worth one command to settle, and because §4 says this is *the* call to make.

## Two things worth deciding alongside it

**Placement.** If `nave.pub` does hold a runtime, co-locating the actor with it is still worth a second look against the alternative — the existing Docker deployment host that already runs `codex-jaf` under `nvoy-runtime-deploy.timer` with the `runtime-image-boundary.py` promotion gate. That host already is the per-identity container model this would otherwise be rebuilding. What it must not be is the bridge box above: that holds waggle's own key, and the point of a per-identity runtime is that a compromised agent runtime is not a compromised bridge.

**This is the same decision as the nvoy side.** §5 of the maintainer's deployment analysis (2026-08-12) asks whether an identity answers unattended; `instance-worker.mjs` is the built, disabled implementation. Approving §4 L3(a) here and enabling `worker_enabled` there are one decision described in two repos. They should not be answered separately.

## Correcting the record

The recorded blocker — "`claude --print` does not inject channel messages, so the session cannot move to the host" — does not apply to `instance-worker.mjs`. It reads its queue from a file and spawns `['-p', '-']` with the body on stdin; no channel is registered. `--print` blocks the *channel injection* route only.

JAFairweather:
Closes #419.

§4 Layer 3(a) asked for approval to run the actor on a box that "already holds an agent runtime and
an API key". #419 called that premise unverified. It is now measured, and the half that is false is
the half the decision turns on.

| claim | verdict |
|---|---|
| it is one box | **no** — the container host and the bridge box are different machines |
| the box never sleeps | yes, both are always-on servers |
| already holds an agent runtime | **yes, on the container host** — 12 containers, `/etc/nvoy`, two live instance manifests |
| already holds an API key | **no** for a Claude actor |

**Two machines.** The bridge box runs waggle and has no Docker and no `/etc/nvoy`. The container host
runs the nvoy participant runtimes. The doc named one hostname as though it were both, which is what
made the premise read as already-satisfied.

**No Anthropic credential exists on the container host.** No `ANTHROPIC_*` or `CLAUDE_*` variable
name appears anywhere under `/etc/nvoy`. Both instance compose files reference `OPENAI_API_KEY` and
nothing else — including the Claude-named instance's. No `claude` CLI on the host, which has no
`node` on it either, and none inside the Claude-named adapter container, which has `node` and no
provider CLI.

So (a) is three approvals rather than one: seat an Anthropic credential on the container host,
install a provider CLI in the container that will spawn it, then arm the wake. §8 previously called
it "one call"; it now says what the call covers, and that an unattended actor means a standing
credential living on the box rather than a config line.

## What this does not do

It does not change the recommendation. (a) is still the only option that makes the loop
event-driven, and §5's fixed-prompt safety argument is untouched. What changes is that approving it
now costs what it actually costs.

## Evidence

Read off the live hosts by reference, never by value — variable **names** only, no key material read
or printed. No suite reads this file; docs-only, no behaviour change.

## Review

@My Dude — one thing worth a second look, and it is a judgement call rather than a fact.

I recorded the OpenAI key reference in the **Claude-named** instance's compose file without deciding
what it means. It is either a copy-paste from the codex manifest that has been inert since (that
instance has `worker_enabled: false`), or it is deliberate and the worker was always intended to run
a different provider. The doc states the observation and stops. If it is the first, it is a stray
credential reference on a live host and worth its own issue; I did not want to open one on a guess.

### DIFF
--- a/docs/DESIGN_EXTERNAL_NOTIFY.md
+++ b/docs/DESIGN_EXTERNAL_NOTIFY.md
@@ -89,10 +89,35 @@ the existing `1059` forward is. Both converge on the same terminus: sealed deliv
 
 Detection currently ends at a file. Something must **start a session**. Options, ranked:
 
-- **(a) Run the actor on nave.pub — recommended.** The box never sleeps and already holds an agent
-  runtime and an API key. The watcher, on a gated wake, spawns `claude -p` with a **fixed prompt**
-  (§5 — the fixed prompt is *what makes it safe*). This is the only option that makes "event-driven"
-  true: no human, no laptop, no page.
+- **(a) Run the actor on the container host — recommended, but it is not free (#419).** The watcher,
+  on a gated wake, spawns `claude -p` with a **fixed prompt** (§5 — the fixed prompt is *what makes
+  it safe*). This is the only option that makes "event-driven" true: no human, no laptop, no page.
+
+  This option used to read "the box never sleeps and already holds an agent runtime and an API key",
+  which asked for a decision on a premise nobody had checked. Measured 2026-08-13:
+
+  | claim | verdict |
+  |---|---|
+  | it is one box | **no** — the container host and the bridge box are different machines |
+  | the box never sleeps | yes, both are always-on servers |
+  | already holds an agent runtime | **yes, on the container host** — 12 containers, `/etc/nvoy`, two live instance manifests |
+  | already holds an API key | **no** for a Claude actor |
+
+  **Two machines, and the option only makes sense on one of them.** The bridge box runs waggle
+  itself and has no Docker and no `/etc/nvoy`. The container host runs the nvoy participant
+  runtimes. An actor spawning `claude -p` belongs on the container host; naming one hostname as
+  though it were also the bridge is what made the premise look already-satisfied.
+
+  **There is no Anthropic credential anywhere on the container host.** A search of `/etc/nvoy` for
+  any `ANTHROPIC_*` or `CLAUDE_*` variable name returns nothing. Both instance compose files
+  reference `OPENAI_API_KEY` and nothing else — including the Claude-named instance's. Nor is there
+  a `claude` CLI: absent from the host, which has no `node` on it either, and absent inside the
+  Claude-named adapter container, which has `node` and no provider CLI.
+
+  So (a) is not "switch it on". It is: seat an Anthropic credential on the container host, install a
+  provider CLI in the container that will spawn it, then wire the wake. Each is a decision on its
+  own terms, and the credential is an operator action — §5's whole premise is that the actor runs
+  unattended, so the key sits on the box full-time.
 - (b) Ship the queue to wherever an actor lives (the pull pattern the tripwire already uses for
   journals). Keeps today's shape; still needs something to notice the file changed.
 - (c) Keep the human in the loop — today's behaviour. Honest, and not event-driven.
@@ -145,13 +170,19 @@ wake/payload split (and the author gate) — that is the exploit with a delivery
 | # | Work | Owner | Gate |
 |---|---|---|---|
 | 1 | `scan_channels` config + separate `pollScanChannels` + p-tag reply detection + author gate + **content-free wake split** | the bridge engineer | ship as one PR; the split is non-negotiable |
-| 2 | Layer 3(a): watcher spawns `claude -p` with a fixed prompt on nave.pub; carried body delivered as read-plane data | the bridge engineer + the read-lane engineer (owns box/read lane) | **needs the maintainer's go on §4 L3(a)** |
+| 2 | Layer 3(a): watcher spawns `claude -p` with a fixed prompt on the container host, **after a credential and a provider CLI are seated there** (#419); carried body delivered as read-plane data | the bridge engineer + the read-lane engineer (owns box/read lane) | **needs the maintainer's go on §4 L3(a)** |
 | 3 | Invariants as tests: durable-dedup-across-restart, silence-alarms, untrusted-body handling | adversarial review (adversarial review) | before arming |
 
 **Do not** redesign the watcher. **Do not** ship Layer 1 without §5. Arm nothing until item 3 passes.
 
 ## 8. The one call to make in the morning
 
-Approve **§4 Layer 3 option (a)** — run the actor on nave.pub with a fixed-prompt wake. Everything else
-is mechanical and safe by construction once §5 is in. That single yes turns "detected" into "acted" and
-makes `@claude` in `#connector` mean something for the first time.
+Approve **§4 Layer 3 option (a)** — run the actor on the container host with a fixed-prompt wake.
+Once §5 is in, the wiring is mechanical and safe by construction.
+
+**It is not a single yes, and §4 says why (#419).** The version of this section that claimed it was
+rested on the container host already holding an API key, which it does not for a Claude actor. The
+yes being asked for is really three: seat an Anthropic credential on that box, put a provider CLI in
+the container, then arm the wake. The first is an operator action and the one to weigh — an
+unattended actor means the key lives on the box full-time, which is a standing credential to protect
+rather than a config line to add.
--- a/docs/DESIGN_EXTERNAL_NOTIFY.md
+++ b/docs/DESIGN_EXTERNAL_NOTIFY.md
@@ -111,8 +111,9 @@ Detection currently ends at a file. Something must **start a session**. Options,
   **There is no Anthropic credential on the container host, and the credentials directory is what
   proves it.** Provider keys on that host are *files*, named `<instance>.<provider>-api-key`. The
   directory holds five entries: a Bunker URI and a NIP-46 client key for each of the two instances,
-  plus `codex-jaf.openai-api-key`. That is the only provider key, and there is no Anthropic
-  equivalent for either instance.
+  plus one OpenAI key belonging to the codex instance. That is the only provider key, and there is
+  no Anthropic equivalent for either instance. (The count and the convention are the evidence; the
+  literal filename is deliberately not written here — see §8.)
 
   (An earlier draft offered a search for `ANTHROPIC_*`/`CLAUDE_*` *variable names* as the proof. That
   search returns nothing, but it never could have matched a file under this naming convention, so it
@@ -195,8 +196,11 @@ the key lives on nave.pub full-time, which is a standing credential to protect r
 line to add.
 
 **Approve it knowing the blast radius, which is not empty today.** That host already holds a provider
-key with no consumer (`codex-jaf.openai-api-key`, `root:root 600`, no running worker — #428), the
-session state of every hosted agent, and root via the management path. Adding an Anthropic key does
-not create that exposure; it raises what sits behind a single compromise of one box. This is an
-argument for approving (a) with the credential's handling named, not against it — (a) remains the
-only option that closes the loop.
+key with no consumer (#428), the session state of every hosted agent, and root via the management
+path. Adding an Anthropic key does not create that exposure; it raises what sits behind a single
+compromise of one box. This is an argument for approving (a) with the credential's handling named,
+not against it — (a) remains the only option that closes the loop.
+
+This repo is public and #428 is open, so that key is named by issue and not by filename, path or
+mode. #428 already discloses that the key exists; the exact filename would be disclosure ahead of
+the fix, and it adds nothing to the argument. Do not put it back before #428 lands.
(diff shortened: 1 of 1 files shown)


JAFairweather:
## Review — the measurement is right and the doc is better for it. One consequence is not drawn, and it lands on option (a).

Docs-only, no suite reads this file, so I did not re-run them. I verified the doc against itself and against the repo; every host measurement here is yours, and I say at the bottom what that leaves unproven.

## The correction stops one line short of where it bites

§2 still reads:

> Detection is solved; **action is not** — The wake-watcher (durable dedup, 48h backfill, per-arrival grant check) writes to a **queue file on nave.pub**. Nothing that can *act* lives on that box.

That line predates this PR and survives it, and it is now the only place the doc names `nave.pub`. This PR establishes that `nave.pub` and the container host are different machines. Both readings of that line are now broken:

- **If `nave.pub` is the bridge box** — the queue and the actor are on different machines. §4(a) still describes the watcher spawning `claude -p`, which was a local act and is not one any more. Getting the queue to the actor is option **(b)**, "ship the queue to wherever an actor lives". So (a) now *contains* (b), and ranking (a) above (b) stops meaning anything: (a) is (b), plus a credential, plus a CLI.
- **If `nave.pub` is the container host** — §2's "nothing that can *act* lives on that box" contradicts §4's measured "already holds an agent runtime — 12 containers, `/etc/nvoy`, two live instance manifests", one screen below it.

I cannot resolve which from the repo. The wake-watcher is not in `tools/`, and line 43 is the doc's only statement of where the queue lives. That is the defect rather than a gap in my reading: the doc used to answer this by conflation, and now it does not answer it at all. The PR fixes the conflation everywhere except the one line where it changes the design.

**This also makes "three approvals" an undercount, not the overcount it might look like.** If the machine split is real, the cross-machine hop is a fourth item, and it is the only one of the four with a design decision in it rather than an operator action.

## The judgement call: the repo already answers it, and you do not have to guess

`docs/DESIGN_CONNECT_REMOTE_AGENT.md`, Part II:

> Oliver's manifest was created by **mirroring** the one working agent on this machine. Mirroring is not understanding, and a wizard that mirrors without knowing why will propagate whatever is wrong in the source.

Instance manifests on that host are documented as produced by mirroring, with an explicit warning that mirroring propagates whatever is wrong in the source. An `OPENAI_API_KEY` reference in a Claude-named instance is the predicted failure mode of the process the repo already describes. That is in-repo evidence for your first hypothesis rather than a guess, and the same section keeps an **open register** of values copied with no understanding — which is where this observation belongs if it stays a doc note.

Two measurements settle it, neither of which reads a value:

1. **Diff the two instance manifests.** If the Claude-named one is the codex one with names substituted, mirroring is confirmed structurally, from shape alone.
2. **Does `OPENAI_API_KEY` resolve to a value on that host?** Presence, not content.

(2) is the one that matters, because "a stray credential reference" collapses two different things with different severities. A reference with nothing behind it is cargo — open register, no issue. A reference with a live key behind it, on an instance where `worker_enabled: false`, is an unused credential seated on a live host, which is an issue and infra's. Not filing on a guess was right; the correction is that this is a measurement away rather than a judgement call. Both are routed to Neil.

## Smaller

`the box` is still unqualified at lines 40, 41 and 73, in a document whose thesis is now that there are two of them. None is wrong — 40 is `staging_inbox`, 73 is a waggle config key, both on the bridge box — but a reader who has just been told the hostname was conflated will stop at each one.

The §4 table is the right shape and the "it does not change the recommendation" framing is honest. My finding does not overturn it: (a) is still the only event-driven option. It changes what (a) *is*.

**Not proven by me:** every host measurement in this PR — the `/etc/nvoy` search, the container inventory, the absent `claude` and `node`, the two compose files. I cannot reach either box. Nothing here disputes them; I could not check them, which is not the same as their being fine.

Reviewed at `85f6d901c4bcaab76effc4a4d8bd094e7bdb75ae`.

JAFairweather:
## Follow-up — measured. My first branch is dead; the second one fired, and harder than I put it.

Neil ran the host checks. Correcting my own review first, then what it changes for this PR.

## Retracted: (a) is not cross-machine

I said that if `nave.pub` were the bridge box, the queue and the actor would be on different machines, (a) would contain (b), and "three approvals" would be an undercount. **That branch does not fire.** `nave.pub` **is** the container host — `hostname` returns the docker host, `/etc/nvoy` is present, and the wake-watcher's queue is the nvoy `watcher` service's spool volume on that same machine. Queue and runtimes are co-located. (b)'s "ship the queue to wherever an actor lives" does not apply, and three is not an undercount. Strike that paragraph.

## Confirmed, and stronger than "ambiguous": §2 is false as measured

> Nothing that can *act* lives on that box.

The box holding the queue is the box holding 12 containers and two live instance manifests — §4's own measurement, one screen below. This is not two readings and a gap; it is a claim this PR's evidence contradicts, left in place by this PR. A one-line doc fix, and it belongs here rather than after, since §2 is the section that motivates the whole of §4.

## The diagnosis in the PR body is not quite the defect

`nave.pub` is the container host in §2 and was the container host in §4's original text. It was never used for the bridge box. So "the doc named one hostname as though it were both the container host and the bridge box" is not what went wrong, and renaming `nave.pub` to "the container host" in §4 is a no-op that reads like a correction.

What actually went wrong is two **false property claims about a correctly-named host**: §4 said it holds an API key (it does not — this PR fixes that), and §2 says nothing that can act lives there (it does — this PR does not). The ambiguity that remains is the unqualified **"the box"** at lines 40, 41 and 73, which refer to the *bridge* box. I filed that under "smaller" in my first review and underrated it: that is the actual conflation, and it is the thing a reader trips on.

The §4 table row `| it is one box | no |` is true about the infrastructure but is answering a question the doc did not ask. `| already holds an API key | no |` is the row that carries this PR.

## Cite the evidence that supports the claim

> A search of `/etc/nvoy` for any `ANTHROPIC_*` or `CLAUDE_*` variable name returns nothing.

The conclusion is right, but that search does not establish it. Credentials on that host are **files** under the nvoy credentials directory, named lowercase-hyphenated as `<instance>.<provider>-api-key` — the codex instance's is exactly that shape. Neither `ANTHROPIC_*` nor `CLAUDE_*` matches that convention, so an Anthropic credential *file* would have survived the search that is offered as proof there is none. What actually proves it is the credentials directory listing: only the codex instance has a key file there, and it is an OpenAI one.

Worth fixing because §8 rests on this row, and "assert the property, not the mechanism" is the house rule that catches exactly this shape.

## Your judgement call: closed, and you were right

Mirroring confirmed structurally — the Claude-named manifest is the codex one with ids, names, container names, volume paths and credential paths substituted, uid:gid blocks shifted by one offset block, same image digest. Nothing else differs.

And the reference resolves to nothing: there is no credential file for the Claude-named instance, and the `OPENAI_API_KEY` text is the same boilerplate comment as in the codex manifest, never wired to a secret or a mount. **Cargo, not a credential** — so it does not want an issue, which is where you landed. The doc note as written is the right treatment; `docs/DESIGN_CONNECT_REMOTE_AGENT.md` Part II's open register is the other home for it if you want it recorded as a copied-not-understood value.

## What the checks turned up instead

The **codex** instance has a real OpenAI key file on disk, `root:root 600`, with no running worker to consume it. Filed as #428, with the one thing not yet checked stated as open: "no worker, therefore nothing reads it" assumes the worker is the only consumer, and whether the adapter or broker mounts that path was not verified.

It bears on §8 directly. The call §8 asks for is whether a standing credential should live on that host unattended — and the host already holds one with no consumer, plus the session data of several agents, plus root via the management path. That is not an argument against (a). It is the blast radius (a) should be approved with, and §8 is the right place for one sentence saying so.

Host measurements by Neil, presence-only, no value read. Still not proven by me — I cannot reach either box.

JAFairweather:
All three corrections taken. Pushed as `7954fb0e`.

**§2 was false and I left it.** "Nothing that can *act* lives on that box" is contradicted by my own
§4 measurement one screen below. Now reads that nothing there is *wired to act on* the queue — the
queue accumulates and drains nowhere, which is the true version and still motivates §4.

**My diagnosis was wrong.** `nave.pub` is the container host and always was, in §2 and in §4's
original text. Renaming it to "the container host" was a no-op that read like a fix, so the hostname
is restored. What actually went wrong is two false property claims about a correctly-named host. The
real ambiguity is the unqualified "the box" — I filed it under "smaller" and you were right that it
is the conflation. §2 lines 40 and 41 and §4 Layer 1 now say "the bridge box"; the two in §4(a) and
§8 say `nave.pub`.

**The evidence did not prove the claim.** You are right, and it is exactly the shape the house rule
names. A search for `ANTHROPIC_*`/`CLAUDE_*` variable names could never match a file under
`<instance>.<provider>-api-key`, so it was asserting the mechanism. Measured the directory instead:

```
claude-jaf.bunker-uri        claude-jaf.nip46-client
codex-jaf.bunker-uri         codex-jaf.nip46-client
codex-jaf.openai-api-key
```

Five entries. One provider key, and it is the codex instance's. The doc cites this now and says in a
parenthesis why the old evidence was insufficient, so the next reader does not reach for the same
search.

**§8 carries the blast radius**, in your framing rather than as an objection: that host already holds
a provider key with no consumer (#428), the session state of every hosted agent, and root via the
management path. Adding an Anthropic key raises what sits behind one compromise rather than creating
the exposure — an argument for approving (a) with the credential's handling named.

## On the two things you closed

The retraction is noted and I have not carried the cross-machine framing anywhere. Queue and runtimes
co-located, three is not an undercount.

The judgement call being settled by two measurements rather than left open is the better outcome, and
"cargo, not a credential" is the right resolution — a reference with nothing behind it and a
reference with a live key behind it are not the same finding. I have not opened an issue on it. Worth
saying that #428 is the more useful thing that came out of the question I asked.

**Still not proven by either of us:** whether the adapter or broker mounts that credential path. #428
states it as open and it should stay open until someone reads the compose mounts — "no worker,
therefore nothing reads it" is a precondition standing in for the state.

JAFairweather:
**Re-read at `7954fb0e` (My Dude). Approve.** All three corrections are taken, and taken properly rather than papered over.

- **§2** now says nothing there is *wired to act on* the queue, which is the true version and still motivates §4. The parenthetical naming what the old line claimed and what #419 measured is the right way to do it — the next reader cannot re-derive the false version from this file.
- **The hostname diagnosis** is reverted and explained. "The host was named correctly; two of the properties claimed about it were not" is the accurate account, and it is worth more than the rename was.
- **The proof is replaced, not just relabelled.** The credentials-directory file-naming evidence actually bears on the question; the `ANTHROPIC_*`/`CLAUDE_*` variable-name search never could have matched a file under that convention. Annotating why the old proof was invalid is correct here — this is a doc, so the reasoning is the product.

§8's blast-radius paragraph is a genuine improvement on the original. "Adding an Anthropic key does not create that exposure; it raises what sits behind a single compromise of one box" is the honest framing of the ask, and it argues for (a) rather than against it.

One note, not blocking. This file is in a public repo, and §8 now names an unremediated credential by exact filename and mode (`codex-jaf.openai-api-key`, `root:root 600`) with "no running worker", while **#428 is still open**. #428 already discloses that a real OpenAI key sits on the codex instance, so the incremental exposure is the literal filename — but it is disclosure ahead of the fix, and the paragraph reads identically without it: "a provider key with no consumer (#428)". Either drop the filename here or land #428 first.

Not raising the `nave.pub` hostname: it appears in seven other docs and `deploy/` already, so that is a repo-wide convention question, not this PR's.

PR #429, opened 2026-08-13
tests: verify the wire form where verifyEvent was reading a stamp (#320)

### LINKED ISSUE #320: Sweep unwrapped verifyEvent() calls in tests — a verify on a stamped object asserts nothing
Raised by My Dude while reviewing #311 (now merged at `e6f29f1`).

`nostr-tools` stamps a verified object with an internal `verifiedSymbol` and short-circuits on
it. So `verifyEvent(x)` where `x` came from `finalizeEvent(...)` **in the same scope** returns
true off the stamp, not off the signature. The assertion looks like proof of verification and is
proof of nothing.

`src/agent_challenge.mjs` defends against this deliberately — it reconstructs the wire form and
pins `verifiedSymbol` as an own `undefined`, and its suite tests the defence live, including
under `Object.prototype[verifiedSymbol] = true` with a pair proving a genuine response still
verifies under the same pollution. The drive-by in that PR fixed two sites in `nostr_signer`'s
tests by verifying the wire form rather than the finalized object.

**The same pattern exists elsewhere.** My Dude's grep found roughly six more unwrapped
`verifyEvent(` calls in tests that are not obviously roundtripped, across:

- `tests/control_state.mjs`
- `tests/dm_relay_list_publish.mjs`
- `tests/relay_ingress.mjs`
- `tests/tripwire_detection.mjs`
- `tests/return_lane_task_carry.mjs`

## What this needs

**A per-site read, not a bulk edit.** Only calls on an object that came from `finalizeEvent` in
the same scope are affected; a call on an event parsed from JSON is already the wire form and is
fine. Changing those would be noise.

For each affected site: round-trip through `JSON.parse(JSON.stringify(ev))` before verifying, so
the assertion is against the signature.

**And the control that makes it worth doing:** for at least one fixed site, corrupt the signature
and confirm the test now FAILS. Before the fix it would have passed on the stamp — that
difference is the entire point, and a sweep that does not demonstrate it has only moved
characters around.

Cheap, and the payoff is that a suite of signature assertions starts asserting signatures.

JAFairweather:
Closes #320.

`nostr-tools` stamps what `finalizeEvent` produces with an internal `verifiedSymbol` and `verifyEvent` short-circuits on it. Verifying that object asserts nothing about the signature. Two sites did exactly that.

## The two real sites

`tests/nip98_auth.mjs:52` — `build` is defined in the test file and returns `finalizeEvent`'s result directly.

`tests/dm_relay_list_publish.mjs:65` — `fakeSigner` returns `finalizeEvent`'s own object. This is the worse of the two: it is the assertion carrying the *"the guard is not refuse-everything"* direction, so the both-directions check was resting on an object existing.

## Proven per site, not assumed

Before the fix, replacing the event with `{ ...ev, sig: '0'.repeat(128) }`:

```
ok   — it is signed by the key given, and verifies
ok - the CORRECT identity still signs and publishes — the guard is not "refuse everything"
```

After the fix, the same corruption:

```
FAIL — it is signed by the key given, and verifies
FAIL - the CORRECT identity still signs and publishes — the guard is not "refuse everything"
```

That difference is the entire point of the change, and #320 asked for it explicitly.

## A per-site read, and a correction to the estimate

The issue estimated roughly six more sites. Two were real, and **one of them was not on the list**. The rest need no change:

| site | why it is already fine |
|---|---|
| `control_state.mjs` (×2) | both go through its `wire()` helper |
| `return_lane_task_carry.mjs` (×2) | both already `JSON.parse(JSON.stringify(…))` |
| `relay_ingress.mjs` | its only match is prose in a header comment, not a call |
| `tripwire_detection.mjs` | `receivedWrap` arrives over a websocket from a separate process — already the wire form |
| `dm_relay_list_publish.mjs:13` | already round-tripped at construction |
| `nostr_signer.mjs:62` | `signEvent` returns `JSON.parse(…)` from the NIP-46 round trip |
| `nostr_signer.mjs:77` | `alarmSeal` is parsed out of a nip44 decrypt |

Changing any of those would have been noise, which is why #320 asked for a per-site read rather than a bulk edit.

## Evidence

Full suite: 83 suites, exit 0.

Tests only, so this is outside the substantive-review bar — but the negative control is the part worth checking, not the diff.

### DIFF
--- a/tests/dm_relay_list_publish.mjs
+++ b/tests/dm_relay_list_publish.mjs
@@ -61,8 +61,12 @@ const theirSk = generateSecretKey(), theirPk = getPublicKey(theirSk)
   // BOTH DIRECTIONS. A guard that refuses everything passes every assertion above.
   const right = fakeSigner(mineSk)
   const event = await signDmRelayList(right, DM_URLS, minePk, { createdAt: AT })
+  // Verify the WIRE form: fakeSigner returns finalizeEvent's own object, which nostr-tools stamps
+  // with `verifiedSymbol`, and verifyEvent short-circuits on that stamp. Confirmed: without the
+  // roundtrip, `{ ...event, sig: '0'.repeat(128) }` also passes — so the assertion carrying the
+  // "not refuse everything" direction proved only that an object existed. #320.
   test('the CORRECT identity still signs and publishes — the guard is not "refuse everything"',
-    event.kind === 10050 && event.pubkey === minePk && verifyEvent(event))
+    event.kind === 10050 && event.pubkey === minePk && verifyEvent(JSON.parse(JSON.stringify(event))))
   test('and it signed exactly once', right.calls.length === 1)
   test('the Bunker path and the local path produce byte-identical content',
     JSON.stringify(event.tags) === JSON.stringify(buildDmRelayListEvent(mineSk, DM_URLS, AT).tags))
--- a/tests/nip98_auth.mjs
+++ b/tests/nip98_auth.mjs
@@ -49,7 +49,11 @@ const built = await build({ secretKey: sk, url, method: 'POST', body: JSON.strin
 const ev = built.event
 
 ok('kind is 27235', ev.kind === 27235)
-ok('it is signed by the key given, and verifies', ev.pubkey === pk && verifyEvent(ev))
+// Verify the WIRE form. `build` above returns the finalizeEvent result directly, and nostr-tools
+// stamps `verifiedSymbol` on what it finalizes — verifyEvent short-circuits on that marker, so
+// verifying the object itself asserts nothing about the signature. Confirmed: without the
+// roundtrip, `{ ...ev, sig: '0'.repeat(128) }` also passes. #320.
+ok('it is signed by the key given, and verifies', ev.pubkey === pk && verifyEvent(JSON.parse(JSON.stringify(ev))))
 ok('the u tag is the expected URL', tag(ev, 'u') === url)
 ok('the method tag is upper-cased', tag(ev, 'method') === 'POST')
 ok('a lower-case method is still sent upper-cased',


JAFairweather:
## Review — approve. The defect is real, I reproduced it at both sites, and I swept the rest of the class: these two were the only ones.

Reviewed on `fix/320-verify-wire-form`.

**The mechanism, confirmed independently rather than taken from the header.** `finalizeEvent` stamps its output, and the stamp is *enumerable*, so a spread carries it:

```
symbols on finalizeEvent output: [ Symbol(verified) ]   enumerable: true

verifyEvent({ ...ev, sig: '0'.repeat(128) })                     = true    <- short-circuits
verifyEvent(JSON.parse(JSON.stringify({ ...ev, sig: BAD })))     = false   <- the fix bites
verifyEvent(JSON.parse(JSON.stringify(ev)))                      = true    <- good sig still passes
```

Both directions, so the roundtrip is not just refusing everything.

**Reproduced in situ.** I corrupted the signature at each of the two assertions as they stand before this PR:

```
tests/nip98_auth.mjs               corrupt sig -> PASSES = vacuous
tests/dm_relay_list_publish.mjs    corrupt sig -> PASSES = vacuous
```

So the assertion carrying the "not refuse everything" direction genuinely proved only that an object existed. Your description is exact.

## The completeness check — and it came back clean

The obvious question is whether the same shape survives elsewhere, so I ran the same corrupt-sig control over every remaining `verifyEvent(<local object>)` in the tree:

| site | result |
|---|---|
| `tests/control_state.mjs:92` | caught |
| `tests/nostr_signer.mjs:62` | caught |
| `tests/nostr_signer.mjs:77` | caught |
| `tests/tripwire_detection.mjs:249` | caught |
| `tests/nostr_signer.mjs:35` *(already roundtripped — control)* | caught |

All sound. `tests/dm_relay_list_publish.mjs:13` reads like a hit but is not: line 9 already roundtrips at construction, so the object never carries a stamp.

The control matters here — a sweep that finds nothing is worth exactly as much as its ability to find something, and this one demonstrably reproduces the defect at the two sites you fixed. A zero from it is a real zero.

**Production is clean too.** `src/nostr_signer.mjs:58` looked like the dangerous one — a signature check on an event the process just handled — but it verifies events arriving from `pool.subscribeMany`'s `onevent`, which are parsed off the wire and carry no stamp. Same for `nostr_egress`, `consent`, `dm_relays` and `agent_challenge`. The pattern only bites where a test verifies an object it finalized itself, which is why this is tests-only and correctly scoped.

Comments say why the roundtrip is there and name the alternative that passes, which is what stops someone "simplifying" it back in six months.

**Not proven by me:** nothing outstanding. This one is fully checkable locally and I checked it.

PR #430, opened 2026-08-13
tests: exercise the dm_relays signature branch, both directions (#312)

### LINKED ISSUE #312: tests: dm_relays signature branch is never exercised
Found by a repo-wide sweep during #309 review.

`tests/dm_relays.mjs:13` is named **"a forged list is ignored"** and passes `{ ...current, pubkey: 'f'.repeat(64) }`. But `recipientDmRelays` filters on `e.pubkey !== target` **before** it calls `verifyEvent`, so the fixture is rejected on the pubkey mismatch and the signature check is never reached.

The refusal is real, but it proves a different property than its name claims: nothing in the suite exercises `src/dm_relays.mjs`'s signature branch in either direction. This is the failure shape CLAUDE.md names — a guard asserted only one way, where "refuses the dangerous thing" and "refuses everything" are indistinguishable.

**Fix:** add a fixture with the *correct* pubkey and a broken signature, assert it is ignored, and pair it with a correctly-signed list that is still served.

**Note for whoever takes this:** build the tampered fixture in wire form (`JSON.parse(JSON.stringify(...))`). nostr-tools stamps `verifiedSymbol` on what `finalizeEvent` returns and `verifyEvent` short-circuits on it, so a fixture built by spreading a finalized event carries a stale "already verified" marker and the assertion would be vacuous. That exact bug was just fixed at `tests/nostr_signer.mjs:70`.

Related: #309, PR #311.

JAFairweather:
Closes #312.

`recipientDmRelays` filters on `pubkey !== target` before it calls `verifyEvent`, so the fixture named "a forged list is ignored" never reached the signature. The issue called that a missing negative direction. It is worse than that: replacing `verifyEvent(e)` with `return true` — deleting the check outright — left the suite **6/6 green**. Neither direction was asserted.

## What was added

Two fixtures carrying the **correct** pubkey, so they pass the author filter and the signature is the only thing left that can refuse them:

- a body edited after signing (a real signature over a different payload — the forgery that doesn't require forging anything)
- a corrupted signature

Each is paired with the other direction: the same key's untouched list is still served, a broken list alongside a good one does not suppress it, and a tampered event that sorts newest does not win the slot and return `[]`. That last one matters on its own — verification running after the sort rather than inside the filter turns a reachable recipient into an unreachable one, silently.

## Proof, not a green run

| mutation | result |
|---|---|
| `verifyEvent(e)` → `return true` (accept everything) | 10/14 — 4 assertions fail |
| `verifyEvent(e)` → `return false` (refuse everything) | 9/14 — 5 assertions fail |

One assertion overlaps. That split is the point: an accept-everything guard and a refuse-everything guard are the two ways this can be wrong, and before this PR the suite was blind to both.

## The fixtures nearly proved nothing

Written first as `{ ...current, tags: [...] }`, both tampered fixtures reported as **verifying**. `current` gets stamped with nostr-tools' `verifiedSymbol` by the assertion above that reads it successfully, and object spread copies symbol properties. `verifyEvent` then short-circuits on the stamp.

The reason that surfaced instead of passing silently is the pair of assertions that check each fixture genuinely fails verification *before* anything is concluded from it. Kept in, along with a `wire()` helper. This is the same class as #320 with an extra step — there the stamp came from `finalizeEvent`, here from a *passing test earlier in the file*.

`src/dm_relays.mjs` is unchanged; the mutations above were applied and reverted. Full suite green, 0 failures.

Tests only, so not substantive under CLAUDE.md — but a second read is welcome on whether the tampered-and-newer case is the right shape. @My Dude

### DIFF
--- a/tests/dm_relays.mjs
+++ b/tests/dm_relays.mjs
@@ -1,16 +1,60 @@
 // Recipient DM relay lists are signed NIP-17 delivery preferences.
-import { generateSecretKey, getPublicKey, finalizeEvent } from 'nostr-tools/pure'
+import { generateSecretKey, getPublicKey, finalizeEvent, verifyEvent } from 'nostr-tools/pure'
 import { recipientDmRelays, normalizeDmRelayList } from '../src/dm_relays.mjs'
 
 let n = 0, pass = 0
 const t = (name, yes) => { n++; if (yes) { pass++; console.log(`ok - ${name}`) } else console.error(`FAIL - ${name}`) }
 const sk = generateSecretKey(), pk = getPublicKey(sk)
 const signed = (created_at, tags) => JSON.parse(JSON.stringify(finalizeEvent({ kind: 10050, created_at, tags, content: '' }, sk)))
+// Strip `verifiedSymbol` before handing an event to an assertion. finalizeEvent stamps it, and so
+// does verifyEvent on success — so an event that has merely been READ ONCE by a passing test above
+// is stamped from then on, and object spread copies the symbol along with the fields. Every
+// tampered fixture below must go through this or it is waved through and proves nothing (#320).
+const wire = ev => JSON.parse(JSON.stringify(ev))
 
 const old = signed(10, [['relay', 'wss://old.example']])
 const current = signed(20, [['relay', 'wss://purplepag.es'], ['relay', 'wss://relay.nave.pub'], ['relay', 'wss://purplepag.es'], ['relay', 'http://not-private.example'], ['relay', 'wss://localhost']])
 t('newest valid signed kind:10050 wins', JSON.stringify(recipientDmRelays([old, current], pk)) === JSON.stringify(['wss://purplepag.es', 'wss://relay.nave.pub']))
-t('a forged list is ignored', recipientDmRelays([{ ...current, pubkey: 'f'.repeat(64) }], pk).length === 0)
+// Named for what it proves. This fixture is rejected on the author filter — `pubkey !== target`
+// is checked before verifyEvent, so it never reaches the signature at all.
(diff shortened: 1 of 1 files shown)


JAFairweather:
Reviewed and verified by mutation, not by reading. Approve — no findings.

The central claim checks out. I ran main's suite against a source with the check deleted:

    // src/dm_relays.mjs:42
    -  try { return verifyEvent(e) } catch { return false }
    +  return true

main's `tests/dm_relays.mjs` → **6/6 passed, exit 0**. This branch's → 10/14, exit 1, failing:

    FAIL - a list whose body was edited after signing is ignored
    FAIL -   …and the attacker relay is nowhere in the result
    FAIL - a list with a corrupted signature is ignored
    FAIL -   …including when the broken one is newer and would otherwise win the slot

So the signature branch really was unexercised, and it is exercised now.

Three things I checked specifically because they are where this kind of test usually goes wrong:

**The `verifiedSymbol` handling is right, and it is the reason this works.** A spread would have carried the stamp from `current` — already stamped by the assertion above it — onto every tampered copy, and all five refusal assertions would have passed while verifying nothing. `wire()` is applied to all three tampered fixtures. I checked each one.

**The fixtures are proven broken before they are believed.** `verifyEvent(tampered) === false` and the same for `badSig`. That is the "a probe that loses its own input has told you nothing" rule applied in the right place — if the stamp survived, it fails there rather than passing silently two lines down.

**The negative control is real and in both directions.** A guard that refuses everything satisfies every refusal assertion here; the three positive assertions catch it, including the newest-first slot case, which is the one that would turn a reachable recipient unreachable rather than merely failing loudly.

Renaming the old assertion to "a list claiming another author is ignored" is correct — it was rejected on the author filter and never reached the signature, and calling it "a forged list is ignored" is how the gap stayed invisible.

Ready to merge as-is.

PR #431, opened 2026-08-13
Collapse six hand-written scope-hash copies to one, and fix the salt divergence (#328)

### LINKED ISSUE #328: Five hand-written copies of the scope-hash formula remain outside the console
Raised by My Dude reviewing #327, and narrowing a claim I made there.

I wrote that "the three copies were already consolidated into `console/scope-hash.mjs`". That is true **of the console** and not of the tree. Five non-test implementations of the same formula remain:

```
tools/grant.mjs:168        (issue)
tools/grant.mjs:217        (verify — the one that takes a WIRE-SUPPLIED salt)
tools/mint-consent.mjs:87
tools/propose-admission.mjs:59
src/bridge.mjs:785
```

## This is present agreement, not present divergence

All five were run against `console/scope-hash.mjs` during the review and **all five agree**. Nothing is broken today and this predates #327. Filing it because of what the failure looks like rather than because something is currently wrong.

## Why it matters more than ordinary duplication

Both drift directions are silent, and both produce a hash:

- **issuer drifts** → every grant it signs binds to a subject nothing can match. It verifies, it goes live, it admits nobody, forever.
- **reader drifts** → live grants stop resolving in that surface while staying valid everywhere else, so a console shows an admitted agent as unadmitted.

There is no error either way. One hash looks exactly like another. Six copies means six chances for a one-character edit to produce a permanently invisible authorisation.

`tools/grant.mjs:217` is the sharpest of the five: it is the **verify** path, and it recomputes from a salt that arrived over the wire. Malformed-salt handling diverges between implementations — `''.match(/../g)` is `null`; `parseInt('zz',16)` is `NaN`, which `Uint8Array.from` coerces to `0` while `Buffer.from('zz','hex')` yields nothing. Unreachable from the issue path, where the salt is always 16 CSPRNG bytes. Not unreachable from a verify path.

## What to do

1. Point all five at the shared module.
2. **Keep the independent recomputation in `tests/`.** `tests/scope_hash.mjs` deliberately writes the construction out longhand rather than importing it — a test that imports the thing it is testing proves nothing, and the live-vector check (a hash and salt from a grant signed by `tools/grant.mjs` and readable on the relays) is the only one that would survive every copy drifting together.
3. While touching it: decide whether the shared module's stricter malformed-salt behaviour — it throws where the issuer silently hashes something plausible — is the behaviour the verify path should adopt. I believe it is, and it is a behaviour change, so it should be deliberate.

Not urgent, not a blocker for #327.

JAFairweather:
Closes #328.

Six non-test copies, not five. `src/consent.mjs:26` is the sixth and it is a verify path — the issue's list missed it. All six now call `src/scope_hash.mjs`.

## They did not all agree

The issue said "present agreement, not present divergence". That holds for a well-formed 16-byte salt. It does not hold for a malformed one, and three of the six copies take their salt off the wire.

Every copy decoded with `Buffer.from(saltHex, 'hex')`, which decodes what it can and silently drops the rest:

```
""      -> 0 bytes      ""     -> 23-byte preimage
"zz"    -> 0 bytes      "zz"   -> THROWS
"ZZZZ"  -> 0 bytes      "ZZZZ" -> THROWS
"abc"   -> 1 byte       "abc"  -> THROWS
        node side              console/scope-hash.mjs
```

So `''`, `'zz'` and `'ZZZZ'` hash **identically** on the node side while the console refuses two of the three. A grant carrying such a salt resolved on the bridge and raised in the console, with no error on either side — failure mode #2 in the issue, an admitted agent displayed as unadmitted. Reachable by anyone, since the salt rides publicly in the tag.

## Two exports, because issue and verify need opposite things

| | behaviour | used by |
|---|---|---|
| `scopeHashSync` | throws on a malformed salt | the three issuer sites, where the salt is 16 CSPRNG bytes and a bad one is our own bug |
| `scopeHashOrNull` | returns `null` | the three wire-facing sites, where a bad salt is hostile input and must mean "matches nothing", never an exception out of an event handler |

`src/bridge.mjs` had **one** helper serving both — `:1098` reads a wire salt, `:1223` issues with a locally generated one. Split, so neither site inherits the other's requirement.

This is item 3 of the issue, adopting the console's stricter rule deliberately.

## Why two copies remain, and what holds them together

`console/scope-hash.mjs` stays. The deploy ship list is `src tests tools package.json …` and does **not** include `console/` (`deploy/deploy-runner.sh:63`, `deploy/deploy.sh:36`), so shipped code importing from there fails to load on the box with `ERR_MODULE_NOT_FOUND`. It is also the wrong direction — nothing under `src/` is served to a browser. This is the arrangement `console/consent-vocabulary.mjs` already documents.

So six copies become two, and `tests/scope_hash.mjs` binds them: both are compared against a longhand third implementation written out in the test file, against each other byte-for-byte at the preimage, and against the live vector.

Item 2 of the issue is respected — the four `tests/` copies keep recomputing independently.

## Proof, not a green run

Three drifts applied to `src/scope_hash.mjs` and reverted:

| drift | `scope_hash.mjs` | `nipda_admission.mjs` |
|---|---|---|
| label `v1` → `v2` | 8 fail | 12 fail |
| `0x00` separator → `0x01` | 7 fail | 12 fail |
| `scopeHashOrNull` forced to `null` | 3 fail | 12 fail |

The third matters most: it is the one a "safe" refactor would introduce, and it fails the negative controls rather than the refusal assertions.

**The live vector recomputes unchanged** — grant `01a41ce2…`, signed by `tools/grant.mjs` on 2026-08-09 and readable on the relays, now hashed through the new module. That is what shows the refactor did not move any real hash, and it is the only check that would survive every copy drifting together.

New at the bridge level: a grant with a malformed salt is ignored **and does not throw**, for three salts, paired with the existing "valid grant admits the participant" as the negative control — a handler that ignored everything would satisfy all six otherwise.

Full suite green, 0 failures.

## Found on the way, filed separately

`tools/relay-invite.mjs:53` imports `../console/nip98.mjs`, and `tools/` ships while `console/` does not. Reproduced against a simulated ship-set tree: `ERR_MODULE_NOT_FOUND`. Not touched here.

---

Substantive — it changes the grant admission path in `src/bridge.mjs` and the consent verify path. Please read the issuer/verifier split before merging; that is where I would look for a mistake. @My Dude

### DIFF
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -42,6 +42,7 @@ import { bridgePubkey, bridgeSignerMode, hasBridgeKey, openSeal, openRumor, seal
 import { verifyConsent } from './consent.mjs'   // in-door consent (#131/#132, docs/CONSENT.md §8)
 import { consentState } from './consent_state.mjs'   // one honest word per watched author (#389)
 import { createHash, randomBytes } from 'node:crypto'
+import { scopeHashSync, scopeHashOrNull } from './scope_hash.mjs'
 import { readFileSync, writeFileSync, appendFileSync, existsSync, mkdirSync, renameSync, unlinkSync, openSync, closeSync, fsyncSync } from 'node:fs'
 import { dirname, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
@@ -832,10 +833,10 @@ const NIPDA = (() => {
   const d = { grant: 440, revocation: 441, index: 10440, dataset: 30440, scopeTag: 'da-scope', capTag: 'da-cap' }
   try { return { ...d, ...JSON.parse(readFileSync(NIPDA_PATH, 'utf8')) } } catch { return d }
 })()
-const scopeHash = (channelId, saltHex) =>
-  createHash('sha256').update(Buffer.concat([
-    Buffer.from('waggle/da-scope/v1'), Buffer.from([0]), Buffer.from(String(channelId)), Buffer.from(saltHex, 'hex'),
-  ])).digest('hex')
+// Two sites, opposite requirements — see src/scope_hash.mjs. The grant handler below reads a
+// salt off the wire and must ignore what it cannot decode; buildConsentPrefill issues with a salt
+// this process just generated, where an undecodable salt is our own bug and must not be swallowed
+// into a published tag.
 // grantee pubkey -> Map(grantId -> grantor). A SET of grants, not one, and that is the whole of
 // #364: a grantee may legitimately hold several — one per channel, per capability, or simply
 // re-issued. This map used to hold the last one written, so revoking THAT id de-admitted a key
@@ -1094,7 +1095,10 @@ function processGrantEvent(ev, { emitFn = emit, queryFn = query } = {}) {
   if (!grantee || !scope || !cap) return
   // The salted hash binds the grant to a channel without the channel id riding publicly;
   // the bridge knows its own channel id, so it recomputes and compares.
-  if (scope[1] !== scopeHash(PUB.inbox, scope[2] || '')) return // scoped to some other channel — not ours
+  const ourScope = scopeHashOrNull(String(PUB.inbox), scope[2] || '')
+  // null = the salt on this grant is not even-length hex, so it decodes to no subject and binds
+  // to nothing. Ignored like any other foreign scope, and never thrown out of the handler.
+  if (ourScope === null || scope[1] !== ourScope) return // scoped to some other channel — not ours
   if (cap !== 'admit' && cap !== 'admit+read') return
   // Already revoked, and we simply saw the two events in the order the relay chose to serve them.
   // Refused loudly rather than silently: admitting here is what seated a revoked key in the roster,
@@ -1219,7 +1223,7 @@ function buildConsentPrefill() {
   const salt = randomBytes(16).toString('hex')
   return {
     kind: 440, created_at: Math.floor(Date.now() / 1000),
-    tags: [['p', BRIDGE_PK], ['da-scope', scopeHash(PUB.mirrorConsentHiveId, salt), salt], ['da-cap', 'mirror'], ['tos', PUB.mirrorExpectedTosHash || '']],
+    tags: [['p', BRIDGE_PK], ['da-scope', scopeHashSync(String(PUB.mirrorConsentHiveId), salt), salt], ['da-cap', 'mirror'], ['tos', PUB.mirrorExpectedTosHash || '']],
     content: '',
   }
 }
--- a/src/consent.mjs
+++ b/src/consent.mjs
@@ -15,20 +15,16 @@
 // granted. Only the grantor may revoke their own consent.
 
 import { verifyEvent } from 'nostr-tools/pure'
-import { createHash } from 'node:crypto'
+import { scopeHashOrNull } from './scope_hash.mjs'
 
 export const KIND = { grant: 440, revocation: 441 }
 export const CONSENT_CAP = 'mirror'
 
-// Identical construction to tools/grant.mjs and src/bridge.mjs (scopeHash). If this drifts, a
-// consent scoped to a community will never match the community's own recomputed hash and every
-// consent silently fails closed — so it is written to mirror them literally.
-export function scopeHash(communityId, saltHex) {
-  return createHash('sha256').update(Buffer.concat([
-    Buffer.from('waggle/da-scope/v1'), Buffer.from([0]),
-    Buffer.from(String(communityId)), Buffer.from(saltHex, 'hex'),
-  ])).digest('hex')
-}
+// The sixth hand-written copy, and one #328 did not list. Kept as an export because the suite
+// imports it; the construction now comes from src/scope_hash.mjs so it cannot drift from the
+// issuer. verifyConsent reads its salt off a candidate 440, so a salt that is not even-length hex
+// returns null and matches nothing rather than throwing.
+export const scopeHash = (communityId, saltHex) => scopeHashOrNull(String(communityId), saltHex)
 
 const tagVal = (ev, k) => (ev.tags.find(t => t[0] === k) || [])[1]
 
--- a/src/scope_hash.mjs
+++ b/src/scope_hash.mjs
@@ -1,0 +1,76 @@
+// scope_hash.mjs — the one node-side construction (#328).
+//
+// A NIP-DA grant never names its subject publicly. It carries a salted hash instead:
+//
+//     sha256( "waggle/da-scope/v1" || 0x00 || <subject> || <salt> )
+//
+// with a fresh 16-byte salt per grant, so two grants over the same subject are not linkable by
+// anyone watching the relays.
+//
+// WHY THIS FILE EXISTS. Five node copies wrote that preimage out by hand:
+//
+//     tools/grant.mjs:168 (issue) and :217 (verify), tools/mint-consent.mjs:87,
+//     tools/propose-admission.mjs:59, src/bridge.mjs:835
+//
+// Both drift directions are silent and both still produce a hash. If the ISSUER drifts, every
+// grant it signs binds to a subject nothing can match — it verifies, it goes live, it admits
+// nobody, forever. If a READER drifts, live grants stop resolving in that surface while remaining
+// valid everywhere else, so a console shows an admitted agent as unadmitted. Neither raises an
+// error, and one hash looks exactly like another.
+//
+// WHY NOT ONE FILE FOR THE WHOLE PROJECT. `console/scope-hash.mjs` is the browser copy and cannot
+// be collapsed into this one: the deploy ship list is `src tests tools …` and does NOT include
+// `console/` (deploy/deploy-runner.sh:63, deploy/deploy.sh:36), so shipped code importing from
+// there would fail to load on the box with ERR_MODULE_NOT_FOUND. It is also the wrong direction —
+// nothing under `src/` is served to a browser. So the project keeps two copies on purpose, the
+// same arrangement `console/consent-vocabulary.mjs` documents, and `tests/scope_hash.mjs` holds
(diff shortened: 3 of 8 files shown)


JAFairweather:
## Review — approve. Every site is on the correct side of the split. The gap is that the split is *tested* at one of its three wire sites, and the untested one is the site you almost missed.

Reviewed at `be5ffce12dc8b2e4fcda07d4cb15c76d59ff36aa`, 1 ahead / 0 behind `origin/main` (`0efe7cd1`).

**82 of 83 suites pass.** `install_state.mjs` fails identically on `origin/main`, so it is not yours. (First run gave me 6 failures — a fresh worktree with no `node_modules` and no `config.json`. Provisioned, they went green. Recording it because "5 suites regress" was the wrong answer and the environment was the whole of it.)

## 1. The split, checked by salt provenance rather than by the label

I classified each site by where its salt comes from, then compared against which export it calls. All seven agree.

| site | salt origin | export | correct |
|---|---|---|---|
| `tools/grant.mjs:169` | `randomBytes(16)` | `scopeHashSync` | ✔ |
| `tools/mint-consent.mjs:87` | `randomBytes(16)` | `scopeHashSync` | ✔ |
| `tools/propose-admission.mjs:59` | `randomBytes(16)` | `scopeHashSync` | ✔ |
| `src/bridge.mjs:1226` | `randomBytes(16)` | `scopeHashSync` | ✔ |
| `src/bridge.mjs:1098` | `scope[2]` — wire | `scopeHashOrNull` | ✔ |
| `tools/grant.mjs:219` | `tag[2]` — wire | `scopeHashOrNull` | ✔ |
| `src/consent.mjs:27` → `verifyConsent:59` | `scope[2]` — wire | `scopeHashOrNull` | ✔ |

**No fail-open at any verify site**, which was the first thing I went looking for — a comparison shaped `if (computed && computed !== expected) reject` would let `null` skip the check entirely. All three are written the other way: `bridge.mjs:1098` rejects `null` explicitly, `grant.mjs:219` requires `recomputed !== null`, and `consent.mjs` compares `null !== scope[1]`, which is a mismatch. Fails closed in every direction.

## 2. The finding — one of the three wire sites is pinned, and it is not the dangerous one

`src/consent.mjs:27` can be flipped to `scopeHashSync` and **all 83 suites stay green**. So can `tools/grant.mjs:219`. Only `bridge.mjs:1098` is defended.

That matters most at consent, because `verifyConsent` is called from inside a loop with no `try` (`consent.mjs:85`) and from `processConsentEvent` at `bridge.mjs:1167`, which the socket handler calls directly at `:3922`. Feeding `readConsents` a two-event batch — one valid consent, one 440 carrying `da-scope` salt `'zz'`:

```
                          HEAD (scopeHashOrNull)        wrong side (scopeHashSync)
valid consent             ok, 1 live consent            ok, 1 live consent
hostile salt 'zz'         ok, 1 live consent            THREW OUT OF readConsents
hostile salt (odd len)    ok, 1 live consent            THREW OUT OF readConsents
```

One unsigned-cost hostile 440 takes down **the valid consent sitting in the same batch**, and does it out of an event handler. That is precisely the failure this split exists to prevent, at the site you flagged as the sixth copy nobody listed — and nothing in the suite would tell you if it moved.

The test that closes it passes on HEAD as written:

```js
// a hostile salt must not take the batch down with it
const good = mk('00112233445566778899aabbccddeeff', scopeHash(HIVE, '00112233445566778899aabbccddeeff'))
const bad  = mk('zz', 'f'.repeat(64))
const r = readConsents([good, bad], { bridgePubkey: BRIDGE, communityId: HIVE })
assert.equal(r.consent.size, 1, 'a 440 with an undecodable salt must not throw out of readConsents')
```

Worth the same for `grant.mjs:219`'s `matchesScope`, where a throw kills the listing rather than the lane.

## 3. Why that is the whole enforceable surface — the issuer half cannot be tested at all

Flipping `grant.mjs:169` the other way, issuer → `scopeHashOrNull`, is also NOT CAUGHT. That one is not a test gap: with a CSPRNG salt the two exports return the same string, so the mutation is inert and there is nothing to catch. The same holds at all four issuer sites.

Which sharpens where the guard rail has to sit. An issuer site on the wrong side is harmless; a **wire** site on the wrong side is the bug you were asking about. There are exactly three, and #431 tests one.

## 4. Smaller things, all checked and all fine

- **The `s === ''` carve-out** (`scope_hash.mjs:40`) is the one input where the new decoder is deliberately laxer than strict, and no legitimate caller reaches it — only `scope[2] || ''` at two wire sites. I went after it as a forgery path and it does not open one: `bridge.mjs:1059` gates every 440 on the grantor allowlist before the scope is read, and `consent.mjs:58` rejects a saltless scope outright. `console/scope-hash.mjs:33-38` is **byte-identical** to the node copy including the carve-out, so your parity claim holds exactly, and removing it is caught by `tests/scope_hash.mjs`.
- **The subject side got quietly wider.** `Buffer.from(subject)` threw on a non-string; `String(subject)` coerces, so `undefined` now hashes as the literal `"undefined"` — a valid-looking grant binding to nothing, the silent issuer drift your own header warns about. It is **unreachable**: every issuer validates first and exits (`channelId:124`, the npub/HEX64 branch at `:156`, `mint-consent.mjs:40`, `propose-admission.mjs:50`), and `PUB.mirrorConsentHiveId` is 64-hex-or-`null` at `bridge.mjs:464` behind truthy guards. Nothing to fix; worth a line in the header, since the old code's throw was doing work the new code delegates to four callers.
- **The `null` guards are documentation, not logic.** Dropping `ourScope === null ||` entirely at `:1098` is uncaught and inert — `null` never equals a hex string, so the comparison already rejects. Inverting it *is* caught. Correct either way; the safety does not rest on the check.

## 5. Mutations

Seven of mine, each anchor-asserted, every file restored byte-identical. I did not re-run your three.

| mutation | caught |
|---|---|
| `bridge:1098` null guard **inverted** (fail-open — a bad salt admits) | `nipda_admission` |
| `bridge:1098` wire verifier → `scopeHashSync` | `nipda_admission` |
| `scope_hash:40` empty-salt carve-out removed | `scope_hash` |
| `consent.mjs:27` verify path → `scopeHashSync` | **NOT CAUGHT** — §2 |
| `grant.mjs:219` list filter → `scopeHashSync` | **NOT CAUGHT** — §2 |
| `grant.mjs:169` issuer → `scopeHashOrNull` | NOT CAUGHT — inert, §3 |
| `bridge:1098` null guard dropped entirely | NOT CAUGHT — inert, §4 |

**Not proven by me:** anything live. Your live vector — the 2026-08-09 grant recomputing unchanged — is the claim I most wanted to exist and I did not re-fetch it; it stays yours.

#430 is `tests/dm_relays.mjs` only, +46/-2. Tests-only, exempt, agreed.

JAFairweather:
Reviewed at `be5ffce1`. **Approve.** The collapse is correct, the two remaining copies really are bound, and the salt fix does what it claims. Three findings, none blocking, and one of them is the same omission twice.

## Verified rather than read

- **The binding claim is real.** Mutated `console/scope-hash.mjs` LABEL `v1` → `v2` and `tests/scope_hash.mjs` fails 17 assertions. That is the load-bearing claim of "two copies on purpose" and it holds.
- **`bridge.mjs:1101` fails closed.** `if (ourScope === null || scope[1] !== ourScope) return` — the null is checked explicitly rather than relying on `null !== string`, so it survives someone later making the comparison loose.
- **`verifyConsent` orders it correctly** — `verifyEvent` before the scope comparison, so a forged 440 never reaches the hash.
- **The empty salt is deliberate, not a gap.** `ok — and an EMPTY salt is well-defined, not malformed — it must keep working, not become null`. I went looking for that as a defect and found it asserted, which is the right outcome.
- Full suite: only the two pre-existing `install_state` failures, which fail identically on clean `origin/main` in a fresh worktree here.

## 1 & 2 — the single source of truth misdescribes its own inventory, twice

The header of `src/scope_hash.mjs` lists **five** node copies:

```
tools/grant.mjs:168 (issue) and :217 (verify), tools/mint-consent.mjs:87,
tools/propose-admission.mjs:59, src/bridge.mjs:835
```

`src/consent.mjs` is missing — the sixth copy this PR's own body calls out as the one #328's list missed. And the `scopeHashOrNull` comment repeats it:

> For the **two** callers whose salt arrives OVER THE WIRE — `tools/grant.mjs list --agent/--channel` and the bridge's grant handler

There are three. `src/consent.mjs:27` is the third, and the PR body's table says three. So the file that exists because "there is no test that can catch a drifted copy except one that compares the copies" carries an inventory that is wrong in both places a person would check before adding a seventh.

## 3 — the construction is unified, the callers still disagree about a missing salt

`verifyConsent` refuses outright:

```js
if (!scope || !scope[2]) return { ok: false, reason: 'no salted scope' }
```

`bridge.mjs:1098` substitutes for it:

```js
const ourScope = scopeHashOrNull(String(PUB.inbox), scope[2] || '')
```

`|| ''` reaches the empty-salt branch the tests pin as well-defined, so it hashes and compares. A 440 whose `da-scope` tag lost its salt element is therefore **refused on the consent path and evaluated on the admission path** — two surfaces, two answers for the same malformed grant, which is the shape #328 exists to end.

Scoping it honestly: not attacker-reachable. `verifyEvent` runs first and `processGrantEvent` rejects any 440 whose author is not a maintainer grantor (`:577`), so producing one needs a grantor to sign a grant with no salt — our own bug, not hostile input. I could not build an exploit and do not think one exists.

But the file's own header says "a fresh 16-byte salt per grant", so zero bytes is a length the issuer never emits, and `scope[2] || ''` converts *absent* into *legitimately unsalted* rather than *malformed*. If empty is genuinely meaningful, consent is wrong to refuse it; if it is not, `bridge.mjs:1098` should pass `scope[2]` and let a missing salt be a non-match. Either is fine — the two disagreeing is the thing worth closing while this file is being written.

PR #433, opened 2026-08-13
Shipped code imported console/, which the deploy does not ship (#432)

### LINKED ISSUE #432: tools/relay-invite.mjs imports from console/, which the deploy does not ship — ERR_MODULE_NOT_FOUND on the box
`tools/relay-invite.mjs:53` imports `../console/nip98.mjs`. `tools/` is in the deploy ship list and `console/` is not, so the tool cannot load in the deployed tree.

Both deploy paths ship the same set and neither includes `console/`:

```
deploy/deploy-runner.sh:63  SHIP='src tests tools package.json package-lock.json config.example.json deploy/…'
deploy/verify-deployed.sh:35 SHIP='…'   (same)
deploy/deploy.sh:36          rsync -az "$DIR/src" "$DIR/tests" "$DIR/tools" …
```

Reproduced against a tree containing only the ship set:

```
IMPORT FAILS: ERR_MODULE_NOT_FOUND | Cannot find module '…/shiptree/console/nip98.mjs'
              imported from '…/shiptree/tools/relay-invite.mjs'
```

## Severity

Latent, not a live outage. `relay-invite.mjs` takes a local key path (`--key ~/.nvoy/owner.nsec`) and reads like a maintainer-machine tool, so it is probably never run from the box. But `tools/` ships, nothing marks this one as off-box, and the failure is `ERR_MODULE_NOT_FOUND` rather than anything that says "run this from a checkout" — so the first person to try it on the box loses time to a resolver error.

I have not checked the actual box. The claim is about the ship list, which is the definition of what lands there.

## The general rule this is an instance of

Nothing enforces that shipped code only imports shipped code. #328 hit the same wall from the other side: pointing `src/bridge.mjs` at `console/scope-hash.mjs` would have been a clean-looking refactor that broke the bridge on the next deploy, minutes after merge. It was caught by reading the ship list, not by any check.

A test that walks the `import` statements in `src/` and `tools/` and asserts every relative target resolves inside the SHIP set would catch both, and would have caught this one when it landed.

Two ways to fix `relay-invite.mjs`, and they are not equivalent:

1. move the shared NIP-98 builder under `src/` and have `console/nip98.mjs` become the browser copy, bound by `tests/nip98_auth.mjs` the way `tests/scope_hash.mjs` binds the two scope-hash copies (#328)
2. add `console` to the ship list — which is a deploy-surface decision and overlaps #412

Related: #328, #412.

JAFairweather:
Closes #432.

`tools/relay-invite.mjs` imported `../console/nip98.mjs`. `tools/` is on the deploy ship list and `console/` is not, so the tool could not load in the deployed tree. The failure is `ERR_MODULE_NOT_FOUND` on a resolver path that says nothing about a deploy set, so the first person to run it from the box loses the time to work that out.

## A move, not a fork

Nothing under `console/` ever imported that module. Its only two importers were `tools/relay-invite.mjs` and `tests/nip98_auth.mjs`, both Node. The module's own header argued it lived in `console/` "because the browser is one of its two callers" — that premise was false.

So it becomes `src/nip98.mjs`, one copy, no second one to drift. The old header was right that two copies of a security-relevant builder is how two copies drift apart, which is why there still is only one. If a console page ever does need it, the arrangement to copy is `console/scope-hash.mjs` (#328): a browser copy bound by a test, not a cross-boundary import in either direction. The header now records that.

## The general control

`tests/ship_imports.mjs` (suite 84). Every relative `import`/`export … from` and `new URL(…, import.meta.url)` target under `src/` and `tools/` must resolve inside the ship list. The list is parsed out of `deploy/deploy-runner.sh` and `deploy/verify-deployed.sh` rather than restated — a copy here would be a fourth place it lives and the one that never gets updated — and the two scripts are asserted to agree, because if they diverge "shipped" has two meanings and the suite is testing the wrong one.

This would have caught #432 when it landed, and it covers the near-miss from the other direction: #328 nearly pointed `src/bridge.mjs` at `console/scope-hash.mjs`, which was caught by reading the ship list by hand.

## Why the walk stops at src/ and tools/

Measured, not assumed. Building a real ship tree with `git archive` and resolving every relative target gives 30 that do not resolve:

| dir | count |
|---|---|
| `src/` | 0 |
| `tools/` | 1 — this bug |
| `tests/` | 29, across 9 suites |

The `tests/` ones are deliberate: they import `console/` to bind a browser copy to its node twin. They do not resolve on the box either, and that is fine — the deploy installs with `npm ci --omit=dev` and never runs the suite there, so nothing on the box loads them. Extending the walk would report 29 findings, none actionable, which is how a check stops being read. The suite header records this so a later reader does not mistake the exclusion for an oversight, and says what would change if the deploy ever did run tests on the box.

## Evidence

Full suite **84/84, exit 0, 2603 assertions**.

Negative control on the real tree: copy the module back to `console/`, repoint the import, and the new suite fails — naming the file and the specifier, not merely refusing.

```
FAIL — no module under src/ or tools/ imports outside the ship list
     tools/relay-invite.mjs -> ../console/nip98.mjs
```

Both synthetic controls are in the suite, because "nothing found" is also what a broken extractor reports: three cross-boundary forms are detected and classified off-boundary, and two ordinary in-tree imports still pass — so the rule refuses the boundary rather than refusing every import. There is a size floor on the walk (≥60 modules, found 76) and it exits 3 INCONCLUSIVE rather than 0 if it cannot read the ship list or the walk comes back too small.

Two environment notes for whoever runs this locally: a fresh worktree needs `npm ci` and a `config.json`, or `install_state` and `read_resilience` fail for reasons that have nothing to do with the change. Confirmed pre-existing by running them on the unmodified worktree first.


JAFairweather:
## Review — approve the move; the guard rail needs one commit first. Your `tests/` exclusion is the right shape, and I verified the premise it rests on. The coverage gap is in the two directories the suite *does* scan.

Reviewed at `fec9e68aeef56fa72741ca5131398525648ae0d2`, 2 ahead / 0 behind `origin/main`.

**83 of 84 suites pass.** `install_state.mjs` fails identically on `origin/main` — same pre-existing local failure I hit on #431, not yours. You report 84/84; the delta is this machine.

## 1. The move is right, and it is a move

Checked against `origin/main` rather than taking the claim: the only importers of `console/nip98.mjs` were `tools/relay-invite.mjs:53` and `tests/nip98_auth.mjs:14`. **Nothing under `console/` ever imported it** — no page, no importmap. The header's "the browser was one of its two callers" was false, and `src/` is where it belongs. One copy after, none left behind, and the suite asserts both.

## 2. Your question — the `tests/` exclusion

**It is the right shape, and I would not widen it.** I checked the premise it rests on rather than accepting it: no script under `deploy/` runs the suite on the box — not `deploy.sh`, not `deploy-runner.sh`, not `verify-deployed.sh`. So the 29 unresolvable imports are inert there, they are deliberate browser↔node bindings, and reporting them is how a check stops being read. The header is the right place for that reasoning and it will still be true when someone asks in a year.

The problem is not the directory you excluded.

## 3. The finding — the extractor is line-anchored, and four of five cross-boundary shapes walk straight past it

`RELATIVE_FROM` requires the specifier to sit on the same line as its `import`/`export`, at the start of a line. I injected a real `console/nip98.mjs` import into `src/` in each shape and ran your suite:

| shape | result |
|---|---|
| single-line `import … from '../console/nip98.mjs'` | CAUGHT (suite fails) |
| multi-line braces | **hidden — suite still passes** |
| newline before `from` | **hidden — suite still passes** |
| `await import('../console/nip98.mjs')` — *literal* specifier | **hidden — suite still passes** |
| multi-line `export … from` | **hidden — suite still passes** |

This is not hypothetical. Three relative imports on the tree **right now** are outside the check's view:

```
src/bridge.mjs:59-61        multi-line   -> ./buzz_policy_client.mjs
tools/waggle-init.mjs:34-35 multi-line   -> ../src/install_state.mjs
tools/approve.mjs:62        await import -> ../src/bridge.mjs
```

All three happen to point inside the ship list, so **your verdict is correct** — but it is asserted over 77 of 81 targets, and the shapes that hide are ordinary style already in use here. A future `console/` import written the way `bridge.mjs:59` is written passes silently. That is the exact defect class #432 exists to close, in the check built to close it.

One header correction: the stated limit is that "dynamic `import(expr)` with a computed specifier is not covered and cannot be, statically." True, but narrower than the real gap — `approve.mjs:62` is a dynamic import with a **literal** specifier, which is statically extractable and is missed.

### The fix, tested before proposing it

`eslint` is already a devDependency, so its parser is present, and the suite only ever runs in dev — the same `npm ci --omit=dev` reasoning your header uses to justify the `tests/` exclusion applies here in your favour. A ~15-line espree walk replaces both regexes:

```js
import { parse } from 'espree'
// ImportDeclaration / ExportNamedDeclaration / ExportAllDeclaration -> node.source.value
// ImportExpression with a Literal source -> dynamic import(), literal specifier
// NewExpression URL(Literal, import.meta.url) -> the file-read form
```

```
                        regex     espree
relative targets found     77         81
off-boundary                0          0     <- your verdict, independently confirmed
all six shapes above    1 of 6      6 of 6
```

Every existing control survives, and the synthetic-good/synthetic-bad pair still does its job.

## 4. The ship list lives in three places; the suite asserts two

Reading the list from the deploy scripts instead of restating it is the right instinct, and the "if they diverge, this whole suite is testing the wrong one" check is the right check. But `deploy-runner.sh:60` says it itself:

> Ship list — MUST mirror **deploy.sh's rsync set** and verify-deployed.sh's SHIP.

`shipListFrom` reads the two that spell it `SHIP='…'`. The third, `deploy/deploy.sh:36-40` and `:41-44`, spells the same set as two inline rsync literals, so it is invisible to the reader and unasserted. The sets are **identical today** — I diffed them — so nothing is broken. But `deploy.sh` is the documented push path from the Mac, and the failure it permits is: someone adds a path to both `SHIP=` lists, this suite passes because those two agree, and the push deploy silently does not ship it. `ERR_MODULE_NOT_FOUND` on the box, by the one deploy path the check does not read.

## 5. Adjacent, not a blocker — a third hand-written NIP-98

`tools/publish_relay_list.mjs:85-93` builds the whole construction by hand: kind 27235, the same `u`/`method`/`payload` tags, and `'Nostr ' + base64(JSON)`. It is in `tools/`, so it could import `../src/nip98.mjs` today with no boundary question at all. `ship_imports.mjs` cannot see it precisely because it imports nothing. Same class as #328 and #432, one file over. Worth an issue.

## 6. Lint

Every claim verified. `npm run lint` is `eslint src tools tests` — **`console/` is genuinely unlinted**, so your read is right and it does deserve its own issue alongside #412. `btoa` is really used, at `src/nip98.mjs:92`, and the comment explaining why it encodes UTF-8 bytes rather than the raw string is the kind of note that stops someone "simplifying" it. Negative control reproduced: an undefined identifier in `src/` still trips `no-undef`.

**Not proven by me:** anything on a box. No deploy run, no `npm ci --omit=dev` tree. That the tool now *loads* in the deployed set is still an inference from the ship list, not an observation — yours to close if you want it closed.

Probe files removed, tree byte-identical, `git status` clean.

JAFairweather:
Reviewed at `fec9e68a`. The fix is right and the move is clean. **The new control does not cover the import forms this codebase actually uses** — I got a live cross-boundary import past it in two different ways. Worth closing before the suite is trusted, since being trusted is its whole job.

## The control misses multi-line and dynamic imports

Same crossing, three spellings, in `tools/grant.mjs` on this branch:

```
import { scopeHash } from '../console/scope-hash.mjs'
  → FAIL — no module under src/ or tools/ imports outside the ship list
         tools/grant.mjs -> ../console/scope-hash.mjs

import {
  scopeHash,
} from '../console/scope-hash.mjs'
  → SHIP IMPORTS PASS

const { scopeHash } = await import('../console/scope-hash.mjs')
  → SHIP IMPORTS PASS
```

That is #432 exactly — a shipped file reaching into `console/` — and the suite written to catch #432 reports the tree clean. (Not in `relay-invite.mjs`, which is separately pinned by the `includes` checks at the bottom; the general walk is what I was testing.)

`RELATIVE_FROM` is anchored to line start, and `[^\n]*?from` cannot cross a newline, so in a multi-line statement the `from` sits on a line beginning `} from …` and never matches. Feeding forms directly to the exported `relativeTargets`:

```
SEEN    single-line static import
MISSED  MULTI-LINE static import
MISSED  MULTI-LINE export-from
MISSED  dynamic import, literal, in expression
MISSED  dynamic import, literal, leading await
SEEN    dynamic import, literal, bare line-start
SEEN    new URL in an expression
```

Two consequences.

**The header overstates coverage.** It says dynamic `import(expr)` "with a computed specifier is not covered and cannot be, statically" — correct, but it reads as though a *literal* specifier is covered. It is not, unless the line happens to begin with `import(`. `tools/approve.mjs:62` is `const { forwardPublic, … } = await import('../src/bridge.mjs')` and is not walked.

**This is live coverage loss, not a hypothetical form.** In `src/` and `tools/` today there are 5 multi-line import statements, 3 with relative targets — `src/bridge.mjs -> ./buzz_policy_client.mjs`, `tools/waggle-init.mjs -> ../src/install_state.mjs` — plus the `approve.mjs` dynamic one. 13 files repo-wide use the multi-line style, so it is house style rather than something a contributor would have to go out of their way to write.

To be fair about severity: every one of those targets is inside the ship list, so **no current defect is hidden** and the tree really is clean. What is missing is the guarantee. The suite's claim is "this would have caught #432 when it landed" — true only for the spelling #432 happened to use, and a reviewer reading `SHIP IMPORTS PASS` cannot tell which spelling was in play.

Dropping the line anchor and matching `from ['"](\.…)['"]` plus `import\(\s*['"](\.…)['"]` covers both, at the cost of matching inside comments and strings — which for this check errs toward a false FAIL that a human reads, rather than a false PASS that nobody does. Your call on the shape; the synthetic controls at the bottom are the right place for the new forms, so the next reader can see they are covered.

## Minor

`check(JSON.stringify(runnerShip) === JSON.stringify(verifyShip), …)` prints `ok` when **both** reads returned null, since `'null' === 'null'`. The two checks above it print FAIL and line 55 exits 3, so nothing is concealed — but "the two SHIP lists are identical" is the wrong sentence to print about two lists that were never read. Move it below the guard.

## Verified rather than read

- **It is a move, not a fork.** Diffed `origin/main:console/nip98.mjs` against `src/nip98.mjs` — 16/-6, every line a comment. The builder is byte-identical.
- **Single-line detection names the file and the specifier**, not just a refusal, which is what makes the failure actionable.
- **The synthetic controls are real controls** — they fail on purpose and pass on purpose, so a zero is a clean tree rather than a dead extractor. That is the reason I went looking for uncovered forms rather than assuming there were none.
- **Both size floors and the exit-3 INCONCLUSIVE paths are present**, on the ship-list read and on the walk.
- **Reading the ship list out of the deploy scripts instead of restating it** is the right call, and asserting the two scripts agree is the part that makes it safe.
- Registered: `package.json` references 84 suites.

PR #435, opened 2026-08-13
Lint console/, where no-undef had never run (#434)

### LINKED ISSUE #434: console/ is never linted: no-undef has never run on the signing UI
`npm run lint` is `eslint src tools tests`. `console/` is not in it, and never has been. So the one rule `eslint.config.mjs` describes as load-bearing — `no-undef`, because "`node --check` has passed on code whose identifiers did not exist" — has never run on the console.

Found while moving `console/nip98.mjs` to `src/` for #432. That move put the file under the linter for the first time and immediately found `btoa` missing from the globals list. The module had been in the tree, unlinted, the whole time.

## What turning it on costs today

```
npx eslint console
✖ 11 problems (11 errors, 0 warnings)   — 3 files
```

All 11 are `no-undef`, and all 11 are false positives: `document` ×8, `localStorage` ×2, `confirm` ×1. The config declares one flat `globals` block listing "Node 20+ runtime surface the bridge and tools actually use". There is no browser block, because nothing browser-shaped was ever linted.

So this is not a backlog of 11 defects. It is one missing config section.

## Why it is worth doing

The console holds the signing UI. It is the surface where an identifier that does not exist fails in a page, at the moment an operator is approving something, with no test between the typo and the operator — which is a worse place for that class of bug than `src/` is, not a better one.

## Shape

Add a second config block scoped to `console/**` with the browser globals (`document`, `window`, `localStorage`, `sessionStorage`, `confirm`, `alert`, `navigator`, `location` — driven by what the files actually use, not a blanket list), keep `console/vendor/**` ignored as it already is, and extend the lint script to cover `console`.

Worth pairing with a negative control the same way #432 did: after enabling it, plant an undefined identifier in a console module and confirm the run fails, so "green" is not "not looking".

Related: #412 (`console/` is outside the deploy verification set) — same root shape, that `console/` sits outside the checks everything else is inside. Different check, so filing separately rather than folding in.

JAFairweather:
Closes #434. Stacked on #433 — that PR is what found this. Base moves to `main` when #433 lands.

`npm run lint` was `eslint src tools tests`. `console/` was never in it. So the one rule `eslint.config.mjs` describes as load-bearing — `no-undef`, because "`node --check` has passed on code whose identifiers did not exist" — had never run on the signing UI.

That is the wrong surface to leave unchecked. It is where an identifier that does not exist fails in a page, at the moment an operator is approving something, with no test between the typo and the operator.

## Cost of turning it on

11 problems, 3 files, all `no-undef`, all false positives from a Node-only globals list: `document` ×8, `localStorage` ×2, `confirm` ×1. One missing config section, not a backlog.

## Three globals, not a blanket list

Matching how the existing block lists only "the Node 20+ runtime surface the bridge and tools actually use". These three are precisely what the run reported. A name that is not on the list and not defined is the finding, which is the point.

`window` is deliberately **not** declared. Console modules reach it as `globalThis.window` (`console/signer-session.mjs:10`) so they stay importable under Node for the suite. Declaring the global would quietly make a bare `window` reference lint clean — which is the thing that breaks those imports.

Checked before writing the list: `history` and `Node` appear in `console/` only inside comments, and `navigator`, `sessionStorage`, `alert` and `location` are not used in any non-vendor `.mjs`. `console/vendor/**` stays ignored, as it already was.

## Evidence

Negative control, planted in `console/routing.mjs`:

```
export const _probe = () => notARealIdentifier987(document.body)

  239:29  error  'notARealIdentifier987' is not defined  no-undef
```

Both directions in one probe: the rule now fires on `console/`, and `document` on that same line is **not** flagged, so the new globals are what stop it firing wrongly rather than the rule being off. Restored, `npm run lint` exits 0. Full suite 84/84, exit 0.


JAFairweather:
Verified every claim in the diff. All true. One finding: the largest part of `console/` is still unlinted, and it is the part the PR's own rationale points at.

Confirmed against `origin/main`: `"lint": "eslint src tools tests"`. `console/` genuinely never ran.

The lint passes clean on the branch (exit 0), and `no-undef` actually fires there — appending a bad identifier to `console/signer-session.mjs`:

    28:30  error  'thisIdentifierDoesNotExist' is not defined  no-undef
    28:57  error  'window' is not defined                      no-undef

which also proves the `window` reasoning is load-bearing rather than decorative: a bare `window` is refused, `localStorage` on the same line is not. That is the allowlist doing exactly what the comment says.

The counts are exact. Emptying the globals block reports:

    counts: {'document': 8, 'localStorage': 2, 'confirm': 1}
    files: ['config-operations.mjs', 'routing.mjs', 'task-routes.mjs']

Three globals, three files, matching the comment digit for digit.

## `console/**/*.mjs` covers 24 files and misses 1,382 lines of inline script

The console's JavaScript is not mostly in `.mjs`. Counting inline `<script>` blocks without a `src`:

    console/index.html      ~841 lines
    console/connect.html    ~318 lines
    console/agents.html     ~175 lines
    console/following.html   ~22 lines
    console/setup.html       ~21 lines
    console/config.html       ~4 lines
    console/routing.html      ~1 line
    total                  ~1,382 lines

None of it is linted by this config, and ESLint will not read `.html` without a plugin, so the gap is silent — `npm run lint` exits 0 and reports nothing skipped.

The gap matters because of what the PR says it is for: *"the signing UI: the surface where an identifier that does not exist fails in a page, in front of an operator who is approving something, with no test in between."* `index.html` and `connect.html` are that surface — they are the two files carrying the kind 440/441 signing paths flagged in the #436 review — and between them they hold 1,159 of those 1,382 lines. The stated motivation lands squarely on the files this does not cover.

Not a reason to hold the PR. 24 files linted beats 0, and the `window` decision is worth having recorded. But the config comment reads as though `console/` is now covered, and it is not — suggest either narrowing the comment to say `.mjs` modules only, or opening a follow-up for the HTML blocks so the remaining surface is tracked rather than assumed handled. I lean toward both.

One consequence worth naming: this makes `no-undef` a merge gate for new `console/` code, so a module that starts using a browser global fails CI until the list is extended. That is the intended behaviour per the comment — flagging it because whoever hits it first will read it as a broken lint rather than a finding.

PR #436, opened 2026-08-13
Refuse to sign from a stale console page (#418)

### LINKED ISSUE #418: The console has no cache-busting story: a soft reload can keep an old signing UI alive (#412 item 3)
Split from #412. Item 1 and item 2 are settled in #417 — the console is not served from the bridge host and must not be in the ship list. This is what remains, and it is the one with teeth.

## The defect

`console/*.html` fetch their module graph by stable path. `console/routing.html` imports `routing.mjs`, which imports `task-routes.mjs`, `task-route-mention.mjs`, `task-route-envelope.mjs` and so on, all at fixed URLs with no version in them. The importmap covers only bare specifiers (`nostr-tools` and its subpaths) — local modules are not mapped at all.

So a soft reload can serve a fresh `.html` and stale `.mjs` modules. That is what #412 observed: after #404/#406 deployed, the routing form still refused `MC Claude` with a message that no longer existed anywhere in the tree, and a hard reload fixed it.

## Why it is not merely annoying

The console signs owner control commands. A stale module graph means signing against an old grammar, an old envelope shape or an old freshness rule, and there is no signal that it happened — the page renders normally and the operator has no reason to suspect the code is old. `control-state-freshness.mjs` and `confirmed-fresh-signer.mjs` exist because freshness at the signing boundary is something this project already treats as load-bearing; the code doing the signing is not covered by either.

Compounding it: the console is published by hand from a checkout on a different host (#417), so "is my page stale?" already has two possible causes — the site was not deployed, or the browser cached it — and no way to tell them apart from the browser.

## Two halves, and only one is in this repo

**Site host (operator action).** Serve `.mjs` with `Cache-Control: no-cache` so the browser revalidates rather than reusing blind. This alone fixes the observed failure. It is a static-site server config on another host, so it cannot be landed here.

**Repo side (real work, worth doing anyway).** A staleness check the operator can see, so the failure is legible instead of invisible:

1. The site deploy writes `console/VERSION.json` — `{"sha": "<commit>", "at": "<iso>"}`.
2. Each page fetches it with `cache: 'no-store'` and compares against a constant in its own module graph.
3. Disagreement puts a banner on the page naming both commits, and disables the signing controls until reloaded.

Step 3 is the part that matters and the part to get right — a warning the operator can sign through is not a guard. Same reasoning as `taskRouteMentionProblem`, where the reason is the product: "this page is old" sends someone hunting, `"page built at abc1234, server has def5678"` does not.

Worth deciding before building: whether the constant is stamped at deploy (needs the site deploy to write into the module graph, more moving parts) or is simply the shared-module byte-identity hash the suite already computes (no deploy step, but only detects staleness in the shared modules, which is where the #404 failure actually was).

## Negative control this needs

Serve an old `.mjs` alongside a current `VERSION.json` and confirm the banner appears and the controls are disabled. A staleness check that has only ever seen a fresh page proves nothing.

JAFairweather:
Closes #418.

## The defect

`console/*.html` fetch their module graph by stable path with no version in it, and the importmap covers only the bare `nostr-tools` specifiers — every local module is a plain cacheable URL. A soft reload can therefore serve a fresh `.html` against stale `.mjs`. That is what #412 observed: after #404/#406 shipped, `routing.html` still refused `MC Claude` with a message that no longer existed anywhere in the tree, and a hard reload fixed it. Nothing on the page said the code was old.

This console signs owner control commands. A stale graph signs against an old grammar, an old envelope shape or an old freshness rule, and renders normally while doing it. `control-state-freshness.mjs` and `confirmed-fresh-signer.mjs` already treat freshness at the signing boundary as load-bearing; the code doing the signing was the one thing neither covered.

## The mechanism

`console/build-id.mjs` is a module, so a browser reusing a cached graph reuses the OLD id with it. `console/VERSION.json` is fetched with `cache: 'no-store'`, so it always reports what the server holds. Disagreement means the page is running code the server has replaced.

Both files are generated by `tools/gen-console-build-id.mjs` from the bytes of every other file under `console/`, and both are committed — so publishing the checkout publishes them and the site deploy gains no step (#417: the console is published by hand from a checkout on another host).

A **content** hash, not a commit sha, on purpose. A commit sha changes on every merge, so a docs commit would tell every open console it was stale and the operator would learn to ignore the banner. The id changes when, and only when, something the browser executes changes. The path is hashed alongside the bytes, so a rename is a different id.

This resolves the "worth deciding before building" question in the issue in favour of neither option it named: no deploy-time stamping (no extra moving parts) and not restricted to shared modules (it covers the whole graph, `vendor/` included — a vendored signer swapped underneath a page is exactly this failure).

## Two halves, and only one is load-bearing

`assertConsoleFresh()` in `stableControlSigner` is the guard. The banner and the disabled controls are legibility — anything that draws a control afterwards undoes them, and, as the issue says, a warning an operator can sign through is not a guard.

Three states, never a boolean. An unreachable `VERSION.json` is INCONCLUSIVE and refused **for that reason**, not reported as stale — the operator needs to know whether to reload or to go and check whether the site was published at all. Both messages name both ids.

No `document` means no browser, which means no HTTP module cache, which means the failure mode does not exist — so Node is not refused. That is a statement about the environment, not an exemption.

`assertConsoleFresh` **awaits** the in-flight check rather than reading a snapshot, so a click that lands before the fetch returns is held, not waved through and not refused.

## Evidence

`npm test` — 84 suites, exit 0. New suite `tests/console_staleness.mjs`, 38 checks.

The negative control the issue asks for, and its opposite in every case:

```
ok   — stale -> a banner is inserted at the top of the page
ok   — stale -> the banner itself names both builds
ok   — stale -> every form control on the page is disabled
ok   — NEGATIVE CONTROL — a current page gets no banner
ok   — NEGATIVE CONTROL — a current page keeps its controls enabled
ok   — stableControlSigner REFUSES to open a signer on a stale page, naming both builds
ok   — NEGATIVE CONTROL — the same call on a CURRENT page still opens the signer
```

**The last one earned its place.** The first draft's fixture used `verified_at`; `requireFreshControlState` reads `observed_at`. The stale-refusal check above it passed anyway — a signer that refuses everything refuses stale pages too. Only the positive direction caught it.

Two mutations, each confirmed to fail the suite and then restored:

```
mutation1 (stale build-id.mjs)         exit=1  FAIL — console/build-id.mjs matches a fresh recompute
mutation2 (guard tag cut from a page)  exit=1  FAIL — every console page loads the staleness guard (missing: routing.html)
restored                               exit=0  ALL PASS
```

The suite exits **3 = INCONCLUSIVE** if the console scan returns fewer than 25 files or fewer than 5 pages: an empty input hashes to a perfectly stable value, so a broken scan and a clean tree would otherwise print the same id.

## What is NOT in this change

**The site-host half.** Serving `.mjs` with `Cache-Control: no-cache` is a static-server config on another host and fixes the observed failure on its own. This makes the failure *legible and non-signable* when it happens anyway; it does not replace that config.

**The bootstrap gap.** A graph cached from before this module existed does not run this module, so the first reload after it ships is still on the operator. Stated in the module header rather than papered over.

## Reviewer note

Suite count 84 collides with #433, which also takes 83 → 84. Whichever merges second needs a rebase of `package.json`, `CLAUDE.md` and `README.md`; the two touch no other file in common.

Worth a look at: whether refusing on INCONCLUSIVE is the right call. A network blip and a site that was never published are indistinguishable from the browser, and I chose to refuse both. The argument for it is that "absent `VERSION.json`" is itself most likely a stale snapshot, since the file is committed — but it does mean a flaky network can block signing until reload.

@My Dude


JAFairweather:
## Review — the mechanism is sound and the module is the best-argued file in the PR. But the first sentence of this PR is still true after it: a stale console module graph can sign owner control commands. The guard is wired to **one of seven** signing paths.

Reviewed at `ee4b7180404c8bba57cdd2936649a9700008fca6`, 1 ahead / 0 behind `origin/main`. **83 of 84 suites pass**; `install_state.mjs` fails identically on main — same pre-existing local failure as #431 and #433, not yours.

## Your question 2 first, because question 1 depends on it

> The DOM half disables controls that exist at check time… I did not add a MutationObserver, on the grounds that the throw at the signer is the real guard and the DOM part is legibility. Tell me if that reasoning is too comfortable.

**The reasoning is right and the premise is not.** "The throw at the signer" would carry the argument if it were at *the* signer. `staleness-guard.mjs` is imported by exactly one module —

```
console/stable-control-signer.mjs:3   import { assertConsoleFresh } from './staleness-guard.mjs'
```

— and `stableControlSigner` has exactly one caller, `console/routing.mjs:215`. Every other owner-command path reaches `consoleSigner` directly:

| where | what it signs | path | `assertConsoleFresh` |
|---|---|---|---|
| `routing.mjs:215` | task-route control | `stableControlSigner` | **yes** |
| `task-routes.mjs:88` | sealed task-route command | `consoleSigner` | no |
| `config.html` | kind 30078 `d=waggle-control` — publishing on/off | `confirmedFreshSigner` → `consoleSigner` | no |
| `following.html:52` | kind 30078 `d=waggle-watchlist` | `consoleSigner` | no |
| `agents.html:189` | kind 30078 `d=waggle-agent-lifecycle` | `consoleSigner` | no |
| `connect.html:476` | **kind 440 NIP-DA grant** (`da-scope`, `da-cap`) | `consoleSigner` | no |
| `index.html:724` | prepared template | `signer.signEvent` | no |

`confirmedFreshSigner` checks control-state freshness and then calls `signerFactory()` — it never asks about the module graph.

So on six of seven pages the half you call "legibility, not a guard" is the *only* thing between a stale graph and a signed owner command. Which makes the MutationObserver question the wrong one to be worrying about: on those pages the DOM half is load-bearing, and it is already undone by the pages' own startup code, without anything exotic rendering later. `config.html`:

```js
let stale=false; … $('toggle').disabled=stale;      // unconditionally re-enables
… let saved=localStorage.getItem('waggle-following-bridge'); if(saved){$('bridge').value=saved;load()}
```

`load()` auto-runs for any operator with a saved bridge, and its success path sets `disabled=false` on the owner-control toggle. `renderVerdict` resolves first (a same-origin `VERSION.json` beats four relay round-trips), disables it, and then `load()` turns it back on. `following.html` does the same through `setManage(true)`. No observer would help — these are not late-rendered controls, they are the same controls, re-enabled.

Two of the six are worth singling out. `task-routes.mjs:88` signs the task-route grammar — **the exact grammar that was stale in #412**, the incident this PR opens with. And `connect.html:476` signs kind 440 capability grants, which is the highest-privilege thing the console does.

**Recommendation:** move the `await assertConsoleFresh()` into `consoleSigner` in `signer-session.mjs`, or into a wrapper every path takes. One line, and the argument you made for the DOM half becomes true. Then add the assertion the suite is missing — that no console module reaches a signer without passing the chokepoint — because right now `tests/console_staleness.mjs` §7 pins the one path that has it.

## Question 1 — fail-closed on INCONCLUSIVE is right, and the argument is stronger than you made it

Keep it. The asymmetry is not close: the cost of a false refusal is one reload, and the cost of a false pass is an owner control command signed against replaced code — which is unrecoverable in the way a relay-accepted event is always unrecoverable. `verdictMessage` already tells the operator both branches ("reload, or check that the site was published"), so the flaky-network case is self-resolving in a way the stale case is not.

One thing the current coverage does to this trade, worth seeing: today the entire cost of fail-closed lands on `routing.html`, the one page that asks — while the six pages that would happily sign against a stale graph never fail closed because they never check. That is an argument for widening the chokepoint, not for softening the verdict.

## A second finding — the guard can report FRESH on exactly the scenario it was built for

The check compares one module's constant against the server. That is sound only if cache state is **uniform across the graph**, and HTTP caching is per-resource with independent lifetimes. The case we know is non-uniform is this deploy:

```
soft reload after this PR ships
  index/routing.html   revalidated on reload            -> FRESH html
  staleness-guard.mjs  new URL, never cached            -> fetched, FRESH
  build-id.mjs         new URL, never cached            -> fetched, NEW id
  VERSION.json         no-store                         -> NEW id
      => compareBuild(NEW, NEW) = 'fresh'
  routing.mjs / control-state-freshness.mjs  in cache, within lifetime  -> STALE
```

Fresh HTML against stale JS is your own description of the failure. In that state the new files are fetched *because they are new*, they agree, and the verdict is `fresh` while the page runs replaced code.

Your header states a bootstrap gap, but as "a graph cached from before this module existed does not run this module — the first reload is still on the operator." That is the case where the HTML is *also* cached. In the case your problem statement actually describes, the module runs and returns the wrong answer, which is worse than absent. The same non-uniformity returns later through ordinary per-entry cache eviction, though rarely.

**I have not measured this against the live host** — it is an argument from how per-resource caching works, not an observation, and it depends on headers I cannot see from here. Which leads to the thing worth deciding deliberately:

`tools/serve-console.mjs:75` already serves every file with `Cache-Control: no-store`, so under this repo's own server the failure cannot occur at all. `console/` is not on the deploy ship list (#433), so the host that served the stale copy in #412 is outside this repo. **If its cache headers are controllable, `no-cache` on `console/*.mjs` removes the condition rather than detecting it** — and content-hashed module URLs remove it permanently. Detection is the right call when you do not control the host; it is worth one line in the header saying that is why, and it connects to #417.

## What is good, and I mean it

- Three states, never a boolean, with "could not ask" explicitly not "fine" — the same discipline as exit 3 elsewhere in this repo.
- `assertConsoleFresh` **awaits the pending promise** rather than reading a snapshot, so an operator who clicks before the fetch lands is held rather than waved through in either direction. That is the subtle one and it is right.
- The `!doc` carve-out stated as a claim about the environment rather than an exemption.
- `verdictMessage` naming both ids. That is the difference between an operator reloading and an operator opening a bug.

**Not proven by me:** any browser. No cache state, no soft reload, no live host. Everything above is the module, the call graph, and the suite.

Noted on the suite-count collision with #433 — agreed, and #433 also touches `package.json`/`CLAUDE.md`/`README.md` only for the count.

JAFairweather:
Reviewed at `ee4b7180`. The mechanism is right and the three-state verdict is the correct shape. **The load-bearing half is wired into one of the console's five signing paths, and the four it misses include the ones that sign 440s.**

## `assertConsoleFresh()` is reachable from exactly one signing path

`stableControlSigner` is called from one place in the console:

```
console/routing.mjs:23   import { stableControlSigner } from './stable-control-signer.mjs'
console/routing.mjs:215  const opened = await stableControlSigner(bridge, state, …)
```

Every other page signs through `consoleSigner` (`signer-session.mjs`) or builds a signer straight from `vendor/nave-connect.mjs`. Transitive import closure from each entry point:

```
CANNOT REACH staleness-guard  <- signer-session.mjs          (2 modules in closure)
CANNOT REACH staleness-guard  <- confirmed-fresh-signer.mjs  (4 modules in closure)
REACHES      staleness-guard  <- stable-control-signer.mjs   (6 modules in closure)
```

So the guard cannot fire on those paths — not "does not today", cannot, there is no edge to it.

What those paths sign:

| page | signer | signs |
|---|---|---|
| `routing.mjs` | `stableControlSigner` | **guarded** |
| `index.html` | `signerFromSession` / `nip07Signer` direct | **kind 440, kind 441** |
| `connect.html` | `consoleSigner` | **kind 440** |
| `config.html` | `confirmedFreshSigner` → `consoleSigner` | kind 30078 |
| `agents.html` | `consoleSigner` | kind 30078 |
| `following.html` | `consoleSigner` | kind 30078 |

Grant issuance and revocation are the two least recoverable things this console does, and they are on the unguarded side. A stale `index.html` signing a 440 against an old scope-hash rule or an old tag grammar is the exact scenario in your opening paragraph.

## The suite's page check is satisfied by the half you call not-a-guard

Section 8 asserts every page contains the string `staleness-guard.mjs`. That is the module's auto-install block, which starts the fetch and renders the banner. It is true and it is worth having — but the PR's own argument is that the banner "can be undone by any code that renders controls afterwards" and that "an operator can sign through a banner." By that standard, `every console page loads the staleness guard` currently reports coverage the signing path does not have. Section 7 tests the teeth in both directions and does it well; it just tests the one path that has them.

**The fix looks small.** `consoleSigner` is the real chokepoint — `routing.mjs:215` already passes it as `signerFactory`, so putting `await assertConsoleFresh()` there covers four pages and leaves routing.mjs guarded twice. `index.html` needs its own, since it does not go through `consoleSigner` at all. Then section 8 can assert what it means: no page reaches `signEvent` without the guard in its closure.

## Verified rather than read

- **The generator is deterministic and the commit is current.** Ran `tools/gen-console-build-id.mjs` on a clean tree — `1009bbfe3714a8cf5979655f5caa92fd (34 files)` and `git status` came back empty. That matters more than it looks: a generator that drifted from its committed output would make the banner fire on a current page, and the PR is right that an operator who learns to ignore the banner is worse off than one who never had it.
- **The content-hash-not-commit-sha argument holds** for the same reason, and the path is hashed with the bytes so a rename is a new id.
- **The three-state verdict fails closed.** `assertConsoleFresh` throws on anything that is not `fresh`, so an unreachable `VERSION.json` refuses to sign rather than assuming current. `fetchServerBuild` funnels every failure — throw, `!ok`, bad JSON, non-string `build` — to `null`, and `inconclusive` carries a different message from `stale`, so the operator is told which of the two happened.
- **Awaiting `pending` rather than reading a snapshot** is the right call for a click that lands before the fetch does.
- **The bootstrap gap is stated, not papered over** — a graph cached from before this shipped does not run this module.
- Baseline `ALL PASS`; nothing modified in the worktree.

JAFairweather:
Pushed `986d0e5e` — the finding is fixed, both halves of it.

**The guard.** `consoleSigner` is the chokepoint, so it goes there; that covers `connect.html`, `config.html`, `agents.html` and `following.html`, and leaves `routing.mjs` guarded twice, which is harmless since both awaits resolve the same in-flight check. `index.html` never goes through `consoleSigner`, so it got its own, placed before the signer is asked for anything rather than around the publish — a stale graph means the scope-hash rule, the tag grammar and the capability vocabulary that assembled `pendingTemplate` are all cached old copies, so the operator would be approving a grant built by code they cannot see.

Under Node there is no `document` and so no HTTP module cache to go stale, and `assertConsoleFresh` returns `fresh` rather than throwing — which is why this does not drag sockets into the suite.

**Section 8 replaced, not patched.** It asserted every page contains the string `staleness-guard.mjs`, which is the banner — the thing this suite's own argument says an operator can sign through. It passed the entire time 440s and 441s were signed from a closure with no edge to the guard. Section 9 asserts the property instead: every entry point that can reach `signEvent` must also reach the guard, by transitive import closure.

**Proven in both directions.** On the pre-fix console the new check fails and names all seven:

```
FAIL — every signing path can reach the staleness guard (8 checked; UNGUARDED:
       agents.html, config.html, connect.html, following.html, index.html,
       confirmed-fresh-signer.mjs, signer-session.mjs)
```

After the fix, 8/8 guarded. That is the check earning its place rather than being asserted into existence.

Three things I did deliberately in the walker, each because the obvious version would have been useless:

- It reads multi-line and dynamic imports. A walker that only sees single-line `from '…'` returns a clean closure for a file it never opened — the exact blind spot #433 found in the ship-list walk, and it would have made this section quietly vacuous.
- It carries a size floor. `the closure walk resolved 49 module edge(s)` fails if the walk resolves nothing, because a walk that reads nothing reports everything clean.
- A synthetic entry that signs and imports nothing is asserted to come out unguarded, so the check cannot pass by marking everything guarded.

`node tools/gen-console-build-id.mjs` re-run, since touching `console/` moves the id — the suite caught that too, which is the generator check doing its job.

`npm test`: 84/84, exit 0. Lint clean.

**This needs an outside read.** I wrote both the review and the fix, so there is no second pair of eyes on it yet — and per CLAUDE.md this touches a signing path, which is squarely "substantive". Routing to @My Dude.

JAFairweather:
## Outside read — the fix is right, the test that holds it in place is not

You asked me to try to break the closure walker rather than read it. I broke it. Twice, and one of
the breaks is the original defect walking back in with the suite fully green.

Taking the good news first, because it's load-bearing: **the fix itself is correctly wired.** I
traced all five signing paths independently rather than trusting the commit message —

| path | reaches a guard **call** via |
|---|---|
| `agents.html`, `connect.html`, `following.html` | `consoleSigner` → `await assertFresh()` |
| `config.html` | `confirmedFreshSigner` → `signerFactory = consoleSigner` |
| `index.html` (kind 440/441) | its own `await assertConsoleFresh()`, before `signEvent` |
| `routing.mjs` | `consoleSigner` **and** `stableControlSigner:18` |

`consoleSigner` really is the chokepoint, and putting the guard there was the right call. Your
diagnosis of the gap — no edge to the guard at all, not merely unguarded — matches what I see.

Everything below is about section 9.

All results are from the branch tree at `986d0e5e`, extracted and run standalone.
Baseline reproduces: **49 edges, 8 entry points, ALL PASS.**

---

### 1. Section 9 does not catch the defect it was written for

Strip the guard **call** from `index.html` and leave the **import** in place — that is the original
bug exactly, kind 440 and 441 signed with no freshness check on the path:

```
-    await assertConsoleFresh()
     const signed = await signer.signEvent(pendingTemplate)
```

Then regenerate the build id (or section 1 fires on changed bytes and masks the result):

```
$ node tools/gen-console-build-id.mjs && node tests/console_staleness.mjs
ok   — the closure walk resolved 49 module edge(s), so it is reading imports
ok   — found 8 entry point(s) that can reach signEvent
ok   — every signing path can reach the staleness guard (8 checked)
ok   — NEGATIVE CONTROL — a signing entry with no guard in its closure is detected as unguarded

ALL PASS          node exit=0, 42/42
```

Section 8 asserted *string in file*. Section 9 asserts *module in import closure*. The property that
actually failed is *the guard is awaited on the path to `signEvent`* — that's **control flow, not
import graph**. You moved along the right axis and stopped one step short of the thing that broke.

And it is now structurally unable to fail. Post-fix, all 8 signing entries have the guard in their
closure, so no signing code added to any existing page can ever be flagged — which is precisely the
most likely future regression. The check has become a tautology about today's wiring.

### 2. The decoy negative control is inert

Delete the detection outright — not subtly, entirely:

```js
const unguarded = []     // was: canSign.filter(e => e.name !== GUARD && !e.closure.has(GUARD))
```

```
ALL PASS          node exit=0
```

The decoy never goes through the pipeline. It's hand-built with `closure: new Set()`, so the
assertion reduces to "does `/\.signEvent\s*\(/` match a string containing `.signEvent(`" and "is an
empty Set empty" — both true by construction, independent of `closureOf`, `canSign` and `unguarded`.
It cannot fail for the reason it names.

Real version: push the decoy into `entries` **before** `canSign` is computed, and assert it comes
out in `unguarded`. Same three lines, runs the actual pipeline.

### 3. The size floor works — credit where it's due

Gut `closureOf` to return an empty Set and the floor fires properly, with names:

```
FAIL — the closure walk resolved 0 module edge(s), so it is reading imports
FAIL — every signing path can reach the staleness guard (5 checked; UNGUARDED: agents.html,
       config.html, connect.html, following.html, index.html)
```

That's the #433 blind spot genuinely covered. Of your two anti-vacuity devices, the floor earns its
place and the decoy doesn't.

### 4. The obvious fix is also vacuous — don't take it

My first instinct was "require a guard *call* in the closure, not just the module." I implemented it
and ran it before recommending it:

```
### post-fix tree          UNGUARDED by call-in-closure : (none)
### guard call stripped    UNGUARDED by call-in-closure : (none)      <-- still misses it
```

`index.html`'s closure reaches a file that calls the guard even when its own 440 path doesn't.
Recording this so nobody re-derives it: once every page's closure reaches the guard, **no static
import- or text-level check can discriminate.** Section 9's ceiling is structural, not a bug in the
walker.

---

### What does work: the seam you already built

`consoleSigner` takes `assertFresh` as an injectable option. That makes the chokepoint assertable
behaviourally, socket-free, no `document` needed — and unlike section 9 it fails when it should.
Written and run against your branch:

```js
// stale -> refuses, and refuses BEFORE the session signer is restored
const err = await consoleSigner({
  storage, assertFresh: async () => { throw new Error('stale console: aaa vs bbb') },
  parse: () => ({ type: 'nip07' }),
  restore: async () => { restoreCalled = true; return { signEvent: async () => ({}) } },
}).catch(e => e)
check(err instanceof Error && /stale console/.test(err.message), 'a stale page refuses at the chokepoint')
check(restoreCalled === false, '  …and it refuses BEFORE the session signer is restored')
// NEGATIVE CONTROL — assertFresh resolves -> a signer comes back and the session WAS restored
```

```
4/4 passed
```

And it is not vacuous — remove `await assertFresh()` from `signer-session.mjs` and it drops to
**2/4** with named failures. The ordering assertion is the one I'd insist on: guard-before-restore
is the entire reason it sits at the top of that function, and it's cheap to state.

That covers four of five paths. Section 7 already covers `stableControlSigner` the same way. So the
suite would then assert the property behaviourally everywhere except one place —

### The one path that stays untestable is the 440/441 path

`index.html`'s guard call lives inside an inline `addEventListener` in the HTML, so there's nothing
importable to drive. The least recoverable operation this console performs is the one the suite
cannot assert, and section 9 gives no cover there for the reasons above.

Extract the `confirm-go` handler body into a module and it becomes testable exactly like the
chokepoint. I'd do it: eyes already missed this specific path once, which is why this PR exists.
If you'd rather not grow the diff, then say plainly in the module header that the 440/441 call site
is eye-reviewed only — but don't let section 9 imply otherwise.

### Mechanics

- **Not rebased** — 5 behind `origin/main`, 2 ahead. Suite count 84 now collides with **#449** as
  well as #433; all three take 83 → 84. Second to land rebases `package.json`, `CLAUDE.md`,
  `README.md`, `docs/GETTING_STARTED.md`.
- **CI was pending** when I looked (run `31728478761`), so 84/84 is your local claim, not something
  I confirmed. I verified `tests/console_staleness.mjs` standalone only — no worktree from this seat.
- The build-id coupling means any `console/` edit fails section 1 until
  `tools/gen-console-build-id.mjs` is re-run. Correct by design, but it cost me a confusing run and
  will cost the next reviewer one too. One line in the suite header would fix that.

### On your INCONCLUSIVE question

Refusing is right, and your own argument is the strongest one: `VERSION.json` is committed, so
absent is more likely a stale snapshot than a real absence. The flaky-network cost is bounded — a
reload clears it — and the alternative fails open on a signing path. Keep it.

---

**Verdict: don't merge yet, and not because the fix is wrong.** The wiring is correct on all five
paths and I'd merge that today. But the test standing behind it passes with the original defect
reinstated, and its negative control passes with the detection deleted — so merging now ships a
correct fix with no regression cover on a signing path, which is the situation #418 was filed about.

Add the behavioural chokepoint test and make the decoy run through the pipeline. Then it's a merge.

JAFairweather:
**One must-fix, then approve.** Reviewed at `3c612d52`. Full `npm test` green, 92 suites, exit 0.

## Must-fix — a literal NUL byte makes `tools/gen-console-build-id.mjs` unreviewable

Line 53 writes the hash domain separator as a **raw 0x00 byte** inside the string literal, rather than as an escape. It is at byte offset 2626, and it is the only control byte in the file:

```js
h.update(rel, 'utf8')
h.update('<a literal 0x00 here>')      // should be ''
h.update(createHash('sha256').update(readFileSync(join(dir, rel))).digest('hex'), 'utf8')
```

git classifies the file binary as a result. The PR diff for it reads:

```
 tools/gen-console-build-id.mjs | Bin 0 -> 3523 bytes
```

so none of its ~100 lines are reviewable in this PR, and `git grep` answers `Binary file HEAD:tools/gen-console-build-id.mjs matches` with no line and no content. The generator of the id the whole guard rests on is the one file a reviewer cannot read.

CLAUDE.md: *"Never put an invisible character literally in source — use `\uXXXX`."* A NUL as a probe key already cost us a mutation run that proved nothing.

The fix is one byte and costs nothing else. I applied the `` escape and checked: the file becomes text, `node tools/gen-console-build-id.mjs` leaves `console/` byte-identical (it hashes `console/` only, and this file is not under it), and `console_staleness` stays green. **No regeneration needed** — the build id does not move.

## The ordering fix is genuine, and §9 to §11 was the right move

You were right that section 9 could not do it, and moving the two lines into `guarded-signer.mjs` is what makes the ordering assertable rather than merely visible. Eight mutations, all killed by the assertion that names them:

| mutation | killed by |
|---|---|
| `index.html` signs directly again | §11, naming the call site and line 731 |
| `signFresh` signs first, asserts after | "…and the signer is never ASKED" |
| `consoleSigner` asserts after the session restore | "consoleSigner REFUSES on a stale page (it returned a signer)" |
| `stableControlSigner` guard dropped | "REFUSES to open a signer on a stale page, naming both builds" |
| INCONCLUSIVE read as fresh | "an unknown page is refused, for the unknown reason and not the stale one" |
| never-checked waved through | "a browser page that never ran the check is refused, and told so" |
| `cache: 'no-store'` dropped | the no-store assertion |
| mismatched ids called fresh | "differing ids -> stale" |

§1's recompute pin is what closes the hole I would have looked for next — edit anything under `console/` without regenerating and both files still agree with each other while the server holds new code. It fails instead.

## A method note, because my first pass on those eight was worthless

All eight came back killed on the first run — and every one was killed by §1's build-id pin, not by behaviour, because a mutation under `console/` invalidates the recompute. It read as a clean sweep. Your header says so at lines 35-39; I read it afterwards. Regenerate between applying the mutation and running the suite, or the suite kills everything for the wrong reason.

Your judgement on the seventh mutation is right too — moving the `await` past the `localStorage` read in `consoleSigner` is equivalent, not surviving: the read is synchronous and signs nothing. Moving it past `restore()` is a different mutation, and that one is killed.

Fix the NUL and this is an approve from me.

JAFairweather:
## Still blocked on the one byte

Re-read at `3c612d52`. The signing-guard work is unchanged from my approve-conditional, and the new commit ("Assert the guard is AWAITED, not merely imported") is a real strengthening. But the must-fix is not applied.

`tools/gen-console-build-id.mjs` still carries a **literal NUL at line 53**, byte offset 2626 — `h.update(<raw NUL in a string literal>)`.

`grep` and `awk` will both lie to you about this, and mine did. My first probe was `grep -c $'\x00'` and it reported **zero**, because zsh cannot carry a NUL through `$'…'`. My second was an `awk` `index($0, sprintf("%c",0))`, which reported the byte on all 75 lines, because `sprintf("%c",0)` is the empty string and `index` returns 1 for everything. `od -c` and a `readFileSync` byte scan are the only reads that told the truth — one NUL, offset 2626, line 53. Same trap that cost a mutation run here on 2026-08-09.

The consequence is unchanged and it is the whole point. Git still classifies the file binary, so this PR's diff for it reads:

```
Binary files /dev/null and b/tools/gen-console-build-id.mjs differ
```

The generator of the build id that the entire staleness guard rests on is the one file in this PR a reviewer cannot read.

I re-verified the fix costs nothing. Replaced the raw byte with the six-character `\u0000` escape and, in the worktree:

- the file becomes text — `readFileSync(...).includes(0) === false`
- `node tools/gen-console-build-id.mjs` exits 0 and leaves **`console/build-id.mjs` and `console/VERSION.json` byte-identical**. `'\u0000'` evaluates to the same one-character string, so the hash does not move and nothing needs regenerating.
- `node tests/console_staleness.mjs` → `ALL PASS`

One byte. Then this is an approve.

PR #437, opened 2026-08-13
One bridge key, one place, written only after it verifies (#322)

### LINKED ISSUE #322: The console asks for the bridge key on every page — only two of five remember it
Reported from the live console: Config prepopulates the bridge public key, Agents does not, so
the same 64-hex key has to be pasted again on a page you reached by clicking a tab.

Measured on `origin/main`, references to the shared `waggle-following-bridge` localStorage key:

| page | persists the key |
|---|---|
| `config.html` | yes |
| `following.html` | yes |
| `agents.html` | **no** |
| `routing.html` | **no** |
| `index.html` (Access) | **no** |

Three of five pages ask for something the console already knows. It reads as five unrelated
tools that happen to share a titlebar rather than one console for one hive.

## Fix

Read and write the same key on every page that takes a bridge public key, on the same
save-on-successful-load rule the two working pages already use — persist after a load that
verified, not on keystroke, so a typo is never what gets remembered.

`index.html` is slightly different: its field is the **grantor**, not the bridge, and it fills
from sign-in when you sign in. Worth confirming whether those are the same value in practice
before sharing one key between them — if they are not, it needs its own key rather than a wrong
prefill, which would be worse than an empty box.

Small, but it is friction on the exact path an owner walks when checking who has access.

JAFairweather:
Closes #322.

## The measurement disagrees with the report, and the difference is the fix

The issue reads as three of five pages never having been taught to remember the bridge key. They had been. `agents.html` persisted it perfectly well — under `waggle-agents-bridge`, its own name for the value every other page stored as `waggle-following-bridge`.

| page | before | |
|---|---|---|
| `config.html` | `waggle-following-bridge` | |
| `following.html` | `waggle-following-bridge` | |
| `routing.html` | `waggle-following-bridge` | via `routing.mjs` — the report's table says no; it has done since #404/#406 |
| `agents.html` | `waggle-agents-bridge` | **the split** |
| `index.html` | — | grantor field, not a bridge field |

So the console was not forgetting; it was writing the answer somewhere the next page did not look. That matters for the shape of the fix: from the operator's chair a split namespace and a missing feature look identical, and the natural remedy for a missing feature is to teach the page to remember — which is how you get a fourth key.

`console/bridge-key-store.mjs` owns the name now, and the suite asserts that **no other file under `console/` names a bridge storage key at all**. That is the assertion that stops this returning on the next page; "agents.html works now" would not have.

The old name is still read and migrated to the shared one on first read. Dropping it would empty the box for exactly the operator who reported this.

## A second defect, not in the report

All four pages wrote the key immediately after parsing the field:

```js
const bridge = hex($('bridge').value)
localStorage.setItem('waggle-agents-bridge', bridge)   // before any relay has answered
```

A well-formed key for a hive that does not exist was therefore remembered and prefilled on every later visit, outliving the error message that had just rejected it. The issue describes the correct rule — persist after a load that verified — as what the two working pages already did. They did not; none of the four did.

`rememberBridgeKey(bridge, verified)` now requires the verified control state and checks it matches the key. That makes the ordering **structural** rather than a convention four pages are each trusted to follow: a page that calls it too early has nothing to pass, so it cannot persist.

## index.html stays out, deliberately

The issue flagged this as worth confirming before sharing one key. `index.html`'s field is the **grantor** — whose approvals you want to see, i.e. the owner's key — not the bridge's. They are different values, so prefilling one from the other would put a confidently wrong 64-hex string in front of an operator, which is worse than an empty box. The suite pins the separation, so a later "helpful" prefill has to argue with a test.

## Evidence

`npm test` — 84 suites, exit 0. New suite `tests/console_bridge_key.mjs`, 24 checks.

```
ok   — only bridge-key-store.mjs names a bridge storage key (33 console sources scanned)
ok   — NEGATIVE CONTROL — the same scan DOES find every storage name inside the store itself
ok   — every page with a bridge field reaches the shared store
ok   — NEGATIVE CONTROL — pages with no bridge field do not read the key (connect.html, index.html, setup.html)
ok   — agents.html's old key is still read, so nobody loses their prefill
ok   — and it is migrated to the shared name on first read
ok   — NEGATIVE CONTROL — a verified load DOES persist the key (or this guard would just refuse everything)
ok   — no verified state -> nothing is written: a typo cannot outlive the error that rejected it
```

Two mutations, each confirmed to fail and then restored:

```
A  a page re-invents its own storage name    exit=1  FAIL — offenders: agents.html
B  rememberBridgeKey moved back to parse     exit=1  FAIL — agents.html: the key is written after the state is verified, not on parse
restored                                     exit=0  ALL PASS
```

The suite exits **3 = INCONCLUSIVE** below 20 console sources or 5 pages: a scan that reads nothing finds no offending literal either, and prints the same clean result as a tidy console.

## Reviewer notes

**Not proven in a browser.** Everything here is asserted against the exported functions and the page sources under Node. `npm test` has no browser, and the four inline module edits — two of them in the single-line minified modules in `config.html` and `following.html` — are the part I would most want a second pair of eyes on. Each edit was anchor-asserted (unique match required) rather than pattern-replaced.

**Worth challenging:** `rememberBridgeKey` returns `false` rather than throwing when the verified state is missing. That keeps a caller mistake silent — the page just stops remembering, which is the original symptom. I chose it because throwing inside `load()` would surface as a spurious error banner on an otherwise successful load, and the static ordering check in the suite is what catches the caller mistake instead. Tell me if that trade is wrong.

**Suite count 84 collides with #433 and #436.** Whichever merge last need a rebase of `package.json`, `CLAUDE.md`, `README.md` and `docs/GETTING_STARTED.md`. This one also touches `console/agents.html`, `config.html`, `following.html` and `routing.mjs`; #436 touches every page's `</body>` line only, so those two should merge cleanly apart from the counts.

@My Dude


JAFairweather:
## Review — approve. The reframe is the right call and the module is correct. Two findings, one of which reintroduces your own defect 2 through the migration path.

Reviewed at `7d6cf5a19424c37b2c913bca741c5b0e55c432a2`. **83 of 84 suites pass**; `install_state.mjs` fails identically on `origin/main` — the same pre-existing local failure as #431, #433 and #436.

Refusing "teach agents.html to remember" because it produces a fourth key is the right read, and measuring before fixing is what turned a five-row table into a one-line namespace bug. Keeping `index.html` out on the grounds that a confidently wrong 64-hex prefill beats an empty box is also right, and pinning it so a later "helpful" prefill has to argue with a test is better.

## 1. `loadBridgeKey` launders an unverified value into the canonical key

Your defect 2 is that a well-formed key for a hive that does not exist got remembered and prefilled forever. `rememberBridgeKey` now refuses that. But `loadBridgeKey` **writes**, at line 40, with no verification at all:

```js
if (old && HEX64.test(old)) {
  value = old
  try { storage?.setItem(BRIDGE_KEY_STORAGE, old) } catch { /* private mode */ }
```

```
legacy store holds only  waggle-agents-bridge = aaaa…aaaa   (well-formed, hive does not exist)
loadBridgeKey()      ->  returns aaaa…aaaa
after the call       ->  waggle-following-bridge = aaaa…aaaa      <- canonical, unverified
rememberBridgeKey(same value, no verified state) -> false, writes nothing
```

The guarded door refuses the value; the reading door promotes it. And this is not an edge case — it is the whole legacy population **by construction**. Every `waggle-agents-bridge` value in existence was written by the old `agents.html` line you just deleted, which wrote *before any relay answered*. So the values eligible for migration are precisely the unverified ones, and migration widens the blast radius from one page to all four.

Fix is the `setItem`, not the read:

```js
if (old && HEX64.test(old)) { value = old; break }   // prefill from it; let a verified load promote it
```

Tested: the legacy key still prefills, the canonical key stays unwritten, and `rememberBridgeKey` promotes it on the first verified load. **It costs exactly one assertion** — `console_bridge_key.mjs:129`, "and it is migrated to the shared name on first read" — which pins the behaviour deliberately. So this is a disagreement about the design, not an oversight, and I'd argue the assertion should become "the legacy value prefills but is not promoted until a load verifies it", which is strictly your own stated principle applied to the one path that currently exempts itself.

## 2. The namespace guard checks two literals, not "a bridge storage key"

§1's header names the threat exactly right — "a split namespace is indistinguishable from a missing feature, and the natural fix for it is a fourth key." The scan does not cover that:

```js
const NAMES = [store.BRIDGE_KEY_STORAGE, ...store.LEGACY_BRIDGE_KEY_STORAGE]
const offenders = files.filter(f => !f.endsWith(STORE)).filter(f => NAMES.some(n => …includes(n)))
```

It asks whether any file mentions `waggle-following-bridge` or `waggle-agents-bridge`. It cannot ask whether a file names a bridge storage key. I added a fifth page inventing `waggle-connect-bridge` — the precise scenario the header describes — and **the suite passes clean**.

This is cheap to close because the console is already tidy: after this PR there is exactly **one** literal `localStorage` key left outside the store, `waggle.accesslist.view` (2 uses). So an allowlist scan over `localStorage.(get|set|remove)Item('…')` in `console/` is a handful of lines and catches an invented name rather than a reused one.

## Your question 1 — the unverified inline edits

This is the part I could actually take off your hands, and the specific risk of anchor-editing minified single-line modules is broken syntax or an edit landing in the wrong scope. Both are now checked:

```
ok    agents.html:90      10323 bytes
ok    config.html:4        4821 bytes
ok    connect.html:217    14350 bytes
ok    following.html:35    6131 bytes
ok    index.html:341      45575 bytes
5 inline module blocks parsed, 0 failed
```

And both minified edits are placed correctly, downstream of the verification throw, passing the right object:

```js
// config.html    …if(!s)throw Error('No fresh valid signed Waggle configuration was found.');live=s;rememberBridgeKey(b,s);…
// following.html …if(!winner)throw Error('No fresh valid signed state…');activeBridge=bridge;activeState=winner;setManage(true);rememberBridgeKey(bridge,winner);…
```

`valid(e,b)` and `validState` each require `state.bridge === bridge`, so the identity check inside `rememberBridgeKey` cannot be vacuous at either site.

**This does not make it browser-proven** — no page was loaded, no `localStorage`, no relay. But the failure mode you were actually worried about when editing minified code is now excluded mechanically rather than by reading. Worth making that a standing control: espree is already a devDependency, `console/` has no parse guard today, and #433's `ship_imports.mjs` deliberately does not scan HTML. Ten lines, and the next inline edit is covered.

## Your question 2 — `false` rather than a throw

**The trade is right, and your safety net is real** — I checked rather than taking it on trust. §3 is a genuine structural ordering check, and it binds:

| mutation | result |
|---|---|
| move `rememberBridgeKey` above the verified-state pick (`routing.mjs`) | CAUGHT, 2 checks |
| drop the verified-state requirement from the store entirely | CAUGHT, 4 checks |
| a fifth page invents its own storage name | **NOT CAUGHT** — finding 2 |

A silent `false` is the correct choice here for the reason you gave: the alternative puts an error banner on a *successful* load, and the caller mistake it would report is a static property that a test can hold better than a runtime can. `persisting.length === 4` also means a new page that persists forces the count to be re-decided rather than sliding in.

One residue worth knowing about: a page that *reads* the store but never persists is invisible — §2 only asks whether a page reaches the store, and §3 counts exactly the four that persist. That page would present as the original symptom, silently. Small, and I would not hold the PR for it.

**Not proven by me:** any browser, any real `localStorage`, any relay. The store's behaviour above is executed under Node against a fake storage; the page edits are parsed and read, not run.

Noted on the triple suite-count collision with #433 and #436 — last to merge rebases `package.json`, `CLAUDE.md`, `README.md`, `GETTING_STARTED`.

JAFairweather:
Reviewed at `7d6cf5a1`. The diagnosis is right, the store is right, and the suite is one of the better ones in this repo — both directions on every guard, size floors, a real INCONCLUSIVE path. **One finding, and it is on the assertion the PR names as the load-bearing one.**

## The scan catches the two names it already knows, not "a bridge storage key"

The header says the suite asserts that *no other file under `console/` names a bridge storage key at all*, and that this "is what stops this coming back on the next page." What it implements is `NAMES.some(n => src.includes(n))` where `NAMES` is the canonical name plus the enumerated legacy one. A page that invents a *new* name is invisible.

Added one line to `console/routing.mjs` on this branch:

```js
localStorage.setItem('waggle-routing-bridge', 'a'.repeat(64))   // valid JS, node --check clean
→ ALL PASS
```

Against the same line using the sanctioned name:

```js
localStorage.setItem('waggle-following-bridge', 'a'.repeat(64))
→ FAIL — only bridge-key-store.mjs names a bridge storage key (33 console sources scanned; offenders: routing.mjs)
```

So the scan works, and it is blind to precisely the shape that caused #322. `waggle-agents-bridge` is caught today only because it has already been discovered and written into `LEGACY_BRIDGE_KEY_STORAGE`. Had this suite existed when `agents.html` was written, it would have passed. The next page does not have to argue with a test to invent a fourth key — it only has to pick a name nobody has enumerated yet, which is what the last one did.

A literal-shape scan closes it: any `['"]waggle-[\w-]*bridge[\w-]*['"]` outside the store. I checked it would not be noisy — `console/` has 6 other `localStorage` call sites, all in `index.html` (`waggle.accesslist.view`, a session key, a dismiss key), and none of them match. A broader "any `localStorage` in console/ outside the store" would flag all 6 and stop being read, which is the failure mode `ship_imports` reasons about in #433.

## Not findings, having checked

- **Junk under the canonical name blocking legacy migration.** `loadBridgeKey` skips the legacy branch whenever the canonical slot is truthy, so a non-hex value there hides a good legacy key and returns `''`. Unreachable from the console: all four pages write through `hex()`, which throws on anything that is not 64-hex or an npub, so the `setItem` never runs. Only hand-edited storage gets there. Mentioning it because the module takes a `storage` argument and may outlive that guarantee.
- **The ordering check is textual, not execution order** — `indexOf('newestFreshControlState(')` vs `indexOf('rememberBridgeKey(')`. It cannot prove a page verified before it persisted. The suite says so itself and points at the runtime guard as the real enforcement, which is the honest framing; the textual check earns its place by catching a call moved back to the top of `load()`.

## Verified rather than read

- **`agents.html` really did persist under its own name.** Read `origin/main:console/config.html` — `hex()` throws before the `setItem`, so the write path was sound and only the *name* diverged. The PR's central claim, that this was a split namespace rather than a missing feature, holds, and it is the reason the fix is a shared module instead of teaching three more pages to remember.
- **The store's guards fail in both directions** — `w1` proves a verified load does persist, so `w2`/`w3`/`w4`/`w5` are refusing the dangerous thing rather than refusing everything. That is the pairing that was missing in the 2026-08-01 slot-validator outage.
- **Private-browsing paths return rather than throw**, on both read and write.
- **`index.html` exclusion is correct and pinned.** The grantor key is a different value; prefilling it would be a confidently wrong 64-hex string.
- Registered in `package.json`; baseline `ALL PASS`, tree restored clean after both mutations.

PR #438, opened 2026-08-13
An empty agent roster is not an empty hive (#321)

JAFairweather:
Closes nothing — #321 stays open on the design question. This lands the one part of it that is settled regardless of the answer, quoting the issue:

> Whichever is chosen, the empty state on the Agents page should not read as a bare "no agents" when grants exist — that is the sentence that made this look broken.

## The defect

`console/agents.html` rendered *"Nobody is in this hive yet."* whenever the lifecycle roster was empty. An owner who had just issued a 440 grant read that as a broken bridge and went looking for a fault that was not there. A grant admits a key to the community; the roster is a different registry that only this page writes to. #392 already models the case as `grant_no_row` — "the agent can act, the roster denies it exists, because a grant does not create a row."

## The change

The empty state now names its own scope, says what a grant does without a row, and links to Access:

> No agent has been seated through this page. That is not the same as an empty hive: a 440 grant lets a key act without creating a row here, so check **Access** for keys that already hold one. Seating an agent here is a decision that widens what it can reach — the table below says exactly how far.

This does **not** pick between the three options in #321 (derive rows from grants / seed a row in `processGrantEvent` / keep them separate and say so). That is the maintainer's call and the page reads correctly under any of them.

## Evidence

`tests/console_agent_empty_state.mjs` (11 checks) extracts the branch from the page and executes it against a fake DOM, so every assertion is on the sentence an operator actually reads — a script that drifted from what it renders would pass a grep and fail here.

- The banned strings are exact claims of absence, not keywords. `empty hive` appears in the *correct* text, inside "not the same as an empty hive"; banning the vocabulary would have forbidden the fix and passed the bug. That is asserted directly.
- Negative controls: the banned-phrase check is shown flagging the reported sentence, and the link scan is shown finding zero links in a branch that renders none.
- Bound to `src/registry_reconcile.mjs` — if `grant_no_row` is renamed or dropped, the model behind this text has moved and this suite fails rather than the drift reaching an owner.

Mutation control: restoring the reported sentence fails the suite with **exit 1** and names the reason (`found: "Nobody is in this hive"`). An earlier draft failed that mutation with exit **3** instead — the render floor was set near the length of a real sentence, so the 27-character bug tripped it as INCONCLUSIVE. Correct refusal, misleading reason; floor lowered to 10, which is what a floor is for.

`npm test` 84/84 green, `npm run lint` clean.

## Notes for review

- Suite count 83 → 84, including the stale-base threshold in `CLAUDE.md`. **#433, #436 and #437 each also claim 84** — whichever lands last needs a rebase of `package.json`, `CLAUDE.md`, `README.md` and `docs/GETTING_STARTED.md`.
- **#437 also touches `console/agents.html`.** Different function (`draw` here), but expect a conflict marker if they land close together.
- Separate finding, not fixed here: `src/registry_reconcile.mjs` is imported by nothing but its own test. The reconciler from #392 computes findings that no console page renders. Filing that as its own issue.


JAFairweather:
## Review — approve. Your question 1 answers itself in your favour: the anchor is brittle, and it is never silent. Question 2 has a real hole — the ban is case-sensitive, and the most likely replacement sentence walks through it.

Reviewed at `a19c9ca68d54b1739479dc1edc80fe7dfb617645`, 1 ahead / 0 behind `origin/main`.

The copy is right. It names its scope in the first clause, states what a grant does without a row, and sends the owner to the page that would actually answer their question. It reads correctly under all three options #321 is still open on, as you say.

## Question 1 — brittle, yes. Silent, no. That distinction is the whole answer.

I broke the anchor four ways, each one a refactor someone would make without thinking about this test:

| harmless refactor | result |
|---|---|
| `if (!state.agents.length)` → `if (state.agents.length === 0)` | exit 3, ANCHOR MISS |
| `if (! state.agents.length)` — whitespace | exit 3, ANCHOR MISS |
| rename the loop variable `a` → `agent` | exit 3, ANCHOR MISS |
| `for (const a of (state.agents ?? []))` — defensive | exit 3, ANCHOR MISS |
| hoist the copy to a module const, reintroduce the claim in it | exit 1 (uncaught `ReferenceError`) |
| **control:** restore the reported sentence | exit 1, names the reason |

So the anchor breaks on 4 of 4 — it is as fragile as you suspected — but `inconclusive()` exits **3**, and `npm test` is a single `&&` chain, so exit 3 stops the build and CI's `run: npm test` fails. The refactor costs someone a re-anchor. It cannot buy them a false all-clear. That is the repo's exit-3 convention doing exactly the job it exists for, applied to a test rather than a tool, and it is the reason this design is fine.

**One change worth making anyway.** The close anchor is `'\n  for (const a of state.agents) {'`, which encodes two facts that have nothing to do with the branch under test: the indentation, and the loop variable's name. Renaming `a` to `agent` is the most likely refactor on that list and it is the one with the least connection to what you are asserting. Brace-match forward from `OPEN` instead, or anchor the close on the branch's own `return\n  }`. That removes rows 3 and 4.

Rows 1 and 2 I would leave. If someone edits the empty-roster condition itself, coupling this test to that edit is correct — they *should* re-read it.

## Question 2 — the list is case-sensitive, and `'This hive has no agents.'` passes

`rendered.includes(c)` against exact claims. I ran twelve sentences a future edit would plausibly reach for:

```
  caught           "Nobody is in this hive yet."
  caught           "This hive has no agents yet."
  caught           "No agents have been admitted."
  *** PASSES ***   "This hive has no agents."
  *** PASSES ***   "No one is in this hive yet."
  *** PASSES ***   "This hive is empty."
  *** PASSES ***   "The hive is empty."
  *** PASSES ***   "Nobody has joined this hive yet."
  *** PASSES ***   "There are no members in this hive."
  *** PASSES ***   "This hive has admitted nobody."
  *** PASSES ***   "Nothing here yet."
  *** PASSES ***   "No keys have been admitted to this hive."
```

Eight of twelve. `'This hive has no agents.'` is the one I would bet on someone actually writing, and it misses on both counts — `'No agents'` is capital-N, and `'no agents yet'` needs the trailing "yet".

Your reasoning for exact claims over keywords is right, and the fix keeps it. Lower-case both sides and widen the set slightly. Tested before proposing:

```js
const ABSENCE_CLAIMS = [
  'nobody is in this hive', 'nobody has', 'no one is in this hive', 'no one has',
  'this hive has no', 'this hive is empty', 'the hive is empty', 'hive has admitted no',
  'no agents', 'there are no', 'nothing here yet', 'no members', 'no keys have',
]
// rendered.toLowerCase().includes(c)
```

**13 of 13 bad sentences caught, zero false positives on the shipped text.** The shipped copy survives because it says "No agent has been seated" — singular — and "an empty hive" rather than "the hive is empty". That is a narrow escape and worth a word in the header, because it means the correct sentence and a banned one are one character apart. Your existing `'empty hive'` control already pins the half of that which matters.

## The finding — the silent pass is in the scope, not the anchor

The suite extracts one branch and renders it, so an absence claim anywhere else on the page is invisible. I added one to the status line:

```js
if (!state.agents.length) $('status').textContent = 'Nobody is in this hive yet.'
```

**Suite passes clean, ALL PASS.** The page now makes exactly the claim this PR removed, and the check built to stop that says nothing.

This is not a stretch, because the `list` element already has a second writer:

```
agents.html:229    $('list').textContent = 'Disconnected — state unavailable.'
```

Same DOM node the empty branch writes into, outside the check's view. That line is correct today — it names the connection, not the hive — but it is where the next version of this bug lands.

Cheap complement, and it is cheap precisely because your claims are exact strings rather than keywords: scan the whole file source for them as well as the rendered branch. A few lines, no new machinery, and it catches the sentence reappearing anywhere on the page — including in a branch the anchor cannot reach. The fake-DOM render stays the primary check; this is the belt to it.

## Questions 3 and 4

**3.** Confirmed, 84 in this branch's `package.json`, colliding with #433, #436 and #437. Last to merge rebases `package.json`, `CLAUDE.md`, `README.md`, `GETTING_STARTED.md`.

**4. No interaction.** #437's four hunks in `console/agents.html` are at lines 91, 204–211, 213 and 258. The anchored region is 147–170. No overlap, so neither anchor is touched whichever lands first, and #438's ~9 inserted lines only shift #437's later hunks — an ordinary 3-way merge. Merge order does not matter here.

## What is good

`inconclusive()` exiting 3 with "This is NOT an all-clear" printed on the line below is the thing that makes the rest of this PR safe, and it is why my answer to question 1 is "keep it". A brittle anchor that fails loud is a maintenance cost; a brittle anchor that exits 0 is the bug this repo keeps writing tests to avoid. Also the 4000-byte floor on the file read, the `rendered.length < 10` floor, and three negative controls including the one that proves `'empty hive'` survives the detector — that last one is what stops the fix and the ban from being written to each other rather than to the property.

**Not proven by me:** any browser. No page was loaded; the branch is executed under Node against a fake DOM, same as yours.

**84 suites: 83 pass, 1 fail** — `install_state.mjs`, which fails identically on `origin/main`. Same pre-existing local failure as #431, #433, #436 and #437, not yours. Probe files removed, `console/agents.html` byte-identical, `git status` clean.

PR #440, opened 2026-08-13
Pick the agent from the signed roster instead of pasting an npub (#413)

### LINKED ISSUE #413: Agent identity should be picked from the signed roster the console already fetches, not pasted as an npub
The Agent channel route panel asks the operator to paste an npub into **Agent identity**. The console already holds the list it could offer instead, and throws it away.

`freshBridge()` (`console/task-routes.mjs:37`) fetches the bridge's signed `kind:30078` `waggle-control-state`, verifies the signature, checks the bridge key and the freshness, and then returns `bridge` — the hex string — discarding the state object. That object carries `state.agents`, an array of `{ pubkey, status, label }`, validated on the agents page at `console/agents.html:49-51`. It is the same event, from the same fetch, already verified.

So a picker needs no new round trip, no new trust assumption and no new endpoint. It needs `freshBridge()` to return the state it already parsed.

## Why pasting is worse than it looks

An npub is 63 characters of base32 with no checksummed prefix a human reads. The failure mode is not a rejected paste — it is a paste that validates, signs, publishes, and routes a channel's mentions to **the wrong admitted agent**. `hex()` will happily decode a well-formed npub belonging to someone else. Nothing downstream can catch it, because routing to a different admitted participant is a legitimate operation.

The panel also asks for **Authorized sender** as a second npub, with the same exposure.

## Shape

Replace the free-text field with a select over `state.agents`, rendered `label — npub1abc…xyz` and sorted with live agents first. Keep a free-text fallback for a key that is not yet in the roster, since that is a real case, and label it as such rather than as the default path.

Two things to get right:

- **Status must be visible in the row, not filtered silently.** A paused or revoked agent should be selectable-but-marked rather than absent, or the operator hunts for an agent that is there and cannot be picked. The agents page already has the status vocabulary (`console/agents.html`, WIRE values); reuse it rather than growing a second translation.
- **The label is not the display name.** It is tempting to prefill **Mention handle** from `label`, and mostly that would be right — but `label` is owner-set and constrained to printable ASCII (`/^[\x20-\x7e]{1,64}$/`), while the mention must match what Buzz writes into the channel body from the member's `display_name`, which since #404 may hold any letter including non-ASCII. So a label cannot always carry the name that has to match. Prefill from it if you like, but the operator must be able to correct it, and the field must not be locked to the picker.

## Provenance

Raised by the maintainer after establishing the `MC Claude` route by hand. Related: #412 (the console is outside the deploy verification set), #408 (the mention grammar the handle field validates against).

JAFairweather:
Closes #413.

## The exposure

The task-route panel asked for an npub in a text box. The failure mode is not a rejected paste — 63 characters of base32 have no prefix a person reads, so a **wrong** npub decodes cleanly, signs, publishes, and routes a channel's mentions to a *different admitted agent*. Nothing downstream can catch it: routing to another admitted participant is a legitimate operation. The only place it is catchable is before the paste.

`Authorized sender` was a second npub box with the same exposure.

## The fix costs nothing to trust

`freshBridge()` already fetched the bridge's signed `kind:30078` control state, verified the signature, checked the bridge key and the freshness — then returned the bridge hex and threw the state away. It now returns both. Same event, same checks, no new round trip, no new endpoint.

`console/agent-roster.mjs` owns the roster shaping and the status vocabulary. `agents.html` declared its own `STATUSES`/`STATUS_LABEL`; it imports them now. The suite asserts no other file under `console/` declares either — two translations of one wire value drift in exactly the direction where `revoked` reads as something reassuring.

Three things the issue asked for, each pinned by a test:

- **Nothing is filtered.** A paused or removed agent is selectable and marked, never absent. An agent that is on the roster and not in the list sends the operator hunting for an identity that is right there.
- **The free-text field stays the value `manage()` reads.** The picker writes into it. An off-roster key — a real case — is the same code path, not a fallback that rots.
- **The mention handle is suggested, never locked.** A `label` is owner-set printable ASCII; the `display_name` it must match may hold any letter since #404. So the label is a guess and a typed value is an answer, and the suggestion only fires on an empty field.

## Evidence

`tests/console_agent_roster.mjs`, 31 checks. Full suite 84/84, `npm run lint` clean.

Three mutation controls, each failing with the reason named rather than merely failing:

| mutation | what fails |
|---|---|
| filter non-admitted agents out of the roster | `every agent survives the sort (got 2 of 4)` |
| `freshBridge` returns a bare bridge key again | `freshBridge returns the state it parsed…` |
| the label overwrites a typed mention handle | `the mention is only suggested when empty…` |

Negative controls: the sort output is asserted to differ from its (deliberately reversed) input, so it is not a pass-through; the validator is shown dropping a malformed row, so "nothing dropped" is a measurement; and the vocabulary scan is shown finding the declaration inside the module that owns it.

**The suite caught a bug in its own fixture.** The first draft keyed agents off `'1'.repeat(64)`, so `uppercase hex is refused` passed its input through `toUpperCase()` unchanged — a probe that loses its own input, reporting a pass for a case it never ran. It showed up as a FAIL against a correct implementation. Keys are lettered hex now.

## Notes for review

- **`npx eslint console/` reports `'document' is not defined` ×5 and `'confirm' is not defined` ×1 on `task-routes.mjs`.** These are browser globals in a browser module; the shared config has no browser env for `console/`, which is #434 / PR #435. One of those errors is pre-existing on `main`; mine add more of the same shape. Left alone deliberately so this does not collide with #435.
- Suite count 83 → 84, threshold included. **#433, #436, #437, #438 also claim 84** — last to land rebases `package.json`, `CLAUDE.md`, `README.md`, `docs/GETTING_STARTED.md`.
- **`console/agents.html` is also touched by #437 and #438**, and `console/config.html` (which loads `task-routes.mjs`) by #436 and #437. Different regions, but expect markers.


JAFairweather:
## Review — request changes. Question 2 is a yes, and it is worse than you framed it: reloading the roster does not clear the stale value either. Question 3 has the same defect without switching bridges at all. And question 1's premise does not hold — the bridge does not gate on what the picker displays.

Reviewed at `af505d2265ff044fc68eb8ed0aa300afbf909254`.

The module is right and the reasoning in its header is right. `freshBridge()` returning the state it already parsed is the correct way to make this free, `agent-roster.mjs` owning the vocabulary closes a real drift, and refusing to lock the mention field because a label is ASCII and a `display_name` since #404 is not is exactly the distinction that matters. The fixture bug you caught — `'1'.repeat(64)` making the uppercase-refusal check test nothing, because digits have no case — is the "probe that loses its own input" lesson landing in real work.

The findings below are all in the panel, not the module.

## Question 2 — yes. Demonstrated, and a reload does not fix it.

I drove the real `console/task-routes.mjs` under a fake DOM with two genuinely signed control states, B1 and B2:

```
[1] bridge = B1, roster loaded
  options offered            Choose an admitted agent | My Dude — npub14w46h2a…cf6zts

[2] operator picks the agent from B1's verified roster
  route-participant          A1  ✔
  route-mention              "My Dude"

[3] operator changes the bridge field to B2 — nothing else touched
  route-participant STILL    A1  <-- STALE
  route-mention STILL        "My Dude"  <-- B1 label
  picker still displays      My Dude — npub14w46h2a…cf6zts
  picker still SHOWS as selected   My Dude — npub14w46h2a…cf6zts

[4] what B2's verified roster actually contains
  B2 roster                  Dennis — npub1ehxumnw…kccvaq
  is A1 on B2's roster?      NO — the stale value is off-roster for B2
  route-participant after reload   A1 — still stale, reload did NOT clear it
```

Three things there, in increasing order of how much they bother me.

The field `manage()` reads keeps A1. `freshBridge()` re-reads `$('bridge')` at line 39 every time, so the signature and freshness checks correctly follow the operator to B2 — and the participant does not. Nothing anywhere listens to that field; I checked every console source, and there is no `oninput`/`onchange` on `$('bridge')` at all. `routing.mjs:237` also writes it, so it moves under this panel without the panel knowing.

Step 3 is the part I would call a defect rather than an omission. The select still renders B1's roster and still shows A1 as the chosen option, which is a positive assertion that this key is on the verified roster for the bridge currently in the box. The whole value of this PR is that a picked value carries more authority than a pasted one. Here that authority is real, and attached to the wrong bridge.

Step 4 is the one that decides it. Reloading the roster is the obvious remedy, it is the only affordance you give, and it repopulates the dropdown while leaving `route-participant` holding A1. So after the reload the operator sees a correct B2 roster above a stale B1 key — and the select no longer has an option matching it, so there is nothing on screen tying the value in the field to anything.

**What actually happens on publish** is not catastrophic, and I want to be accurate about that: `applyTaskRouteCommand` at `src/bridge.mjs:2915` refuses with `participant is not admitted` when the key is not in `grantSet`. So the off-roster case fails closed at the bridge. The case that does not fail closed is the one where A1 holds a live grant on both hives, which is ordinary for a crew running more than one — then the route activates against B2, pointing at an agent the operator picked for B1. That is the original bug this PR closes, reached through the picker.

**Suggested fix.** Record which bridge the roster was loaded for, and have `manage()` refuse when it disagrees with `$('bridge')`. That keeps the free-text path intact — which is the property §4 of the suite correctly pins — and closes only the stale-pick path, with a named reason. Clearing the pickers on a bridge change is worth doing too, but on its own it is cosmetic; the refusal is the guard.

## Question 3 — suggest-on-empty is the wrong rule, and it does not need two bridges

Same harness, one bridge, two agents on the roster:

```
[1] operator picks "My Dude"        participant AGENT_A   mention "My Dude"
[2] operator changes their mind, picks "Dennis"
                                    participant AGENT_B  (updated)
                                    mention     "My Dude"   <-- STALE, still agent A

  => signed route would be: participant=Dennis, mention="My Dude"
     an "@My Dude" in channel wakes Dennis.
```

Changing your mind is the most ordinary thing an operator does in a dropdown, and it produces a route whose mention names a different agent than the one it wakes. The mention field reads as correct, which is why this is worse than leaving it blank.

The rule you want is not "suggest when empty", it is **"suggest unless the operator has typed something"** — and those are only the same thing on the first pick. Remember the last value you suggested; overwrite when the field still holds it, leave it alone otherwise. That preserves your stated principle exactly — a typed handle is an answer and a label is a guess — and fixes the case where the previous *guess* is being treated as an answer.

Worth noting this lands next to #414: two routes both claiming `@My Dude` is now the arbitration path, and one of them would be there by accident.

## Question 1 — "the bridge is the authority" is true, but it is not the authority over what you are displaying

You chose visible-and-pickable on the grounds that the console should not invent a second gate. The gate the bridge actually applies is `grantSet.has(route.participant)` — NIP-DA grants. What the picker renders is the lifecycle roster's `status`. These are two independent lanes:

- `grantSet` is written only by 440 grants and cleared only by 441 revocations (`bridge.mjs:1073-1080`).
- `status` is written only by `applyLifecycle` (`bridge.mjs:2719-2733`), whose own comment calls it "the one place an operation becomes a row change". It never touches `grantSet`.

So neither predicts the other, in either direction:

| picker shows | grant live? | bridge verdict on upsert |
|---|---|---|
| `— removed` | yes (only `agent_revoke` ran) | **accepted** |
| `— removed` | no (441 revoked) | refused |
| in | no (441 revoked) | refused, despite reading as fine |

A `revoked` mark therefore does not mean the route will be refused, and an unmarked agent does not mean it will be accepted. That is the substance of my answer: **selectable-but-marked is defensible, but not on the reason you gave** — the mark is advisory, and it is ambiguous about the exact thing that decides the outcome.

This is your own #438 insight pointing the other way. `agents.html` now says a grant lets a key act without creating a roster row; the corollary is that a roster row can outlive what makes the key act, and this picker treats the roster as the authority on both.

I would keep them selectable and drop the claim that the bridge covers it. If you want the marks to mean something, the honest version is to say what will happen — but that needs grant state the console does not have, so the cheaper move is a note that the roster shows lifecycle status and admission is decided separately.

## Question 4 — the count is the known part; the real collision is #437, in code

Tested rather than predicted, with `git merge-tree` against each head:

```
#440 + #437 -> CONFLICT   CLAUDE.md, README.md, console/agents.html
#440 + #438 -> CONFLICT   CLAUDE.md, README.md, package.json
#440 + #436 -> CONFLICT   CLAUDE.md, README.md, package.json
```

#436 and #438 are count-only, as you expect. **#437 conflicts in `console/agents.html`** — both PRs add an import immediately after the `control-state-freshness.mjs` line, and #440 deletes the `STATUSES`/`STATUS_LABEL` block directly below it. That is a content conflict needing a real read, not a rebase of doc files. #438's `agents.html` edit is at the empty-roster branch and merges clean, so only #437 is affected.

## Suite

**84 suites: 83 pass, 1 fail** — `install_state.mjs`, which fails identically on `origin/main`, same as every PR this week. My batch runner also flagged `watchlist.mjs`; it passes 29/29 standalone in 0.5s, so that was my runner, not your branch.

`console_agent_roster.mjs` is a good suite — the reversed-input control on the sort and the malformed-row control on the filter are both real, and §5's negative control on its own scan is the right instinct. It has no coverage of any of the three findings above, because all of them are panel state across time and every assertion in §4 is a static read of the source. That is the gap worth closing along with the fixes.

**Not proven by me:** any browser. The harness above is a fake DOM I wrote; the control states are genuinely signed and `freshBridge()` verified them, but no page was loaded and no relay was contacted.

JAFairweather:
Fixed in e9d00913. All three findings taken as written.

**Q2 — the stale pick.** You were right that reload does not clear it, and the suggested fix is what landed: the roster now records which bridge it was loaded for, and `manage()` refuses on a disagreement with `$('bridge')` before the round trip and before anything is signed. Named reason, so the operator is told which of the two things is stale. The field also clears what the picker wrote, by provenance — a key typed over the pick is left alone, so the free-text path §4 pins is untouched. The listener is the courtesy; the refusal is the guard, because `routing.mjs` writes that field programmatically and fires no event.

Reloading now clears too — `loadRoster()` drops the picked values when the bridge it loaded for differs from the last one.

**Q3 — suggest-unless-typed.** Your rule, verbatim: remember the last suggestion, overwrite while the field still holds it, leave it alone otherwise. The change-of-mind case was reachable with one bridge, as you said.

**Q1.** Dropped the claim. `grantSet` and `status` are written by different code and neither predicts the other, so the marks are advisory — the panel says so now, and the comment in `agent-roster.mjs` says why rather than asserting the mark decides.

**On the suite.** §4 pinned the string `if (!$('route-mention').value.trim()`, and that string *was* the bug — pinning it is why a regex could not have caught this. Replaced with §6, which drives the real handlers against a fake DOM and an in-memory relay serving two signed control states. No sockets.

Mutation-checked three ways, each failing by name: the old empty-only rule, the guard removed, and clearing without checking provenance. The third survived its first run — my negative control pasted a key when nothing had been picked, so the branch it existed to exercise was never reached. Fixed, and that is why the sequence in §6 is written the long way.

84 suites green, lint clean. `console/` is not in the lint script's paths; the six pre-existing `no-undef` errors in `task-routes.mjs` are unchanged by this.

I wrote both the PR and this fix, so it still needs an outside read before merge.

JAFairweather:
Reviewed at `e9d00913`, driving the real handlers rather than reading them. The three earlier findings are addressed as described, and §6 replacing §4's string-pinning is the right call — the negative controls in it are real ones. One blocker.

## Blocking — picking an unnamed agent leaves the previous agent's mention

`console/task-routes.mjs:201`

```js
$('route-participant').value = event.target.value
picked.participant = event.target.value
const label = (option.textContent || '').split(' — ')[0]
if (!label || label === 'unnamed agent') return   // <- returns after the key, before the mention
```

The early return fires *after* the participant key is written and *before* the mention reconciliation, so both `route-mention` and `picked.mention` keep the previous pick's label.

Reproduced against the real module with §6's harness, one unnamed agent added to the served state:

```
after picking Dennis:
  participant = 12121212…   mention = "Dennis"
after changing the pick to the UNNAMED agent:
  participant = 34343434…   mention = "Dennis"
```

That is the failure this PR exists to close — a mention naming one agent beside another agent's key — reached through a row the module treats as first-class: `validAgent` accepts `label: null` (asserted at `tests/console_agent_roster.mjs:48`) and `agentOptionText` renders it (asserted at `:95`).

Nothing downstream catches it. `Dennis` is valid mention grammar, and the confirm at `task-routes.mjs:178` renders `@Dennis for npub1…` — the mention reads correct and the npub is exactly the string a person cannot check, which is the premise of the whole change.

Fix is in the same branch: when there is no suggestion to make, clear our own guess instead of returning.

```js
if (label === 'unnamed agent') {
  const current = $('route-mention').value.trim()
  if (!current || current === picked.mention) $('route-mention').value = ''
  picked.mention = null
  return
}
```

(`!label` is unreachable — `agentOptionText` always emits a non-empty name — so it can go.)

## §6 cannot see it

All three agents in `served` (`tests/console_agent_roster.mjs:217-221`) are labelled. The section written to catch rule changes behaviourally has no unnamed row, so the one branch still carrying the bug is never entered. Add an unnamed agent and assert the change of mind in both directions: labelled → unnamed clears, unnamed → labelled suggests.

## Provenance is compared untrimmed

`task-routes.mjs:207-210` stores the raw label and compares it against `.value.trim()`. `LABEL = /^[\x20-\x7e]{1,64}$/` permits a trailing space, so a label like `Kerouac ` never round-trips equal, and on the very next pick the picker reads its own guess as the operator's answer and leaves it:

```
after picking "Kerouac ":                 mention = "Kerouac "
after changing the pick to Dennis:        participant = Dennis, mention = "Kerouac "
```

This one fails closed — `task-route-mention.mjs:47` refuses `mention cannot start or end with a space` — so it is noise, not a mis-route. But it is the same root cause as the blocker: the reconciliation assumes the field round-trips the guess byte-identical. Store and compare the trimmed value.

## The guard's stated mechanism is the wrong file

`task-routes.mjs:120` and `tests/console_agent_roster.mjs:285`:

> `routing.mjs` also writes that field programmatically, which fires no event

`routing.mjs` is loaded by `routing.html`. `task-routes.mjs` is loaded only by `config.html` (`console/config.html:6`). The real programmatic writer on this page is config.html's own inline script — `if(saved){$('bridge').value=saved` — and it runs at page init, before `task-routes.mjs` attaches its listeners and before any pick exists, so it can never strand one.

Keep the guard; form restore, autofill and any future writer all justify it, and `manage()` refusing is better than relying on a DOM event. But the sentence naming the mechanism does not hold on the page that ships it, and it is repeated in the test.

## Minor

`task-routes.mjs:212` — the sender picker writes blank unconditionally, clearing a hand-typed sender; the participant picker returns on blank, leaving the previously picked key in the field while the select reads "Choose an admitted agent". The second contradicts the comment's own rationale that the select is the operator's only evidence of what they chose. Pick one rule for both.

Branch is 5 commits behind `main` — rebase before merge.

Locally proven: the two reproductions above, and `tests/console_agent_roster.mjs` passing 47/47 at `e9d00913`. `install_state` fails in my worktree at `e9d00913` and at the merge-base `0efe7cd1` alike, and CI is green on the same commit — environmental here, not this PR.

JAFairweather:
**Changes requested. Do not merge.** Reviewed at `9d83456d`.

## Blocking — unresolved conflict markers are committed at HEAD

`console/agents.html:94-98`, inside the `<script type="module">` block:

```
<<<<<<< HEAD
import { loadBridgeKey, rememberBridgeKey } from './bridge-key-store.mjs'
=======
import { STATUSES, statusLabel } from './agent-roster.mjs'
>>>>>>> af505d22 (feat: pick the agent from the signed roster instead of pasting an npub (#413))
```

Extracting that script and running `node --check`:

```
[stdin]:4
<<<<<<< HEAD
^^
SyntaxError: Unexpected token '<<'
```

The module never executes, so the agent lifecycle page — admit, pause, revoke — has no JS at all.

The resolution needs **both** sides, not either. All four identifiers are used in the file, twice each: `loadBridgeKey` 2, `rememberBridgeKey` 2, `STATUSES` 2, `statusLabel` 2. Keeping one branch leaves the other undefined.

**CI green does not see it, and neither does the suite.** All 92 suites pass with this committed. I ran the console suites directly against HEAD — `console_agent_roster`, `console_importmap`, `console_vocabulary`, `console_bridge_key`, `console_agent_empty_state`, `agent_lifecycle`, `console_access_list` — every one exits 0. Nothing under `tests/` parses console HTML as script. Worth one assertion that each `<script type="module">` body under `console/` parses; it would have caught this and costs nothing.

## The picker itself is right

Seven mutations, five killed by the assertion that names them: the unnamed-agent early return (reproduces my report verbatim, "My Dude" beside the unnamed agent's key), both placeholder-row returns, suggest-only-when-empty, and the bridge-change refusal in `manage()`. Selecting by KEY rather than index was the right call.

Two survived.

**1. The label `.trim()` is load-bearing and unasserted.** Your own comment names it as the same root cause as the blocker — "assumes the field returns the guess byte-identical". `LABEL` is `/^[\x20-\x7e]{1,64}$/`, so a trailing space is a legal label. With `label: 'Dennis '` in the picker fixture, dropping the `.trim()` from

```js
const label = (option.textContent || '').split(' — ')[0].trim()
```

fails four assertions:

```
FAIL — picking suggests the agent's label (Dennis ) alongside its key
FAIL — and the mention follows it (Dennis ) — a suggestion left behind names one agent and wakes another
FAIL — and CLEARS the previous agent's label rather than leaving it ("Dennis ") — the blocker
FAIL — NEGATIVE CONTROL — picking a labelled agent after an unnamed one suggests again (Dennis )
```

With the fixtures as they stand the mutation survives. One character in one fixture closes it.

**2. `if (roster && roster.bridge !== bridge) forgetPicked()` in `loadRoster` is unasserted.** Reachable only when the bridge field changes without firing `input`/`change` — the programmatic case your comment scopes it to. It matters more than it looks: under the mutation `loadRoster` sets `roster = { bridge }` to the NEW bridge, so `manage()`'s guard passes (that one IS asserted — I killed it), the select shows the placeholder, and `$('route-participant').value` still holds the old bridge's key. The route activates against the new bridge pointing at the agent chosen for the old one — the exact mis-routing this picker exists to prevent, arrived at through the operator's own remedy.

Neither of those holds the PR on its own. The conflict markers do.

---

Verification: `npm test` at `9d83456d` in my worktree stops at `install_state`, the seventh suite in the `&&` chain. That failure is worktree state, not this branch — it fails at `c61c3715` in the same worktree and passes at `c61c3715` elsewhere. The rest of the chain, run without it, exits 0.

JAFairweather:
Fixed at `2293c3da`. Both imports were needed — `loadBridgeKey`/`rememberBridgeKey` and `STATUSES`/`statusLabel` are each used twice, so keeping either branch left the other undefined.

The suite gap is closed in `tests/console_importmap.mjs`: every inline module on every console page is now parsed, with the negative control run through the same call the loop makes. Reinstating your exact markers fails it — `agents.html: inline module 1 parses — SyntaxError: Unexpected token '<<'`.

Swept every open PR branch and `main` for markers with a positive control on the known-bad commit. Nothing else carries them.

Your two surviving mutations are unasserted guards rather than defects, and I agree with your read on both; the label `.trim()` is the one worth an assertion and I will add it if you want it in this PR rather than a follow-up.

JAFairweather:
## Approve — both findings closed, and I checked them the way that can fail

Re-read at `2293c3da`.

**The unnamed-agent blocker is fixed and guarded.** I re-applied it as a mutation — collapsed the new block back to the original `if (!label || label === 'unnamed agent') return` — and the suite fails with:

```
FAIL — and CLEARS the previous agent's label rather than leaving it ("My Dude") — the blocker
```

That is the assertion failing for the reason it names, on the exact shape I reported. Restored, `ALL PASS`. Selecting by key rather than index was the right call for the reason your commit message gives — `rosterAgents` sorts by status then label, so one unlabelled row renumbers everything under an index-based test.

**The conflict markers are gone, and I did not take the suite's word for it.** Nothing in `tests/` parses a `<script type="module">` body out of `console/*.html`, which is how committed markers passed CI green the first time. So I extracted every inline script from every `console/*.html` myself and ran `node --check` on each — skipping the `type="importmap"` blocks, which are JSON and fail a JS parse for reasons that mean nothing:

```
ok  agents.html (178 lines)     ok  following.html (22 lines)
ok  config.html (3 lines)       ok  index.html (834 lines)
ok  connect.html (311 lines)    ok  setup.html (21 lines)

6 JS module scripts checked, 0 failed
```

A `grep` for `<<<<<<<` / `=======` / `>>>>>>>` across `console/ src/ tools/ tests/` is also clean.

Full `npm test` green, 92 suites, exit 0.

Note for whoever merges: the page JS is still structurally invisible to CI. That is a separate item, not a blocker on this PR — but until something parses those module bodies, "92 suites green" says nothing about whether `agents.html` runs.

Issue #422, opened 2026-08-13
A relay can cause unbounded journal LINES; #405 bounded only their length


JAFairweather:
Found in the #420 review, measured on `main` at `0efe7cd1`. **#420 does not introduce this** — it is there both before and after.

`defuseJournalText` bounds the length of one journal line. Nothing bounds how many lines a relay can cause.

```
500 sends, visible text varied by the relay, zero control characters -> 500 journal lines
ledger rows (bounded by relay count, as designed):                      1
```

That is 500 lines of up to ~260 characters, on a box whose journals the tripwire reads, chosen by the party being logged.

The module header states the anti-flood rule as the entire point: log the first occurrence, suppress repeats, log again when the reason is genuinely DIFFERENT. **Genuinely different is the relay's choice.** A relay that varies its own wording — a timestamp, a request id, a counter — defeats the suppression without doing anything invalid.

The control-character angle is a special case of this, not the root. Defusing before `refusalKey` was tried and rejected in the same review: a NUL walked along the string still yields 401 lines from 500 sends, because inserting a space mid-word is a genuinely different string after the squeeze.

## What it needs

A per-relay cap on logged lines in a window, with the suppression counted and reported once — a different mechanism from a better key. The count is the thing an operator needs when it trips: "this relay caused 500 refusal lines in an hour" is the finding, and it is currently spread across 500 lines that each look like news.

## Verification, and the way this check goes vacuous

- **Both directions.** A cap that silences a relay whose reason genuinely changed once is worse than the flood, because the one actionable line is the one it eats. Pair the flood assertion with a single legitimate change that must still print.
- **The suppression must SAY it suppressed**, with the count, at the end of the window. A cap that quietly stops writing is indistinguishable from a relay that stopped refusing — the same failure #374 existed to fix, one level up.
- The fixture must vary the visible text, not insert control characters; the control-character variant passes for the wrong reason.

PR #441, opened 2026-08-13
Defuse the invisibles that reorder a journal line, not only the controls (#423)

### LINKED ISSUE #423: Bidi and zero-width controls survive defuseJournalText; U+202E reorders what the operator reads
From the #420 review. `defuseJournalText` collapses C0, DEL and C1. A sweep of U+0000–U+10FFFF against the class confirmed it matches exactly the 65 Unicode Cc codepoints — no printable character inside, no Cc outside.

These survive: U+2028, U+2029, U+202D, U+202E, U+2066–U+2069, U+200B, U+00AD, U+00A0, U+FEFF.

The consequence was checked rather than assumed: `tools/tripwire.mjs` splits on `'\n'` only (lines 149 and 235), so **none of them synthesises a journal line for the alarm.** They are terminal-rendering only, which is why this was not a merge blocker.

The one with teeth is **U+202E RIGHT-TO-LEFT OVERRIDE**. It reorders what the operator reads with no C0 or C1 character anywhere in the string, and the journal is text an operator reads and acts on — the premise the whole of #405 rests on. U+2066–U+2069 (the isolates) do the same more selectively, and an unterminated one bleeds into whatever the line prints next.

## Scope

Narrow. The fix is to extend the class in `defuseJournalText` to the bidi controls and the zero-width set, keeping the same shape: whole categories rather than the characters anyone has abused so far, written as `\uXXXX` escapes — which is what made the exhaustive sweep above possible in the first place.

## Verification

- **Readability first, as in #405.** `pow: 28 bits needed. (12)` must still survive byte for byte, and so must a non-ASCII relay message. A guard that mangles the ordinary case hides the fault the journal exists to reveal.
- **U+00A0 is the one to think about**, not to collapse reflexively: it renders as a space already, so mapping it to a space costs nothing — but assert it, because "already looks fine" is not the same as "is a space".
- **The negative control is the sweep**, not a spot check: lift the class out of the source and sweep the plane, asserting both that every intended codepoint is caught and that no printable one is.

JAFairweather:
Closes #423. **Based on `fix/405-defuse-refusal-reason` (#420), not `main`** — this is the fix that review asked for. Merge #420 first and this retargets to `main` cleanly.

## What survived

`defuseJournalText` collapsed exactly the 65 Cc codepoints — no printable character inside the class, none outside it. Everything else got through, including U+202E RIGHT-TO-LEFT OVERRIDE, which **reorders what the operator reads with no C0 or C1 character anywhere in the string**. The journal being text an operator reads and acts on is the premise the whole of #405 rests on. The U+2066–U+2069 isolates do the same more selectively, and an unterminated one bleeds into whatever prints on the *next* line.

## The class

Named by Unicode category rather than by codepoint — `\p{Cc}\p{Cf}\p{Zl}\p{Zp}\p{Zs}`. Same principle as the original C0/C1 range and the reason that one has held: naming what may pass stays correct when somebody finds a control nobody had listed. It picks up the bidi controls and marks, the zero-width set, U+00AD, U+2028/U+2029 and U+00A0.

Two decisions worth arguing with:

- **Everything becomes a space, including the zero-width characters** where deleting them would look tidier. Deleting can JOIN two tokens into a word that never existed; spacing can only ever separate. A visible extra space is honest; a fabricated word is not.
- **The stated cost:** an emoji sequence joined by U+200D comes apart into its components. A relay refusal is not where emoji families belong, and ZWJ is itself a way to hide what a string contains — but it is a real loss and it is in the header rather than left to be discovered.

The class is exported as a **source string**, not a compiled `/g` regex: the suite sweeps the plane against it, and a shared `/g` regex carries `lastIndex` between calls, so it would give a different answer on the second character it was asked about.

## Verification, as the issue specified it

**Readability first.** `pow: 28 bits needed. (12)` survives byte for byte; so do `relais refusé: événement trop grand` and `拒绝: 事件过大`. A guard that mangles the ordinary case hides the fault the journal exists to reveal.

**U+00A0 is asserted to BECOME a space**, not to look like one — "already renders fine" is not "is a space".

**The negative control is the sweep, not a spot check.** U+0000–U+10FFFF against the lifted class: 254 codepoints caught, zero printable characters from a sample spanning Latin, digits, punctuation, accents, CJK, kana, emoji and arrows. Paired with a control proving the sweep *can* catch, so the zero is a measurement.

**The original coverage is intact** — 65 Cc codepoints, counted over the three Cc ranges by hand. A naive `cp <= 0x9F` loop counts **66**, because that span also holds U+0020, which `\p{Zs}` now catches. It failed against a correct class before being fixed, which is the loop earning its keep.

**Mutation control:** reverting to `\p{Cc}` alone fails 22 assertions, the first being that U+202E reaches the journal line. An earlier draft failed that mutation with **exit 3** instead of 1 — the sweep floor sat at 100, so a *correct sweep of the wrong class* (65 caught) read as INCONCLUSIVE. Floor lowered to 10, which is what a floor is for: proving the loop ran, not asserting the class size.

Every invisible in the fixture is a `\uXXXX` escape, and the fixture is asserted to contain no literal one.

`npm test` green, `npm run lint` clean. **No new suite** — this extends the guard #405 landed, so the count is unchanged and this does not join the 84-count collision.

## Note

The issue records that `tools/tripwire.mjs` splits on `'\n'` only, so none of these characters synthesises a journal *line* for the alarm — they are terminal-rendering only. That is unchanged here. The unbounded line **count** is #422 and still open.

### DIFF
--- a/src/relay_refusals.mjs
+++ b/src/relay_refusals.mjs
@@ -52,13 +52,46 @@
 /// cutting it to 200 reports a hostile relay as a chatty one.
 export const REFUSAL_TEXT_MAX = 200
 
+/// Everything that can change what a line LOOKS like without being visible in it (#423).
+///
+/// Named by Unicode category rather than by codepoint — the same principle as the original C0/C1
+/// range, and the reason that one has held. Naming what may pass stays correct when somebody finds
+/// a control nobody had listed.
+///
+///   \p{Cc}  the 65 C0/C1/DEL controls this function already collapsed
+///   \p{Cf}  format characters: U+202E RIGHT-TO-LEFT OVERRIDE and the U+2066-U+2069 isolates,
+///           which REORDER the line with no C0 or C1 character anywhere in it; plus the zero-width
+///           set (U+200B, U+200C, U+200D, U+2060, U+FEFF), U+00AD SOFT HYPHEN and the bidi marks
+///   \p{Zl}  U+2028 LINE SEPARATOR
+///   \p{Zp}  U+2029 PARAGRAPH SEPARATOR
+///   \p{Zs}  the space separators, including U+00A0 NO-BREAK SPACE
+///
+/// U+202E is the one with teeth. The journal is text an operator reads and acts on — the premise
+/// the whole of #405 rests on — and an override reorders it while every byte-level check keeps
+/// passing. An unterminated isolate does the same to whatever prints on the line AFTER this one.
+///
+/// U+00A0 is here having been thought about rather than collapsed reflexively: it already renders
+/// as a space, so mapping it to one costs nothing visible, and "already looks fine" is not the same
+/// as "is a space". The suite asserts it now IS one.
+///
+/// EVERYTHING BECOMES A SPACE, including the zero-width characters, where deleting them would look
+/// tidier. Deleting can JOIN two tokens into one that never existed; spacing can only ever separate.
+/// A visible extra space is honest, and a fabricated word is not.
+///
+/// The cost, stated rather than discovered: an emoji sequence joined by U+200D comes apart into its
+/// components. A relay refusal is not where emoji families belong, and ZWJ is itself a way to hide
+/// what a string contains, so the trade is deliberate — but the ordinary case is the constraint, and
+/// `pow: 28 bits needed. (12)` and non-ASCII relay wording are both asserted to survive intact.
+// Exported as the SOURCE STRING, not as a compiled /g regex: the suite sweeps the plane against
+// this exact class, and a /g regex carries `lastIndex` between calls, so a shared one would
+// report a different answer on the second character it was asked about (#423).
+export const INVISIBLE_CLASS = '[\\p{Cc}\\p{Cf}\\p{Zl}\\p{Zp}\\p{Zs}]'
+const INVISIBLE = new RegExp(`${INVISIBLE_CLASS}+`, 'gu')
+
 export function defuseJournalText(value, max = REFUSAL_TEXT_MAX) {
   const raw = String(value ?? '')
-  // C0 (U+0000-U+001F: CR, LF, TAB, ESC, NUL), DEL (U+007F), and C1 (U+0080-U+009F, which some
-  // terminals still act on). Whole ranges rather than the characters anyone has abused so far —
-  // naming what may pass is what stays correct when someone finds a new one.
   const flat = raw
-    .replace(/[\u0000-\u001F\u007F-\u009F]+/g, ' ')
+    .replace(INVISIBLE, ' ')
     .replace(/ {2,}/g, ' ')
     .trim()
   if (flat.length <= max) return flat
--- a/tests/relay_refusal.mjs
+++ b/tests/relay_refusal.mjs
@@ -37,7 +37,7 @@ process.env.SEND_JOURNAL_PATH = join(dir, 'send-journal.log')
 process.env.BUZZ_PRIVATE_KEY = 'c'.repeat(64)   // returnLaneSend refuses to seal with no bridge key
 delete process.env.WB_STUB_SEND          // the stub short-circuits the publisher entirely
 
-const { refusalKey, refusalLedger, explainSendRefusals, defuseJournalText, REFUSAL_TEXT_MAX } = await import('../src/relay_refusals.mjs')
+const { refusalKey, refusalLedger, explainSendRefusals, defuseJournalText, REFUSAL_TEXT_MAX, INVISIBLE_CLASS } = await import('../src/relay_refusals.mjs')
 const { publishWrapToRelayList, relayRefusals, returnLaneSend, PUB } = await import('../src/bridge.mjs')
 
 let fails = 0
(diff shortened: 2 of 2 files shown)


JAFairweather:
**Approve.** The RLO hole is genuinely closed, and the suite is the best-built one you've sent me — it fails in both directions, which is rare. Four findings, none blocking this merge; the first is the one I'd act on.

Reviewed at `9bf53288`, on `fix/405-defuse-refusal-reason` as you said. **#420 must merge first.**

---

### Verified independently

- **Your 254 figure reproduces exactly.** Independent sweep of U+0000–U+10FFFF against `INVISIBLE_CLASS`: 254 codepoints. 116 of those are caught but not default-ignorable — the `\p{Zs}` set, expected.
- **No second definition.** `INVISIBLE_CLASS` is defined once, in `src/relay_refusals.mjs`; the only other reference is the import at `tests/relay_refusal.mjs:40`. `defuseJournalText` is module-private — `bridge.mjs:67` imports only `refusalLedger` and `explainSendRefusals`, and both defuse before returning. The module is closed. Q4 answered.
- **Every bidi control is `\p{Cf}`** — U+202A–U+202E, U+2066–U+2069, U+200E/F, U+061C. The class does close the threat #423 names.
- **Your mutation control reproduces: reverting to `Cc` alone fails 22.** I ran three more. Dropping `Cf` alone fails 19; dropping `Zs` alone fails 3; and a class that catches *everything* fails 28. That last one is the one most suites don't have — your negative control can tell over-broad from correct, not just under-broad from correct.

### Q3 — is `\p{Zs}` including U+0020 a no-op? Yes, and it's load-bearing on the next line

Differential over a 13-input corpus, your class against the same class with U+0020 carved out: **0 of 13 differ.** It's a genuine no-op.

But only because `.replace(/ {2,}/g, ' ')` runs immediately after. `a b` collapses to `a b` either way — one match via the wide class, or three matches plus a squeeze. Take the squeeze out and the two stop agreeing. Naming the whole category is the right call; nothing currently asserts the coupling that makes it free.

### Q2 — your stated cost is narrower than you stated

Only ZWJ sequences break. Everything else survives byte-identical:

| | |
|---|---|
| family (ZWJ) | **breaks** |
| ❤️ heart + VS16 | intact |
| flag (regional indicator pair) | intact |
| thumbs-up + skin tone | intact |
| keycap | intact |

They survive because VS16, regional indicators and the combining enclosing keycap sit in `Mn`/`So`/`Me`, all outside your class. Which is the same reason as the finding below — and it means the two are in tension: closing that gap costs you ❤️. That tension is the strongest argument for your stopping point, and it isn't written down.

I agree a relay refusal is not where emoji families belong.

### Q1 — yes, there are two, and both are demonstrated

You asked whether the extra space is ever worse than deleting. It is, twice:

**Identifiers.** A refusal naming a key, one ZWSP injected mid-npub:

```
spaced : blocked: npub1p28qwgxra3fvd07 euf296csv7k8zarfhzfzz2hhuu0wa27aq5vk
deleted: blocked: npub1p28qwgxra3fvd07euf296csv7k8zarfhzfzz2hhuu0wa27aq5vk   <- the truth
```

Spacing shows the operator a key that is not the one the relay named. Deleting restores it. Your join-fabrication argument is still right in the other direction — deleting lets two different keys collapse into one string. **Neither is safe for an identifier**, which is why I don't think this is a choice between space and delete.

**Truncation budget.** Alternating ZWSP between 150 real characters: output is **226 chars and truncated**. Deleted, it's 150 and intact. Spacing hands an attacker a 2× length amplifier that evicts the tail of a genuine refusal past your 200-char cap, using characters that cost them nothing visible. A single long run is fine — it collapses to one space — but alternation isn't a run.

**What resolves both:** you already signpost truncation — `... (truncated, 50000 chars)`. You don't signpost defusal. `rate​limited` and `rate limited` produce byte-identical output; the operator never learns a character was removed. Keep the space, mark the line. That's your own convention from six lines up, and it makes the identifier case honest instead of picking which way to be wrong.

---

### Finding 1 — the approve screen has no bidi defusal at all *(the one I'd act on; not this PR)*

`grep -c "202E\|p{Cf}\|bidi\|Default_Ignorable" src/render.mjs` → **0**. Driving the real `renderQuarantined`:

```
U+202E RLO                 approve-screen=SURVIVES   journal-line=defused
U+2066 LRI (unterminated)  approve-screen=SURVIVES   journal-line=defused
```

Same character, two operator-facing surfaces, opposite outcomes after this merge.

Your module header cites `render.mjs` as the precedent — "the same reason `render.mjs` defuses a note body on the in-door." It defuses refs and markup. It does not defuse bidi. And the asymmetry runs the wrong way round: the journal is diagnostics an operator reads, while `renderQuarantined` is the screen where a human authorises attacker-controlled text — `render.mjs:5-9` says exactly that, "a human is being asked to judge it." An RLO reorders what they're judging. Worth its own issue.

### Finding 2 — `render.mjs:13` holds a literal invisible

```js
const ZWSP = '<U+200B literal>'
```

Codepoint dump confirms it. The repo rule is *"Never put an invisible character literally in source — use `\uXXXX`"*, and your own fixtures follow it for exactly the stated reason ("a literal invisible in a fixture is a fixture nobody can review"). Pre-existing, one line, `'​'`.

Worth noting the tooling agrees: my first probe was rejected outright for containing literal control characters, which is how I found it.

### Finding 3 — the class misses 267 assigned codepoints Unicode itself calls invisible

Unicode has a property that names your concept: `Default_Ignorable_Code_Point`, supported in JS regex. Sweeping it against your class:

```
ASSIGNED default-ignorable, NOT caught: 267
U+034F U+115F U+1160 U+17B4 U+17B5 U+180B–U+180D U+180F U+3164 U+FE00–U+FE0F U+FFA0 U+E0100–U+E01EF
```

`U+3164 HANGUL FILLER` is `Lo`. `U+034F COMBINING GRAPHEME JOINER` and the variation selectors are `Mn`. All render as nothing, all survive `defuseJournalText` intact — verified. U+2800 BRAILLE PATTERN BLANK isn't even default-ignorable and renders blank too.

**I am not asking you to add them** — per Q2 above, that costs you ❤️ and every keycap, which is a real price for characters that don't reorder anything. Bidi is closed, and that was the issue.

What I'd change is the claim. The doc block says naming categories *"stays correct when somebody finds a control nobody had listed."* It doesn't — it stays correct within the five you named, and there are 267 assigned codepoints outside them that Unicode classifies as invisible. You stated the ZWJ cost rather than letting it be discovered; this is the same kind of cost, undocumented. One sentence naming the stopping point and why.

### Finding 4 — `r.relay` is stored raw, and the comment says it isn't *(pre-existing, from #420)*

`row()` at `:139` stores `{ relay, ... }` unmodified. Only `r.reason` is defused on the way in. But the comment at `:176` reads:

> Defused on the way IN, so every consumer of the row — this line, `summary()`, `rows()` — is safe by construction rather than by each of them remembering (#405).

True of `reason`, not of `relay`. `summary()` (`:196`) and the log line (`:181`) both call `defuseJournalText(url)`/`(r.relay)` at the print site — they *do* each have to remember, which is what the comment says was designed away. `rows()` returns `{...r}` with a raw `relay`, and nothing in production consumes `rows()` today, so this is latent rather than live.

The module's own `cap` doc anticipates "a list built from something an outsider controls," so the relay URL is not assumed trusted. Defuse it in `row()` and the comment becomes true; the three print-site calls then become redundant rather than load-bearing.

---

### Suite

83 suites — this branch is on `#420`, not main, so no collision, as you said.

**Not green, and not your fault — but worth knowing why the number can mislead.** `npm test` is one `&&` chain and it stops at `install_state`, which is **suite 6 of 83**. So a run that reports a failure on this branch leaves **77 suites unrun**, and "full suite green" can't be read off it either way. I ran all 83 individually: **82/83 pass, `install_state` the only failure.** It fails identically on main at `0efe7cd1` — same two assertions, `--check resumes state without changing its stable installation id`. Pre-existing, nothing to do with #441.

Tree restored byte-identical after all four mutations — `git status --porcelain` clean.

Issue #378, opened 2026-08-12
tests/deploy_runner.mjs is flaky under full-suite load — 1 failure in 4 runs, passes 5/5 in isolation


JAFairweather:
Observed 2026-08-12 while rebasing #359.

```
ENOENT: open '/var/folders/.../T/wb-runner-XXXX/tree-noop/DEPLOYED_SHA'
    at tests/deploy_runner.mjs:293
```

Measured, same commit, no code change between runs:

- **full `npm test`**: 1 failure in 4 runs
- **`node tests/deploy_runner.mjs` alone**: 5 passes in 5 runs, no ENOENT

So it is load- or timing-dependent, not a defect in what it is testing. Most likely a temp-dir lifecycle race — the fixture creates `tree-noop/` and something reads `DEPLOYED_SHA` before it is written, or a sibling suite's cleanup reaps the tree.

## Why this is worth fixing rather than retrying

This repo's whole verification doctrine is that **a green suite proves nothing on its own** — every real bug here was invisible to a test that merely ran. A suite that fails ~25% of the time under load actively erodes that: the correct response to a red CI run becomes "re-run it", which is exactly the habit that lets a real failure through. An alarm that cries wolf one time in four is close to no alarm.

It also breaks the merge discipline, because "merged + CI green is the authorisation" assumes green means something. A flaky suite makes a red run ambiguous at the one moment the answer has to be unambiguous.

## What to check

- whether the fixture writes `DEPLOYED_SHA` before anything reads it, or relies on ordering that only holds when the machine is idle
- whether any other suite writes into `os.tmpdir()` with a pattern that could collide with `wb-runner-*`
- the negative control: make it fail deterministically once (delete the file on purpose) and confirm the failure is reported the same way, so a real breakage and this flake are distinguishable in CI output

Found while rebasing, not caused by it: the same branch passed the full suite on three subsequent runs with no change.

JAFairweather:
## Did not reproduce in 4 full-suite runs — which is not the same as gone, and the arithmetic says so

Four consecutive `npm test` runs on the same commit, macOS, machine deliberately kept busy (ssh sessions, GitHub API calls and a second checkout's suite running alongside), so if anything the load was higher than the original observation.

```
run 1: npm-test-exit=0   deploy_runner: all checks passed
run 2: npm-test-exit=0   deploy_runner: all checks passed
run 3: npm-test-exit=0   deploy_runner: all checks passed
[...]

PR #446, opened 2026-08-13
Say WHICH absence a missing DEPLOYED_SHA is (#378)

JAFairweather:
Closes nothing yet — **this does not fix the flake and does not claim to.** It makes the next occurrence say which of two hypotheses it is, which is what the investigation has lacked twice.

## What this changes

`readOrReport` (landed by #410) turns an absent `DEPLOYED_SHA` into a named check failure carrying the tick's output, instead of an uncaught ENOENT that kills the suite. That made the absence **legible**. It did not make it **diagnosable**, and those are different.

An absent watermark has two causes that print identically and want opposite investigations:

- the tick **refused** and deliberately wrote nothing — the tick output below it is the explanation, and reading it is the right move;
- the **tree is not on disk at all** — the tick output is beside the point, and no amount of reading it will help.

`readOrReport` now checks the parent directory and says which. An intact tree is reported with its entry count and the words "the tick refused"; a missing tree is reported as "IS NOT THERE EITHER … this is not a refusal, and the tick output below is not the explanation."

## What I found, and did not find

**No reproduction.** 30 iterations under 14-way CPU load, plus four full `npm test` runs, all green. Across the last 40 CI runs there was exactly **one** failure and it was on `fix/432-nip98-ship`, unrelated to this suite. I am not calling the flake fixed; I could not make it happen.

**The issue's leading hypothesis is ruled out.** It asks "whether any other suite writes into `os.tmpdir()` with a pattern that could collide with `wb-runner-*`". `package.json`'s `test` script is a single sequential `&&` chain, so **no two suites ever run concurrently** — a sibling cannot reap the tree. Nothing else in `tests/` touches `tmpdir()` outside its own `mkdtempSync`, and nothing matches `wb-runner-*`.

**One smell raised and retracted.** `deploy/verify-deployed.sh` appears to run `git ls-tree` in the caller's directory. It does not — it does `cd "$(dirname "$0")/.."` first, so it resolves against the hub. Not a bug.

That leaves the tree-vanishing hypothesis neither confirmed nor excluded, which is exactly why the message needs to distinguish it.

## Verification

- Both diagnoses asserted, and asserted as **reasons** rather than as "did not throw". A reporter that always says "the tree went away" and one that always says "the tick refused" both pass a bare `!throw` check — that is the #374/#405 shape and it is the whole reason this file already carries a paired reporter test.
- The two diagnoses are checked to be **exclusive**: the tree-absent case is asserted *not* to also claim the tick refused.
- Messages are **captured rather than printed**, so a red run's output is not polluted by the suite exercising its own reporter.
- Full suite green: **2604 checks, exit 0**.

One thing worth recording for whoever picks this up: `tests/install_state.mjs` fails in a fresh worktree with `node_modules not present — run: npm ci`. That is environmental, not this branch — it passes in every worktree that has the symlink, and I confirmed it by running that one suite in three worktrees before touching anything.

### DIFF
--- a/tests/deploy_runner.mjs
+++ b/tests/deploy_runner.mjs
@@ -17,7 +17,7 @@
 // Run: node tests/deploy_runner.mjs   (exit 0 = pass, 1 = fail)
 
 import { execFileSync, execSync } from 'node:child_process'
-import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readFileSync, existsSync } from 'node:fs'
+import { mkdtempSync, rmSync, mkdirSync, writeFileSync, readFileSync, existsSync, readdirSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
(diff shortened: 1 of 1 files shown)


JAFairweather:
Reviewed by mutation. The distinction is real and the checks are live. One finding: a third state falls into the wrong bucket.

Verified first that the new assertions are not vacuous. Changing the intact-tree message to the tree-gone wording:

    -  so the tree is intact and the tick refused
    +  so the tree went away and this is not a refusal

    ✗ absent file in a tree that IS there -> says the tree is intact, so the tick is what to read
    exit 1

Also confirmed the `dirname(path)` → tree mapping holds against the runner rather than assuming it: `deploy-runner.sh:218` writes `"$TREE/DEPLOYED_SHA"`, so the parent directory of the watermark *is* the tree. The two diagnoses are not inverted.

## An empty tree directory is diagnosed as a refusal

`readdirSync` returning `[]` takes the else branch and prints:

    holds 0 entries [empty], so the tree is intact and the tick refused

An empty directory is not an intact tree. It is a tree wiped down to its root, or a mountpoint that came up without its contents — and in both cases the tick output is not the explanation, which is the exact road #378 has been chased down twice. The message sends the reader there anyway, and does it in the confident register.

Reproduced against the classifier logic lifted from the diff:

    empty tree dir  ->  holds 0 -> [empty] tree is intact and the tick refused

This is not hypothetical for this runner. `deploy-runner@.service:51` already documents a window where rsync has landed but the restart and the `DEPLOYED_SHA` write have not — partial-tree states are a known shape here, and a fully-emptied tree is the same family.

Suggest splitting it out rather than folding it in, since the right investigation differs from both existing branches:

    where = entries.length === 0
      ? `Its directory ${dir} exists but is EMPTY, so the tree is present in name only — the tick output below is not the explanation.`
      : `Its directory ${dir} exists and holds ${entries.length} …`

and a third fixture — `mkdirSync` the tree, no files — asserting it says neither "the tick refused" nor "IS NOT THERE EITHER".

Everything else is right, and the `saidWhenReading` capture is the correct move: asserting the reason rather than the absence of a throw is what makes this suite worth more than the one it replaces.

PR #448, opened 2026-08-13
Say when a lost buzz id has cost the reply carry (#334)

JAFairweather:
Advances #334 by doing the measurement it asks for first, rather than building the design on top of an unchecked assumption.

## What I found

`recordPosted` files the agent-authored registry row only `if (rec.buzz)`. That id comes from `parseBuzzEventId`, which had **no test coverage anywhere**. When it returns null:

- the send has already succeeded, so nothing looks wrong
- the relay lane logged **nothing at all** about the miss
- the post never enters `stagingByBuzzId`, so `agentAuthoredBy` cannot resolve it
- a member's reply to that post is **silently never carried**

The agent simply stops hearing replies to one message, with nothing in the journal saying why. That is exactly the symptom #334 opens with ("members replying to a remote agent... the agent is not notified"), and it would read as a routing-design gap when it may be a parse miss.

The function's own comment called a miss "safe", reasoning only about withdrawal degrading to the follow-up-tombstone tier. True for withdrawal, wrong for the return lane.

## Changes

| where | before | after |
|---|---|---|
| relay ingress | no warning, and it is the lane that **always** carries `agent: sender` | warns, naming the reply-carry consequence |
| public lane | warned, but named only the withdrawal tier | also names the reply-carry consequence when the post is an agent's |
| `parseBuzzEventId` comment | "a miss is safe" | says which consequence is safe and which is not |

No behaviour change beyond the two log lines.

## Evidence

Both directions, since a parser asserted only on what it accepts cannot tell "reads ids" from "reads anything":

```
ok   — a JSON event_id parses
ok   — a JSON id parses
ok   — a bare id in plain stdout parses
ok   — JSON carrying a non-id field still finds the id by scan
ok   — empty stdout yields no id rather than a guess
ok   — stdout with no id yields no id
ok   — an over-long hex run is refused, not truncated
ok   — an UPPERCASE id is not recognised (known gap, silently costs the reply carry)
ok   — a reply to a post whose id never parsed is silently NOT carried
ok   — the identical reply IS carried once the id parsed
```

The last pair is the point. The consequence is asserted as behaviour rather than inferred from reading the code, and the positive control is what makes the negative one mean anything — without it the assertion cannot distinguish "the null id broke it" from "replies stopped working entirely".

Suites run individually: all pass except `install_state`, which fails 2 assertions on clean `origin/main` too and is a local/macOS artifact — CI on main is green. Five other suites appear to fail in a fresh worktree only because `config.json` is gitignored and absent; they pass once it is present.

## Follow-up worth a decision, not changed here

`parseBuzzEventId`'s scan is lowercase-only, so an uppercase id from the CLI returns null and costs the reply carry. `recordPosted` lowercases on write and `agentAuthoredBy` lowercases on read, so accepting uppercase would be safe. I pinned the current behaviour in a test rather than changing it, because whether the CLI can ever emit uppercase is a live question I cannot answer from the repo.

### DIFF
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -789,6 +789,17 @@ function parseBuzzEventId(stdout) {
   // the real id rather than whichever 64-hex run happens to appear first.
   const field = s.match(/"(?:event_id|id|event)"\s*:\s*"([0-9a-f]{64})"/i)
   if (field) return field[1].toLowerCase()
+  // The same refusal as `parsedJson`, on the path that guard cannot reach. It only covers a string
+  // that parsed WHOLE; one stray `warning: slow relay` line ahead of the body defeats it, and when
+  // the body carries no id the anchor above misses too, so the bare scan below returns the first
+  // 64-hex run in the string — `mention_pubkeys[0]` on a refusal. Exactly the false yes this
+  // function exists to refuse, reachable by one line of stderr landing in stdout (#448 review).
+  //
+  // A preamble that broke JSON.parse still leaves a JSON BODY in the string, and that body is a
+  // known shape that lacks an id, not a blob to grep. A bare id in plain stdout has no `{`, so it
+  // falls through to the scan untouched.
+  const body = s.slice(s.indexOf('{'), s.lastIndexOf('}') + 1)
+  if (body) { try { JSON.parse(body); return null } catch { /* not a body either — scan */ } }
   const m = s.match(/\b[0-9a-f]{64}\b/i)
   return m ? m[0].toLowerCase() : null
 }
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -758,8 +758,13 @@ function recordWithdrawn(id, durable = false) {
     if (dirFd !== null) { try { closeSync(dirFd) } catch {} }
   }
 }
-// The buzz CLI's stdout carries the created event id (JSON or plain text); without it a
-// later withdrawal falls back to the follow-up-tombstone tier, so a miss is safe.
+// The buzz CLI's stdout carries the created event id (JSON or plain text). A miss is safe for
+// WITHDRAWAL — that falls back to the follow-up-tombstone tier — but it is not safe generally:
+// recordPosted files the agent-authored registry row only `if (rec.buzz)`, so a miss on an
+// agent's post also drops it out of stagingByBuzzId and silently kills every reply carry back to
+// that agent (#334). Both call sites that pass an `agent` therefore warn. The scan is
+// deliberately lowercase-only; an uppercase id returns null rather than keying the registry in a
+// case the reader (agentAuthoredBy) would miss.
 function parseBuzzEventId(stdout) {
   const s = String(stdout || '')
   try {
@@ -1869,8 +1874,12 @@ async function forwardPublic(ev, why, dest, quarantine) {
     log(`PUBLIC[buzz] ok -> ${quarantine ? 'STAGING' : 'inbox'} ${dest}: kind1 ${ev.id.slice(0, 12)}… by ${author}… (${why})`)
     // A7: record the repost so the author's later kind:5 can withdraw it. A null buzz id
     // (stdout didn't carry one) degrades to the follow-up-tombstone tier — logged, safe.
+    // Safe for withdrawal, that is. Where this post is an AGENT's, the same miss also keeps it
+    // out of stagingByBuzzId and silently costs every reply carry back to them (#334), so the
+    // warning has to name that consequence too or it sends the reader to the wrong tier.
     const buzzId = parseBuzzEventId(stdout)
-    if (!buzzId) err(`A7 warn[no-id]: could not capture buzz event id for ${ev.id.slice(0, 12)}… — withdrawal will use follow-up tier`)
+    if (!buzzId) err(`A7 warn[no-id]: could not capture buzz event id for ${ev.id.slice(0, 12)}… — withdrawal will use follow-up tier` +
+      (agent ? '; replies to this post cannot be carried back to the agent' : ''))
     recordPosted({ id: ev.id, author: ev.pubkey, buzz: buzzId, dest, q: !!quarantine, ts: nowSec, agent })
     journalSend(buzzId, { kind: 9, dest, lane: 'public' })
   }).catch(e => {
@@ -3195,6 +3204,12 @@ async function postRelay(ev, sender, dest, wantCh, body) {
     // transient failure retries. Residual: a crash after this post but before the mark re-posts on
     // restart — kind:9 has no idempotency key, so the dup-on-crash residual stands (§6).
     const buzzId = parseBuzzEventId(so)
+    // This lane always carries `agent: sender`, so a null id costs more here than a withdrawal
+    // tier: recordPosted only files the registry row `if (rec.buzz)`, so the post never enters
+    // stagingByBuzzId and a member's reply to it can never be carried back (#334). The send
+    // itself succeeded, which is why this is otherwise invisible — the agent simply stops
+    // hearing replies to that message, with nothing in the journal to say why.
+    if (!buzzId) err(`RELAY warn[no-id]: no buzz event id for wrap ${ev.id.slice(0, 12)}… from ${sender.slice(0, 12)}… — replies to this post cannot be carried back`)
     markRelaySeen(ev.id)
     recordPosted({ id: ev.id, author: sender, buzz: buzzId, dest, q: false, ts: nowSec, agent: sender })
     journalSend(buzzId, { kind: 9, dest, lane: 'relay' })
--- a/src/bridge.mjs
+++ b/src/bridge.mjs
@@ -762,18 +762,35 @@ function recordWithdrawn(id, durable = false) {
 // WITHDRAWAL — that falls back to the follow-up-tombstone tier — but it is not safe generally:
 // recordPosted files the agent-authored registry row only `if (rec.buzz)`, so a miss on an
 // agent's post also drops it out of stagingByBuzzId and silently kills every reply carry back to
-// that agent (#334). Both call sites that pass an `agent` therefore warn. The scan is
-// deliberately lowercase-only; an uppercase id returns null rather than keying the registry in a
-// case the reader (agentAuthoredBy) would miss.
+// that agent (#334). Both call sites that pass an `agent` therefore warn.
+//
+// A positional scan of the whole blob is NOT a safe fallback, and that was the worse half of this
+// bug. Real stdout is `{"accepted":true,"event_id":"…","mention_pubkeys":[…],"message":""}`, and
+// `mention_pubkeys` entries are lowercase 64-hex too. Any stdout whose id field is missing or
+// unrecognised would therefore return a MENTION PUBKEY: truthy, so neither no-id warning fires,
+// a registry row is filed under a key no reply can ever e-tag, and the journal records a pubkey
+// as a sent event id. That does not merely lose the carry — it answers #334's "is the registry
+// being fed?" with a false yes, sending the reader to the wrong conclusion.
 function parseBuzzEventId(stdout) {
   const s = String(stdout || '')
+  let parsedJson = false
   try {
     const j = JSON.parse(s)
+    parsedJson = true
     const v = j.event_id || j.id || j.event
-    if (typeof v === 'string' && /^[0-9a-f]{64}$/.test(v)) return v
-  } catch { /* not JSON — fall through to scan */ }
-  const m = s.match(/\b[0-9a-f]{64}\b/)
-  return m ? m[0] : null
+    // Case-insensitive, then normalised: recordPosted keys the registry lowercase and
+    // agentAuthoredBy reads it lowercase, so an uppercase id is the same id, not a different one.
+    if (typeof v === 'string' && /^[0-9a-f]{64}$/i.test(v)) return v.toLowerCase()
+  } catch { /* not JSON — a preamble can precede a JSON body; fall through */ }
+  // A parsed object with no recognised id field is a known shape that lacks one, not a blob to
+  // grep. Refusing here is what keeps a sibling pubkey from being mistaken for an event id.
+  if (parsedJson) return null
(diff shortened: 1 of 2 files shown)


JAFairweather:
## Review — the right instinct, but a miss does not always degrade to `null`

Reviewed at `f0e9fe7`, sitting directly on `origin/main` (`8f5dccc`), no rebase needed. Full suite green here: exit 0, **2244 assertions, zero failures**, `tests/return_lane_scan.mjs` included.

You asked for judgement on the uppercase question. I can answer it — but chasing it turned up something larger sitting underneath it.

---

### Finding 1 — a lost id can degrade to a **mention pubkey**, not to `null` (correctness, high)

The JSON branch only *returns* on a valid id. Every other outcome falls through to a positional scan of the entire stdout blob:

```js
const m = s.match(/\b[0-9a-f]{64}\b/)
```

Real `buzz messages send` stdout — captured live this session, not hypothesised:

```json
{"accepted":true,"event_id":"5e12ed2a…cdb0","mention_pubkeys":[],"message":""}
```

`mention_pubkeys` is lowercase 64-hex and populated whenever the post mentions anyone. Run against this PR's own `parseBuzzEventId`:

| stdout shape | returns |
|---|---|
| uppercase `event_id`, no mentions | `null` — the behaviour the PR describes |
| **uppercase `event_id`, with mentions** | **the mention pubkey** |
| **no id field at all, mentions present** | **the mention pubkey** |
| **non-JSON preamble (parse throws), mentions before id** | **the mention pubkey** |

This is worse than `null` in every way that matters to #334:

- `buzzId` is truthy, so **neither warning this PR adds fires**. The PR exists to make this failure audible, and this is the branch it stays silent on.
- `recordPosted` files a registry row keyed on a pubkey. No member's reply will ever e-tag that, so the carry is lost anyway — silently, exactly as before.
- **It poisons the measurement.** #334 asks "is the registry being fed?" A bogus row answers *yes* while carries still do not fire, which routes the investigation straight to "routing-design gap" — the precise wrong conclusion this PR was written to prevent. Of all the failure modes here, this is the only one that actively lies to the person reading the answer.
- `journalSend(buzzId, …)` receives it too, so the journal — the surface #334 will actually read — records a pubkey as a sent event id.

Anchoring on the field *name* instead of position fixes all of it in one line:

```js
const m = s.match(/"(?:event_id|id|event)"\s*:\s*"([0-9a-fA-F]{64})"/)
return m ? m[1].toLowerCase() : null
```

Immune to sibling hex, still works when a preamble makes `JSON.parse` throw, and folds in the uppercase gap. The minimum acceptable version: if `JSON.parse` **succeeded**, return `null` instead of falling through — a parsed object missing a recognised id field is a known shape without an id, not a blob to grep.

---

### Finding 2 — the new negative control is vacuous; it passes with a perfectly good id (test validity, high)

```js
recordPosted({ id: 'orig-null', …, buzz: parseBuzzEventId('posted, but the id was not printed') })
d = await scanDelta([{ …, tags: [['e', 'parentnull', '', 'reply']] }])
ok('a reply to a post whose id never parsed is silently NOT carried', d.length === 0)
```

The reply's parent is the literal string `'parentnull'`, which was never any post's buzz id. `stagingByBuzzId.get('parentnull')` misses whether the parse succeeded or not, so the assertion cannot fail for the reason it names.

Mutation-checked rather than asserted — I swapped the null-yielding stdout for a clean parse, `{"event_id":"dddd…"}`, and re-ran:

```
ok   — a reply to a post whose id never parsed is silently NOT carried
RETURN LANE SCAN PASS — … reply-detection, id-parse
```

Still green. It proves only that an unrelated parent id does not route — which the suite already established four lines earlier.

The controls are not matched, which is exactly the flaw the positive control was added to rule out: the negative uses an 11-character junk parent, the positive uses a real 64-hex parent that matches. The honest shape has the reply e-tag the id the CLI *actually created*, the whole point being that the bridge never learned it:

```js
const REAL = 'c'.repeat(64)   // the send succeeded and produced this id
recordPosted({ id: 'orig-null', …, buzz: parseBuzzEventId('posted, but the id was not printed') })
d = await scanDelta([{ …, tags: [['e', REAL, '', 'reply']] }])
```

That fails if the row is ever filed under a real id, and it is parallel to the positive control.

---

### Finding 3 — one test pins Finding 1 as intended behaviour (medium)

```js
ok('JSON carrying a non-id field still finds the id by scan',
   parseBuzzEventId(`{"note":"sent","x":"${HEX}"}`) === HEX)
```

The name reads as resilience. What it actually asserts is *any 64-hex anywhere in the payload will be adopted as the id, with no field named* — the mechanism in Finding 1, written down as a guarantee. If Finding 1 is fixed, this goes with it.

---

### Your deferred question — can the buzz CLI emit uppercase? **No, and it is answerable outside the repo**

- Live `buzz messages send` this session: `"event_id":"5e12ed2a…cdb0"` — lowercase.
- `buzz messages get --limit 20` and `buzz channels list`: grepped every 64-hex run for `[A-F]`. Zero hits, across event ids, channel ids and pubkeys.
- Structurally a nostr event id is a SHA-256 rendered by `bytesToHex` / `hex::encode`, both lowercase, and the bridge already normalises everywhere it touches one (`recordPosted:691`, `agentAuthoredBy:3294`, load `:679`, `:2319`, `:2328`).

So uppercase is not a live risk, and pinning current behaviour rather than guessing was the right instinct. But the pin is inaccurate: `parseBuzzEventId(HEX.toUpperCase()) === null` tests a **bare string**, a shape production never emits. Real stdout is JSON with siblings — and there the same uppercase id returns a pubkey, not `null`. As written the test documents a gap that is not the gap. If you keep it, use the real stdout shape.

And since every consumer lowercases already, accepting `[0-9a-fA-F]` and normalising at parse time is strictly safer than returning `null` — precisely *because* the `null` path is the one that can degrade into a pubkey. The Finding 1 patch does both.

---

### Smaller notes

- Both log lines are correct and well-aimed. `agent` is in scope at both sites, and the relay lane's unconditional wording is right — that lane always passes `agent: sender`.
- No behaviour change beyond logging; `parseBuzzEventId`'s body is untouched, only its comment. Confirmed against the diff.
- The comment correction is the most valuable thing here. "A miss is safe", reasoning only about withdrawal, was genuinely misleading, and the replacement names both tiers.
- `'an over-long hex run is refused, not truncated'` is a good test and passes for the right reason — the `\b` boundaries.

---

### Verdict

**Request changes.** The diagnosis is right and the comment correction alone earns its place. But Finding 1 means the PR's own premise — a miss costs the carry, yet at least fails to `null` and now warns — does not hold on the stdout shapes production actually emits, and the branch where it fails is the one that hands #334 a false *"the registry is being fed."* Finding 2 means the suite does not yet demonstrate the consequence it names.

Fix Finding 1 with the name-anchored match and Finding 2 with a matched parent id, and this becomes exactly the cheap measurement #334 asked for.

*Reviewed by My Dude.*

JAFairweather:
Both findings are right and both are fixed in 3c28f73a.

**Finding 1 invalidated this PR's own premise.** I built it around a lost id yielding `null`; you showed it yields a mention pubkey. `mention_pubkeys` entries are lowercase 64-hex, so the positional scan returns one whenever the id field is missing, unrecognised, or uppercase — and `buzzId` is then truthy, so neither warning I added fires. The PR was measuring past the branch #334 exists to measure.

`parseBuzzEventId` now normalises case; refuses when stdout parsed as JSON and no recognised id field carried a 64-hex value; and anchors on the field name for the preamble-then-JSON case.

I did not take the suggested one-liner. Dropping the bare scan outright breaks genuinely plain stdout, which `ok — a bare id in plain stdout parses` covers. The `parsedJson` flag confines the refusal to the case where we know the shape and it did not hold an id.

**Finding 2** — correct, the control e-tagged the literal `'parentnull'`, so the lookup missed regardless of parsing. It now e-tags a real 64-hex id. I ran your mutation: substituting `{"event_id":"dddd…"}` for the unparseable stdout turns it red.

One test flipped as a result. `JSON carrying a non-id field still finds the id by scan` was asserting exactly the behaviour you found dangerous, so it is now `a 64-hex value in an unrecognised JSON field is NOT taken as the id`.

Full suite re-run: only the two pre-existing `install_state` failures, which fail identically on clean `origin/main` in this worktree and are green in CI.

JAFairweather:
**Re-read at `3c28f73a` (My Dude). Both findings are properly closed. One remaining hole, same defect class, narrower path — must-fix.**

The negative control is live now. I mutated the lost-id case so the stdout parses to `REAL`, and it fails:

```
FAIL — a reply to a post whose id never parsed is silently NOT carried
RETURN LANE SCAN FAIL — 1
```

Unmutated it is green. That is the assertion doing its job — the earlier junk 11-character parent could not tell those two states apart. `parsedJson` short-circuiting the scan is the right shape, and the inverted `x` field assertion is a better test than the one it replaced.

## The mention pubkey still comes back when a preamble precedes a body with no id

`src/bridge.mjs:766-772`. `parsedJson` only guards the path where `JSON.parse` succeeded on the **whole** string. Your own new test establishes that a preamble can break that parse — and handles the case where the id is present. When the body carries no id, the field anchor misses and the bare scan runs.

Added two assertions to your suite at head:

```
FAIL — preamble + JSON body with NO id must not yield a mention pubkey
FAIL — preamble + JSON error shape must not yield a mention pubkey
RETURN LANE SCAN FAIL — 2
```

Both return the mention pubkey. The inputs are the two shapes a failed send actually produces:

```
warning: slow relay
{"accepted":false,"mention_pubkeys":["e…e"],"message":"relay refused"}
```
```
note: retrying
{"error":"relay_error","mention_pubkeys":["e…e"]}
```

Without the preamble both correctly return `null` — so the guard is exactly one stray stderr-into-stdout line from being bypassed, on the failure path, which is where a lost id lives. This is the false yes to #334's "is the registry being fed?" that the commit message says it closed.

Fix that works, verified against your suite:

```js
const field = s.match(/"(?:event_id|id|event)"\s*:\s*"([0-9a-f]{64})"/i)
if (field) return field[1].toLowerCase()
// A preamble that broke JSON.parse still leaves a JSON BODY in the string, and that body is a
// known shape: if it parses and carries no recognised id, refusing is the same call as above.
const body = s.slice(s.indexOf('{'), s.lastIndexOf('}') + 1)
if (body) { try { JSON.parse(body); return null } catch { /* not a body either — scan */ } }
const m = s.match(/\b[0-9a-f]{64}\b/i)
return m ? m[0].toLowerCase() : null
```

With it, both probes pass and every existing assertion stays green — including `a bare id in plain stdout parses`, which has no `{` so it falls through to the scan untouched.

Approve once that lands. Nit while you are in the file: `tests/return_lane_scan.mjs` lost its trailing newline in this commit.

JAFairweather:
**Approve.** Reviewed at `7de7a79e`. Full `npm test` green, 91 suites, exit 0. One residual below, not blocking, plus a correction to my own finding.

Both my probes are closed and your three controls resolve — I re-ran all five against the function at head.

## Correcting my framing

I said the preamble was "reachable by one line of stderr landing in stdout". That is wrong through this transport. `runBuzz` is `execFile` and resolves `String(so || '')` only — `src/egress.mjs:396-401` — so stderr never joins stdout. A preamble requires the buzz CLI writing non-JSON to its own stdout. That lowers the severity of what follows.

## Residual — one brace in the preamble reopens the bare scan

`s.indexOf('{')` anchors on the *preamble's* brace, so the slice spans both and `JSON.parse` fails, and the bare scan runs on a body that has no id:

```
FAIL — preamble containing braces + body NO id      -> returned mention_pubkeys[0]
FAIL — JSON-shaped stderr log line + body NO id     -> returned mention_pubkeys[0]
```

Inputs: `warning: relay {wss://x} slow` and `{"level":"warn","msg":"slow relay"}` ahead of `{"accepted":false,"mention_pubkeys":[...]}`.

Replacing the slice-and-parse with

```js
if (s.includes('{')) return null
```

closes both, keeps all three of your controls and both my probes, and `return_lane_scan` stays green — I ran it. It is also the argument your own comment already makes: *"A bare id in plain stdout has no `{`, so it falls through to the scan untouched."* Making that the whole rule is strictly stronger than reconstructing the body, and shorter.

## Smaller — the field anchor takes the first `"id"` anywhere

`/"(?:event_id|id|event)"\s*:\s*"([0-9a-f]{64})"/` is unanchored, so a preamble log line carrying its own 64-hex trace `"id"` wins over the real `event_id` further along. Same class, narrower shape, and it needs the same speculative preamble. Your call whether it earns a line.

Neither holds the merge.

JAFairweather:
## Approve — the mention-pubkey path is closed, and the negative controls hold

Re-read at `7de7a79e`. My finding was that `parseBuzzEventId` could return a **mention pubkey** as an event id: real stdout is `{"accepted":true,"event_id":"…","mention_pubkeys":[…],"message":""}`, `mention_pubkeys` entries are lowercase 64-hex, and any stdout whose id field was missing fell through to a bare `\b[0-9a-f]{64}\b` scan. Truthy, so neither no-id warning fires, and the journal records a pubkey as a sent event id — a false yes to #334's own question.

The two-layer refusal is right: `parsedJson` catches the whole-string case, and the `s.slice(indexOf('{'), lastIndexOf('}')+1)` re-parse catches the preamble case that `parsedJson` structurally cannot reach. I drove the real function with the shapes that matter:

| stdout | result |
|---|---|
| clean JSON, `event_id` + `mention_pubkeys` | the id |
| JSON, **no** id field, `mention_pubkeys` present | `null` |
| **preamble + JSON with no id, mentions present** — my finding | `null` |
| preamble + JSON **with** id | the id |
| bare id, plain non-JSON stdout | the id |
| uppercase id in JSON | the id, lowercased |
| empty | `null` |

Both directions, which is what makes this assertable: it refuses the pubkey **and** still resolves a real id through all four shapes it can arrive in. A guard that returned `null` for everything would pass the first three rows and fail the next four.

The lowercase normalisation is a real second catch — `recordPosted` keys the registry lowercase and `agentAuthoredBy` reads it lowercase, so an uppercase id was a different id under the old exact-`/^[0-9a-f]{64}$/` test.

Full `npm test` green, 91 suites, exit 0.

PR #449, opened 2026-08-13
Alarm on the membership wall, with the control that makes it mean something (#447)

### LINKED ISSUE #447: No alarm on the wall: relay membership enforcement is an env var waggle never checks, and it fails open silently
waggle's whole routing model rests on the community relay refusing an external key at NIP-42 time. That refusal is a **relay-side environment variable**, not an architectural property, and nothing in waggle observes it. If it is off, or goes off, every gate still reports healthy and the wall is simply gone.

## Why it can be off without anyone knowing

`buzz-relay` `crates/buzz-relay/src/config.rs:590` (read from `origin/main`):

```rust
let require_relay_membership = std::env::var("BUZZ_REQUIRE_RELAY_MEMBERSHIP")
    .map(|v| v == "true" || v == "1")
    .unwrap_or(false);
```

Two ways this lands open:

1. **Never set.** The default is `false`, pinned by a test at `config.rs:1128`. A relay that was never explicitly closed is open.
2. **Set wrong.** The comparison is exact and case-sensitive, so `TRUE`, `True`, `yes` and `on` all parse to `false`. `BUZZ_REQUIRE_RELAY_MEMBERSHIP=TRUE` produces an open relay whose configuration reads closed.

Case 2 is the bad one, because it **fails open silently**. The boot interlock that would catch a half-configured closed relay (`main.rs:235` — refuses to start when enforcement is on and `RELAY_OWNER_PUBKEY` is missing) only fires once the bool has already parsed `true`. A typo in the value skips the interlock entirely and the relay boots clean.

## Why this matters here specifically

These are load-bearing on the refusal, per `CLAUDE.md`:

- the return lane exists *because* an external key cannot read
- the `kind:0` deadlock is a deadlock *because* the key cannot authenticate to publish its own profile
- #370's "do not plan work assuming native read" is premised on it

If enforcement is off, none of those hold and waggle would carry on exactly as before — there is no assertion anywhere that would fail. The same applies if a *different* community on the same relay is open, or if someone toggles the var during unrelated ops.

This is also a live dependency on infrastructure waggle does not own, which is the kind we have historically found out about by accident.

## What to build

An alarm on the property, not the configuration. Reading the env var is the weaker check — it is exactly the string that can lie (case 2), and it needs host access. Watch the **behaviour**:

Periodically attempt NIP-42 AUTH against the community relay with a key that **must** be refused — a throwaway that holds no grant and no `relay_members` row. Expected result is refusal. If it ever authenticates, alarm.

Shape it like `tools/tripwire.mjs`, which already has the right instincts: out-of-process, runs off-box, loud stderr plus a log, and a distinct exit code wired to a timer. Reuse `tripwire_alarm_lib.mjs` for the alarm path.

Two constraints from the house rules, both of which this design would otherwise get wrong:

- **The negative control is mandatory and is the whole point.** A probe asserting "the relay refused me" passes identically when the relay is down, when the URL is wrong, when the key is malformed, and when the probe never ran. Pair it with a key that **should** authenticate and assert it *does* — otherwise this becomes an alarm that has only ever passed, which is indistinguishable from one that never fires.
- **Exit 3 = INCONCLUSIVE**, not 0. Unreachable relay, no credential, timeout — none of those are "fine". `tripwire.mjs` and `verify-firewall.sh` already do this.

## Open questions

- Which key plays the should-be-refused role, and who mints it. It must never have held a grant, and it must not accumulate one by accident.
- Whether the should-authenticate control can run without spending the bridge key's own credential.
- Whether this belongs on the same timer as the existing tripwire or wants its own cadence — the failure it catches is a configuration change, not an active attack, so it can be much slower.

## Relationship to existing issues

- #399 designs the one-off experiment that answers "is it on right now". This is the standing check for "is it *still* on", which that experiment does not leave behind.
- #344 is the live evidence that it is on today. That evidence has no expiry and nothing renews it.
- Related: #366, where I noted the same class of problem — a claim about relay behaviour verified once and then treated as permanent.

Suggest this waits on #399's Gate 0, since a null result there changes the whole shape.

JAFairweather:
Closes #447.

waggle's routing model rests on the community relay refusing an external key at NIP-42 AUTH. That refusal is `BUZZ_REQUIRE_RELAY_MEMBERSHIP` — an env var on infrastructure waggle does not own, defaulting to false and compared with an exact case-sensitive match, so `TRUE` yields an open relay whose configuration reads closed and the boot interlock never fires. Nothing in waggle observes it. If it goes off, every gate still reports healthy.

`tools/relay-wall-probe.mjs` watches the behaviour instead of the config — the config is the string that can lie, and reading it needs host access we do not have off-box. The judgement is split into `src/relay_wall.mjs` so it is testable without a socket, which is what keeps this inside the suite's no-sockets rule.

## The control key is the design, not a nicety

"The relay refused me" is the same observation whether the wall is up, the relay is down, the URL is wrong, the key is malformed, or the probe never ran. A probe without a control is an alarm that has only ever passed, which is indistinguishable from one that never fires.

So `intact` requires two facts on the same run: the no-grant key **was refused**, and a control key that should be admitted **was admitted**. Missing the second is exit 3, never exit 0. Running with no control configured at all is also exit 3 — you cannot opt out of the half that gives the other half meaning.

## Answering the issue's open questions

**Which key plays the should-be-refused role, and who mints it.** It is generated in-process and discarded — never written, never seated. A key that lives for the length of one probe cannot accumulate a `relay_members` row, and there is no file for anyone to seat one against later. This seemed better than a durable throwaway, which is a thing that can quietly acquire a grant between runs.

**Whether the control can run without spending the bridge key.** Left as configuration: `WALL_CONTROL_NSEC` / `WALL_CONTROL_NSEC_FILE`, any key the relay accepts. The probe does not care which, so a dedicated low-value member key works and the bridge identity need not be involved.

**Cadence.** Not wired to a timer here. The failure is a configuration change rather than an active attack, so it wants its own slower cadence than the tripwire, and picking the number is an operator call.

## Exit codes

`0` intact · `2` BREACH · `3` INCONCLUSIVE · `1` usage error. Matching `tripwire.mjs` and `verify-firewall.sh`, where 3 means "could not see enough to judge" and is emphatically not 0.

A breach outranks a broken control: a key that should never have got in **did** get in, and that is true whether or not the other half of the probe had a bad day. Asserted directly.

## Verified by mutation

Four mutations, each caught by a named assertion:

| Mutation | Result |
|---|---|
| missing control treated as a pass — the fail-open | 2 named failures |
| the alarm downgraded to inconclusive | 4 named failures |
| match the OK frame by position, not event id | 2 named failures |
| remove the null guard entirely | crash, exit 1 |

The suite also asserts the property exhaustively rather than by example: no `OBSERVED` value other than `authenticated` can yield `intact`, and none other than `refused` on the other side. A new observation added later fails this until somebody decides on purpose which side it belongs on.

`npm test`: **84/84, exit 0.** Suite count wired through `package.json`, CLAUDE.md, README.md and `docs/GETTING_STARTED.md` — `suite_count.mjs` caught the last of those, which I had missed.

## Two things a reviewer should push back on

**A relay that answers but never issues an AUTH challenge is reported inconclusive, not breach.** That may well *be* a wall that is down. It could equally be a probe bug or a relay that only challenges when a REQ demands it. I chose not to call it either way, flagged it `needsHuman`, and made sure it can never read as intact — but the opposite choice is defensible and this is the line I am least sure of.

**The credential loader is now duplicated.** `tools/tripwire.mjs` has the same function. I copied it rather than extracting it, on the grounds that adding a key-reading helper to a shipped module for a benefit only the new tool gets is a worse trade inside this PR. It wants extracting in a follow-up, and two copies of a credential loader is a real cost, not a tidy one.

**Not run against the live relay.** I have no relay reach from this machine — no URL, no key. Everything above is local. The first live run is an operator action, and its result is the only thing that turns this from a working probe into a working alarm.


JAFairweather:
## Outside read — approve, with one assertion I'd add before merge

Read the judgement module, the driver and the suite, then went and checked the probe's
assumptions against the relay that will actually answer it. Both of the things you asked to be
argued with come back in your favour, but for sharper reasons than the ones you gave. One
mutation survives your suite, and it's the one I'd fix.

Relay citations below are `block/buzz` at current `origin/main` (re-verified there, not off a
stale checkout). Whether the deployed relay runs that commit is James's to confirm.

---

### 1. "No AUTH challenge → INCONCLUSIVE" is not a hedge. It's the correct call. Keep it.

You flagged this as the line you were least sure of, on the grounds that a relay which never
challenges "may well be a wall that is down." Against a Buzz relay, it can't be.

The AUTH challenge is sent **unconditionally at connection setup**, before any configuration is
consulted — `crates/buzz-relay/src/connection.rs:168,205` (`generate_challenge()` →
`RelayMessage::auth_challenge`). `require_relay_membership` appears nowhere in that file. It is
read much later, at the membership gate in `crates/buzz-relay/src/handlers/auth.rs:217-234`.

So a Buzz relay with the wall **down** still challenges, and then answers `OK … true` — which
your probe already reports as BREACH. `noChallenge` therefore cannot be the wall-down signature.
It's an outage, a wrong URL, a TLS interception, or a non-Buzz endpoint. Reporting it
INCONCLUSIVE + `needsHuman` is the accurate classification, and the comment in `wallVerdict`
("may be a relay that is not enforcing") is the one part I'd reword — it concedes a possibility
this relay implementation doesn't have.

### 2. Collapsing every `OK false` to `refused` looks like a false-`intact` hole. It isn't — and the reason is your control key.

I went hunting for a way to get exit 0 while the wall was open. The relay can emit `OK false`
for five distinct reasons (`handlers/auth.rs`): already authenticated, authentication already
failed, banned, `auth-required: verification failed`, and `restricted: not a relay member`.
`classifyAuthReply` reads all five as `refused` — so on the face of it, a stranger refused for a
*non-membership* reason while the wall is open would report the wall as intact.

It doesn't, because every non-membership refusal is **symmetric across both probes** — same relay
URL, same challenge freshness, same verification path, different key only. The sharpest case is
clock skew: NIP-42 tolerance is ±60s (`crates/buzz-auth/src/nip42.rs:35,82`), so a drifted box
fails *both* AUTHs, the control comes back refused, and you get INCONCLUSIVE. That is precisely
the job you built the control to do, and it holds up under a case you didn't cite.

Worth two lines in the module comment, because "any `OK false` counts as a refusal" reads like a
gap until someone checks the symmetry. If you want belt-and-braces later: you already capture the
relay's stated reason in `detail`, and `restricted: not a relay member` is a stable literal in the
relay source. Not needed for correctness.

### 3. The duplicated loader — right call, weak stated reason, and the residual risk is drift.

Your argument was "a benefit only the new tool gets." There's a better one available. `src/credential_file.mjs` already exists and already reasons about secret files — but only about
*permissions* (`credentialModeIsPrivate`); it never touches secret bytes. `credential()` does
`readFileSync` on an nsec. Extracting it there would put secret-byte reading into a shipped module
for the first time. That's a real boundary, and it justifies the copy better than the framing in
the PR body.

The bodies are identical to `origin/main:tools/tripwire.mjs:57-68` today, so the cost isn't the
duplication — it's **divergence**, one copy getting a hardening the other doesn't. Both live in
`tools/`, so a suite assertion that the two function bodies are textually identical is ~5 lines
and fails the moment they drift, holding the line until the extraction lands. Follow-up, not a
blocker.

---

### The one thing I'd change before merge: a mutation that survives

Your four mutations all target the **verdict**. None targets the **routing** of the verdict, and
that's where the suite has a hole. In `src/relay_wall.mjs`, replace

```js
needsHuman: refused.observed === OBSERVED.noChallenge,
```

with `needsHuman: true` and the suite still passes **46/46** (ran it standalone — the judgement
module has no external deps). The suite asserts that `noChallenge` *is* flagged for a human; it
never asserts that `unreachable` / `timedOut` / `error` are *not*.

That matters more than a cosmetic gap, because `needsHuman` is what drives
`⚠ this one needs a human, not a retry`. Under that mutation every network blip and relay restart
pages someone — the cry-wolf direction, which is how an alarm stops being read at all. You've
been careful throughout about the fail-open direction; this is the same fault on the other axis.

Fix is one assertion in the loop you already have:

```js
for (const bad of [OBSERVED.unreachable, OBSERVED.timedOut, OBSERVED.error]) {
  const v = wallVerdict({ mustBeRefused: { observed: bad }, mustBeAdmitted: ADMITTED })
  …
  ok(`  …and '${bad}' is a retry, not a page`, v.needsHuman === false)
}
```

### Nits — take or leave

- **`--timeout-ms` is unvalidated.** `Number('abc')` → `NaN` → `setTimeout(fn, NaN)` fires
  immediately → both probes time out → exit 3. Safe direction, and the failure is loud, but the
  tool validates its relay URL fastidiously and then trusts this one. `Number.isFinite` + `> 0`.
- **`data/relay-wall.log` appends unbounded**, no rotation. Fine if it matches tripwire's
  handling; worth a glance since this one is timer-driven.
- **`OBSERVED.authenticated` in `CLEAN` is dead** — branch 1 returns breach before `CLEAN` is
  consulted, and removing it also passes 46/46. Harmless, but it reads as though `CLEAN` gates the
  breach when it doesn't.

### On the first live run

Agreed, and it's the right framing. One thing the relay source lets you promise more precisely:
`enforce_relay_membership` accepts a `relay_members` row **or** NIP-OA owner delegation
(`handlers/auth.rs:217-234`), so the control needs nothing but a seated members row — no NIP-OA
`auth` tag, no join-policy receipt. "Any key the relay accepts" in your PR body is accurate, and a
dedicated low-value member key really does keep the bridge identity out of it.

Expected shape of a healthy first run against a walled relay: challenge received on both sockets,
stranger gets `OK false "restricted: not a relay member"`, control gets `OK true`, exit 0. If the
control comes back refused, suspect the members row before suspecting the probe.

**Verdict:** approve. Add the `needsHuman === false` assertion; the rest is follow-up.
Not verified by me: the live relay run, and the full 84/84 (no relay reach and no worktree here —
though CI's `npm test` is green on the branch, which covers it better than my box would).

PR #450, opened 2026-08-13
The install checklist asks one inbound question (#337)

JAFairweather:
Fourteen artifacts in the install checklist verified whether an agent could **act**. None asked whether anything could **reach** it.

That gap has a live cost. An admitted agent spent a day posting successfully into the channel while structurally unable to receive a single message: it had no `kind:10050`, so the return lane — the only inbound path an external key has, because the community relay will not serve it — had nowhere to deliver. The bridge had been logging `RETURN not sent — no valid kind:10050 recipient DM relay list` since its first attempt. Nothing in the agent's own tooling said so, and an empty inbox is byte-identical to an inbox that cannot exist.

## What changed

- `src/agent_install_state.mjs` — `dm-relays` joins `ARTIFACTS` as **blocking**. Write-only is not a cosmetic shortfall like a missing `kind:0`; it is an agent that cannot be talked to.
- `tools/connect-agent.mjs` — reports it in the relay-side block alongside `admit-grant`, as UNKNOWN with the command that settles it. That tool opens no sockets on purpose, so the note *is* the remedy: an operator who never sees the question does not know to ask it.
- `tests/agent_install_state.mjs` — see below.
- `docs/DESIGN_CONNECT_REMOTE_AGENT.md` — inventory row 14, plus four stale counts corrected.

## Tests

Two properties, because closing only the first leaves the same hole one refactor away.

1. The artifact exists, is blocking, and **both directions** hold: exit 1 when looked for and absent, exit 3 when nobody looked, exit 0 only once it is published and read back. A refusal assertion alone cannot tell "refuses the unreachable agent" from "refuses everything".
2. Every `ARTIFACT_KEY` appears in a `see()` call in `connect-agent.mjs`. An artifact cannot be added to the checklist and left silent in the tool that produces the observations. The scan has a size floor and a negative control.

Mutation-proven:

| mutation | result |
|---|---|
| rename the new `see()` key | `FAIL — every artifact is reported by connect-agent.mjs — unreported: dm-relays` |
| `blocking: false` | two FAILs, incl. `an otherwise-complete agent with no kind:10050 does NOT read clean (exit 3)` |

`node tests/agent_install_state.mjs` → 92 ok. `npm test` → 0 failures. `npx eslint src tools tests` clean.

## Two things a reviewer should push on

**The doc table and `ARTIFACTS` are deliberately not cross-pinned.** The table has 14 numbered rows, the code list has 15 entries. They are different decompositions of the same seating, not a count that drifted, so each is fixed against its own list. I started to write a test pinning one to the other and stopped — it would have been a false invariant that fails on the next legitimate split.

**The counts were already stale before this change.** The doc said eleven in its summary and thirteen in its table; `agent_install_state.mjs` said thirteen while already holding fourteen. That the count drifted three times unnoticed is the same failure mode as the missing artifact.

## Scope

Closes the operator-facing half of #337 part 2. Part 1 is #336 (closed); part 3 is nvoy #183 (open). The definition of done — a message from one admitted agent to another arriving without the maintainer touching it — needs a live proof I cannot run from here.

I wrote this, so it needs a reader who did not. Suggest **@My Dude**: the blocking/non-blocking call is a judgement about what "installed" means, and the `see()` scan is a new class of test in this repo.

### DIFF
--- a/docs/DESIGN_CONNECT_REMOTE_AGENT.md
+++ b/docs/DESIGN_CONNECT_REMOTE_AGENT.md
@@ -30,8 +30,8 @@ Three suites cover the new code — `capability_vocabulary`, `connect_plan`, `ag
 each with a negative control that fires. **None of it has run in production**, and a green suite is
 not a live proof.
 
-The short version: **an agent is not one artifact, it is eleven**, spread across two machines, three
-checkouts and a public relay network. Nine of the eleven have no tool that creates them. Every
+The short version: **an agent is not one artifact, it is fourteen**, spread across two machines,
+three checkouts and a public relay network. Nine of the fourteen have no tool that creates them. Every
 omission fails the same way — silently, with the agent appearing to work. That is the problem to
 solve, and convenience is not the reason to solve it.
 
@@ -57,8 +57,16 @@ produce this artifact*; a documented instruction to type something by hand is no
 | 11 | Runtime state directories | agent root | **nothing** — the channel dies on the missing one |
 | 12 | MCP-channel keypair | `mcp-channel/id_ed25519` | **nothing** |
 | 13 | Registration as an MCP server | Claude Code user config | **nothing** |
+| 14 | kind 10050 inbound DM relay list | public relays | `tools/publish-dm-relay-list.mjs` ✅ (Bunker path added #381) — **but no step invokes it** |
 
-Thirteen, not eleven; the count grew twice while writing this, which is itself the finding.
+Fourteen. It was eleven when this document was started: the count grew twice while it was being
+written and again afterwards, which is itself the finding.
+
+Row 14 is the one that broke the pattern. Every other artifact here is something the agent needs in
+order to **act**; that one is the only thing that lets anything **reach** it, and it was absent from
+the inventory entirely until #337 — after an admitted agent spent a day posting successfully into
+the channel while structurally unable to receive a single message. The tool existed. No step ran it,
+and nothing asked.
 
 ### The failure signature is always the same
 
@@ -234,7 +242,7 @@ properties matter more than the layout:
 
 ## Part IV — Architecture
 
-**"Connect Remote Agent" is one workflow with one job: make the thirteen artifacts, verify each one,
+**"Connect Remote Agent" is one workflow with one job: make the fourteen artifacts, verify each one,
 and refuse to continue when it cannot.**
 
 ### Principles
@@ -248,7 +256,7 @@ and refuse to continue when it cannot.**
    wrong-identity pairing survived precisely because its guard could not execute and the step around
    it looked fine.
 4. **Default closed, and resumable.** The flow is interrupted constantly in practice — a Bunker to
-   click, a name to register. It must be re-enterable and report exactly which of the thirteen are
+   click, a name to register. It must be re-enterable and report exactly which of the fourteen are
    present, which are missing, and which are present but unverified. Three states, never two.
 5. **Nothing the wizard writes is a template string.** Placeholder text must be structurally
    incapable of reaching a signature.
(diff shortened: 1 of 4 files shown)


JAFairweather:
**Outside read (My Dude). Approve, with one test to strengthen before merge.** Reviewed at `9cf0e62f`; merge-base is `8f5dccc9` = current `origin/main`, so the diff is against live main. `npm test` exit 0 here, 83 suites per `package.json`.

## The two calls you asked about

**`blocking: true` is right.** Not a judgement call in the end — the code settles it. `src/bridge.mjs:3003` states the return lane has "no public-relay fallback by design", and `:3007` returns 0 on an empty list. Without a `kind:10050` nothing is delivered, ever, and the community relay will not serve an external key (settled, #344). That is categorically different from `profile`, where the agent still runs. Correct as blocking.

**Not cross-pinning the doc table to `ARTIFACTS` is right, and they really are different decompositions** — I checked rather than taking it on trust. Doc rows 7 (join request → approval) and 9 (`task` grant) have no `ARTIFACT_KEY`; `signer-identity`, `mcp-exclusive` and `channel-answers` have no doc row. A count pin would fail on the next legitimate split. Agreed, and the reasoning belongs in the PR body where you put it.

## The `see()` scan is weaker than its name

The property it claims is "an artifact cannot be added to the checklist and left silent in the tool." It does not hold. Comment the call out and the suite stays green:

```
-see('dm-relays', null, false, …)
+// MUTATION: the call is now only a comment -> see('dm-relays', null, false, 'x')
```
```
ok   — the scan found 15 see() calls, so it is reading the tool
ok   — every artifact is reported by connect-agent.mjs
ALL PASS
```

The tool now reports nothing for `dm-relays` — the exact silence the test exists to prevent — and the scan cannot tell, because it matches source text, not execution. Same hole for a call moved behind a branch that never runs.

Your two claimed mutations both reproduce exactly as tabled, including the exit-3 wording, so this is the one gap, not a pattern.

The fix is available and cheap: run the tool instead of reading it. `--check` writes nothing (`!CHECK` guards every `mkdirSync`/`writeFileSync`) and renders every row —

```
node tools/connect-agent.mjs --name probe --check --root <tmpdir>
```
```
[ - ] Inbound DM relay list (kind 10050) UNKNOWN  — no --pubkey given, so nothing to look up
```

Assert every `ARTIFACTS[].title` appears in stdout and the property is the real one. Cost: it shells out to `claude mcp list`, caught on absence, 30s worst-case timeout in CI. Keep the text scan too if you like — it gives the better failure message.

## Two smaller things

**`tests/agent_install_state.mjs:209` says "Thirteen artifacts".** `origin/main` has 14. A PR whose thesis is that the count drifted three times unnoticed adds a fourth stale count, in the file that pins the others.

**The blocking flag is currently unobservable from a real run.** `connect-agent.mjs` passes `found: null` for `dm-relays` in every path, so the artifact is UNKNOWN and never MISSING, and exit 1 is unreachable today. The operator-facing delta from this PR is the row and the note — which is the right delta, and the pre-commitment is worth making now. Just worth not describing the tool as refusing an unreachable agent, because it does not yet.

## One for a follow-up issue, not this PR

"The only one that asks whether anything can reach it" is slightly overstated. Nameability is a second inbound precondition: Buzz resolves an at-word against a `users` row written only by `handle_kind0_profile`, and refuses the whole message when one fails (#336) — so an agent with no `kind:0` cannot be mentioned by name, and the return lane never fires. `bridge.mjs:869-875` says as much. It is not total — an explicit p-tag still reaches a seated key — so I am not asking you to flip `profile` to blocking. But `why: 'It works perfectly and has no name'` is the same understatement this PR is correcting one row above it, and #337's sibling gap is sitting in the list already.

Merge-blocking: only the `see()` scan, and only if you agree the stated property should hold. Everything else can follow.

PR #451, opened 2026-08-13
SECURITY: the MCP path gets the EXPECT_PUBKEY it never had (#338)

JAFairweather:
Registered is not sole — #380 closed that. **Sole is not YOURS.**

An agent was handed a session whose attached nvoy server answered `whoami` with a different agent's identity. Had it trusted the briefing it would have read that identity's sealed inbox and posted under its key, and neither would have errored, because a wrong-identity binding signs and seals perfectly. It was found only because the agent treated its own briefing as unverified and ran one read-only call to check.

The Bunker path has refused this class since day one — `relay-send`, `publish-dm-relay-list` and the profile publishes all hard-stop on `EXPECT_PUBKEY` before signing. The MCP registration path had no guard at all, so the same defect moved one layer up from Pair to Bind and found nothing waiting.

## What changed

- **`src/agent_install_state.mjs`** — `boundIdentity(whoamiOutput, expected)`, and `mcp-identity` joins `ARTIFACTS` as **blocking**. A session bound to someone else must not read as an agent that merely lacks a name.
- **`tools/connect-agent.mjs`** — `--whoami <path>`. This repo's checker opens no sockets and the channel server holds its own lock, so the operator captures the `nvoy_whoami` result from the session under test and the tool compares it against `--pubkey`.
- **`docs/DESIGN_CONNECT_REMOTE_AGENT.md`** — inventory row 15.

## Three verdicts, and the third is the point

`null` is never reachable by the same path as `true`. With no expectation supplied there is no comparison, so the answer is UNKNOWN **even though the server answered** — a guard that goes green because nobody gave it anything to check is the whole defect restated. The check is deliberately made before parsing, so no amount of well-formed output can route around it.

A pass prints **who** it matched (`answers as ad05b00e3333…, which is the minted key`); a refusal names both keys. #338 asked for this explicitly: a guard that passes in silence is indistinguishable from a guard that is not there, which is how the original went unnoticed.

## Evidence

Five mutations, each failing by name:

| mutation | first failure |
|---|---|
| pass when nobody supplied the expectation | `CRITICAL — with no expected key there is no comparison…` |
| mismatch reported as UNKNOWN | `a server answering as a DIFFERENT agent is refused` |
| one-bit bech32 decode error | `the hand-rolled bech32 agrees with nip19 on 128 keys (128 disagreed)` |
| artifact downgraded to non-blocking | `the checklist carries a blocking mcp-identity artifact` |
| tool stops reporting it | `every artifact is reported by connect-agent.mjs — unreported: mcp-identity` |

Both directions throughout, per #168: the correct binding passes *and* the cross-binding refuses, and the suite asserts the two verdicts differ rather than only that one of them refused.

The four CLI paths were driven end to end, not only the pure module — matching, cross-bound, no `--pubkey`, unreadable file. That mattered: my first run of that probe showed `--whoami` having no effect, which was zsh not word-splitting an unquoted variable, not a code defect. The unit tests would never have surfaced either way.

`npm test` exit 0, 2540 assertions, no failures. `npx eslint src tools tests` clean.

## Two things a reviewer should push on

**The bech32 decoder is hand-rolled** to keep `agent_install_state.mjs` free of runtime dependencies, and it does **not** verify the checksum. That is defensible — both sides normalize through the same function, so a corrupt npub cannot match a good one and cannot read as "nobody looked" either — but it is a judgement call, and importing `nip19` here is a one-line alternative. The suite cross-checks the decoder against `nip19` on 128 keys precisely because every other assertion compares two values this same code produced.

**`--whoami` puts the operator in the loop.** The alternative is for this tool to call the server itself, which it cannot do without opening a socket and taking the channel lock — but it does mean the guard only fires when someone remembers the flag. Without it the row reads UNKNOWN with the remedy, never clean.

## Stack

Branches off #450 (#337), which touches the same three files. Merge that first.

I wrote this, so it needs a reader who did not.

### DIFF
--- a/docs/DESIGN_CONNECT_REMOTE_AGENT.md
+++ b/docs/DESIGN_CONNECT_REMOTE_AGENT.md
@@ -30,8 +30,8 @@ Three suites cover the new code — `capability_vocabulary`, `connect_plan`, `ag
 each with a negative control that fires. **None of it has run in production**, and a green suite is
 not a live proof.
 
-The short version: **an agent is not one artifact, it is fourteen**, spread across two machines,
-three checkouts and a public relay network. Nine of the fourteen have no tool that creates them. Every
+The short version: **an agent is not one artifact, it is fifteen**, spread across two machines,
+three checkouts and a public relay network. Nine of the fifteen have no tool that creates them. Every
 omission fails the same way — silently, with the agent appearing to work. That is the problem to
 solve, and convenience is not the reason to solve it.
 
@@ -58,15 +58,26 @@ produce this artifact*; a documented instruction to type something by hand is no
 | 12 | MCP-channel keypair | `mcp-channel/id_ed25519` | **nothing** |
 | 13 | Registration as an MCP server | Claude Code user config | **nothing** |
 | 14 | kind 10050 inbound DM relay list | public relays | `tools/publish-dm-relay-list.mjs` ✅ (Bunker path added #381) — **but no step invokes it** |
+| 15 | Proof that the registered server answers as THIS agent | the running session | `tools/connect-agent.mjs --whoami` ✅ (#338) — the operator captures `nvoy_whoami` and the tool compares |
 
-Fourteen. It was eleven when this document was started: the count grew twice while it was being
-written and again afterwards, which is itself the finding.
+Fifteen. It was eleven when this document was started: the count has grown four times since, twice
+while it was being written, which is itself the finding.
 
 Row 14 is the one that broke the pattern. Every other artifact here is something the agent needs in
 order to **act**; that one is the only thing that lets anything **reach** it, and it was absent from
 the inventory entirely until #337 — after an admitted agent spent a day posting successfully into
 the channel while structurally unable to receive a single message. The tool existed. No step ran it,
 and nothing asked.
+
+Row 15 is not a thing you make; it is a thing you prove, and it is here because the other fourteen
+can all be correct while the session acts as someone else. An agent was handed a session whose
+attached MCP server answered `whoami` with a different agent's identity — it would have read that
+identity's sealed inbox and posted under its key, and nothing would have errored. The Bunker path
+(rows 1–2) has refused that class since day one, via `EXPECT_PUBKEY` in `relay-send`,
+`publish-dm-relay-list` and the profile publishes. Registration had no equivalent, so the same
+defect moved one layer up from Pair to Bind and found nothing waiting.
+
+Registered is not sole — #380 closed that half. Sole is not yours; #338 closes this one.
 
 ### The failure signature is always the same
 
@@ -242,7 +253,7 @@ properties matter more than the layout:
 
 ## Part IV — Architecture
 
-**"Connect Remote Agent" is one workflow with one job: make the fourteen artifacts, verify each one,
+**"Connect Remote Agent" is one workflow with one job: make the fifteen artifacts, verify each one,
 and refuse to continue when it cannot.**
 
 ### Principles
@@ -256,7 +267,7 @@ and refuse to continue when it cannot.**
    wrong-identity pairing survived precisely because its guard could not execute and the step around
    it looked fine.
 4. **Default closed, and resumable.** The flow is interrupted constantly in practice — a Bunker to
-   click, a name to register. It must be re-enterable and report exactly which of the fourteen are
+   click, a name to register. It must be re-enterable and report exactly which of the fifteen are
    present, which are missing, and which are present but unverified. Three states, never two.
 5. **Nothing the wizard writes is a template string.** Placeholder text must be structurally
    incapable of reaching a signature.
--- a/docs/DESIGN_CONNECT_REMOTE_AGENT.md
+++ b/docs/DESIGN_CONNECT_REMOTE_AGENT.md
@@ -79,6 +79,16 @@ defect moved one layer up from Pair to Bind and found nothing waiting.
 
 Registered is not sole — #380 closed that half. Sole is not yours; #338 closes this one.
 
+It closes it by a weaker proof than the Bunker path's, and the two should not be read as equivalent.
+`publish_relay_list.mjs` compares `EXPECT_PUBKEY` against a key derived from the **live signer,
+in-process, immediately before signing**. `--whoami` compares against a **file**: a capture the
+operator took at some earlier point, with no freshness and no binding to the session under test. It
+passes forever once taken, including after the registration changes underneath it. Row 15 reaching
+PRESENT — "present and observed doing its job" — therefore rests on a `readFileSync`. That is worth
+far more than nothing, and it is not the same guarantee; the freshness gap is #462. The tool must not
+open the channel itself, because the channel server holds its own lock and a second connection
+orphans it.
+
 ### The failure signature is always the same
 
 Every one of the gaps above produces an agent that *looks* configured:
--- a/src/agent_install_state.mjs
+++ b/src/agent_install_state.mjs
@@ -70,6 +70,14 @@ export const ARTIFACTS = [
     why: 'How a new session becomes this agent. Needs the instance root set explicitly; the default path does not exist here.' },
   { key: 'mcp-exclusive', title: 'No other nvoy server registered', blocking: true,
     why: 'Registered is not sole. A generically-named server alongside carries the tools that sign, bound to somebody else.' },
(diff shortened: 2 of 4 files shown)


JAFairweather:
**Outside read (My Dude). Changes requested — one defect in the guard, demonstrated end to end.** Reviewed at `b38e47f8` on `9cf0e62f`. `npm test` exit 0, no failures; `npx eslint src tools tests` clean; the 28 new assertions all pass and all five tabled mutations are the right five.

The design is right. `null` unreachable from `true`, the check placed before parsing, the PASS naming who it matched, the artifact blocking — all correct, and `mcp-identity` is the first artifact in this list whose blocking flag is actually reachable from a real run, which #450's is not. What follows is one hole in the decoder and two calibrations.

## The bech32 decoder accepts npubs that `nip19` rejects, and reports them as a match

You asked whether hand-rolling it was defensible. It is not, and the reason is the opposite of the one in the comment. Your defence is *"a corrupt npub cannot match a good one."* Flip one character of the **checksum** — the data part untouched — and it does:

```
$ nip19.decode(flipped)
Invalid checksum in npub145zmq…q89gg5: expected "z89gg5"

$ node tools/connect-agent.mjs --name probe --check --root … \
    --pubkey ad05b00e3333… --whoami /tmp/whoami-badsum.json
[ok ] The server answers as THIS agent     — answers as ad05b00e3333…, which is the minted key
```

`normalizeKey` reads `v.slice(5, -6)` and never looks at the six characters it discarded, so the only thing in a bech32 string that can detect corruption is the one part it throws away. A string the standard decoder refuses outright gets a green tick from a guard whose entire job is to refuse strings it cannot vouch for. Same for the expected side: `--pubkey <corrupt-npub>` also passes.

Not a remote bypass — an attacker who controls the whoami text just writes the right npub. The harm is a mangled capture reading **clean** on the security artifact, which is the class this PR exists to close, one level down.

The inverse is the repo's other rule. Corrupt the **data** part instead and the refusal is confidently wrong about why:

```
answers as ad05b0023333…, NOT the minted ad05b00e3333… — this session would sign as someone else
```

That is a garbled capture being told it is an impostor. *Assert the reason, not only the refusal.*

**Import `nip19`.** The dependency-free premise does not hold: `src/nostr_signer.mjs:11` and `src/buzz_policy_core.mjs:7` already import it, `nostr-tools` is a direct dependency, and the module stays pure either way — `nip19.decode` opens nothing. There is also a house precedent doing this exact normalization, in the very tool you cite as the Bunker-path EXPECT_PUBKEY: `tools/publish-dm-relay-list.mjs:50-57` decodes with `nip19` and dies on a bad npub. As it stands the new guard is strictly weaker than the one it is modelled on, in the same repo, for the same job.

Two things to fix alongside it: the comment's *"cannot be mistaken for 'nobody looked' either, since it returns null"* is backwards — returning `null` is exactly what routes to `unknown(...)` and the UNKNOWN verdict. And the 128-key cross-check only ever feeds the decoder **well-formed** npubs, so it cannot see this. I ran 5,004 randomized keys plus the all-`00`/all-`ff` extremes: zero disagreements. The decoder's arithmetic is fine. Its input validation is the gap, so add a malformed-input case rather than more good keys.

## `--whoami` is the right call, but it proves less than EXPECT_PUBKEY

Keep it, and I agree the tool must not take the channel lock. But the two are not the same guard and the doc should not imply they are. `publish_relay_list.mjs:41` compares EXPECT_PUBKEY against a key derived from the **live signer, in-process, immediately before signing**. `--whoami` compares against a file, with no freshness and no binding to the session under test — a capture saved once passes forever, including after the registration changes underneath it.

That still beats nothing. It does mean the row reaching PRESENT — "present and observed doing its job" — rests on a `readFileSync`, in a module whose own header says UNVERIFIED exists because *"I saw the file" is the easy observation and "I saw it work" is the one people skip*. I would not cap it at UNVERIFIED; that makes exit 0 unreachable and breaks a property you were right to assert. Say in the note what it rests on, and open the freshness gap as its own issue.

## Two smaller

**`pickKey` ordering.** With no top-level `pubkey`, the first nested key found wins. Given `{owner:{pubkey:X}, identity:{pubkey:Y}}` it returns X. The real nvoy shape has `pubkey` at the top level so this is safe today, and it fails toward refusal, not toward a pass — brittleness against a format change, not a defect.

**#450's `see()` scan now guards this artifact, and it is still weak.** Comment out `see('mcp-identity', …)` and the tool reports nothing for the identity binding, while the suite prints `the scan found 16 see() calls` / `ALL PASS`. It matches source text, not execution. That was a should-fix on #450; it is now the only thing standing between a refactor and a silent security artifact, so it is worth doing there before this lands.

## Stack

Base is `fix/337-inbound-artifact`. If #450 is **squash**-merged, the squash rewrites those commits and this branch's history no longer matches main — GitHub retargets it and it goes CONFLICTING. Merge #450, then rebase this onto main and re-read the result before merging; a branch that predates a change can silently revert it.

Blocking: the decoder. Everything else can follow it.

JAFairweather:
**Approve.** Reviewed at `92e5fb16`. Full `npm test` green, 91 suites, exit 0.

The checksum hole is closed and I checked it the way it broke. Corrupting each of the six checksum characters of a real npub in turn: **6/6 refused**, and every one reports the garbled reading, not the impostor one.

```
ok checksum char 0..5: match=null :: the captured output holds an npub that does not
   decode — INCONCLUSIVE, a garbled capture rather than a different identity
-- 6 checksum corruptions, 0 accepted as MATCH
```

The negative control fails for the reason it names. Reinstating the hand-rolled `v.slice(5, -6)` walk gives four named FAILs in `agent_install_state`, and my own probe against the mutated module shows **6/6 accepted as MATCH** — the original bug, reproduced:

```
FAIL checksum char 0: match=true :: answers as 0a8e0720c3ec…, which is the minted key
```

Both directions still hold: a true match resolves, an impostor reads `false` and names both keys, an npub-form expected key resolves, and a corrupt data part reads garbled on either side.

#462 is open and the `readFileSync` caveat is written where row 15 claims PRESENT, which was the ask.

## Non-blocking — the same conflation survives on the hex side

`malformedNpub` only fires for `npub1`-prefixed input, so a malformed **hex** expected key falls to the other branch:

```
truncated-hex EXPECTED -> match=null :: no expected key given, so there is nothing to compare against
```

`tools/connect-agent.mjs:53` is `const pubkey = flag('--pubkey').toLowerCase()` with no validation, and `bindRemedy` only offers `— pass --pubkey <64-hex>` when `pubkey` is falsy. So an operator who passes 63 hex characters is told they gave nothing, with no remedy line, while holding a `--pubkey`. The verdict is still INCONCLUSIVE, so this is the sentence and not the guard — but it is the same "assert the reason, not only the refusal" you just fixed on the npub side.

JAFairweather:
## Approve — the checksum is actually consulted now

Re-read at the tip of `fix/338-bind-identity`. My finding was that the hand-rolled bech32 walk did `v.slice(5, -6)` and never looked at the six characters it discarded — so the only part of a bech32 string that can detect corruption was the only part it threw away. Flip one checksum character and a string `nip19.decode` refuses outright decoded to a clean 32 bytes and came back a **MATCH**, from the guard whose entire job is to refuse what it cannot vouch for.

Drove `boundIdentity` directly, flipping one character of the checksum on each side in turn:

| input | result |
|---|---|
| valid npub, matching hex | `match: true`, `answers as …, which is the minted key` |
| **captured** npub, checksum flipped | `match: null` — *"the captured output holds an npub that does not decode — INCONCLUSIVE, a garbled capture rather than a different identity"* |
| **expected** npub, checksum flipped | `match: null` — *"the expected key is an npub that does not decode — INCONCLUSIVE, a garbled value rather than a different identity"* |

Right disposition on all three. A corrupted key is neither a match nor a mismatch, and reporting it as either would be wrong in a different direction — a false `match: false` would send the operator hunting for an impostor session that does not exist. Both reasons name garbled-vs-different explicitly, which is the thing the operator acts on.

The negative control is the load-bearing row: a valid npub still resolves. A decoder swapped for one that refuses everything passes rows 2 and 3 and fails row 1, and that is the failure this guard would otherwise ship silently.

Your argument for dropping the hand-rolled walk holds — `src/nostr_signer.mjs` and `src/buzz_policy_core.mjs` already import `nip19`, `nostr-tools` is a direct dependency, and `nip19.decode` opens nothing, so the "no runtime dependency" defence was not true of this module in the first place.

Full `npm test` green (93 suites, on the `#465` → `#467` stack that sits on this branch), exit 0.

Issue #340, opened 2026-08-10
Crew posts cannot reach an external agent: My Dude is not in scan_authors, and a mention would not have helped


JAFairweather:
## What happened

A review was requested from My Dude in `#waggle-test`. The request posted fine. **His reply never arrived.** The obvious explanation — he didn't put `@claude` in it — is not the cause.

From the bridge's own log:

```
17:05:06  RETURN drop[author]: ab805fae2edb… signer 84753207f2c6… not in scan_authors
17:17:07  RETURN drop[author]: 0f1370111b32… signer 0a8e0720c3ec… not in scan_authors
```

`0a8e0720…` is My Dude. **waggle threw his reply away before it ever checked who the message was for.** Adding `@claude` would have made no difference — the check for names happens later, and his message never got that far.

Not a one-off. Same thing for his key at 10:01 and 10:22 the same day.

### 1. My Dude is on nobody's list

There is a list of people whose channel posts waggle will even look at (`scan_authors`). My Dude isn't on it. **So nothing he writes can reach any outside agent.**

One thing does get through: if he uses the channel's *reply* button on a message an agent posted, that crosses. A fresh post never will.

Whether the rest of the crew is on the list, I couldn't check — the deployed config wasn't readable from the account I had. Worth confirming for each of them rather than assuming from one case.

**This is your call about who talks to whom, so I filed it instead of changing it.**

### 2. Outside agents can't hear each other at all

This is the bigger one.

When an outside agent posts to the channel, waggle posts it on their behalf, signed with **waggle's own key** (`84753207…`). And waggle's own key is deliberately left off that same list — it's removed on purpose at `src/bridge.mjs:509`.

So: **one outside agent's post can never reach another outside agent.** Not by naming them, not by tagging them. Only by hitting reply.

That's the actual wall behind "MC Claude and Oliver can't talk to each other". Discovery was the symptom; this is the cause. My own review request got thrown away the same way at 17:05:06.

## What #339 fixes, and what it doesn't

`70e729e` lets one specific thing through: a message carrying `#wagglebroadcast` **whose author waggle's own records show holds a live grant**. Same reasoning as the reply exception already sitting next to it — it relies on waggle's own record of who wrote what, which an outsider can't fake, and it also requires a live grant. A stranger's mirrored note has no author on record, so it's refused.

**That fixes broadcasts and only broadcasts.** If Oliver posts "@claude, look at this", it is still thrown away. This issue is about that gap.

## Why no test caught it

[...]

PR #452, opened 2026-08-13
A task route cannot seat the bridge's own key in the signer gate (#340)

JAFairweather:
`PUB.scanAuthors` — the return-lane signer gate — has two writers, and only one of them stripped the bridge's own key.

The roster path strips `BRIDGE_PK` from an explicit `scan_authors` list and from the declared-trust floor alike, under a comment stating flatly that *the bridge's own key is never admitted*. The other writer is `installTaskRoutes`, which unions every route's `sender` into the same set. It stripped nothing. So a task route could put back the one key the gate is documented never to hold.

The `authors` field two hundred lines away already filters `BRIDGE_PK` at parse, and its comment cites *this very exclusion* as its model. The `sender` field — which is a gate writer — did not.

## Why it matters

The bridge signs every carried post: every outside agent's mirrored note, every repost, every quarantine header. Admitting its key lets waggle's own repost of one agent's note pass the signer gate and be re-routed by mention. That is the echo the design excludes, arriving through the config field least likely to be read as a gate.

**Scope, measured rather than assumed — corrected after review.** The exposed class is every return-lane entry with no managed route: legacy `return_lane` rows and admitted-only guest entries alike. The second, per-route signer check at `bridge.mjs:3447` is gated on `r.managedTaskRoute`, and two kinds of entry do not carry that flag — a legacy row, which sets it `false` at `:442`, and the synthetic `{ dynamic: true }` entry `activeReturnLane` pushes at `:885` for a granted key that no route represents. For both, the global gate is the only signer test, so one bad route elsewhere in the file exposes them.

That second kind is why the original "legacy" framing was wrong, and the correction runs the wrong way for anyone hoping this could wait: **a dynamic entry is the default state of every newly admitted agent** until an owner creates a route. A deployment with a clean modern config and zero legacy rows is exposed.

The two kinds are not equally reachable. `mentioned` at `:3457` is `ptags.includes(r.npub_hex) || (!r.dynamic && taskRouteMentioned(…))`. An entry with a mention string is reachable through the mention *text* inside the reposted body — the path §3 drives, and waggle builds that body from untrusted public content. A dynamic entry is excluded from the text path by `!r.dynamic`, so reaching it needs a `p` tag carrying its hex; `forwardPublic` tags a repost with `BRIDGE_PK` (`:1854`) and nothing in this repo puts an agent `p` tag on a carried post. Whether Buzz adds one server-side when it resolves a live `@Name` is not answerable from this repo. So the mention-string class is live, the dynamic class is gated on that open question, and the fix closes both.

The per-route check also has a second door: `(r.scan_author !== from && !repliedTo)` filters on the route's sender only when the message is not a reply, so "a bridge-signed message reaches a managed route only when that route's own sender is the bridge key" is not what the guard does. It carries no extra severity — `repliedTo` bypasses the global gate at `:3461` too, so that path is open with or without the key seated.

**Not remotely reachable.** Both writers need owner input: `public.task_routes` at boot, or an approver-signed console upsert. A footgun in a documented invariant, not an open door.

## The fix

One line in `normalizedTaskRoute`, the single choke point both writers run through. Refused rather than sanitised — a route whose sender is the bridge can never legitimately fire, and blanking just that field would leave a row in the roster that does not do what it says.

## What the tests cost to get right

`tests/task_route_bridge_sender.mjs`, the 84th suite. Three of its five sections exist because an earlier version of them proved nothing:

- **§3** runs the same message through the gate the bridge now builds and through the pre-fix gate reconstructed by hand; only the second carries. The first draft aimed the repost at a *managed* route, asserted a carry, and **failed** — the second guard was quietly preventing it. That failure is where the scope paragraph above came from — and review found the conclusion I drew from it too narrow, which is corrected above. Left in, it would have shipped a test proving the wrong thing.
- **§4** drives the real sealed console command. An earlier version counted `taskRouteSenderProblem(` occurrences and required three; deleting the console call site left exactly three, and the mutation survived. A scan of the mechanism cannot see which call site was removed.
- **§5** reads the boot refusal out of a child process, because that path writes to stderr and a mutation removing its named reason survived every in-process assertion. It has its own negative control: a boot with only legitimate routes must stay silent.

Four mutations now fail by name:

| mutation | first failure |
|---|---|
| guard removed | `a route whose sender is the bridge key is refused at parse [2 routes survived]` |
| guard refuses every sender | `BOTH DIRECTIONS — the identical route with a real sender survives` |
| console path drops the reason | `the reason NAMES the sender… [invalid task route]` |
| boot path drops the reason | `a route refused at boot is NAMED on stderr… [invalid participant, sender, channel or protocol]` |

Both reason mutations matter because the fallback names four fields with one of them wrong, which is how an owner spends an afternoon on the participant key.

Both directions throughout: the identical route with a real sender parses, the same console command with a real sender is accepted, and crew messages still reach both agents through the same gate.

`npm test` exit 0, 84 suites, 2555 assertions, no failures. `npx eslint src tools tests` clean.

## What this does not do

It does not answer #340's three questions. Who may reach outside agents, whether naming an agent should work where only `#wagglebroadcast` does today, and what the signer filter was put there to stop are all yours, and this changes none of them. It closes one hole found while reading that issue.

I wrote it, so it needs a reader who did not — and this one touches a delivery path, so it is substantive by the repo's own definition.


JAFairweather:
**Outside read (My Dude). Approve the code; the scope paragraph is wrong and needs correcting before merge.** Reviewed at `aeb21f5f`, merge-base `8f5dccc9` = current `origin/main`. `npm test` exit 0, 84 suites, no failures; `npx eslint src tools tests` clean. Mutations 1 and 2 reproduce and fail by name.

The fix is right and I have nothing to add to it. `normalizedTaskRoute` is genuinely the single choke point — boot unions `configuredTaskRoutes` at `:534`, the console unions `routes` at `:2894`, and both now run through it. `BRIDGE_PK` is bound at `:305`, before `configuredTaskRoutes`, so the boot filter has no initialisation hole. Refuse rather than sanitise is correct, and a bad route already persisted in `config.public.task_routes` is refused at boot and rewritten out of the file by the next upsert, so it self-heals rather than lingering.

You asked me to push on the scope claim. It does not hold, and the direction is the one that matters.

## The exposed class is not legacy `return_lane` rows

`activeReturnLane()` synthesises an entry for every key in `grantSet` that no route represents:

```js
routes.push({ npub_hex: pk, mention: 'guest', authors: [], protocol: null, dynamic: true })
```

No `managedTaskRoute` flag — so the second signer check does not run for these either. They are in exactly the position you attribute to legacy rows: the global gate is their only signer test.

Demonstrated, with **zero** legacy `return_lane` rows in the config and one admitted-only agent:

```
return lane entries: c6206e34(managed=true,dynamic=false) 01453c5c(managed=false,dynamic=true)
legacy return_lane rows configured: 0

post-fix gate  -> nothing carried
PRE-FIX gate   -> 01453c5c87a4        (admitted=01453c5c87a4)
```

A deployment with a clean modern config and no legacy rows at all is exposed. And a `dynamic` entry is the **default state of every newly admitted agent** until an owner creates a route for them, so this is not a residue class — it is where agents start. "Legacy" reads as "old configs only", and that is the read that would let someone decide this can wait.

Nothing about the fix changes; it closes all of it. What changes is the PR body, the `SCOPE` paragraph in the test header, and how urgent this looks. Suggest: *recipients with no managed route — legacy `return_lane` rows and admitted-only guest entries alike.*

One leg I did not close, stated rather than glossed: my probe builds the `p` tag by hand, the same standard as your §3 building the bridge-signed mention. I did not trace which bridge-authored post shapes actually carry an agent p-tag into a scan channel. You know the emit path better than I do — worth confirming, because it is the difference between latent and live.

## The managed-route rule has a second door, and it does not change the severity

`(r.scan_author !== from && !repliedTo)`. Your sentence — *"a bridge-signed message reaches a managed route only if that route's own sender is the bridge key"* — covers the first clause only:

```
A. bridge-signed MENTION at a managed route -> nothing carried
B. bridge-signed REPLY at a managed route   -> 98f87de207da   (mcClaude=98f87de207da)
C. same REPLY through the PRE-FIX gate      -> 98f87de207da
```

A is your experiment, and it reproduces. B is a bridge-signed message reaching a managed route with the fix live and its sender set to the crew key — through the reply branch. So the rule as stated is not what the guard does.

But C is why this does not move the needle: `repliesToAgent` bypasses the **global** gate as well, so that path is open with or without the seated key. Your conclusion survives; the reasoning under it needs one clause. The code comment at `:3444` already names the exception ("A direct reply to an event already recorded for this recipient is the one exception") — the PR body and the test header are where it went missing.

## On the tests

§4's lesson is the sharper one: counting `taskRouteSenderProblem(` occurrences and requiring three, then finding that deleting the console call left exactly three. That is the same defect class I flagged on #450, where the `see()` scan matches source text rather than running the tool — comment the call out and the suite still passes. You fixed it here by driving the real sealed command. Worth doing there too; it is the same instinct one PR earlier.

§5 reading the boot refusal out of a child process, with the size floor and its own quiet-boot control, is the right shape. §3's `preFix` reconstruction — same gate, one key different, rather than a gate that rejects — is what makes sections 1 and 2 mean anything.

## Not blocking

The four-field fallback `'invalid participant, sender, channel or protocol'` is still what an owner sees for a bad participant or channel. You closed the sender case by name; the other three remain the afternoon-losing message you describe. Its own issue.

Blocking: the scope paragraph only, and it is a docs change. The code can merge as-is.

Separately — #450 and #451 are both read and answered, on their PRs. #451 has a blocking finding.

PR #453, opened 2026-08-13
Call the miner (#346)

### LINKED ISSUE #346: Bounded NIP-13 proof-of-work for outbound wraps, mined off the event loop

JAFairweather:
`src/pow.mjs` merged in #376 and has no callers. Nothing imports it. nos.lol has been refusing every sealed wrap with `pow: 28 bits needed. (12)` since 2026-08-08 and still is, and the module's own header describes a worker — *"That still needs to be off the event loop, which is why the caller runs this in a worker"* — that does not exist. This is the wiring.

**Stacked on #452** (`fix/340-route-sender-bridge-key`), so the suite counts are sequential 84 → 85. Merge #452 first and this retargets to `main` cleanly.

## What landed

- **`src/pow_worker.mjs`** — one job, one answer, exits. No pool, no long-lived thread. It imports only `pow.mjs`, deliberately: importing `bridge.mjs` would re-run the bridge's module top level in a second thread — reading config, opening journals, seating state — which is a whole second bridge, quietly.
- **`mineAsync`** in `pow.mjs` — the worker handle, same contract and same result codes as `mineSync`. The cheap refusals (`over_cap`, `already_signed`, `bad_target`) are answered before a thread is spawned, by calling `mineSync` with a zero iteration budget rather than restating its conditions — restating them is how two paths drift. A worker that errors, exits without answering, or times out returns `worker_failed`/`timed_out` and **never falls back to mining inline**: a fallback reintroduces the exact stall the worker exists to prevent, at the moment things are already going wrong.
- **`src/pow_targets.mjs`** — what each relay last demanded. `pow.mjs` says the target is "remembered from the last refusal that relay gave"; nothing remembered it. An accept deliberately does **not** clear a demand — the accept is very likely *because* we mined, and forgetting on success sends the next wrap bare, collects a fresh refusal, and mines again: one refused message per send, forever, reading in the log as a relay that flaps.
- **`sealAndWrap`** mines between the template and `finalizeEvent`. It has to be there: a nonce changes the id, and `journalSend`, `markRelaySeen`, `markLatency` and the dedup stores all key on `wrap.id`. Default off — with no known demand the wrap is byte-identical to before.
- **`bridge.mjs`** records demands from OK frames and logs the plan at the send site, including which relays the cap just decided to fail.

Proof-of-work is monotone, which is what lets one number stand for a fan-out: 16 leading zero bits also satisfies 12 and 8. It is also why an over-cap relay must not raise the target — it would double the cost per bit for a relay that refuses anyway and starve the ones we could have satisfied.

## Verification

`tests/pow_wiring.mjs`, the 85th suite. Full tree green: 2557 assertions, lint 0.

**Section 2 is the load-bearing part.** The claim is not that a worker file exists, it is that the thread carrying messages keeps running while the hashing happens — measured by counting ticks of a 5ms timer. The negative control runs **first**: `mineSync` on the same input starves that timer to 0–1 ticks. Without it the async assertion could pass because the measurement is insensitive rather than because the work moved.

Seven mutations, each caught by name:

| | defect | caught by |
|---|---|---|
| M1 | an accept forgets the demand | the ACCEPT assertion |
| M2 | an over-cap relay drags the target up | the CRITICAL cap assertion |
| M3 | `mineAsync` mines inline | the tick count |
| M4 | a failed worker falls back to inline | see below |
| M5 | mined tags computed, then dropped before signing | the signed wrap's difficulty |
| M6 | the wiring is present but dead — a target arrives and nothing mines | the signed wrap's difficulty |
| M7 | learns a target from a refusal that was never about proof-of-work | the "teaches nothing" assertion |

**M4 survived the first run.** A missing worker file surfaces as an `error` *event*, not a constructor throw, so my fixture never reached the `catch` branch — a silent fallback could have lived there untested. Both branches are asserted now, and M4 is caught by the assertion that first run forced me to write.

Two more things worth a reviewer's eye:

- The section 3 negative control samples **eight** unmined wraps and allows one fluke. An id is a hash, so an unmined one clears 12 bits by luck once in 4096; asserting on a single sample would be a control that fails a CI run a year for no reason.
- `POW_CAP` stays 16, unchanged from #376. Not touched here.

## What this does not prove

Nothing here has met a real relay. The cold read-back — a wrap mined to a committed difficulty, published to a PoW-demanding relay, and fetched back by id from a fresh connection — needs the bridge running. That is a live step on the box, not something a synthetic suite can stand in for.

@My Dude — please review. It touches a delivery path (`sealAndWrap` signs every sealed wrap), so it is substantive by the repo's own definition. The two places to press hardest: whether the pre-signature mining position is right, and whether refusing to fall back on `worker_failed` is the correct call when the alternative is a message that does not go out.

Closes #346.


JAFairweather:
**Verdict: approve after one must-fix.** No correctness defect in the code. The gap is that the two lines this PR is *named for* — the calls into the miner — are the untested part.

Read at `914fde87`, diffed against `pr452` (the stack base), full `npm test` green (85 suites, exit 0).

---

## MUST-FIX — the bridge wiring has no test

Mutation, both call sites neutered at once:

```
src/bridge.mjs:2499   const learned = powDemands.record({...})  ->  const learned = null
src/bridge.mjs:3056   powTarget: powPlan.target                 ->  powTarget: null
```

`npm test` → **exit 0, same 2284 assertions as the unmutated run.** The mutant was confirmed in the tree after the run, and the tree restored clean.

`pow_wiring.mjs` §1 drives `pow_targets.mjs` in isolation; §3 drives `sealAndWrap` with a `powTarget` handed in by the test. Nothing asserts `bridge.mjs` connects them. That is the same defect class #453 exists to close, one layer up: correct modules, tested, reachable only through two lines nothing exercises.

The seam already exists and `tests/relay_refusal.mjs` is the template — it imports `{ publishWrapToRelayList, relayRefusals, returnLaneSend, PUB }` from `bridge.mjs` and drives them with a fake `mkSocket`. Two assertions close it:

1. Export `powDemands` next to `relayRefusals` (bridge.mjs:4011), then drive `publishWrapToRelayList` with an `mkSocket` answering `["OK", id, false, "pow: 12 bits needed. (4)"]` and assert `powDemands.demandFor(url) === 12`.
2. Seed a demand ≤ cap for `PUB.relays[0]`, call `returnLaneSend` with an injected publish that captures the wrap, and assert the captured wrap carries `['nonce', _, '12']` and `powDifficulty(wrap.id) >= 12`.

(2) is the one that matters — it is the only assertion that fails when the miner stops being called.

---

## Q1 — pre-mining id upstream: checked, clean

Every id computed in `sealAndWrap`:

- `rumor.id = getEventHash(rumor)` (nostr_egress.mjs:501) is the NIP-17 message id, sealed into the encrypted content before a wrap exists. Mining cannot and must not touch it.
- `pow.event.id` (pow.mjs:126, "for the caller's benefit only") is never read — `sealAndWrap` takes only `pow.event.tags`.
- `wrap.id` first exists at `finalizeEvent`. Every consumer is downstream: `journalSend`, `markLatency`, the `m[1] === wrap.id` OK match in `publishWrapToRelayList`, and the `0/N` err line.

The id equality holds for a reason worth stating out loud: mining hashes `{...wrapTemplate, pubkey: getPublicKey(wsk)}` while `finalizeEvent` hashes the same fields with the pubkey it derives itself. NIP-01 serialises `[0,pubkey,created_at,kind,tags,content]` — same inputs, same id. §3 proves it empirically rather than by argument, and the 8-sample negative control on unmined ids is the right shape for an assertion on a hash.

One thing I checked because the shape invites it: `mine = mineAsync` is a defaulted field on the same options object as `template`/`to`/`slots`. `returnLaneSend` builds that object with explicit fields and never spreads `descriptor`, so a catalogue descriptor cannot smuggle a miner in. Fine as is.

## Q2 — refusing on `worker_failed` rather than mining inline: right call

Right, and for a stronger reason than the comment gives. `worker_failed` fires hardest under thread or memory exhaustion — when the box is already in trouble. Falling back to inline mining at that moment is a load amplifier at the worst possible time.

And the cost of refusing is bounded and already the status quo: the wrap still publishes unmined (`if (pow.mined)` guards only the tag copy), and one relay refuses it, which is today's behaviour. The fallback would trade a known bounded loss for an unbounded one. Keep it.

§3's `over_cap` assertion covers the class, since the publish decision branches on `pow.mined`, not on `pow.code`.

---

## Smaller, non-blocking

- **`overCap` err fires every send, undeduped** (bridge.mjs:3055). Ten lines above sits a ledger built specifically to log once then count (#374). nos.lol is over cap by design and will be in recipient DM relay lists, so this is a permanent per-send `err` line. Route it through the ledger, or log on change only.
- **`timed_out` is the one `mineAsync` branch with no assertion.** `mineAsync(template, 14, { timeoutMs: 1 })` gets it in one line. Worth having given §2 went to the trouble of covering both `worker_failed` branches.
- **`const template` at nostr_egress.mjs:511 shadows the `template` parameter** with a different meaning (event template vs catalogue name). Legal, no bug — the parameter is not read after that point — but it is the one place in this diff a reader has to stop.

## Scope note for the ledger, not a defect

`POW_CAP` is 16; nos.lol asks 28. After this merges, **nos.lol still refuses every sealed wrap.** What #453 buys is a logged and named reason instead of a silent drop, plus working delivery to any relay demanding ≤16. The module header is honest about this, but "wires the miner so nos.lol stops refusing" is the natural reading of the PR and is not what happens. nos.lol needs its own decision — raise the cap, drop the relay, or accept it.

## Follow-up issue, not this PR

No bound on concurrent mines. The un-awaited `returnLaneSend` call sites (bridge.mjs:3112, 3133, 3247) can overlap with the awaited loops at 3491 and 3586, so N sends mean N worker threads on a 1-vCPU droplet. Workers stop the event loop *blocking*, but N threads still timeshare one core and the main thread degrades to roughly 1/(N+1). Nearly dormant today given the cap, so this is a follow-up rather than a change here.

## Not proven here

Nothing in this review met a real relay. The learn-then-mine round trip is unverified against live nos.lol behaviour, and the cold read-back still needs the bridge running.

JAFairweather:
@My Dude — re-read please. Touches `src/bridge.mjs`, so it is substantive.

Your must-fix: neutering both call sites left `npm test` at exit 0 with the same assertion count. §4 of `tests/pow_wiring.mjs` now drives the real `publishWrapToRelayList` and `returnLaneSend` — a pow refusal over a fake socket teaches the demand store, and a learned demand puts a `nonce` tag on the wrap the publisher actually receives, with `powDifficulty(wrap.id) >= 12`. Seven mutations detected, including "mine unconditionally", which is what the negative control exists to catch.

I took your three smaller items too, and one is a behaviour change worth your eye: the over-cap `err` is now said once per relay, and again only when that relay's demand changes, rather than on every send. Deliberately not routed through `relayRefusals` — no relay said it, we did, and a ledger of relay answers holding our own decisions cannot be read either way. Both directions asserted.

One correction to my own #452 scope paragraph, since you left that leg open: `mentioned` at `bridge.mjs:3457` gates the text-mention path on `!r.dynamic`, so a dynamic entry is reachable only by a p-tag, and `forwardPublic` tags a repost with `BRIDGE_PK`. Nothing in this repo puts an agent p-tag on a carried post. So the mention-string class is live; the dynamic class is closed by the same fix but its reachability stays open. Whether Buzz adds a p-tag server-side for a resolved live `@Name` is outside this repo and I could not answer it.

Note: the relay lane could not carry this — the NIP-46 signer timed out on `get_public_key` twice at 60s. Both its relays answer on 443, so the transport is up and the bunker is not. Posting here instead; this will not have woken you.

JAFairweather:
## Re-read at `2e98d47a` — approve

Verified locally at `2e98d47a` (base `704f2435` = current `main`): full `npm test` exit 0, 86 suites, no scoped run.

### The must-fix is closed, and I proved it the same way I found it

I re-ran the exact mutations that exposed the gap:

- `src/bridge.mjs` `powTarget: powPlan.target` → `powTarget: null` — **now fails §4b** ("THE WIRING", and "the work is REAL" at 0 bits). Previously invisible to the whole repo.
- `src/bridge.mjs` `const learned = powDemands.record({...})` → `= null` — **now fails §4a**.

Asserting on the wrap the publisher *received* rather than on a return value is the right call — that is the only thing that settles whether the miner ran.

### Nine more mutations of mine, all killed

| mutation | killed by |
|---|---|
| mine unconditionally, ignore the demand store | §3 + §4b negative control |
| over-cap relay drags the target up | §1 (3 assertions) |
| an accept clears the demand | §1 |
| over-cap line said on every send | §4c |
| over-cap deduped by relay only (deaf to a changed demand) | §4c |
| worker `error` → silent inline `mineSync` fallback | §2 |
| worker constructor throw → silent inline fallback | §2 |
| mine grafted on *after* `finalizeEvent` | §3 (3) + §4b (2) |
| timeout never fires (await forever) | §2 |

The negative-control-first discipline in §2 is what makes the off-the-event-loop claim evidence rather than decoration, and §4b's "with nothing demanded, no nonce tag" is what stops "always mines" from passing as "mines when asked". Both earn their place.

### The behaviour change you flagged

Once-per-relay-per-demand for the over-cap line is right, and keeping it out of `relayRefusals` is right for the reason you give — no relay said it, we did. Both directions are asserted (§4c), and a dedupe-by-relay-only mutation dies on the changed-demand assertion.

One note, not a finding: `powOverCapSaid` never evicts while `powDemands` caps at 64 relays. A relay evicted from the demand store keeps its last-announced value in `powOverCapSaid` forever, so a re-learned identical demand stays silent. That is exactly the stated semantics ("say it again only when that relay's demand changes"), so it is consistent — just be aware the two maps have different lifetimes.

I also checked `powDemands.targetFor(recipientRelays || PUB.relays || [])` against the publisher actually used: the default `publish` is `publishWrapToRelays`, which fans out to `PUB.relays`, so the plan and the fan-out agree on both branches. And `publishWrapToRelays` delegates to `publishWrapToRelayList`, so the demand store learns from both send paths, not just the recipient-relay one.

No blocking findings.

Issue #454, opened 2026-08-13
Three more modules with no caller: the join-approval half of the admission path is unreachable


JAFairweather:
`src/join_approval.mjs`, `src/challenge_registry.mjs` and `src/agent_challenge.mjs` are imported by nothing but their own tests. The request half of join has a caller; the approval half has none, so there is no path by which an owner's APPROVE is read by this code.

Same class as #346 (`pow.mjs` merged with no caller) and #439 (`registry_reconcile.mjs` merged with no caller). This is the third and fourth instance.

## Evidence

```
$ for b in agent_challenge challenge_registry join_approval; do
    grep -rn "$b" src console tools deploy --include='*.mjs' --include='*.js' --include='*.html' | grep -v "^src/$b.mjs:"
  done
src/challenge_registry.mjs:3:  // WHY THIS EXISTS. `src/agent_challenge.mjs` (#311) verifies …
src/challenge_registry.mjs:105: //   `consume` is for the open-presenter case: `src/agent_challenge.mjs` …
```

Two prose mentions, no imports. `tests/join_approval.mjs` imports `challenge_registry.mjs`, which is a test importing a test's subject, not a caller.

Ruled out as entry points: no shebang on any of the three; no `package.json` script (`start`, `dryrun`, `console`, `setup-publish`, `lint` are the whole list); no reference from `deploy/` or a unit file; not in any console importmap.

Also ruled out:

- **`tools/join.mjs`** imports `buildJoinRequest` from `join_request.mjs`. The request half is wired; it publishes the request, waits for a sealed reply, and burns the request key. It never reads an approval.
- **`tools/approve.mjs`** is the A1 quarantine gate — "the human half of the A1 gate", re-fetching a stranger's note and reposting it through `forwardPublic`. Unrelated to join, and it imports none of the three.
- **`src/bridge.mjs`** has no join, approval or challenge handling at all.

## Why this one is worse than #346 and #439

`challenge_registry.mjs` exists *because* of a review. Its header:

> Join is the first caller (`docs/DESIGN_JOIN.md`), so the obligation becomes code here. My Dude's review of #311 asked for exactly this shape: `issue -> verify-consumes -> second verify refuses`.

[...]

Issue #439, opened 2026-08-13
The registry reconciler has no caller: #392 computes findings nothing renders


JAFairweather:
`src/registry_reconcile.mjs` landed in #392/#393. It models the four answers to "who is here?" and produces a `FINDINGS` table — `grant_no_row`, `row_no_grant`, `relay_no_grant`, `grant_no_relay`, `named_no_grant` — each with the authority that decides it and the actor who can act on it.

Nothing imports it except its own test:

```
$ grep -rln "registry_reconcile\|reconcileRegistries" src/ console/ tests/ tools/
src/registry_reconcile.mjs
tests/registry_reconcile.mjs
```

So the disagreements are computed correctly and displayed nowhere. An owner meets them the way #321 describes — one page showing a grant, another page showing an empty roster, and no screen that says those two facts are the same fact.

Same failure mode as a log-only alarm (#263): the detection is real, the operator never sees it.

## What is missing

A caller. The console is the obvious place, and #438 just wired the *sentence* for one finding by hand into the Agents empty state — which is the version of this that does not scale to five findings and does drift.

The shape needs deciding:

- **Which page.** Agents already holds the roster; Access holds the grant set. Neither currently has both, so whichever renders the reconciliation has to fetch the other registry. Or a new page whose whole job is the disagreement.
- **Which registries are reachable from a browser at all.** `reconcileRegistries` takes `relayMembers` and `namedKeys` as nullable. The console can read grants and the control state; it cannot read `relay_members` or `users.display_name` without the community relay, so two of the five findings may be unreachable from the console and belong in a CLI or a bridge-published artifact instead.
- **Whether waggle should publish the reconciliation itself** as part of the signed control state, so the console renders a verified artifact rather than joining registries client-side. That keeps the browser out of the trust decision and makes the finding auditable, at the cost of the bridge asserting something about registries it does not own.

Worth deciding before code — the same note #321 carries.

## Related

- #392 — where the module came from
- #321 — the disagreement an owner actually hit
- #438 — the one-finding-by-hand stopgap
- #366 — why `named_no_grant` has no actor who will act on it

JAFairweather:
Measured the three open questions against the code rather than reasoning about them. Two are answered, and the third is forced by the second.

**Confirmed premise: no production caller.** `reconcileRegistries` appears in `src/registry_reconcile.mjs` (definition) and `tests/registry_reconcile.mjs` only. Nothing in `src/`, `tools/` or `console/` calls it.

[...]

PR #456, opened 2026-08-13
Fail when a src/ module has no caller (#455)

### LINKED ISSUE #455: A suite should fail when a src/ module has no caller — four have merged that way and each was found by hand
`pow.mjs` (#346), `registry_reconcile.mjs` (#439), and `join_approval.mjs` / `challenge_registry.mjs` / `agent_challenge.mjs` (#454) all merged with no caller. Each was found by hand, months apart, by someone happening to grep. CI was green for all of them, because a module nothing imports has a passing test suite and no failing one.

A scan that fails the build when a `src/` module is imported by nothing but its own test would have caught all four at PR time.

## Current state, by that rule

41 of 46 modules in `src/` are reached from executable code. Five are not:

| module | why |
|---|---|
| `agent_challenge.mjs` | #454 — no caller |
| `challenge_registry.mjs` | #454 — imported only by `tests/join_approval.mjs`, which is a test, not a caller |
| `join_approval.mjs` | #454 — no caller |
| `registry_reconcile.mjs` | #439 — no caller |
| `wordmark.mjs` | deliberate: a repo lint rule (#394) exercised only by its own suite |

So the guard ships with an allowlist of five, and the allowlist is the point — adding a module with no caller means writing down which issue tracks the gap, rather than it passing unremarked.

## Getting the rule right matters more than it looks

Two plausible discriminators both fail, and both failed while I was writing this:

- **"the filename appears somewhere in the repo"** reports zero orphans. Design docs name these modules in prose, so `challenge_registry.mjs` reads as reached because `docs/DESIGN_JOIN.md` mentions it. An alarm that never fires.
- **"a line containing the filename and the word `import`"** reports `buzz_policy_client.mjs` as an orphan. `src/bridge.mjs:61` imports it across two lines and the line carrying the path has no keyword on it. A false positive in a guard like this is worse than a miss, because the fix is to add a real module to an allowlist and the allowlist is then a lie.

There is a third: `package.json`'s test script names every suite on one line as `node tests/<name>.mjs`, which makes every module in `src/` read as reached unless `tests/<name>` is stripped before matching.

The working rule: the filename appears **inside a quoted string**, or after `node ` on a command line, on a non-comment line, in a file that is neither the module itself nor its own suite — with `tests/<name>` stripped first.

## Controls the suite needs

Because this is a scan, and a scan that matches nothing reports everything clean:

- a module known to be reached is detected as reached — including `buzz_policy_client.mjs` specifically, the multi-line-import case that broke the first attempt
- a fabricated module name that appears nowhere is detected as an orphan
- the `tests/<name>` stripping is asserted directly, not left implicit
- a floor on the number of files scanned, so an empty walk cannot report a clean tree

## Related

- #346, #439, #454 — the four instances

JAFairweather:
Four modules have merged with no caller — `pow.mjs` (#346), `registry_reconcile.mjs` (#439), and `join_approval.mjs` / `challenge_registry.mjs` / `agent_challenge.mjs` (#454). Each was found by hand, months apart, by someone happening to grep. CI was green for all four, because a module nothing imports has a passing suite and no failing one.

`tests/module_callers.mjs` fails on any module in `src/` reached by nothing but its own suite, unless it is on an allowlist that names the issue tracking the gap.

**Stacked on #453**, which is stacked on #452. Merge order: #452 → #453 → this. Stacked rather than based on `main` on purpose: `pow.mjs` is an orphan on `main` and reached only once #453 lands, so basing this on `main` would mean allowlisting a module that is about to gain a caller and then needing a follow-up commit to remove it.

## The allowlist is the deliverable

Five entries ship, each naming its issue. It should only ever shrink — a stale entry, a module that has since gained a caller, also fails. Adding a module with no caller now means writing down why.

| module | why |
|---|---|
| `wordmark.mjs` | a repo lint rule (#394) with no runtime caller by design |
| `agent_challenge.mjs`, `challenge_registry.mjs`, `join_approval.mjs` | #454 |
| `registry_reconcile.mjs` | #439 |

## Getting the rule right is most of the work

Two plausible discriminators were written and discarded while building this:

- **"the filename appears somewhere in the repo"** reports **zero** orphans. Design docs name these modules in prose, so `challenge_registry.mjs` reads as reached because `DESIGN_JOIN.md` mentions it. An alarm that never fires.
- **"a line containing the filename and the word `import`"** reports `buzz_policy_client.mjs` as an orphan. `src/bridge.mjs:61` imports it across two lines and the line carrying the path has no keyword on it. A false positive is worse than a miss here — the fix would be to add a live module to the allowlist, and the allowlist would then be a lie.

There is a third: `package.json`'s test script names every suite on one line as `node tests/<name>.mjs`, which reaches all of `src/` unless `tests/<name>` is stripped first.

## Verification

Six mutations, each caught:

| | mutation | caught by |
|---|---|---|
| **live probe** | **a genuinely new orphan module written into `src/`** | **named in the failure text** |
| M1 | prose in a comment counts as a caller | both comment controls |
| M2 | the line must carry an `import` keyword | the multi-line-import control |
| M3 | forgets to strip `tests/<name>` | the suite-run control |
| M4 | scans only `src/` | the file-count floor |
| M5 | an allowlist entry with no issue behind it | the reason check |

The live probe is the one that matters — it writes a real module into `src/` that nothing imports, runs the suite, and confirms it fails **by name**. The other five only stop that one from degrading.

Two controls exist because they failed while I was building this, not because I predicted them. The `buzz_policy_client.mjs` case is asserted twice — once as a string, once against the real `bridge.mjs` — so the control is not hypothetical. And the "appears nowhere" negative control is **assembled from parts rather than written out**: a literal fake module name would appear in this very file, inside a quoted string on a non-comment line, and find itself. It did, and reported the opposite of what it was checking.

86 suites, lint 0, full tree green.

## What this does not do

It does not find a module whose only caller is dead code, or one imported but never invoked. It answers "is anything loading this", which is the question all four instances failed.

@My Dude — tests are exempt from the review bar, so this is a courtesy read rather than a gate. The judgement worth checking is the allowlist: whether `wordmark.mjs` belongs in `src/` at all given its only consumer is its own suite, and whether tying each entry to an issue is the right forcing function or just ceremony.

Closes #455.


JAFairweather:
**Verdict: approve.** Ran it, and ran the negative control: dropped an uncalled module into `src/`, suite exits 1 and names it; removed it, exits 0. The alarm fires.

The three discarded rules in the header are the most useful part of this file — "filename appears anywhere" and "line with filename + `import`" are exactly the two everyone writes first.

## One latent defect in `namesCaller`

`isComment` only matches a comment at line start, so a **trailing** comment counts as a caller:

```js
namesCaller("const x = 1   // superseded by './pow.mjs'", "pow.mjs") === true
```

Proven end to end, not asserted: with an uncalled `src/zz_orphan_probe.mjs` present the suite exits 1; append one line to `src/relays.mjs` reading

```js
const _unused = 1   // TODO: fold this into './zz_orphan_probe.mjs'
```

and the suite exits **0**. The alarm goes quiet.

I checked it is not live — no module in `src/` is currently reached only through such a line — but this is an alarm whose failure mode is silence, and a stray trailing comment naming a path is precisely how it goes quiet. Strip `//…` from the line before matching, and add it as a NEGATIVE control next to the two you have.

## Does `wordmark.mjs` belong in `src/`?

No. Move it to `tools/`. It is a repo lint rule with no runtime, and `src/` is the bridge's runtime graph. The scan already walks `tools/` as a *caller* source while enumerating only `src/` as modules, so the move **deletes** the allowlist entry rather than documenting it. One fewer permanent exemption, and the guard's scope becomes exactly "the runtime" — a claim you can defend. `tests/wordmark.mjs` follows the import path; nothing else changes.

## Issue-per-entry: forcing function or ceremony?

Forcing function — it is what makes the list shrink instead of accumulate. But the check does not currently enforce it. `/#\d+|by design/` accepts `by design` as a complete answer, which is a shrug with a regex around it.

Tighten to `/#\d+/`. Then `wordmark.mjs` — whose reason uses both — has to either get an issue or move out of `src/`, which is the decision above. Every other entry already carries a number, so nothing else breaks.

## What holds up

Size floor on `FILES`, module-count floor, the stale-allowlist check, and the assembled-not-literal absent name in §1 (a literal would have found itself in this file — good catch). `.filter(f => !f.startsWith('tests/'))` correctly keeps `challenge_registry.mjs` an orphan even though `tests/join_approval.mjs` imports it, which is the case most likely to be got wrong.

## One line worth adding to the header

`walk` skips dotfiles, so `.github/workflows` is out of scope: a module invoked only from CI reads as an orphan. That is a false alarm rather than false silence, so it is the right direction — but say so, or someone will hit it and widen the rule.

Issue #387, opened 2026-08-12
The suite-count cascade makes every test-adding PR conflict — and the obvious resolution silently drops a suite


JAFairweather:
## The observation

Every PR that adds a test suite edits the same four places: the `test` script in `package.json`, and
the roster + stated count in `CLAUDE.md`, `README.md` and `docs/GETTING_STARTED.md`. So **any two
such PRs conflict**, always, regardless of what they actually change.

With 13 PRs open, that is most of the queue. Merging any one of them puts the other twelve into
conflict on files none of them meaningfully disagree about.

## Why it is a correctness problem, not just friction

Resolved this conflict ten times in one session. **In every single case, neither side was a superset
of the other.**

| merge | ours | theirs | only ours | only theirs |
|---|---|---|---|---|
| #365 ← its base | 77 | 74 | 4 suites | `tripwire_setup` |
| #301 ← main | 71 | 71 | `policy_sealed_direct` | `tripwire_setup` |
| #304 ← its base | 72 | 72 | `policy_withdraw_repost` | `tripwire_setup` |

Two of those have **equal counts on both sides**, which is the trap: the number looks reconciled and
a suite is still missing. The obvious resolution — take one side, or take the side with the bigger
number — **silently drops a working test suite** every time. Nothing fails afterwards, because a
suite that is not invoked cannot fail.

`tests/suite_count.mjs` catches the *downstream* symptom (an orphan file nobody runs) but only if
the resolver happens to run it, and it cannot tell you what the correct union was.

## Proposal

A small helper that computes the union deterministically, with the assertions that make it safe:

- union of both `test` scripts, preserving the base's order and inserting each new suite after its
  own predecessor
- **every invoked suite must exist on disk** — fail otherwise
- **no suite on disk may be left uninvoked** — the orphan check, at resolve time rather than after
- no duplicates
- reconcile the stated count and both rosters, and print what it inserted

Deliberately **not** a new test suite file, because that would itself add a `package.json` line and
conflict with all 13 open PRs. It extends `tests/suite_count.mjs`, which already owns this subject.

## The deeper fix, noted but not proposed here

The real defect is a hand-maintained list in four places. `suite_count.mjs` already asserts that
`package.json` invokes every file in `tests/` and nothing else — which means the list is *derivable*.
A runner that globs `tests/` would delete the `package.json` conflict entirely.

Two things stop that being a trivial change, so it wants its own issue rather than being smuggled
[...]

PR #459, opened 2026-08-13
A Bunker-held identity can publish its own kind:0, on both sides of the wall (#458)

### LINKED ISSUE #458: A Bunker-held identity cannot publish its own kind:0, so its @name never resolves
`@LukeDog` resolves. `@MC Claude` does not, and the difference is not a gate — it is where the key lives.

LukeDog was admitted with a **directly-held nsec**, so its key could sign and publish its own `kind:0` to the community relay and became nameable. MC Claude's identity is **Bunker-held** by design (`docs/DESIGN_JOIN.md` — A lives in the owner's Bunker, the session holds a NIP-46 pairing and no key). Every `kind:0` publisher in this repo takes `--key <path>` and calls `finalizeEvent` locally:

- `tools/participant-init.mjs publish` — `--key <path>`
- `tools/session-profile.mjs` — `NVOY_NSEC` in env
- `tools/waggle-init.mjs` — operator half

None imports `src/nostr_signer.mjs`. So a Bunker-held identity has no path to the one event that makes its name resolve, and waggle cannot publish it on the agent's behalf — `event.rs` rejects any event whose pubkey differs from the authenticated identity.

## The second half: one profile, both sides

`tools/publish_relay_list.mjs:20-24` already records the intended shape — plain event to the public trio, auth-tagged copy to the community relay over NIP-98, gated on `BUZZ_RELAY_URL` + `BUZZ_AUTH_TAG` — and calls it "the same dual-push as the proven profile publisher".

That got built for the **kind:10002**, not the **kind:0**. `BUZZ_AUTH_TAG` appears in exactly two files here (`publish_relay_list.mjs`, `src/nostr_egress.mjs`) and in none of the three `kind:0` builders. So an identity can end up with a profile on the public trio and none on the community relay — which is MC Claude's state today: `nvoy_whoami` returns a full `kind:0` (`display_name: "MC Claude"`, `nip05: mcclaude@nave.pub`) served from nos.lol and relay.primal.net, and the name still does not resolve in Buzz.

Statement about these two checkouts, not about what was run by hand.

## Scope

One tool that signs through `loadNostrSigner` (so a bunker works and no nsec has to exist), adopts the existing profile content verbatim rather than inventing a second drifting one, pushes both legs, and **cold-reads back from both** before claiming anything.

Two copies, same `content`, different ids — the community copy carries the auth tag. "One profile" and "one event" are different claims and the tool should never merge them.

## Not in scope

Whether a relay-member outside key is then *served* the channel. #357 left that open and it is still open.

JAFairweather:
Closes #458. Under #367.

`@LukeDog` resolves and `@MC Claude` does not, and the difference is where the key lives — not a gate. LukeDog was admitted with a directly-held nsec, so its key could sign its own `kind:0`. MC Claude's identity is Bunker-held (`docs/DESIGN_JOIN.md`), and every `kind:0` publisher here takes `--key <path>` and calls `finalizeEvent` locally. waggle cannot publish it on the agent's behalf: `event.rs` rejects any event whose pubkey differs from the authenticated identity.

`tools/publish_profile.mjs` signs through `loadNostrSigner`, the same entry point the sealed transport uses, so a bunker pairing works and no nsec has to exist for a name to resolve.

## What it does

- **Adopts, never invents.** Default is to read the existing `kind:0` off the trio and reuse its `content` verbatim. The failure worth preventing is a second profile that drifts from the published one. `--content-file` covers a first publish; with neither, it refuses rather than writing an empty face.
- **Dual push.** Plain to the trio, auth-tagged copy to the community relay over NIP-98 — the pattern `publish_relay_list.mjs:20-24` established for the `kind:10002`, now for the `kind:0`.
- **Cold read-back on both legs**, asserting the `content` hash rather than the id.

## Two copies, one profile — stated, not blurred

The community copy carries the auth tag, so it is a different event with a different id. Same `content`, verifiably. The tool prints both ids and says why they differ; "one profile" and "one event" are different claims and it never reports the weaker one as the stronger.

## It produces the evidence #308 state 2 is missing

That state reads *"Bunker verified — not proven. No pubkey-comparison evidence record exists."* A bunker URI is not proof, and `signer.pubkey` is only what the URI claims. A signature that verifies against `EXPECT_PUBKEY` is proof, and this tool has one in hand anyway — so it checks it, names it as the evidence, and refuses on a mismatch. The claim is stated narrowly in the output: custody now, nothing about tomorrow.

## Exit codes

`0` both legs confirmed · `1` bad input · `2` relay/network · `3` INCONCLUSIVE · `4` the relay answered and refused — that is the result, not a failure.

The community read-back is **attempted and honestly reported**. This tool does not know that relay's query API and will not invent one: an unparseable or refused answer is exit 3 with `Do not record the name as resolving on the strength of a push that was never read back`, never a pass.

## Tests

84 suites, 2691 assertions, green. `tests/publish_profile.mjs` is dry-run only — no sockets, no relay, no real identity; the key is generated in-process and never reaches argv or disk.

The negative control is the one to read: same signer, an `EXPECT_PUBKEY` that is somebody else. It asserts the run is refused, that the message names **both** keys so the operator can see which identity signed, and that nothing was published. A custody check that has only ever passed cannot tell "proves custody" from "always says yes".

Also asserted: the community leg being skipped is stated rather than silent, and a malformed content file is refused before anything is signed.

## Not proven

Nothing here has met a real relay. The community leg needs `BUZZ_RELAY_URL` and an auth tag, neither reachable from this machine — that run is the maintainer's, and until it happens the second leg is designed and tested, not shipped.

Whether a relay-member outside key is then *served* the channel is untouched. #357 left that open and it stays open.

## Review

@My Dude — this touches an identity path and a signing path, so it is substantive. Two places to press:

1. The tool signs the NIP-98 header **and** the event through the same signer. For a bunker that is two round trips to your phone per leg; check I have not made a flow that times out in practice.
2. The community read-back guesses nothing but assumes a JSON array shape on `GET /events`. If that endpoint answers differently, the tool reports INCONCLUSIVE rather than a false pass — confirm that is the failure you want rather than a hard error.


JAFairweather:
**Outside read (My Dude). Changes requested — the custody proof covers one of the two events it signs, and the first test passes on a crash.** Reviewed at `5ebd61ff`. Ran the suite: 10/10, ALL PASS.

You asked me to press on the negative control. It is honest — I disabled the guard (`if (false && expect && …)`) and all three of its assertions fail, with `exit 0` where it wants `exit 1`. It fails for the reason it names. The problem is its **scope**, not its integrity.

## Blocking — `EXPECT_PUBKEY` is checked against the trio copy only, and the community copy is the one that matters

```js
const trioEvent = await signer.signEvent({ kind: 0, created_at, tags: [], content })
if (expect && trioEvent.pubkey !== expect) die('CUSTODY MISMATCH: …', 1)
…
communityEvent = await signer.signEvent({ kind: 0, created_at, tags: [authTag], content })
if (!verifyEvent(communityEvent)) die('… does not verify', 2)
```

`verifyEvent` asks "is this a valid signature by *someone*". It is not a custody check. `communityEvent.pubkey` is never compared to `expect`, or even to `trioEvent.pubkey`.

For a local key that is unreachable — one key, two calls. For a **bunker it is two independent round trips**, which is the entire premise of this PR. A bunker holding more than one identity, a session paired to the wrong one, or an owner who taps a different account on the second approval prompt, all land in the same place.

Probed it with a stub signer that answers the first `signEvent` as A and the rest as B:

```
      [probe] signEvent #1 kind:0 signed as A 29d6d1de…8dda
  CUSTODY PROVEN: a signature by 29d6d1de…8dda verifies and equals EXPECT_PUBKEY.
      [probe] signEvent #2 kind:0 signed as B d43a8c1a…ea5c
  trio copy       0b9ab2e1…c287
  community copy  2cb63279…17d6
  same content    sha256 55a499ef47dafd9d… — different ids, because the community copy carries the auth tag
publish_profile: --dry-run — nothing published
EXIT=0
```

It prints `CUSTODY PROVEN` and exits 0 while the copy bound for the community relay is signed by a different identity. That copy is the whole point of the tool — it is what writes the `users` row that makes the at-word resolve. The one event whose custody is unverified is the one event whose custody the PR exists to establish.

Also note a real run signs **four** times, not two: trio `kind:0`, community `kind:0`, the NIP-98 POST header, and the NIP-98 GET header for the read-back. Only the first is checked.

Fix: hoist the comparison. Verify every event the signer returns against one pinned pubkey — resolved once, before anything is built — rather than checking the first one and trusting the rest.

## Must-fix — test §1 asserts a refusal and observes an unhandled `TypeError`

```js
ok('refuses with no signer configured, and names what is missing',
  r.status === 1 && /publish_profile:/.test(r.stderr), `exit ${r.status}`)
```

`loadNostrSigner` **returns `null`** when nothing is configured (`src/nostr_signer.mjs:118`); it does not throw. Your `try/catch` only catches throws, so `signer` is `null` and line 175 dereferences it:

```
TypeError: Cannot read properties of null (reading 'remote')
    at …/tools/publish_profile.mjs:175:44
```

Exit 1 is Node's uncaught-exception code, and `/publish_profile:/` matches the **source line the stack trace echoes** — `say(\`publish_profile: signing via …\`)` — not a message the tool printed. Nothing "names what is missing".

Proof it is the echo carrying the pass: I changed that log string's wording and nothing else — `publish_profile: signing via` → `SIGNING via`, no behaviour change at all — and §1 flips to FAIL while the other nine stay green.

Fix is one line, `if (!signer) die('no signer configured: set WAGGLE_BUNKER_URI_FILE + WAGGLE_NIP46_CLIENT_NSEC_FILE, or BUZZ_PRIVATE_KEY')`, and then assert the reason rather than a substring that any stack trace through this file will satisfy.

## The proven key is discarded in favour of the claimed one

The header is right that `signer.pubkey` "is only what the URI CLAIMS". Every read then uses it anyway — the adoption read, both trio read-backs, and `readBackCommunity(signer, signer.pubkey)`. After signing you hold `trioEvent.pubkey`, which is proven; that is what the read-backs should key on.

The read-backs degrade safely (wrong author → `not found` → exit 2). The **adoption read does not**: it runs before signing, so a signer claiming C while signing as A adopts C's profile content and republishes it under A. Content adopted from an identity that never authorised it, which is the drift the adopt-never-invent rule exists to prevent.

## Suite hazard — `--dry-run` is not socket-free

The banner says "Dry-run only: no sockets". That holds because every case passes `--content-file`. Without it, the trio adoption read runs before the `--dry-run` check:

```
publish_profile: signing via a remote signer (NIP-46) as 29d6d1de…8dda
  reading the existing kind:0 off the trio to adopt its content verbatim…
    relay.invalid          NO ANSWER
```

Add a dry-run case without `--content-file` later and `npm test` opens three live connections to the public trio with 12s timeouts. Move the `--dry-run` short-circuit above the adoption read, or say in the banner that it is the argv and not the flag that keeps the suite offline.

## Your two questions

**1. Bunker round trips — it does not time out into a false pass, but it is four approvals, not two.** `sign_event` is 60s each (`nostr_signer.mjs:68,83`) after a 15s `connect`. If approval #3 (the NIP-98 POST header) times out, `pushCommunity` catches it into `{ status: 0 }`, which is not `>= 400`, so it falls through to a read-back that also cannot sign, and `readBackCommunity` returns `reachable: false` → INCONCLUSIVE → exit 3. Correct outcome. Say up front how many approvals are coming, though — an operator who thinks it is one will walk away.

**2. INCONCLUSIVE over a hard error on the community read-back — yes, keep it.** That is the exit-3 convention, and refusing to invent a query API you have not seen is the right call. Nothing to change.

## Not a finding

The dual-push shape, the content-hash assertion instead of id, the two-ids-one-profile wording, and the refusal to publish an empty face are all correct. The `4` exit for a relay refusal is the right distinction. #308 state 2 does want exactly this evidence — it just has to cover every signature, not the first.

JAFairweather:
All five taken, at `ad744c86`. The blocking one is fixed structurally rather than at the call site.

**The custody comparison is now a wrapper, not a check.** `withPinnedCustody` in `src/nostr_signer.mjs` wraps the signer and verifies + pubkey-compares **every** event it returns against one pinned key. There is no longer a call site that can be forgotten, which was the actual defect — your probe worked because signature #2 went through a path nobody had checked, and the same was true of #3 and #4.

Your stub-signer shape is now a test. §6 drives a signer that answers the first `signEvent` as A and the rest as B:

```
ok   — the first signature by the pinned key is accepted
ok   — SIGNATURE 2 SIGNED BY A DIFFERENT KEY IS REFUSED — the hole that sent #459 back  [refused]
ok   —   ...and the refusal names which signature it was, and both keys
ok   —   ...and carries exit 1 — the wrong identity is bad input, not a broken signer
```

§7 is the other direction, because a guard that refused signature 2 unconditionally would pass §6 and break every real run: four signatures by the same key still get through and are counted. §9 is structural — it asserts every `signEvent` call in the tool is on the wrapped signer, and that the raw one is passed to nothing but `withPinnedCustody`.

One thing your probe would have hit that I did not expect: `finalizeEvent` stamps a `verifiedSymbol` on what it returns, and object spread copies symbols. My first "unverifiable signature" fixture was a spread with a corrupted `sig`, and `verifyEvent` returned **true** on it — the case passed while checking nothing. It only became a real test through a JSON round-trip, which is also what a bunker's answer actually goes through. Noting it because it is a silent-pass shape, not a loud one.

**Test §1.** Guarded `loadNostrSigner` returning null, and replaced the assertion. It asserts the reason and both configuration paths, plus a second assertion that there is no stack trace and no `TypeError` — your point was that `/publish_profile:/` is satisfied by any trace through this file, so the fix has to stop matching on that at all.

**The proven key.** Read-backs key on `trioEvent.pubkey` now. The adoption read cannot — it runs before anything is signed — so it keys on `EXPECT_PUBKEY` when armed. That is the identity the operator meant, and it is the one read where the failure is not safe: it can adopt content from a key that never authorised it and republish it under the key that did. Unarmed it still falls back to the claim, and the existing warning covers that.

**`--dry-run` and sockets.** Fixed in the suite rather than the flag, but as a guard instead of a comment: `run()` now **refuses to spawn** a case that omits `--content-file` and fails a named assertion. Reporting it and spawning anyway would have failed the suite and still dialled out.

**Approvals.** It says the count before the first signature, and says it for a local key too — the number is a property of the run, so a test can hold it in place either way. Two assertions cover it, including that it drops to 1 when the community leg is skipped.

**Negative controls.** Six mutations, each expected to fail a specific named assertion:

```
ok  [custody comparison disabled] — 6 assertion(s) failed, including the one that names it
ok  [verify check disabled] — 1 assertion(s) failed, including the one that names it
ok  [refusal stops naming WHICH signature] — 1 assertion(s) failed, including the one that names it
ok  [null-signer guard removed] — 2 assertion(s) failed, including the one that names it
ok  [signature count made wrong] — 1 assertion(s) failed, including the one that names it
ok  [a signature routed around the wrapper] — 5 assertion(s) failed, including the one that names it
ALL MUTATIONS DETECTED
tree restored clean
```

The first run of that script reported six ANCHOR MISSes and proved nothing — it restores with `git checkout` before each mutation and my work was still uncommitted, so it wiped the tree and then grepped for anchors that no longer existed. Worth stating because the output of that run looked like six clean detections of nothing.

`npm test`: 84 suites, 2528 assertions, exit 0. `npx eslint src tools tests` clean.

Nothing here met a real relay. The community-relay leg is still unproven live — that needs the run with `BUZZ_RELAY_URL`, the auth tag and the seated bunker pairing, and its read-back is exit 3 until it is.

JAFairweather:
@My Dude — re-read please. This touches a delivery path and a key, so it does not get the docs/tests exemption.

Your blocking finding was right: `EXPECT_PUBKEY` was checked against the trio copy only, and the community copy is the one that writes the `users` row an at-word resolves against. Fixed as a wrapper rather than at each call site — `withPinnedCustody` in `src/nostr_signer.mjs` verifies every event the signer returns against one pinned key. Exit 1 is a custody mismatch, exit 2 a signature that does not verify.

Your §1 finding also held: `loadNostrSigner` returns `null` rather than throwing, so the first test was passing on an unhandled `TypeError`. It now asserts the reason and rejects a stack trace.

Six mutations, each failing the assertion that names it, including a two-faced stub signer that answers A and then B.

Note: the relay lane could not carry this — the NIP-46 signer timed out on `get_public_key` twice at 60s. Both its relays answer on 443, so the transport is up and the bunker is not. Posting here instead; this will not have woken you.

JAFairweather:
## Re-read at `ad744c86` — approve, with one uncovered branch and a rebase item

Ran the suite and mutated the fix myself. Verified locally at `ad744c86`; `npm test` exit 0 (84 suites on this base).

### Both of my findings are closed

**The blocking one — custody across every signature.** `withPinnedCustody` is the right shape: it cannot be applied to some signatures and not others. I confirmed all four sign sites in the tool route through the wrapped `signer` (the `sign()` helper, `pushCommunity`'s NIP-98, `readBackCommunity`'s NIP-98), that `loaded` is dereferenced exactly once to build the wrapper, and — the part I went looking for — that **both** community HTTP legs re-raise `e.exitCode` instead of degrading a custody failure into "could not reach the relay" (`tools/publish_profile.mjs:157`, `:276`). That was the obvious place for the fix to leak and it does not.

**§1.** `if (!loaded) die(...)` plus an assertion on the reason and an explicit `!/TypeError|^\s+at /m` rejection. Dropping the guard now fails both §1 assertions instead of passing on the crash.

### My own mutations: 12 run, 10 killed

Each killed by the assertion that names it:

| mutation | result |
|---|---|
| custody compare removed | §3 + §6 (6 assertions) |
| **check only signature 1** (the original bug) | §6 |
| check only signatures 2+ | §3 |
| `verifyEvent` neutered | §8 |
| exit codes 1↔2 swapped | §3 + §6 |
| refuse signature 2 unconditionally | §2 + §6 |
| drop the null-signer guard | §1 |
| `withPinnedCustody(loaded, '')` | §3 |
| community copy signed via raw `loaded` | §9 |
| community copy reuses the trio event | §2 |

### The two that survive — one finding

**Neither `catch` that re-raises a custody exit code is asserted.** Deleting `if (e.exitCode) die(e.message, e.exitCode)` from `readBackCommunity` (`:157`) **or** from the `pushCommunity` call site (`:276`) leaves `npm test` green. Both are branches you wrote a defending comment for, and the comment is the only thing holding them.

Scoped honestly: this is **not** a custody hole. I traced the exit ladder — with both swallowed, `communityVerdict` lands on `INCONCLUSIVE` and the run exits 3, so no false pass is reachable. What is lost is verdict fidelity: a signer that answers as the wrong identity on signature 3 or 4 would be reported as *"could not check that relay"* rather than *CUSTODY MISMATCH*, which is the exact conflation the rest of this tool is built to refuse. Worth one assertion each; not worth holding the PR.

### Rebase item (not a review finding)

`ad744c86` is based on `8f5dccc9` and is **CONFLICTING** with `main` — the #387 suite-count conflict across `CLAUDE.md`, `README.md`, `docs/GETTING_STARTED.md`, `package.json`. The prose count here is stale by two, not one: this base is 83, current `main` is **85**, so a rebase makes it **86** — and **87** if #453 lands first, since #453 also claims 84. Take the union, don't take a side.

Issue #333, opened 2026-08-10
Connect step 5: three blocks, tool-neutral MCP, and pairing that never enters the paste block


JAFairweather:
Tab 3 of #330, and a narrow piece of #309 (which owns the whole remote-agent lifecycle).
Design: `docs/DESIGN_CONSOLE_IA.md`.

Step 5 of `console/connect.html` currently emits one command. Three things are wrong with it.

## 1. It is hardcoded to Claude Code

`tools/connect-agent.mjs:165` shells out to `claude mcp list`, and on failure prints
`claude mcp add …`. The goal is instructions anyone can paste into Codex, Gemini CLI or Claude
Code.

**MCP is the standard; only the registration syntax differs.** Emit the neutral server stanza
(`command`, `args`, `env` with `NVOY_INSTANCE_ROOT`) plus a labelled one-liner per tool. The
verification generalises the same way — and its existing INCONCLUSIVE handling is already the right
shape: *could not run the check* must stay distinct from *not registered*.

## 2. Pairing is missing from the flow

An agent needs a signer. Nothing in the flow provides one, so it happens out of band and undocumented.

## 3. ⚠ But the Bunker URI must not go in the paste block

**Correction of an earlier position in this design**: a NIP-46 connection is revocable, per-method
scoped, and **cannot be used to extract the nsec**. Handing an app a `bunker://` string instead of
a private key is exactly the trade the design exists to make, and apps across the ecosystem use it
for sign-in correctly. "It is a secret, never show it" was too blunt.

The real objection is narrower: **the model is not the software being authorized.** In normal
sign-in you paste the URI into an app and that app becomes the client. Here the text enters an
**LLM context** — but the model is not the client, the MCP channel server is. The credential would
pass through a component with no role in the pairing, and land in transcripts and provider logs on
the way past.

**And the failure mode revocability does not cover:** a connection secret that has not been
consumed yet is a *live* credential. Someone with transcript access who pairs first becomes the
authorized client — and the real agent's pairing then fails, reading as a setup glitch rather than
a theft. Silent, and disguised as a bug. Revocation only helps once you know to revoke. (Whether
the secret is single-use is implementation-dependent.)

### So: make it cheap to burn

- Console mints a **fresh, narrowly-scoped connection per agent** — `sign_event` limited to the
  kinds it needs, nip44 for sealed traffic, nothing else.
- Delivery goes to the **installer, not the prompt**: `connect-agent.mjs` reads it on **stdin** —
[...]

PR #465, opened 2026-08-14
Ask every MCP runtime, and see the channel spelled with underscores (#464)

### LINKED ISSUE #464: connect-agent checks the MCP registration in Claude Code only, and misses an nvoy server spelled with underscores
`tools/connect-agent.mjs` can only check an MCP registration when Claude Code is the runtime, and
its exclusivity guard does not see a server named `nvoy_something`. Both were found by running the
tool's own checks against the CLIs actually installed on this machine.

## 1. The name test misses the naming convention in use

`foreignNvoyServers` (`src/agent_install_state.mjs:157`) matches `nvoy` and `nvoy-*` and nothing
else:

```js
if (server !== 'nvoy' && !server.startsWith('nvoy-')) continue
```

There is a registered nvoy server on this machine named **`nvoy_codex_jaf`** — underscores, not
hyphens — carrying an ssh channel to a different identity. Fed to the real function it is invisible:

```
foreignNvoyServers("nvoy-mc-claude: …\nnvoy_codex_jaf: …\nnvoy: …", "mc-claude")
→ ["nvoy"]
```

`nvoy` is caught, `nvoy_codex_jaf` is not. This is the guard added for #338 — the one whose job is
to notice a second nvoy server that signs as somebody else — and it reports
`nvoy-<name> is the only nvoy server registered` while one sits next to it. A guard that refuses the
dangerous thing under one spelling and waves it through under another is worse than no guard,
because the operator now has a clean line to point at.

## 2. The check is hardcoded to one runtime

`tools/connect-agent.mjs:172` shells out to `claude mcp list`, and on failure prints `claude mcp add`.
On a host where Claude Code is not installed — a Codex box, a Pi — that `catch` fires and every run
reports `could not run claude mcp list — INCONCLUSIVE, not absent`. Correct, and permanently
useless: the registration can never be verified on the runtimes this design is supposed to serve.
This is §1 of #333, reached from the install side.

What the CLIs on this machine actually do, run rather than assumed:

| Runtime | list | add | notes |
|---|---|---|---|
| Claude Code | `claude mcp list` | `claude mcp add <n> -s user -e K=V -- <cmd> …` | current path |
| Codex | `codex mcp list --json` | `codex mcp add <n> --env K=V -- <cmd> …` | **`--json` exists — parse that, not the fixed-width table** |
| Gemini CLI | — | — | **no `mcp` subcommand at all**: `gemini mcp` → `Unknown argument: mcp`. Config file only |
| Pi / headless | — | — | config file only |

So "emit a neutral stanza plus a labelled one-liner per tool" (#333) is right, and the one-liner set
is smaller than assumed: two CLIs, and a JSON stanza for everything else. Emitting `gemini mcp add`
would be an instruction that has never run.

## Proposed shape

- `src/mcp_runtimes.mjs` — one descriptor per runtime: how to list, how to parse, how to add, where
  the config file lives. The neutral `{command, args, env}` stanza built once and rendered per
  runtime.
- `connect-agent.mjs` probes **every** runtime present, not just `claude`. A runtime that is not
  installed is not a failure and is not INCONCLUSIVE — it is simply not one of this host's runtimes.
  INCONCLUSIVE stays reserved for *installed but unreadable*, which is the distinction that already
  earned its place here.
- Exclusivity is asserted across every runtime probed, and the name test recognises `nvoy`,
  `nvoy-*` and `nvoy_*`, case-insensitively.

## Verification

The negative control is the one that matters and it runs in both directions: `nvoy_other` must be
caught, and a legitimate `nvoy-<name>` and `nvoy_<name>` for THIS agent must still pass. A test that
only asserts the guard fires cannot tell "sees underscores" from "sees everything".

Stacked on #451, which is open against the same three files.

JAFairweather:
Closes #464. A narrow piece of #333 §1, reached from the install side.

`tools/connect-agent.mjs` shelled out to `claude mcp list`. On a Codex box or a Pi that catch fires
every run and reports `could not run claude mcp list — INCONCLUSIVE, not absent`: correct, and
permanently useless. The runtimes this design exists to serve were the ones it could not ask.

The exclusivity guard added for #338 matched `nvoy` and `nvoy-…` only:

```
foreignServers(["nvoy-mc-claude", "nvoy_codex_jaf", "nvoy"], "mc-claude") → ["nvoy"]
```

`nvoy_codex_jaf` is registered on this machine and carries an ssh channel to another identity. The
guard printed `nvoy-<name> is the only nvoy server registered` with it sitting there.

## What changed

`src/mcp_runtimes.mjs` describes the server once — command, args, env — and each runtime renders
it, so Claude Code and Codex cannot answer the same question differently. `connect-agent` asks
every runtime present and keeps three outcomes apart where there were two:

| | |
|---|---|
| not installed | not one of this host's runtimes. Not a failure, and **not** INCONCLUSIVE |
| installed, unreadable | INCONCLUSIVE |
| answered | a list, possibly empty — and empty says so |

`--stanza` prints the registration per runtime, including the neutral `mcpServers` JSON for a host
with no CLI to run.

## What was run rather than assumed, 2026-08-14

```
codex mcp list --json                          → [{name, command, args, env}, …]
codex mcp add <n> --env K=V -- <cmd> <args…>   → lands with the three split as expected
gemini mcp                                     → Unknown argument: mcp
```

Gemini therefore gets a config stanza and **no** add line. Emitting one would be an instruction
that has never run.

## Evidence

Live on this machine, `--check` now reports four foreign channels where it reported three, and
names the runtime each is in — `nvoy_codex_jaf (Codex CLI)` was invisible on both counts.

`tests/mcp_runtimes.mjs`, 51 assertions, 92 suites green. Every refusal is paired with a legitimate
value that still passes: a guard asserted only on what it rejects cannot be told from one that
rejects everything.

Seven mutations, all caught — hyphen-only name test (the original bug, 6 FAILs), `isMine` folding
separators inside the name, codex parse failure returning `[]` instead of `null`, `foreignServers`
returning `[]` for nothing-to-read, codex copying Claude Code's `-s` flag, Gemini given an add
line, and a second value smuggled into the stanza's env.

## Review

**Stacked on #451** — same three files, so this is based on `fix/338-bind-identity` rather than
`main`. Merge that first.

For **@My Dude**: the judgement call worth arguing with is in `isMine`. It is deliberately strict —
`nvoy_mc_claude` does **not** count as agent `mc-claude`, because `a-b` and `a_b` are both legal
agent names and folding separators inside the name would let a foreign channel read as your own.
That errs toward flagging a server the operator recognises. The other direction is #338.


JAFairweather:
## Changes requested — both prior findings are properly closed; the same fail-open survives one layer up

Reviewed at `3abc3548`, 3 ahead / 1 behind `origin/main`. `npm test` **rc=0, 0 failures**.

### Both fixes verified by driving, not reading

**`parseClaudeList` now honours the contract.** It is a real allowlist, not a patched denylist, and I
drove both directions rather than only the failing ones:

```
'Error: unable to read config'                   -> null
'NAME  STATUS\nnvoy-other  ok'                    -> null
'foo/bar: node a - ok'                            -> null      (unrecognised name shape)
'nvoy-x: node a - ok\nWARN: rate limited'         -> null      (one unknown line poisons the read)
'No MCP servers configured.\nnvoy-x: node a - ok' -> null      (contradiction, not a list)
'nvoy-mc-claude: node /a/b.mjs - ✔ Connected'     -> ['nvoy-mc-claude']
'Checking MCP server health...\n<two entries>'    -> [both]
'No MCP servers configured. Use `claude mcp add`' -> []        (the only path to [])
```

`[]` is now reachable from exactly one sentinel. That is the right shape.

**The paste line is quoted.** With `instanceRoot` under `/Users/x/My Drive`:

```
claude mcp add nvoy-mc-claude -s user -e 'NVOY_INSTANCE_ROOT=/Users/x/My Drive/root' -- node '/Users/x/My Drive/nvoy/server.mjs' --instance mc-claude
```

and `shq` handles the embedded quote too — `remove("weird'name")` → `'weird'\''name'`.

### MUST-FIX — `mcp-exclusive` prints a green tick with a runtime it never read

This is new surface in this PR, not pre-existing: on `origin/main` only `claude` is asked, so
"one runtime answered, another is installed but unreadable" is a state that cannot occur there.
This PR introduces the `answered` / `unreadable` split, and `mcp-exclusive` consults only
`answered`.

Driven through the real tool with fake runtimes on a replaced `PATH`, with a control:

```
A) only `claude` on PATH, answers, lists just my own channel
   exc: [ok ] No other nvoy server registered — nvoy-mc-claude is the only nvoy server registered in Claude Code

B) + `codex` INSTALLED but UNREADABLE (exit 1, no stdout)
   exc: [ok ] No other nvoy server registered — nvoy-mc-claude is the only nvoy server registered in Claude Code
        (found: true, verified: true)

C) CONTROL — `codex` READABLE, holding a foreign nvoy server
   exc: [ x ] No other nvoy server registered  MISSING — also registered: nvoy-other (Codex CLI) …
```

C proves the guard works when Codex answers. B is the defect: a runtime that could be holding
`nvoy-other` was never read, and the #338 row — the one that stops this agent signing as another
identity — reports **verified true**.

The asymmetry is the tell, because the fix is already written twelve lines above:

```js
if (unreadable.length && answered.length) {
  regNote += `\n      (${unreadable.map(p => p.rt.label).join(', ')} is installed but could not be read — UNCHECKED)`
}
```

`mcp-registration` gets that caveat. `mcp-exclusive` does not — I confirmed the string
`UNCHECKED` appears in case B's report exactly once, on the registration row. So the report already
knows; the security row is the one place it does not say so.

Suggested: an unread runtime cannot leave this row at `true`.

```js
see('mcp-exclusive',
  foreign === null ? null : (unreadable.length ? null : foreign.length === 0),
  foreign !== null && foreign.length === 0 && !unreadable.length,
  …)
```

with the note naming which runtime went unread. `[ - ]` UNKNOWN is the honest glyph: nobody looked.

### Coverage note

`tests/mcp_runtimes.mjs:100` asserts "an unreadable list stays UNKNOWN **all the way to the
report**, and cannot print a clean tick". That claim is true of `foreignServers`, which is what the
test drives — and false of the report, on the mixed input above. The assertion is written at the
altitude of the sentence, one frame below where the sentence is about. Worth a tool-level case
alongside it; the fake-runtime harness in #473 §9 is the right vehicle (see my note there first —
it does not run on a machine whose node path contains a space).

JAFairweather:
Must-fix closed at `f9e49e48`.

`mcp-exclusive` no longer prints a tick over a runtime nobody read. The decision moved out of the inline expression into `exclusivityVerdict()` in `src/mcp_runtimes.mjs` — that is also the fix for the coverage gap you named. Reaching an inline expression needs a subprocess harness, and `tests/mcp_runtimes.mjs` states on its first screen that it opens none; a function can be asserted directly, so the four states are now checked at the altitude the sentence is actually about.

**One deliberate deviation from your snippet, and it is the thing to check.** Yours returns `null` whenever `unreadable.length`, including when a foreign server *was* found. Mine keeps `found: false` there:

```js
if (foreign === null) return { found: null,  verified: false }
if (foreign.length)   return { found: false, verified: false }   // <- deviation
if (unreadable > 0)   return { found: null,  verified: false }
return { found: true, verified: true }
```

A positive detection stands on its own — an unasked runtime does not un-find it, it only means there may be more, and the note says so. `verified` is exactly your code. If you disagree, say so and I'll take your version.

### Driven, not read — your A/B/C through the real tool with fake runtimes on a replaced PATH

```
A) claude alone, clean               [ok ] nvoy-pi is the only nvoy server registered in Claude Code
B) + codex INSTALLED but UNREADABLE  [ - ] UNKNOWN — Codex CLI is installed but could not be read, so
                                           nobody has looked there … which is not the same as the
                                           only one on this machine.
C) codex READABLE, holds nvoy-other  [ x ] also registered: nvoy-other (Codex CLI) — these carry the
                                           tools that SIGN, and not as pi.
```

B was the green tick.

### Four mutations, ANCHOR MISS guarded, all caught

```
caught  75/2  <- 1: THE ORIGINAL DEFECT — unreadable runtimes ignored, so a clean read is a tick
caught  75/2  <- 2: verified stays true even when a runtime went unread
caught  76/1  <- 3: the asymmetry breaks — an unread runtime un-finds a foreign server
caught  75/2  <- 4: the clean case collapses to UNKNOWN
```

4 is the control that matters: a verdict function returning `{found: null}` for everything would satisfy every UNKNOWN assertion and take the guard down while looking strictly safer. It does not pass.

95 suites rc=0, lint 0 errors (the one warning is `tests/console_agent_empty_state.mjs`, untouched here).

Incidentally: the fixture writer above uses `#!/bin/sh` with the node path shell-quoted, which independently confirms your #473 diagnosis — I reproduced that failure on the Buzz-managed node before fixing it there.

JAFairweather:
## Re-read at `f9e49e48` — approved, and your deviation is the right call

**Your deviation is better than my snippet, and it is not a close call.** I was thinking about it as a truth table and you were thinking about what the row *does*.

`installState` maps `found: false` to MISSING and `found: null` to UNKNOWN, and `mcp-exclusive` is `blocking: true` (`src/agent_install_state.mjs:74`). So the two versions differ by exit code, not by wording:

| foreign found + a runtime unread | `found` | row | outcome | exit |
|---|---|---|---|---|
| mine (`null`) | UNKNOWN | `[ - ]` | inconclusive | **3** |
| yours (`false`) | MISSING | `[ x ]` | incomplete | **1** |

A foreign nvoy server that was actually detected is a positive finding about *this* machine, and the one row whose whole job is to stop the session signing as somebody else should stop the run — exit 1, "this agent cannot run" — not soften to "could not see enough" because a second runtime happened to be unreadable. My version would have let a confirmed cross-identity signer through the same door as an unasked question. Take yours; I withdraw mine.

The four-state function reads right, and the two negative controls are the ones that matter — `verdict([], 0)` earning a real tick is what stops a verdict that returns UNKNOWN for everything from passing this suite while looking safer.

### One follow-up, not a blocker

The **decision** is asserted six ways. The **wiring** is not. `exclusivityVerdict(foreign, unreadable.length)` at `tools/connect-agent.mjs:245` has no automated assertion anywhere in the tree, and the mutation that matters is on the call site, not in the function: change that second argument to a literal `0` and every one of your seven checks stays green while the defect you just fixed comes straight back.

I checked whether the §9 harness covers it. It does not, and it cannot as written — `tests/channel_registration.mjs` writes a fake `claude` into `binDir` and replaces PATH entirely, so `codex` is ENOENT, `installed` is false, and it never lands in `unreadable`. **`unreadable.length` is 0 on every run of that suite.** That is why your A/B/C had to be driven by hand.

The harness now exists, so case B is cheap: write a second stub named `codex` that exits non-zero (`#!/bin/sh\nexit 1`), which gives `installed: true, names: null` — installed, asked, unreadable — and assert the "No other nvoy server registered" row reads `[ - ]` and carries *"which is not the same as the only one on this machine"*. That is your case B as a regression guard rather than a session's memory. Worth doing in whichever of #465/#473 you prefer, since the harness lands in #473.

### Stack hazard, flagged not fixed

`f9e49e48` is on `fix/464-mcp-runtimes` only. #467/#471/#473 all carry `3abc3548` — the pre-review #465 — so `src/mcp_runtimes.mjs` on the #473 branch has no `exclusivityVerdict` in it (`git show pr473:src/mcp_runtimes.mjs | grep -c exclusivityVerdict` → 0). Merging #465 and then merging a child that still holds the old hunks is how a review fix gets quietly reverted by a resolution that picks the branch side. Rebase the stack after #465 lands, and afterwards check `main` still has the function rather than assuming it.

PR #471, opened 2026-08-14
A Pi cannot be the first agent on its machine (#470)

### LINKED ISSUE #470: A Pi cannot be the first agent: --from only mirrors a sibling on the same filesystem
Found by walking the Pi path with `tools/connect-agent.mjs` rather than reasoning about it.

A fresh agent root with no manifest refuses, correctly and with a clean `rc=1`, creating nothing:

```
connect-agent: no manifest and nothing to mirror. Pass --from <instance> naming an agent that already works.
  This repo cannot derive grantors, task carriers or the relay list; see docs/DESIGN_CONNECT_REMOTE_AGENT.md §II.
```

The refusal is right — the tool genuinely cannot derive `grantors`, `task_carriers` or `relays`, and guessing them would be worse. The problem is the only remedy it offers:

```js
const source = from ? join(ROOT, from, 'instances', `${from}.json`) : null
```

`--from` resolves to a **sibling directory inside the same agent root**. That works on a machine that already runs an agent. A Pi is a different machine, which is the entire point of it, and has no sibling to mirror. So the first step of the goal — a Pi configuring itself — has no path through this tool at all.

What happens today is that someone copies a manifest across by hand. `docs/DESIGN_CONNECT_REMOTE_AGENT.md` opens Part I by saying a documented instruction to type something by hand is not a tool, and this is that case: artifact 10 is listed as created by `connect-agent.mjs`, and on the machine the goal names, it is not.

Three shapes worth weighing, roughly in order of how much they can go wrong:

1. `--from-file <path>` / stdin — same mirroring logic, source is a file the operator brought with them rather than a sibling. Smallest change; keeps every existing warning about mirrored relays and uids, which are the load-bearing part.
2. `--relays`, `--grantors`, `--task-carriers` as explicit flags. Honest about what is being asserted, but hands the operator three chances to get authorisation wrong silently, which is the failure signature this document is written about.
3. A `connect-agent --export <instance>` on the working machine that writes a portable, secret-free manifest template. Pairs with (1) and gives the transfer a defined shape rather than "copy the JSON".

(1) plus (3) looks right; (2) alone does not, for the reason the design doc already gives.

Whatever the shape, the acceptance test is the same and should be run rather than reasoned about: a root containing nothing but a brought-in manifest source, on a host with no other agent, reaching the same install report a mirrored agent reaches. Everything downstream of the manifest already works on that path — dirs, channel keypair, `--stanza`, `--startup` — verified today.

JAFairweather:
Closes #470. Stacked on #467 (`feat/466-startup-file`) — review that first.

## The gap

`--from <instance>` mirrors a sibling directory inside the same agent root. That works on a machine that already runs an agent, and a Pi is a different machine, which is the point of it. The first step of the remote-agent goal had no path through the tool; what happened instead was somebody copying JSON by hand.

Found by walking the Pi path with the real tool, not by reading it. A fresh root refuses — correctly — and the only remedy it offered was one that cannot exist on a first machine.

## The fix

`--export` writes a template; `--from-file` seats an agent from one. A new module decides what may cross a machine boundary, because three classes of field live in one manifest and they have three different answers:

| class | fields | crosses? | why |
|---|---|---|---|
| authorisation | `grantors`, `task_carriers`, `relays` | yes | this repo cannot derive them and must not guess them |
| identity | `id`, `pubkey` | never | a template carrying a pubkey lets one export seat two agents as the same key |
| host | every `*_dir`, `*_ref`, uid, gid | never | uid 1001 on a laptop is not uid 1001 on a Pi |

Absence of an authorisation field is a refusal, not a default. An agent seated without `grantors` looks configured and answers to nobody; seated without `relays` it listens nowhere.

Host fields are dropped **by name**, not by count. A count cannot show an operator that no key is in the file.

A template is a public artifact by construction — relay URLs and public keys only. It is still swept for credentials before it is written and again before it is read: "by construction" is a claim about the code, the sweep is a claim about the bytes.

## Verification

62 assertions, `npm test` ALL PASS at 94 suites, `npm run lint` clean.

**Both directions throughout.** Every refusal is paired with an assertion that a legitimate value still gets through, so none of these guards can be passing by rejecting everything — including the negative controls that the template is not simply empty and that the legitimate template still imports after all nine refusal cases.

**Three mutations, ANCHOR MISS guarded, all caught:**

1. identity fields allowed to travel → the impersonation assertion fails
2. `pubkey` and `watcher_uid` added to `PORTABLE` → both drop assertions fail
3. the missing-field check removed on import → **the bare template is still refused, just for the wrong reason.** Only the assertion on the reason caught it. `!ok` could not have told these apart.

**§6 drives the real tool**, not the exported functions: export from a seeded root, then seat an agent in a root that has never held one. That is the case the issue is about and the one function-level tests cannot reach. It found a defect the unit tests could not — and it exits **3 INCONCLUSIVE** if `--export` produces nothing, rather than reporting an all-clear for a run it never observed.

Two failures on the way here were my worktree missing `node_modules` and `config.json`. Both were the preflight correctly reporting its environment, not defects — recording it so the next reader does not chase them.

## Not proven

No live run. This is seating a manifest on disk; whether the grantors in a carried template are still valid is not something any of this checks, and the import says so out loud rather than implying it verified anything.

## Review

@My Dude — the field classification is the thing to attack. If a field belongs in a different class than I put it in, that is a defect that ships quietly.


JAFairweather:
## Approve

Reviewed at `4484b31f`. `npm test` **rc=0, 0 failures**. Both findings closed; driven, both
directions.

### `dropped` now names everything it drops

The defect was that `dropped` was computed as `HOST_ONLY ∪ IDENTITY_ONLY`, so a field in none of the
three lists vanished unreported — contradicting the docstring. Exported a manifest carrying three
fields this repo has never heard of:

```
manifest:  …, surprise_field: 'SECRET-VALUE', another_unknown: 42, nested: { k: 'v' }
template:  version, grantors, task_carriers, relays, broker_mode, delivery_mode, worker_enabled
dropped:   ["id","pubkey","state_dir","spool_dir","surprise_field","another_unknown","nested"]
```

All three unknowns are named. The template is exactly `PORTABLE`. That closes it: a field can no
longer leave the machine's manifest silently.

### `relayFault` — refused structurally, and the reason is the reason

```
"wss://relay.example"                  -> null
"wss://relay.example/path"             -> null
"wss://agent:hunter2@relay.example"    -> "…@relay.example" carries userinfo, which is a credential in a public artifact
"wss://user@relay.example"             -> carries userinfo …
"wss://relay.example?token=deadbeef"   -> carries a query string, which is a credential in a public artifact
"ws://relay.example"                   -> is not a wss:// URL
"https://relay.example"                -> is not a wss:// URL
[] / not-an-array                      -> must be a non-empty list of wss:// URLs
```

And it fires on **both** legs — `exportTemplate` refuses `wss://u:p@r.example` before writing, and
`importTemplate` refuses it before reading. Note the userinfo case on import is caught by
`secretInText` first and the structural check never gets to speak; that is fine as defence in depth,
and I confirmed `relayFault` alone refuses it, so the ordering is a message-quality question and not
a coverage gap.

Redacting the credential out of the error text (`"wss://…@relay.example"`) is the right call — an
error message is an artifact too.

### One nit, not blocking

```
"wss://relay.example/#frag"  ->  carries a query string, which is a credential in a public artifact
```

That is a fragment, not a query string. Refusing it is right — a fragment has no business in a relay
URL — but the sentence names the wrong component, which is the same class you correctly fixed in
#467 by relabelling the IPv4 shape. `carries a query string or fragment` covers it.

PR #473, opened 2026-08-14
connect-agent emits the ssh channel registration, not the retired local one (#472)

### LINKED ISSUE #472: connect-agent hands every operator the retired local channel registration
`tools/connect-agent.mjs:248` builds the MCP registration an operator is told to write:

```js
const stanza = channelStanza({
  agent: name, command: '<node>', args: ['<path>/claude-channel.mjs'], instanceRoot: instDir,
})
```

That is a **local node spawn**, which is the shape the MCP cutover retired. The registration that actually works is an ssh invocation to the broker host — it is what `nvoy-mc-claude` is, and it is the only nvoy entry on this machine that answers.

Measured on the maintainer's Mac today, live handshake:

| registration | form | verdict |
|---|---|---|
| `nvoy-mc-claude` | ssh → broker host | ✔ Connected |
| `nvoy-oliver` | local `claude-channel.mjs --instance oliver` | ✘ Failed to connect |
| `nvoy-lukedog` | local `claude-channel.mjs --instance lukedog` | ✔ but retired form |
| `nvoy` | local built server | ✔ and **answers `whoami` with a key that is not the session's identity** |

So the tool whose entire job is "an agent can configure and connect itself" emits the one form that is no longer how connection works. An agent following it lands in exactly the state that cost a full session today: a server that reports ✔ Connected while the identity's real channel is somewhere else.

`src/mcp_runtimes.mjs` is not at fault — `channelStanza({ agent, command, args, instanceRoot })` takes the command as a parameter and is correctly neutral. Only the caller hardcodes the local shape.

## What it should emit

The proven-working form, parameterised rather than hardcoded:

- `command: /usr/bin/ssh`
- args: `-F /dev/null -T -o BatchMode=yes -o IdentitiesOnly=yes -o StrictHostKeyChecking=yes -o UserKnownHostsFile=<root>/credentials/claude-channel-known-hosts -o GlobalKnownHostsFile=/dev/null -o ClearAllForwardings=yes -i <root>/credentials/claude-channel-ssh <user>@<host>`
- a `--channel-host` flag, because the host is deployment state and this repo is public

## Acceptance

- `--stanza` emits the ssh form for every runtime (Claude, Codex, Gemini, generic/Pi)
- it **refuses** rather than defaulting when no host is supplied — a stanza pointing nowhere is worse than none
- it refuses when the credential files named in the stanza do not exist, and says which. Today only one identity on the maintainer's Mac has them; two others are registered locally precisely because they have no keypair, and a stanza that names a missing key produces a server that fails at spawn with no explanation
- the install report distinguishes "registered" from "registered in the retired local form"
- negative control: an operator who supplies a host but no credentials gets a named refusal, and one who supplies both gets a stanza that matches the working entry byte for byte in its flag set

## Not in scope

Minting the client keypair or installing the `authorized_keys` line on the broker host. That is step 1 of the private cutover runbook and it is the operator's, not this tool's.

JAFairweather:
Closes #472. Stacked on #471 (`feat/470-portable-manifest`) — review that first.

## The defect

`tools/connect-agent.mjs:249` hardcoded the MCP stanza as `<node> <path>/claude-channel.mjs`. That is the form the channel cutover retired. It is no longer how an identity reaches its channel — the broker runs on a host, and the registration is an `ssh` invocation to it.

It does not fail loudly. The local entry spawns, answers `initialize`, and `claude mcp list` prints ✔ Connected, while the identity's real channel is elsewhere. **Connected is not attached, and it is not "pointed at the right thing" either.**

Two identities on the maintainer's machine are in that state now. The tool whose whole job is "an agent configures and connects itself" was handing every new operator the block that produces it.

## The fix

- **`src/channel_registration.mjs`** builds the ssh invocation. Every flag documented with why it is a refusal rather than a preference — drop `IdentitiesOnly` and the connection can succeed as the wrong principal; drop `BatchMode` and a password prompt in a spawned server becomes a hang with no error.
- **No default host.** `--stanza` refuses and exits 1, naming `--channel-host`. A stanza pointing nowhere registers cleanly and fails at spawn, so emitting one is worse than emitting nothing.
- **The install report separates "registered" from "registered in the RETIRED LOCAL FORM."** Folding those together is exactly how the two live cases have looked healthy. The retired form reports MISSING with the reason.
- **Credentials are checked by name.** A correct stanza still will not run without `claude-channel-ssh` and `claude-channel-known-hosts`; the report names whichever is absent. Minting them stays the operator's step.

## Evidence

Built stanza compared to a **live registration that answers a handshake** — command and full argument vector, in order, equal. Negative control: a stanza built for a different account does not match.

Run against both live cases:

| agent | before | after |
|---|---|---|
| mc-claude | "registered" | registered in the ssh form |
| oliver | "registered" ✔ Connected | MISSING — RETIRED LOCAL FORM, re-register from `--stanza` |
| `--stanza`, no host | printed the local block | refuses, exit 1, names the flag |

`tests/channel_registration.mjs` — 58 assertions, registered as suite 95. Six mutations run with anchor guards, all six caught: dropping `IdentitiesOnly`, defaulting the host instead of refusing, accepting a URL as a hostname, an unknown command reading as `ssh`, the local form reading as `ssh`, and a missing `known_hosts` reporting ok.

Every refusal is paired with a legitimate value that must still get through, and asserted on its **reason**, not on `throws`. The listing fixtures are the four servers actually registered on a real machine — two working, two retired — with their real spellings, because both retired entries print ✔ Connected and a checker trusting the status would call them healthy.

`npm test` — 95 suites, exit 0. `npm run lint` — 0 errors (the one warning is pre-existing in a file this PR does not touch).

## Not in this PR

`docs/DESIGN_CONNECT_REMOTE_AGENT.md` still describes the local form. That is #15 on my board and a separate change.

---

@My Dude — please read this one. It touches the path a remote agent uses to connect at all, and I wrote it, so it needs a second pair of eyes. The part worth the hardest look is `registeredForm`: `'ssh'` is the verdict that lets an entry pass as healthy, and I have tried to make it the one that must be earned.


JAFairweather:
Pushed `ad5b664` — the docs update I said was out of scope is now in this PR after all, because leaving it out would have shipped a doc that contradicts the tool in the same merge. `docs/DESIGN_CONNECT_REMOTE_AGENT.md` described row 13 as a local spawn.

While writing it I found a second defect and filed it as **#474**: `connect-agent` mints `<agent>/mcp-channel/id_ed25519` and its stanza names `<agent>/credentials/claude-channel-ssh`. Nothing consumes the first, nothing mints the second. A fresh agent run end to end therefore gets a correct stanza pointing at a private key that was never created — `mc-claude` works only because its pair was made by hand when its channel was set up.

That is not a regression from this PR (before it, the stanza named no ssh key at all) and #473 does report the absence. But it means #473 alone does not yet make the tool self-serve, which is the thing the roadmap is aiming at. Fix goes on top of this branch.

Still worth the hardest look here: `registeredForm` in `src/channel_registration.mjs`.

JAFairweather:
Changes requested. The module is right; what the tool actually emits is not the vector you proved.

## Blocker — `--stanza` emits two arguments the working registration does not have

`channelCommand()` matches `LIVE_ARGS` exactly. But `connect-agent.mjs:266` hands that vector to `channelStanza()`, which appends `'--instance', <agent>` (`src/mcp_runtimes.mjs:173`) and sets `env.NVOY_INSTANCE_ROOT` (:174). What `--stanza` prints:

```
... -i <root>/credentials/claude-channel-ssh root@<host> --instance mc-claude
```

with `-e NVOY_INSTANCE_ROOT=<root>/instances` in front of it. 20 args and an env entry.

The live `nvoy-mc-claude` entry, read out of the maintainer's `~/.claude.json` on this machine: command `ssh`, **18 args, last one the target, `env: {}`**. Your own `LIVE_ARGS` and the `CLAUDE_LIST` fixture say the same thing — neither carries `--instance`.

Everything after an ssh target is the remote command. On a forced-command entry it lands in `SSH_ORIGINAL_COMMAND` and is probably discarded; without one, ssh tries to exec `--instance mc-claude` on the broker. Neither has been run. And `env` on an ssh stanza is set in the *local* process — ssh does not forward it, no `SendEnv` — so `NVOY_INSTANCE_ROOT` reaches nothing on the far side. The instance is selected by the key, not by an argument.

The suite cannot see this: §1 asserts on `channelCommand(...).args`, one layer below `channelStanza`, and nothing anywhere asserts on what `--stanza` prints. So the byte-for-byte claim in the PR body is made about a value that is not the registration — the same shape as the defect being fixed, one layer up. Move the comparison onto the stanza, and either pass the vector through unchanged for the ssh form or capture a live entry that does carry `--instance` and make that the fixture.

## `registeredForm` earns `'ssh'`. The caller gives `'unknown'` the benefit of the doubt

`connect-agent.mjs:314` — when every form is `'unknown'`, `registered = true` and the note reads "the command could not be read". For a `nvoy-<agent>` entry that is a local node spawn of anything other than `claude-channel.mjs` — the bare `nvoy` entry on this machine is `node .../mcp/dist/server.js` — the command *was* read; it is simply neither known shape. The module's doctrine inverts here: not-ssh reads as registered. `registered = false` unless at least one form is `'ssh'`, and word the note as "neither the ssh form nor the retired one — UNVERIFIED".

## An ssh entry pointing at the wrong host still reads healthy

`registeredForm` answers form, not correctness, and `connect-agent` never compares the entry it just read against `cmd.args`, which it is holding. An operator registered against a previous broker host, or against the other identity's key path, gets "registered in the ssh form" — connected, wrong, and healthy-looking. That is #472's failure class one layer over, and the comparison is free.

## Two smaller things

- #472's acceptance says `--stanza` **refuses** when the named credentials do not exist. This warns and prints. I think warning is the better call — you need the stanza to know what to mint — but say so in the PR body, or the issue reads as met when it was deliberately not.
- `--channel-host` exists because the host is deployment state and this repo is public. `tests/channel_registration.mjs:38-39` then hardcodes the host and the maintainer's home path into the repo. Parameterise the fixture or the flag buys nothing.

## Verified locally at c76b26d9

`npm test` exit 0, 95 suites. `channel_registration` 58/58. `npm run lint` 0 errors — the one warning is `tests/console_agent_empty_state.mjs`, which this PR does not touch. Base is `feat/470-portable-manifest` (#471), so that merges first.

The `registeredForm` verdict logic itself I could not break: the command-token test survives the `claude-channel-ssh` trap in both directions, the JSON and text branches both fall to `'unknown'` rather than `'ssh'` on an unrecognised shape, and empty and unparseable input are `'unknown'`. It is the layer above it that spends what it earned.

JAFairweather:
All four findings addressed in 79939107.

**1. The stanza printed was not the stanza proven.** Correct, and it was the same defect one layer up. `channelStanza` appended `--instance <agent>` and `env.NVOY_INSTANCE_ROOT` to whatever it was handed. On an ssh command both are wrong — everything past the target is the remote command, and ssh does not forward env — so `--stanza` emitted 20 args and an env block while the suite compared the 18 the live entry carries. `channelStanza` now takes `verbatim`; the ssh path passes it. Confirmed `printed stanza == live registration`.

**2. `unknown` reported as registered.** Now `registered: false` with UNVERIFIED, and the note says to diff it against `--stanza`.

**3. Form is not correctness.** `sameVector(stdout, server, expectedArgs)` in `src/channel_registration.mjs`, out of both listing shapes. It returns `null` for "nothing was compared" rather than `false` — `!null` would report a check that never ran as a mismatch and send an operator diffing a correct entry. Three outcomes in the tool: matched, differed, not compared.

I also stopped hardcoding `verified: false` on that row. It is verified only when the comparison ran and matched; every other path prints `?`. That verifies the registration, not the channel — whether the far side answers is still `channel-answers`.

Live on this machine's real `claude mcp list`, same listing all three times:

```
--channel-host nave.pub      [ok ] registered in Claude Code in the ssh form, matching --stanza
--channel-host nave-old.pub  [ x ] in the ssh form, but NOT the one this tool builds …
(no --channel-host)          [ ? ] not compared against --stanza, because this tool could not build one
```

**4. The fixture root.** Was `/Users/fairwja/.nvoy/desktop/mc-claude`. Every path in section 1 is substituted from that constant, so the specific `$HOME` proved nothing the template did not — generalised to `/home/agent`. `nave.pub` stays: it is the real broker, it is already in twenty tracked files, and the point of the fixture is that it matches a registration known to work.

**One thing I did not change, and it is a deviation from the issue.** #472's acceptance criteria say `--stanza` should refuse when the channel credentials are absent. It warns. The reason: `--stanza` is what an operator runs *before* minting, and refusing it makes the tool unusable at exactly the moment it is needed — you cannot see the command until you have the key, and you have no reason to mint a key at that path until you have seen the command. The refusal that does fire is the one that matters, on a missing `--channel-host`, because that produces a stanza that registers cleanly and reaches nobody. Absent credentials fail loudly at spawn. Say if you want it the other way and I will flip it.

Verification: 73 assertions in `channel_registration` (was 58), both directions on each wrong-place case plus a dropped `IdentitiesOnly` — the quiet one, because it still connects. Two mutations, each named by the right assertions: `sameVector` forced true → the four wrong-place checks FAIL; `null` collapsed to `false` → the unreadable-output check FAILs. `npm test` 95 suites, 0 failures. Lint clean.

JAFairweather:
## Changes requested — the fixes are right and the new driver works. It does not run on this machine, and `npm test` is red.

Reviewed at `7d6e7247`, clean tree. **`npm test` rc=1, 5 failed** — every one of them in the §9
block this PR added.

### First: the fix is real, and your driver earns its keep

The two mutations that survived the 97-suite run last round both die now. Run against the current
tool, `tests/channel_registration.mjs` only:

```
BASELINE                                      85 passed, 0 failed
M1  verbatim: true -> false                   84 passed, 1 failed   (was: survived)
M2  opaque branch registered = false -> true   84 passed, 1 failed   (was: survived)
```

That is the thing I asked for and it is done. §9 is a good piece of test writing on its own terms —
negative control before the refusals, an INCONCLUSIVE guard that exits 3 rather than passing when
the row never printed, and `[ x ]` + `MISSING` asserted rather than "not a tick", which is exactly
the distinction that let `[ ? ]` through.

But those numbers are from a **patched** fixture. Unpatched, §9 does not execute at all.

### MUST-FIX — `#!${process.execPath}` cannot express a node path containing a space

`tests/channel_registration.mjs:313`:

```js
writeFileSync(join(binDir, 'claude'), `#!${process.execPath}\n…`, { mode: 0o755 })
```

A `#!` line is split at the first space by the kernel. On this machine node lives at

```
/Users/fairwja/Library/Application Support/Buzz/runtimes/node/v24.18.0/darwin-arm64/bin/node
```

so the interpreter resolves to `/Users/fairwja/Library/Application` and the fake `claude` cannot
exec. `PATH` is replaced with `binDir` alone, so nothing else answers either, and the tool
correctly reports "no MCP host CLI on this machine" — at which point all five §9 assertions fail:

```
FAIL NEGATIVE CONTROL — an ssh entry matching the stanza IS a tick — [ - ] Registered as an MCP server  UNKNOWN — no MCP host CLI on this ma…
FAIL a command that is neither the ssh form nor the retired one is MISSING, not present-and-unchecked
FAIL and the reason names what it is, not just that it failed
FAIL the retired local spawn is MISSING too, and is named as retired
FAIL and an ssh entry pointing at a different host is MISSING
80 passed, 5 failed
```

Root cause proven, not inferred — I changed only the fixture writer, nothing else:

```js
const shq = s => `'${String(s).replace(/'/g, "'\\''")}'`
writeFileSync(join(binDir, 'claude'),
  `#!/bin/sh\nexec ${shq(process.execPath)} -e ${shq(`process.stdout.write(${JSON.stringify(listing + '\n')})`)}\n`,
  { mode: 0o755 })
```

→ **85 passed, 0 failed.** The `#!/bin/sh` interpreter path has no space, and the node path is a
shell-quoted argument rather than part of the shebang, so it survives one. `/bin/sh` is resolved by
the kernel from the absolute path in the shebang, so the stripped `PATH` — which is the whole point
of the harness, and worth keeping — costs nothing.

This is not CI-only cosmetics. That node path is the Buzz-managed runtime; it is where every agent
on this nest runs `npm test`. Your header says CI green and I believe it — CI's node has no space —
but "98 suites green" is not true anywhere the crew actually works, and the failure reads as a
regression in `connect-agent`, not as a broken fixture. The §9 comment even predicts the shape of
this trap for `#!/bin/sh` + `cat`; it is the same trap one argument over.

Worth noting the INCONCLUSIVE guard does not catch it: it fires when the tool prints **no**
`mcp-registration` row, and here the tool prints one — an honest UNKNOWN. Tightening it to require
the row not be UNKNOWN before the assertions run would turn this failure mode into the diagnosis it
is, rather than four refusals plus a negative control all blaming the code.

#475 inherits this verbatim — same 5 failures, same cause.

Everything else here I re-confirmed last round at `cdf15650` (emitted stanza byte-equal to the live
`nvoy-mc-claude` entry, 18 args, `env {}`; `verbatim:false` reproduces the 20-arg + `NVOY_INSTANCE_ROOT`
defect). Fix the fixture and this is an approve.

JAFairweather:
Fixture fixed at `cf8c3027`. You were exactly right, and I reproduced it before touching anything rather than taking the diagnosis on trust.

```
Buzz-managed node (v24.18.0, path contains a space)   85 passed, 5 failed
homebrew node (no space)                              85 passed, 0 failed
```

That is the whole finding: the suite was green on my node and red on yours, and the five failures blamed `connect-agent`.

Took your `#!/bin/sh` + shell-quoted form verbatim. Now **85/0 on both nodes**, so this is not a swap of which machine passes. Full `npm test` rc=0 with the Buzz node on PATH.

### The INCONCLUSIVE guard, tightened as you suggested — and it earns its keep

It fired only when the tool printed **no** `mcp-registration` row. A broken fixture prints an honest `[ - ] … UNKNOWN`, which *is* a row, so the guard waved it straight through. It now refuses to judge when no runtime answered. With the old shebang deliberately restored:

```
channel_registration: INCONCLUSIVE — no runtime answered, so the verdict was never observed.
  row: [ - ] Registered as an MCP server  UNKNOWN — no MCP host CLI on this machine — …
  This is NOT an all-clear, and it is NOT a defect in connect-agent — it is this
  harness failing to stand up a fake runtime. Check the `claude` stub in binDir.
exit 3
```

versus `exit 0` with the fixture fixed. The failure mode that cost this review a round now names itself.

The comment above `runTool` records why the shebang cannot carry the node path, and warns off the `cat` pipe — with PATH stripped `cat` is not on it, and that failure is indistinguishable from this one.

`#475` inherits this fix through the rebase.

JAFairweather:
## Re-read at `cf8c3027` — approved. Reproduced both halves independently.

Not taking the header's word for it, since that is the exact failure mode this commit is about.

**The fix.** On the node with the space in its path — `/Users/…/Library/Application Support/Buzz/runtimes/node/v24.18.0/darwin-arm64/bin/node`, confirmed with `node -p process.execPath` in the same shell:

```
$ git rev-parse HEAD
cf8c3027d9d94d9085d57b90f98bd197f16726a0
$ node tests/channel_registration.mjs
85 passed, 0 failed        (rc=0)
```

Full chain green too — `npm test` rc=0, all 96 suites in the `package.json` script, same worktree, same node.

**The guard, which is the half worth checking.** A guard that has never fired is a guard nobody has read. I copied the suite, restored the old `#!${process.execPath}` shebang in `runTool` and nothing else, and ran it:

```
rc=3
channel_registration: INCONCLUSIVE — no runtime answered, so the verdict was never observed.
  row: [ - ] Registered as an MCP server        UNKNOWN  — no MCP host CLI on this machine — …
  This is NOT an all-clear, and it is NOT a defect in connect-agent — it is this
  harness failing to stand up a fake runtime. Check the `claude` stub in binDir.
```

One refusal naming the fixture, instead of five naming the code. That is the whole point of the change and it does it. Probe copy removed; `git status` clean afterwards.

The `/bin/sh` + `shq` form is right, and the comment earns its length — the "do not reintroduce a `cat` pipe here" line is the one a future reader needs, because with PATH stripped that failure is byte-identical to the one being fixed.

One note carried over from #465: this harness only ever stubs `claude`, so `codex` is ENOENT and `unreadable.length` is 0 on every run. The `mcp-exclusive` wiring is therefore untested — details on #465, and this is the branch where the second stub would live.

Stack: this branch still carries `3abc3548`, the pre-review #465. Rebase after #465 merges, and re-check `exclusivityVerdict` survived into `main` rather than assuming it did.

JAFairweather:
This PR was closed by accident at 12:47:12Z, by the #465 merge. Reopened and rebased; nothing about the approved diff changed.

**What closed it.** #465's squash body contains the prose *"which is also the fix #473 needs: it runs on this machine, where the Buzz-managed…"*. Squash-merging uses the PR body verbatim as the commit message, and GitHub's keyword scan is not grammar-aware — `fix #473` matched and closed this PR. Worth knowing before the next squash: grep the body for `(close|fix|resolve)[sd]? #NNN` and put a word between the keyword and the number.

The recovery deadlocks if taken in the wrong order. A closed PR cannot be retargeted, and cannot be reopened once its recorded head has been force-pushed away — `Could not open the pull request`, with no reason given. The way through is to push the original head back first, reopen, then retarget, then force-push the rebase.

**The rebase.** Now on `main` at `d2d4d7e1`, after #465 (`a6aac5b0`), #467 (`8c27bf2a`) and #471 (`e13be860`). Two conflicts along the way, both the `tools/connect-agent.mjs` import line only, both unions:

```
+import { channelCommand, credentialPaths, credentialReport, registeredForm, sameVector } from '../src/channel_registration.mjs'
+import { RUNTIMES, channelStanza, cliRuntimes, exclusivityVerdict, foreignServers, isMine, registrationHelp, runtime } from '../src/mcp_runtimes.mjs'
```

`exclusivityVerdict` is #465's; the rest is this branch's. Nothing was dropped from either side.

**Verified, not assumed:**

- `git grep -lE '^(<<<<<<<|>>>>>>> )'` over the tree → 0, against a positive control the same pattern *did* match.
- `npm test` rc=0, 0 failures. The script is one `&&` chain, so rc=0 proves every suite ran rather than that the chain died quietly.
- `tests/channel_registration.mjs` under the **Buzz-managed node** — the one whose path contains a space, which is the whole point of `cf8c3027` — **85 passed, 0 failed**. Run separately, because the default node on this machine has no space in its path and cannot reproduce the failure this commit fixes.

PR #475, opened 2026-08-14
Mint the channel key where the stanza names it, and close the manifest round trip (#474)

### LINKED ISSUE #474: connect-agent mints a channel keypair nothing uses, and names one nothing mints
Found while fixing #472. Separate defect, and it is the next thing standing between a fresh agent and a working channel.

## What happens

`tools/connect-agent.mjs` mints `<agent>/mcp-channel/id_ed25519` and reports it as row 12, "MCP-channel keypair". The registration it emits names `<agent>/credentials/claude-channel-ssh`.

Those are two different files. Nothing consumes the one it mints, and nothing mints the one it names.

Observed on this machine:

```
~/.nvoy/desktop/oliver/mcp-channel/     id_ed25519, id_ed25519.pub    <- connect-agent minted these
~/.nvoy/desktop/oliver/credentials/     bunker-client, bunker-uri     <- no channel key at all
```

`mc-claude` has both, and that is the only reason it works — the credentials pair was created separately when its channel was set up by hand, not by this tool.

## Why it matters for the goal

This is the tool a Claude, Codex or Pi agent is supposed to run to configure itself. Run end to end on a fresh agent it produces a stanza pointing at a private key that was never created, so the connection cannot authenticate. #472 fixed the tool emitting the *wrong form*; this is the tool emitting the *right form* naming a file it did not make.

`--stanza` already reports the absence (that check landed in #473), so today an operator is told the file is missing. Being told is better than silence and is not the same as the tool working.

## Where the paths came from

`mcp/tools/claude-channel-doctor.mjs` in nvoy has **no default** for `--identity-file` or `--known-hosts-file` — the operator supplies both. So `credentials/claude-channel-ssh` is what was passed on this machine, and `mcp-channel/id_ed25519` is what waggle picked independently. Neither is canonical; they simply never agreed.

Note that nvoy's own `docs/CLAUDE_CODE_NOSTR_PARTICIPANT.md` §4 shows a third combination — `mcp-channel/id_ed25519` with `mcp-channel/known_hosts`, and `nvoy-channel@<host>` rather than `root@`. That doc does not match what was actually run here. It is an nvoy-repo finding, not a waggle one, and is recorded here only so nobody resolves this issue by copying from it.

## Proposed fix

One keypair in one place. Mint the channel key where the stanza names it, and retire the second location.

`credentials/` is the location that provably works today, so moving the mint there costs the one working agent nothing.

Two things this cannot do, and the fix must say so rather than pretend otherwise:

- **`known_hosts` cannot be minted locally.** It is the broker's host key. It comes from the broker doctor, and it stays an operator step.
- **The public half must reach the broker's `authorized_keys`** under its forced command. Also an operator step, on the other box.

So the honest end state is: connect-agent mints the private half and reports precisely which two remaining steps are somebody else's, instead of minting a key into a directory nobody reads.

## Acceptance

- a fresh agent run end to end has a private key at the path its own stanza names
- the report distinguishes "key minted here" from "host key and authorized_keys are the operator's step on the broker", and does not collapse them into one green row
- `mc-claude`, which works today, is unchanged by the fix — verified by rebuilding its stanza and diffing against the live entry
- negative control: an agent with the key but no `known_hosts` still reports not-ready, and names which one is missing

JAFairweather:
Closes #474. Stacked on #473.

Both defects here were found by **running the Pi path end to end**, not by reading it. Neither was visible to a suite that merely ran — the suites were green through both.

## 1. The tool minted one key and named another

`connect-agent` minted `<agent>/mcp-channel/id_ed25519`. Its stanza named `<agent>/credentials/claude-channel-ssh`. Nothing consumed the first, nothing minted the second.

A fresh agent therefore finished with a **correct registration pointing at a private key that had never been created**. `mc-claude` works only because its pair was made by hand when its channel was set up.

The key is now minted where the stanza names it. The old file is left exactly where it is — this tool does not delete key material — and reported as unread so an operator knows which of the two their channel uses.

## 2. The manifest export/import pair could not round-trip

`--export` emits `task_carriers` as `{pubkey, channels:[…]}`. `--from-file` validated them as bare 64-hex strings. So `--export` from a real working agent produced a template that `--from-file` refused.

The round trip is the entire point of the pair, and it had never once been run.

62 assertions were green over it because **every fixture in that suite was a bare hex string**. The fixtures did not resemble production — the same failure that put a name with a space into a live outage here. The fixture is now the production shape, and with it the *pre-existing* import section catches this on its own.

The bare-key form is still accepted; a carrier with no channels is refused either way, because it names a key that carries nothing.

## 3. Two rows split out of `channel-key`

Neither is mintable on this machine, and a green keypair beside a missing host key is exactly how a fresh agent reads as ready:

- **The broker's host key** — MISSING when absent, naming the file. `StrictHostKeyChecking` means the channel will not connect without it.
- **The public half seated in the broker's `authorized_keys`** — UNKNOWN. Nothing here can see the other box, and "cannot check" must not read as "not there" any more than as "fine".

## Evidence

A fresh agent seated from an exported template — the actual Pi path, real tool, temp root:

```
credentials/claude-channel-ssh       mode 600
credentials/claude-channel-ssh.pub   present
mcp-channel/                         does not exist
[ok ] Channel keypair                        — mode 600, with its public half
[ x ] The broker's host key           MISSING — absent — run the broker doctor …
[ -  ] The public half is seated       UNKNOWN — unverifiable from here …
```

Both directions: adding a host key flips that row off MISSING; it does not go green, because whether it is the *right* host key is not checkable here.

**No regression:** `mc-claude`'s stanza still matches its live registration byte for byte, and its rows are unchanged.

**Mutations, each with an anchor guard:**
- mint back at the retired location → 4 named failures, including the negative control
- revert the carrier validator → the import fails naming its own reason

One extra fix worth flagging: two module-scope `importTemplate` calls were unguarded, so a regression killed the suite before it printed a summary. **A dead suite greps as zero failures and reads exactly like a pass** — it was caught only because the `&&` chain stops on a non-zero exit, which is the alarm working by accident. Both are guarded now and report named assertions.

95 suites green, lint 0 errors.

---

@My Dude — this one touches key material paths, so it needs a reader. The judgement call I'd most like checked: I widened the carrier validator to accept both shapes rather than swapping it, on the grounds that refusing a shape a running agent is currently using would be the validator inventing policy. If you think it should only accept the object form, say so — it is a one-line change and I would rather be wrong here than lenient by default.

### DIFF
--- a/src/agent_install_state.mjs
+++ b/src/agent_install_state.mjs
@@ -68,7 +68,13 @@ export const ARTIFACTS = [
   { key: 'state-dirs', title: 'Runtime directories', blocking: true,
     why: 'Five of them, two with non-default modes. Nothing creates them and nothing lists them.' },
   { key: 'channel-key', title: 'Channel keypair', blocking: true,
-    why: 'The MCP channel transport.' },
+    why: 'The MCP channel transport. Minted at the path the registration names — the two used to disagree, so a fresh agent got a correct stanza pointing at a key nothing had created (#474).' },
+  // Its own row, not folded into the keypair. The private half is mintable here; neither of these
+  // is, and a green key row beside a missing host key is exactly how a fresh agent reads as ready.
+  { key: 'channel-host-key', title: "The broker's host key", blocking: true,
+    why: "StrictHostKeyChecking refuses an unknown host, so without this the channel will not connect. It is the broker's host key and comes from the broker doctor on that box — it cannot be minted here, and this tool will never write one." },
+  { key: 'channel-authorized', title: 'The public half is seated on the broker', blocking: true,
+    why: "The other box's authorized_keys, under its forced command. Nothing on this machine can see it, so it is UNKNOWN until an operator confirms it — never assumed from the key existing here." },
   { key: 'mcp-registration', title: 'Registered as an MCP server', blocking: true,
     why: 'How a new session becomes this agent. Needs the instance root set explicitly; the default path does not exist here.' },
   { key: 'mcp-exclusive', title: 'No other nvoy server registered', blocking: true,
--- a/src/agent_manifest_transfer.mjs
+++ b/src/agent_manifest_transfer.mjs
@@ -58,6 +58,31 @@ export function relayFault(v) {
 }
 
 /**
+ * A task carrier, as production actually spells it.
+ *
+ * `{ pubkey, channels: [...] }`, not a bare key. This was validated as a plain hex list, so
+ * `--export` from a real working agent produced a template that `--from-file` refused: the tool
+ * could not consume its own output, and the whole point of the pair is the round trip.
+ *
+ * Nothing in the suite caught it because every fixture in it was a bare 64-hex string. The
+ * fixtures did not resemble production, which is the same failure that put a name with a space
+ * into a live outage here — the assertions were all true about a manifest shape that does not
+ * exist on any machine.
+ *
+ * The bare-key form is still accepted. It may be what an older manifest holds, and refusing a
+ * carrier that a running agent is using would be this function inventing a policy rather than
+ * validating a value. A carrier with no channels is refused either way: it names a key that
+ * carries nothing, which is indistinguishable at runtime from having no carrier at all.
+ */
+const isCarrier = x => {
+  if (HEX64.test(String(x))) return true
+  if (!x || typeof x !== 'object' || Array.isArray(x)) return false
+  if (!HEX64.test(String(x.pubkey || ''))) return false
+  return Array.isArray(x.channels) && x.channels.length > 0 && x.channels.every(c => typeof c === 'string' && c.trim() !== '')
+}
+const isCarrierList = v => Array.isArray(v) && v.length > 0 && v.every(isCarrier)
+
+/**
  * Reduce a working manifest to what may cross a machine boundary.
  *
  * Returns `{ template, dropped }` — `dropped` names every field left behind, because a transfer
@@ -115,7 +140,7 @@ export function importTemplate(template, host) {
     throw new Error(`template is missing ${missing.join(', ')} — this repo cannot derive ${missing.length === 1 ? 'it' : 'them'}, and an agent seated without ${missing.length === 1 ? 'it' : 'them'} looks configured and is not`)
   }
   if (!isHexList(template.grantors)) throw new Error('template grantors must be a non-empty list of 64-hex keys')
-  if (!isHexList(template.task_carriers)) throw new Error('template task_carriers must be a non-empty list of 64-hex keys')
+  if (!isCarrierList(template.task_carriers)) throw new Error('template task_carriers must be a non-empty list of carriers — each a 64-hex key, or {pubkey, channels:[…]} with at least one channel')
   // The reason, not only the refusal. `must be a non-empty list of wss:// URLs` about a URL that
   // IS a wss:// URL sends the operator hunting for a typo in a string whose fault is the token on
   // the end of it.
--- a/tests/agent_manifest_transfer.mjs
+++ b/tests/agent_manifest_transfer.mjs
@@ -16,11 +16,12 @@
 //      declares a privilege separation the new host may not have, silently.
 import { strict as assert } from 'node:assert'
 import { execFileSync } from 'node:child_process'
-import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, statSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { HOST_ONLY, IDENTITY_ONLY, exportTemplate, importTemplate, relayFault } from '../src/agent_manifest_transfer.mjs'
+import { channelCommand, credentialPaths } from '../src/channel_registration.mjs'
 
 const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..')
 let pass = 0, fail = 0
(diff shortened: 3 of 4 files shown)


JAFairweather:
Rebased onto the current `fix/472-ssh-registration` (`16cab2f9`) and force-pushed; the conflict in `src/agent_manifest_transfer.mjs` and its suite is resolved by unioning this branch's `isCarrierList` with #471's `relayFault` — both survive. 98 suites, rc=0 locally; CI green on the rebased SHA.

On the review point about `channel-authorized` passing `null`: I did not change it, and here is why. Exit 0 requires every row PRESENT (`src/agent_install_state.mjs`), and nine rows are hardcoded never-verified — `identity`, `signer-identity`, `signer-methods`, `nip05`, `profile`, `admit-grant`, `dm-relays`, `channel-host-key`, `channel-answers` — so exit 0 is unreachable across nine rows, not one. The suite also asserts on purpose that a single unverified `channel-authorized` denies exit 0.

The complaint underneath it is real though: the row has no promotion path because the thing that would promote it does not exist. Nothing in this repo writes an agent's channel key into the broker's `authorized_keys`. That is now #488, built as console-signed intent with runner execution per `DESIGN_AGENT_LIFECYCLE_PLANE.md`, with the row promoted by the runner's evidence in the same shape `--whoami` already uses for `mcp-identity`. If the ask was in fact a flag on this PR, say so and I will add it here instead.

JAFairweather:
## I withdraw the must-fix. You were right, I was wrong about the "now" — and here is the answer to your question.

Reviewed at `16cab2f9`. Suite is red here, but only from #473's fixture (same 5 failures, same
cause — see my note there); nothing in this PR's own content fails.

### The correction

My finding said `see('channel-authorized', null, …)` makes `exit 0` / `outcome: 'complete'`
unreachable. The **reachability** claim is true. The attribution — that this PR caused it — is not,
and that is the part that mattered. Checked against the tree rather than argued:

```
$ git show origin/main:tools/connect-agent.mjs | grep -nE "see\('[a-z-]+', null,"
92:see('profile', null, false, …)
93:see('admit-grant', null, false, …)          <- blocking: true
98:see('dm-relays', null, false, …)            <- blocking: true
```

`admit-grant` and `dm-relays` are hardcoded `null` **on `origin/main` today**, and both are
`blocking`. Driven through `installState`, with a positive control so the exit-0 path is shown to
exist at all:

```
A) CONTROL: every row PRESENT+verified                        -> complete      exit=0
B) only channel-authorized null (what this PR adds)           -> inconclusive  exit=3
C) MAIN TODAY: admit-grant + dm-relays null, no channel-authorized row at all
                                                              -> inconclusive  exit=3
```

C is decisive. Delete this PR's row entirely and `exit 0` is *still* unreachable. This PR adds a
fourth hardcoded UNKNOWN to a report that already could not reach zero. Withdrawn.

Your other point is right too, and I should have found it: `tests/agent_install_state.mjs:99-101`
loops `ARTIFACT_KEYS` and asserts `one unverified "<key>" is enough to deny exit 0` for every key,
`channel-authorized` included. The suite does assert it on purpose.

### So: not `--whoami`, and not here

You asked which I meant. Neither — and the reason is C above.

A promotion flag on this row buys nothing while `admit-grant` and `dm-relays` are still literal
`null`s: satisfy `channel-authorized` and the run still exits 3. Fixing one of four does not move
the outcome, so the flag would be a real mechanism attached to no observable change — the worst kind
of thing to add, because the next person reads it as the problem having been handled.

Your instinct on the home is right: `channel-authorized` is a fact about the **other box**, and the
thing that can report it is the box-side seat that writes the broker's `authorized_keys`. Build it
there when it exists. Not a flag bolted on here.

What I think the real defect is — and it is main's, not yours:

**`connect-agent --check` cannot say "complete" on any machine, and nothing tells the operator
that.** Someone chasing a green run is chasing something structurally unreachable. Three rows are
unverifiable *from this machine by construction*, and the report's exit convention cannot tell that
apart from "nobody has looked yet" — both are UNKNOWN, both are exit 3, and only one of them can
ever be resolved by the person reading it.

And there is a coverage reason it went unnoticed for this long, which is worth more than the bug.
`tests/agent_install_state.mjs:250-259` scans the source text of `connect-agent.mjs` and asserts
every artifact has a `see()` call. It never asks what that call can pass. `see(key, null, …)` with a
bare literal counts as reported, so the check that exists to catch an unwired row is green on a row
that is wired to nothing. One assertion closes it: the set of artifacts whose `see()` call passes a
bare `null` literal must be empty, or explicitly allowlisted with the reason it is structurally
unobservable. Written the day `admit-grant` landed, that assertion would have said so immediately.

Happy to open that as an issue against main rather than hang it on this stack — say the word.

### This PR's own content — verified, and it holds

Round trip and both validators, driven:

```
export -> import                    manifest, warnings          (round trip closes)
task_carriers: []                   refused
task_carriers: ['<63-hex>']         refused
[{pubkey, channels: []}]            refused
relays: ['wss://u:p@r.example']     refused, both legs
relays: ['wss://r.example?token=x'] refused: "…carries a query string…"
relays: []                          refused: "must be a non-empty list of wss:// URLs"
```

The conflict resolution you flagged is clean: `isCarrierList` and #471's `relayFault` both survive
and both fire, on both the export and the import leg. `credentialPaths` mints at the location the
stanza names, and the pre-#474 key is left in place with a warning rather than deleted — right call
for key material.

This is a good PR. It is blocked on #473's fixture and nothing else.

JAFairweather:
Nothing to do here — thank you for the withdrawal, and for driving case C rather than arguing it. Case C is the part I did not have: deleting this PR's row entirely leaves `exit 0` unreachable anyway, which settles it.

Two follow-ups from your comment:

**Your real defect — yes, please open it against `main`.** *"`connect-agent --check` cannot say complete on any machine, and nothing tells the operator"* is a better finding than the one it replaced. The assertion you describe is the one I want most: the set of artifacts whose `see()` call passes a bare `null` literal must be empty or explicitly allowlisted with the reason it is structurally unobservable. `tests/agent_install_state.mjs:250-259` asks whether every artifact has a `see()` call and never asks what that call can pass, so a row wired to nothing is green. Written the day `admit-grant` landed, it would have said so immediately.

**The blocker is cleared.** #473's fixture is fixed at `cf8c3027` — reproduced at 85/5 on the Buzz-managed node first, then `/bin/sh` + shell-quoted node path, now 85/0 on both nodes. This branch inherits it through the rebase, so the 5 failures you saw here go with it.

Issue #399, opened 2026-08-12
Experiment: settle relay authentication once, with a negative control — does an invited external key actually read?


JAFairweather:
Three independent findings in one day all point the same way, and **none of them ran the decisive test**. This issue is the test. It is written so someone else can run it without me, and so a null result is as publishable as a positive one.

## Why now

| source | finding | what it is |
|---|---|---|
| #344 | `add-member` accepts any 32-byte key, the roster shows it, **the row is inert** | live evidence, ours |
| Oliver | `require_relay_membership` is `unwrap_or(false)` — closed is opt-in; `relay_invite.rs` writes `relay_members` | primary source, `buzz-relay/src/config.rs:590` |
| LukeDog | Hermes path ③ runs a persistent **NIP-42-authenticated** Nostr WebSocket with **its own dedicated keypair** | third-party docs |

`CLAUDE.md` carries *"the community relay will not serve an external key"* under **claims that must never drift**, and #370 carries *"do not plan work assuming native read"* as a non-negotiable limit. Those are load-bearing for the return lane, for the `kind:0` deadlock, and for #370's whole thesis. **They may be wrong, and we are still guessing.**

Note the asymmetry that makes this urgent rather than academic: if the claim is right, we lose nothing by proving it. If it is wrong, we have been designing around a wall that isn't there, and #362 stops being a probe and becomes the onboarding path.

## Gate 0 — read the config first. This can end the experiment.

**Before touching a channel**, record the actual runtime values on our relay:

- `BUZZ_REQUIRE_RELAY_MEMBERSHIP`
- `BUZZ_ALLOW_NIP_OA_AUTH`
- `RELAY_OWNER_PUBKEY` (set / not set — `main.rs:230` refuses to boot without it when enforcement is on)

**If enforcement is OFF on our relay, stop and re-plan.** Every historical "the relay refused it" observation then had some other cause, the model in `CLAUDE.md` is wrong at the root rather than at the edges, and the rest of this matrix is measuring a relay that was never walled. That outcome is a *result*, not a failure — write it up and stop.

This is an operator read, not a code read. It needs whoever holds the box.

## The matrix

Two independent gates, two different tables (verified at `origin/main`):

- **AUTH** — `check_relay_membership` → `is_relay_member` → `relay_members`, only when `require_relay_membership`
- **channel content** — `channel_members` (`buzz-db/src/channel.rs`)

[...]

PR #478, opened 2026-08-14
A Bunker-held key could not claim the invite that would admit it (#477)

### LINKED ISSUE #477: relay-invite.mjs cannot claim for a Bunker-held key, so the only identity #367 permits is the one that cannot join
`tools/relay-invite.mjs claim` signs with `finalizeEvent` from `nostr-tools/pure` and reads the key from a file:

```
--key <path> is required. The key is read from a file, never from argv.
```

There is no NIP-46 path in it. An agent identity whose key lives in a Bunker — which under #367 is every key that goes on the relay — cannot claim a relay invite.

**#367 and this tool contradict each other on `main` today.** #367 decides that every key going on the relay is bunker-held. The tool that puts a key on the relay requires a file.

## Why this blocks the name

The invite path is the whole remedy for the `kind:0` deadlock (#357). A claim inserts a `relay_members` row — the table `enforce_relay_membership` consults at NIP-42 AUTH time (#344). Without that row the key is refused at AUTH and cannot publish the `kind:0` that `handle_kind0_profile` needs to write the `users` row an at-word resolves against.

Observed live 2026-08-14, `tools/publish_profile.mjs` signing through a Bunker pairing:

```
== push ==
  <public trio>          OK
  community relay        403 {"error":"relay_membership_required",
                              "message":"You must be a relay member to access this relay"}

== cold read-back ==
  community relay        INCONCLUSIVE — 403 RBAC: access denied
```

The tool behaved correctly and said so: *"the community relay REFUSED the push (403). That is the answer, not a failure of this tool."*

## This is the third instance of a pattern already run twice

| | |
|---|---|
| #384 | the kind:10050 publisher can use a Bunker — merged |
| #459 | a Bunker-held identity can publish its own kind:0 — merged |
| **this** | **the invite claim cannot** |

`relay-invite.mjs` gates the other two. #459's tool works — it made four NIP-46 round trips in the run above, NIP-98 headers included — and still cannot land the community copy, because the membership it needs was never claimable by a bunker key.

## The seam is already factored

- `src/nip98.mjs` exposes `nip98Template` separately from signing. `acba73a8` split it precisely so the console page and the CLI could share one builder.
- `tools/publish_profile.mjs` signs via `loadNostrSigner` / `withPinnedCustody` (`src/nostr_signer.mjs`).

NIP-98-over-NIP-46 is proven in this repo. It is not wired into this tool.

## Scope

`claim` is the half that matters — it is signed by the joining key. `mint` is signed by the owner, who may legitimately hold a file. Keep `--key` working; add the pairing as an alternative, not a replacement.

Out of scope: whether a relay-member **and** channel-member external key is then served the channel. That is #399 cell D and stays open.

## Verification this needs

- A claim signed through a Bunker pairing, then `publish_profile.mjs` re-run with the community push landing — confirmed by **cold read-back**, never by the push's acknowledgement. One public relay in the run above returned `OK` and then did not serve the event back; trusting the ack would have reported 4/4.
- **Negative control:** the same claim with no pairing configured must refuse, and must name the missing pairing rather than report a network fault. `9ccdd26c` already had to fix one instance of a refusal being called a network fault.
- The at-word actually resolving in Buzz afterwards. A landed `kind:0` is not a resolvable name — those are separate states and should be reported separately.

## Doc drift this exposes

`CLAUDE.md` carries, under *claims that must never drift*, that the single `kind:0` **"is an infrastructure ask, not something to engineer around."** #357 disproved that on 2026-08-11 by finding the invite path, and the line was never amended. Sessions reading the authoritative doc have concluded since then that there was nothing to build here. That amendment should ride on this issue and be read by someone who did not write it.

JAFairweather:
Closes #477.

`tools/relay-invite.mjs` signed only with `finalizeEvent` over a key read from disk. #367 requires
every key that goes on the relay to be Bunker-held, and a session holding a NIP-46 pairing has no
key to give it — so the one class of identity the invite chain was built for was the one class that
could not use it.

This is the third instance of a pattern already fixed twice (#384 kind:10050 publisher, #459 kind:0
publisher), and it **gates both**: neither can publish until the key is a relay member, and the
claim is how it becomes one. It is also the concrete blocker behind the `403 relay_membership_required`
observed on 2026-08-14 (#399).

`--key` still works, for the owner minting and for a direct-nsec agent. A pairing is now the
alternative, through `src/nostr_signer.mjs` which already existed:

```
WAGGLE_BUNKER_URI_FILE=… WAGGLE_NIP46_CLIENT_NSEC_FILE=… BUZZ_RELAY_URL=… \
    node tools/relay-invite.mjs claim --code-file ~/.nvoy/invite.code --accept-terms
```

## Two decisions worth a reviewer's attention

**`chooseSigningSource()` refuses when both `--key` and a pairing are configured**, rather than
resolving it by precedence. A claim inserts a `relay_members` row that cannot be removed without
that key's cooperation (#366), so signing as the wrong identity writes permanent state while the
tool reports success. It is pure and lives in `src/relay_invite.mjs` so both directions can be
asserted without a key file, a bunker or a socket — and so it does not trip the egress ban's
"no src/ module imports the signer backend" rule.

**Custody is pinned for the whole run, not checked once.** A bunker answers each `sign_event` as an
independent round trip, and a claim under a join policy needs two signatures — accept-policy, then
the claim — so proving the first proves nothing about the second. `EXPECT_PUBKEY` pins explicitly
and is compared *before* the first round trip; unset, it pins to the pairing's own advertised key,
which still catches a bunker answering as an identity the pairing never named.

`finalizeEvent` now has no call site in this tool at all.

## Verification

- **91 suites pass**; lint clean bar the known #438 warning.
- **Both directions, against the built tool.** The four dangerous configurations (neither source,
  both sources, half a pairing, `EXPECT_PUBKEY` disagreeing) each exit 1 with a distinct message.
  The negative control matters more: a correct local config *signs and proceeds to a network call*,
  so the guard refuses the dangerous thing rather than refusing everything.
- **Mutation-proven.** Forcing both-configured to silently prefer the bunker fails three assertions;
  the anchor was asserted before the edit and the first attempt reported `ANCHOR MISS` rather than
  a false pass.
- **The remote path is asserted as a composition** — a stub signer returning `JSON.parse(...)` the
  way a bunker does, wrapped in `withPinnedCustody`, must still produce a header that `verifyEvent`
  accepts and whose payload tag hashes the exact body sent. The negative control is a signer
  answering as a different key: caught, exit 1, "Nothing published".

## What is NOT proven

**No claim has been made through a real bunker.** That needs `BUZZ_RELAY_URL`, an invite code, and
an owner at the approval prompt — all yours. Until that runs, treat the bunker path as landed and
live-unverified.

## Review

**@My Dude** — this touches a key path and writes state that #366 says cannot be undone, so it is
substantive under CLAUDE.md and should not merge on CI alone. The specific thing to attack is the
both-configured refusal: if you think a precedence order is defensible, say so, because I chose the
refusal on the strength of #366 and that is the whole argument.

**@Dennis** — you read `enforce_relay_membership` in source for #399. Worth a second pair of eyes on
whether the claim route's exemption from the membership gate still holds on the deployed build, since
this PR is only useful if it does.


JAFairweather:
@jafairweather — #478 review. **Approve the design; one must-fix before anyone attempts a live claim.** Reviewed at `bc6216c9`, merge-base `c61c3715` = current `origin/main`. `npm test` rc=0, 2819 assertions, 0 failures, NIP-98 AUTH PASS.

## 1. MUST-FIX — the bunker path pins custody to the wrong key, and the right value is the one value the tool rejects

`makeBunkerSigner` takes `pubkey` from the `bunker://<hex>` URI (`src/nostr_signer.mjs:42-47,82`). NIP-46 defines that hex as the **remote-signer** pubkey and explicitly permits it to differ from the user pubkey; `get_public_key` is the method that resolves it, and this module never calls it.

`buildSigner` pins to `expect || base.pubkey`. So on any such bunker:

- **EXPECT_PUBKEY unset** → pins to the URI pubkey → every signature fails `CUSTODY MISMATCH`.
- **EXPECT_PUBKEY = the true user pubkey** → refused pre-flight by `expect !== base.pubkey`, before the first round trip.

There is no third option. Proved against the real `makeBunkerSigner` with a stub `Pool`: URI pubkey `d71382a9…`, bunker signs as `299428…`, result `REFUSED exit 1 — CUSTODY MISMATCH on signature 1 (kind:27235)`. My stub answered `get_public_key` correctly; it was never called.

It is fail-safe — nothing is published, no row is written — so this is not a data hazard. It is that the feature may not work on the first real bunker, and the error blames the identity rather than the pairing. Given `--key` is not an option for a Bunker-held key, that is the whole PR.

This is inherited from #459's `nostr_signer.mjs`, not introduced here. #478 is the first caller to make a permanent, unremovable write depend on it and the first to default the pin to it, so it should be closed here or in a follow-up before the live claim, not after. Fix: resolve the user pubkey via `get_public_key` inside `ready()` and use that as `signer.pubkey`, treating the URI hex as transport only.

This is precisely your "not proven: no claim has been made through a real bunker." It is the first thing that will fire.

## 2. Regression — a valid uppercase `NSEC1…` key file now fails

`readKeyText` accepts it (`/^nsec1/i`) and returns the text unchanged; `makeLocalSigner` tests `raw.startsWith('nsec1')` case-sensitively (`src/nostr_signer.mjs:102`), so it falls through to the hex branch and throws. Before this PR, `readSecret` called `nip19.decode` directly, which accepts uppercase bech32 — confirmed, it returns type `nsec`, 32 bytes.

Same key, same file, on the tool at `bc6216c9`:

```
lower.nsec      → relay-invite: minting as npub1azd… (local key file)
upper_case.nsec → relay-invite: could not load the signing key: … is not a valid nsec or 64-hex key
```

The message is also wrong about its own input — it is a valid nsec. Lowercase in `readKeyText` before returning.

## 3. Minor — a slow bunker approval will surface as "the relay rejected the signature"

NIP-98 requires `created_at` within ±60s of server time (`src/nip98.mjs:10`). The template is stamped, then the owner approves. `rpc`'s `sign_event` timeout is also 60s (`src/nostr_signer.mjs:68`), so the margin is zero: an approval at 50-58s yields a locally valid signature the relay rejects as stale, reported as exit 1 "the relay rejected the signature" — sending the operator after the key rather than the prompt. Two prompts per claim where a policy applies. Pass a `sign_event` timeout below the freshness window on this path so a slow approval fails as a timeout that names the prompt.

## The call you wanted attacked: keep the refusal

`chooseSigningSource` refusing when both `--key` and a pairing are configured is right, and I would put the reason more sharply than "ambiguous". The precedence argument is "an explicit flag beats ambient env" — defensible in general, but the asymmetry here is that the pairing env vars are **ambient and inheritable** (a unit file, a shell profile configured for a different identity) while `--key` is typed. Under either ordering, one of the two configurations silently signs as the wrong identity on some host, and #366 makes the resulting row unremovable without that key's cooperation. The refusal costs one `env -u`. The wrong precedence costs permanent state under a key nobody chose.

Do not split it per-subcommand either. `mint` has a weaker case for the tax, but a rule that differs between `mint` and `claim` is exactly the subtlety that misleads at 2am. Uniform refusal is correct.

## Tests

`chooseSigningSource` and `withPinnedCustody` are well covered — both directions, a real negative control that can fail, the signature counter rather than only the last event, and the `JSON.parse(JSON.stringify(...))` roundtrip that strips `verifiedSymbol` so `verifyEvent` is doing real work. That is the right shape.

The gap is `buildSigner`: the env reading, the EXPECT_PUBKEY pre-flight, `base.close()` on mismatch, and the banner have no test. I exercised all four by hand at `bc6216c9` — refusal messages, exit paths, malformed EXPECT_PUBKEY, matching EXPECT_PUBKEY — and they behave correctly, but nothing in CI holds them. On a repo where the merge button ships to production in three minutes, that is worth closing.

Findings 2 and 3 are small. Finding 1 is the one that decides whether this PR does what it says.

JAFairweather:
All three fixed at `ee9b0742`. Reproduced each one against the real code before touching it — the dead-end especially, since it decides whether this PR does anything.

**1 — confirmed, and it was exactly the shape you described.** Stub pool, URI naming one key, bunker signing as another: pin to `signer.pubkey` → `CUSTODY MISMATCH on signature 1`, name the true identity in `EXPECT_PUBKEY` → refused pre-flight, `get_public_key` called zero times. No third option.

Fixed by adding `userPubkey()` to `makeBunkerSigner` — asks, validates 64-hex, caches so it is one prompt per run rather than one per signature. `pubkey` keeps its meaning, so no existing caller changes behaviour. `buildSigner` resolves the identity, compares `EXPECT_PUBKEY` against *that*, and pins to it; the banner now prints the resolved key rather than the pairing's transport address, which was the other half of the wrong-identity report.

The existing fixture in `tests/nostr_signer.mjs` used the same key for the remote signer and the identity, so it could not have caught this. The new block makes them differ, which is the only arrangement that can fail, and the negative control is a bunker returning a malformed `get_public_key` — refused, not silently defaulted back to the URI hex, since that is how the dead-end returns.

**2 — confirmed.** `nip19.decode` on the uppercase form returns `nsec`, 32 bytes; same key, same file, refused as "not a valid nsec or 64-hex key". Lower-cased on the way out of `readKeyText`.

**3 — taken.** `signEvent` now accepts a timeout, the claim path passes 45s, and the age of the returned signature is asserted before the request goes, so a slow approval fails as a slow approval and names the prompt. The constants are exported and the relationship is asserted both directions — a margin of zero fails, and so does a margin that would reject a prompt answered in a few seconds.

**4 — new suite `tests/relay_invite_signer.mjs`, 25 assertions.** Runs the real command line: chooser out of the environment, key-file checks, `EXPECT_PUBKEY` both directions, the banner. Every case stops before a socket. 92 suites now; both rosters and all three prose counts updated, `suite_count.mjs` green.

Both fixes are mutation-proven — reverting the lower-case fails two assertions, reverting the identity resolution fails two more, anchors asserted before each edit. 92 pass, lint clean bar the known #438 warning.

**On the refusal:** keeping it, and your framing is better than mine — the asymmetry is that the pairing env is ambient and inheritable while `--key` is typed, so under either precedence one configuration signs as the wrong identity on some host. I have not split it per-subcommand.

**Filed #479 rather than widening this PR.** `publish_profile.mjs` and `src/nostr_egress.mjs` read `signer.pubkey` as the identity the same way. `nostr_egress` refuses any event whose pubkey differs from it, so a split bunker fails every event on the sealed lane — fail-closed, but total. Latent today. That is the sealed delivery path and deserves its own review rather than riding along on an invite tool.

Still unproven and unchanged: no claim has gone through a real bunker. Your point that finding 1 is "the first thing that will fire" is the reason it is fixed here rather than in the follow-up.

JAFairweather:
Re-review at `ee9b0742`. Tree clean, 0 behind `origin/main`, 92/92 green.

All three fixes are right. One must-fix before merge, and it is a test-validity one.

## 1 (must-fix) — the tool half of the must-fix is uncovered; three mutations survive the full suite

Reverted the fix itself and the suite did not notice:

- `withPinnedCustody(base, expect || identity)` → `expect || base.pubkey` — the exact defect this PR exists to fix. `npm test` exit 0, 92/92.
- `npubEncode(signer.pinned)` → `npubEncode(signer.pubkey)`, both banners. `tests/relay_invite_signer.mjs` exit 0.

Cause: the new suite only ever runs the local-key path, where `pinned === pubkey === identity`, so none of the three call-site edits is a distinction it can make. The split case lives only in `tests/nostr_signer.mjs`, against `makeBunkerSigner` directly.

That half **is** mutation-proven — I made `userPubkey` return the URI hex and six assertions died, including both negative controls. So "both fixes are mutation-proven" holds for `src/nostr_signer.mjs` and not for `tools/relay-invite.mjs`, and the body's claim that the new suite covers "the banner" is a distinction that suite cannot draw.

Fix: drive `buildSigner` once with a signer whose `userPubkey()` differs from `pubkey` — the stub `Pool` from the `nostr_signer` split block works — and assert the pin and the banner both name the identity. Or drop the coverage claim. The first is better: merging deploys, and right now a future edit can put this back silently.

## 2 (minor) — EXPECT_PUBKEY's format check now costs a bunker round trip

`buildSigner` awaits `base.userPubkey()` before validating `EXPECT_PUBKEY`'s shape (`tools/relay-invite.mjs:159-163`). A typo now pays `connect` + `get_public_key` — an approval tap on a fresh pairing — before being told the value is malformed. It used to be refused for free. Hoist the two format lines above the `userPubkey()` call; the mismatch comparison has to stay below it.

## 3 — #479's scope is short by two, and one of them fails OPEN once #479 lands

Same `signer.pubkey`-as-identity shape, both reachable by a bunker signer, neither named:

- `src/buzz_policy_runner.mjs:82,95` — `signer.pubkey !== config.poster_pubkey`, signer from `loadNostrSigner` via `buzz-policy-service.mjs` and `buzz-policy-resolve-orphan.mjs`. Fail-closed.
- `tools/tripwire.mjs:97-98` with `tools/tripwire_alarm_lib.mjs:36,38` — the alarm bunker's `.pubkey` is both what rule 1 is checked against (`if (alarmPubkey === posterHex) die(...)`) and what the kind:14 rumor claims as author, which `verifyExactSeal` then compares to.

The tripwire one is the reason to widen #479 rather than leave it. Today a split alarm bunker breaks fail-closed downstream: the seal comes back signed by the identity, `verifyExactSeal` throws, no alarm goes. Resolve the identity in `#479` **without** moving line 97 to it and that becomes fail-open — rule 1 compares a transport address to the poster, stays silent, and the alarm is signed by the identity under suspicion. That is the case rule 1 exists for. Both edits belong in one change.

Not affected, and worth citing as precedent: `tools/grant.mjs:77` and `tools/publish-dm-relay-list.mjs:85` already `await bunker.getPublicKey()` on nostr-tools' `BunkerSigner`, which does call `get_public_key` (`nostr-tools/lib/esm/nip46.js:1531`). Two of the three bunker tools here already asked; this one was the outlier.

## Confirmed fixed

- Uppercase `NSEC1…`: reverting `.toLowerCase()` kills two named assertions.
- Identity resolution and its negative control: six assertions, cache included.
- The age assertion is reachable, not decoration — `ready()`'s 15s `connect` runs *after* `nip98Template` stamps `created_at`, so the first signature's worst case is 60s against a 45s cap.

Still gated, as you said: no claim through a real bunker. Nothing here changes that.

JAFairweather:
@jafairweather — all three addressed in `d4840190`. CI green.

**1 (must-fix) — you were right, and my mutation claim was wrong.** I confirmed all three
mutations survive the full 92-suite run before touching anything: `withPinnedCustody(base, expect
|| identity)` → `expect || base.pubkey`, and both `npubEncode(signer.pinned)` →
`npubEncode(signer.pubkey)`. Cause is exactly what you named — `tests/relay_invite_signer.mjs`
could only reach the tool by spawning it, and every case a subprocess can drive uses a local key
file, where pairing address, resolved identity and pinned key are the same value. "Mutation-proven"
was true of `src/nostr_signer.mjs` and false of the tool, and the PR body claimed coverage of a
banner the suite could not see.

Fix: the decisions moved to `src/relay_invite_signer.mjs` behind injected deps (`loadBunker`,
`loadLocal`, `pin`) — injection because `src/` may not import the signer backend under the egress
ban. `tools/relay-invite.mjs` now contains zero `signer.pubkey` and zero `npubEncode`. The new
in-process block hands in a split stub (`pubkey: uriHex`, `userPubkey: async () => userPk`) with a
spy `pin` capturing what was pinned. All three mutations now fail: 2, 2 and 3 assertions.

**2 — hoisted.** `EXPECT_PUBKEY`'s format check now runs before anything is loaded or asked, so a
typo costs no connect and no approval tap. Asserted as `asked === 0` — no round trip happened — not
as message text.

**3 — #479 widened**, with your fail-open warning quoted verbatim, because the ordering is the
finding: `tools/tripwire.mjs:97` has to move in the same change as the rest or rule 1 compares a
transport address to the poster and stays silent. Added `src/buzz_policy_runner.mjs:82,95` and
`tools/tripwire_alarm_lib.mjs:36,38`, and cited `tools/grant.mjs:77` /
`tools/publish-dm-relay-list.mjs:85` as the existing precedent.

Nothing here has been run against a real bunker. Whether any pairing in this deployment actually
splits the two keys is still unknown from the repo — that needs one live `get_public_key`, which is
the maintainer's to run.

PR #481, opened 2026-08-15
Let the operator paste a bunker:// URI instead of seating two files (#480)

### LINKED ISSUE #480: relay-invite: the operator cannot paste a bunker:// URI — the pairing must already be two files on disk
`relay-invite` can only reach a Bunker whose pairing is already sitting in two mode-0600 files
(`WAGGLE_BUNKER_URI_FILE` + `WAGGLE_NIP46_CLIENT_NSEC_FILE`). That is fine for a seated agent
runtime, where an administrator put those files there. It is the wrong shape for the operator, who
holds the only key with `owner`/`admin` in `relay_members` and therefore is the only one who can
run `mint` — and whose signer hands them a `bunker://` string to **paste**, not a file to place.

Today the operator's only route is to hand-create two files, one of which must hold a freshly
generated client nsec they have no other reason to own. That is friction in front of the one step
that gates every relay invite, and it pushes toward the thing the tool exists to avoid: putting a
key on disk.

**Why not just take it as an argument or an env var.** A `bunker://` URI carries a `secret`
parameter. `--bunker-uri=<value>` puts it in `ps` output; `BUNKER_URI=…` puts it in the shell
history of anyone who typed it and in `/proc` on Linux. `tools/relay-invite.mjs` already refuses
argv and env for the signing key for exactly this reason, and a URI holding a connect secret is
the same class of material. So the fix is a **TTY prompt with echo off**, and an explicit refusal
if the URI is passed any other way — otherwise the prompt is decoration that the first person in a
hurry routes around.

**Precedent to match, not reinvent** — `tools/grant.mjs:56-66` already solves the client-key half,
with the reasoning in a comment: a NIP-46 bunker authorizes a specific **client keypair**, so a
fresh key every run is a new app the signer has never seen ("Unknown client"). It pairs once,
persists the client key at 0600, and reuses it. Same behaviour belongs here. The URI itself should
NOT be persisted — it holds the connect secret, and the paired client key is what carries the
authorization afterwards.

Two inconsistencies worth settling in the same change rather than leaving as folklore:

- `grant.mjs` takes its bunker URI from `GRANTOR_BUNKER`, an environment variable. That is the
  exact exposure this issue is about. Either it is acceptable there or it is not acceptable here.
- `relay-invite`'s `chooseSigningSource` refuses when `--key` and a pairing are both configured,
  because a claim writes a `relay_members` row that cannot be removed without that key (#366). A
  third source has to enter that same check, or the ambiguity guard develops a hole.

Scope:
- `--bunker` (no value) prompts on `/dev/tty` with echo disabled, for `mint` and `claim` alike.
- Refuse a bunker URI seen in argv or the environment, naming why.
- Client key reused from a 0600 file, generated and saved on first use, with the "approve this app
  once" message `grant.mjs` prints.
- Resolve `get_public_key` and print the npub **before** anything is signed, so the operator sees
  which identity is about to act.
- `chooseSigningSource` gains the third source and keeps refusing every ambiguous combination.

Depends on #477/#478 — the Bunker signing path it extends is not merged yet.

Testing has to cover the split-identity case in-process, not only by spawning the command line.
That is the defect the #478 review found: every case a subprocess can drive uses a local key file,
where the pairing address and the resolved identity are the same value, so three mutations
survived a full green suite.

JAFairweather:
Closes #480. **Stacked on #478** (`fix/477-relay-invite-nip46`) — merge that first; this branch
contains its three commits.

## Why

The operator holds the only key with `owner`/`admin` in `relay_members`, so they are the only one
who can `mint`. Their signer hands them a `bunker://` string to paste. `relay-invite` could only
take a pairing already sitting in two mode-0600 files, which meant creating a client nsec by hand
to avoid putting a key on disk. That is the wrong trade and it stalled the one step the rest of the
chain waits behind.

`--bunker` prompts on the terminal with echo off, for `mint` and `claim` alike. The URI is never
written down. Only the NIP-46 client key persists, so a second run costs no second approval — same
reasoning and same 0600 file as `tools/grant.mjs:56-66`, where a fresh client key each run is an app
the signer has never seen ("Unknown client").

A URI in argv or an environment variable is **refused**, and the message says to rotate the pairing,
because by then the secret is already in `ps` and in shell history. Without that refusal the prompt
is decoration that the first person in a hurry routes around.

`chooseSigningSource` gains the third source. Each ambiguous **pair** is refused by name rather than
resolved by precedence — a claim writes a `relay_members` row that cannot be removed without
whichever key signed (#366).

## Two defects the tests could not see

Both found by running the prompt through a pty, and both invisible to `node --check`, to lint, and
to the whole suite:

**The promise never settled on a closed input.** Ctrl-D or Ctrl-C at the prompt exited 13 on an
unsettled top-level await instead of saying anything. Now it rejects with "the prompt was closed
before a URI was entered — nothing was signed", exit 1.

**The paste was echoed.** A stream opened from `/dev/tty` has no `setRawMode`, so readline could not
take over the echoing and the terminal driver kept doing it — while the prompt printed "nothing will
appear as you paste". A security promise the code did not keep. Fixed by reading `process.stdin` and
requiring a TTY, which also refuses a pipe or a heredoc.

## Evidence

92 suites, `ALL PASS`, exit 0. Lint 0 errors, 1 known warning (#438). 64 assertions in the
signer-selection suite.

The paste path is driven **in-process** through a signer whose `userPubkey()` differs from its
`pubkey` — the arrangement where pinning to the wrong one is visible, and the gap that let three
mutations survive in #478's first attempt.

Five mutations, all detected:

| mutation | result |
|---|---|
| banner drops the `paste` branch | 1 FAIL |
| argv exposure check disabled | 2 FAIL |
| corrupt client key regenerates silently | 1 FAIL |
| `--bunker` + `--key` ambiguity removed | 1 FAIL |
| paste loader wired to `loadLocal` | 1 FAIL |

Plus a negative control — a comment-only mutation — reported SURVIVED, so the harness can say it.
The first attempt at that control did not apply (perl failed to compile on `//` in the pattern);
`ANCHOR MISS` fired and the result was discarded rather than counted.

**The echo suppression has no automated regression test.** Node has no built-in pty and the suite is
node-only, so proving it needs a driver the suite cannot run. It is proven manually, and the
reproduction is:

```
python3 ptydrive.py node tools/relay-invite.mjs mint --bunker    # TYPE_THIS=bunker://…secret=MARKER
```

reading the pty master and grepping for the marker: **0 occurrences**. Two positive controls, because
a probe that has only ever come back clean proves nothing — the same driver against `cat` sees the
marker (2), and against this same prompt with the mute removed sees it (1). Driver in the PR
discussion below rather than in the tree, since an uncommitted-to-CI python file is a maintenance
surface with no runner.

Also confirmed by hand on the valid path: client key written 0600, reused byte-identical on a second
run with no "new key" message, and `grep -rl 'bunker://'` under `WAGGLE_HOME` finds nothing.

## Review

Worth a reader on the prompt itself (`promptHidden` in `tools/relay-invite.mjs`). It is the part with
no CI behind it, and the echo bug is the kind that returns quietly the next time someone touches the
stream handling.

### DIFF
--- a/src/bunker_paste.mjs
+++ b/src/bunker_paste.mjs
@@ -1,0 +1,131 @@
+// bunker_paste.mjs — accept a bunker:// URI the operator pastes, without letting it reach a place
+// that keeps it (#480).
+//
+// The operator holds the only key with `owner`/`admin` in `relay_members`, so they are the only
+// one who can mint an invite — and their signer hands them a `bunker://` string to PASTE, not a
+// pairing to seat in two files. Requiring the files put a key on disk to avoid putting a key on
+// disk, which is the wrong trade and it stalled the one step everything else waits behind.
+//
+// A `bunker://` URI carries a `secret` parameter. That makes it credential material, so:
+//
+//   - it is read from `/dev/tty` with echo off, never from argv (`ps`) and never from an
+//     environment variable (the shell history of whoever typed it, and `/proc` on Linux);
+//   - `findBunkerUriExposure` refuses the run when one shows up in either place anyway. Without
+//     that refusal the prompt is decoration, because the first person in a hurry routes around it
+//     and nothing says otherwise;
+//   - the URI is never written down. What persists is the CLIENT key, which is a different thing:
+//     a NIP-46 bunker authorizes a specific client keypair, so a fresh one each run is an app the
+//     signer has never seen and the connect is refused as "Unknown client"
+//     (`tools/grant.mjs:56-66`, same reasoning, same 0600 file). Persisting it is what makes the
+//     second run cost no approval; persisting the URI would keep the connect secret for no gain.
+//
+// Pure: strings in, verdicts out. No TTY, no filesystem, no socket — those live in the tool. This
+// module also may not import the signer backend (egress ban), which is why the caller builds the
+// signer from what these functions return.
+
+const BUNKER_URI = /^bunker:\/\/([0-9a-f]{64})(\?|$)/i
+
+/**
+ * Is this pasted text a bunker URI that can actually be connected with?
+ *
+ * Every refusal names the part that is wrong. A paste is a manual act — a truncated copy, a
+ * trailing quote from a chat client, or a signer that emitted `nostrconnect://` instead all land
+ * here, and "invalid bunker URI" sends the operator back to re-copy the same string.
+ *
+ * @returns {{uri: string, pubkey: string, relays: string[], hasSecret: boolean}|{error: string}}
+ */
+export function checkPastedBunkerUri(text) {
+  // Signers wrap the string in quotes, and a paste through a terminal can carry stray whitespace.
+  const raw = String(text ?? '').trim().replace(/^["'<]+|["'>]+$/g, '').trim()
+  if (!raw) return { error: 'nothing was pasted' }
+
+  if (/^nostrconnect:\/\//i.test(raw)) {
+    return { error: 'that is a nostrconnect:// URI, which is the OTHER direction — the client ' +
+      'publishes it and the signer connects back.\n' +
+      '  This tool needs the bunker:// string from your signer\'s "connect an app" flow.' }
+  }
+  if (!/^bunker:\/\//i.test(raw)) {
+    const scheme = raw.includes('://') ? raw.split('://')[0] : '(no scheme)'
+    return { error: `expected a bunker:// URI, got ${scheme}://…` }
+  }
+  const match = BUNKER_URI.exec(raw)
+  if (!match) {
+    // The commonest paste failure by far, and it does not look like a failure: the string is
+    // plausible and merely short, so a length-blind check would hand it to the signer and report
+    // the bunker's refusal instead of the truncation.
+    const after = raw.slice('bunker://'.length).split('?')[0]
+    return { error: `the pubkey in the URI is ${after.length} characters, not 64 — the paste looks ` +
+      'truncated. Copy the whole string.' }
+  }
+
+  let url
+  try { url = new URL(raw) } catch { return { error: 'the URI is not parseable — copy the whole string' } }
+  const relays = [...new Set(url.searchParams.getAll('relay').filter(v => /^wss:\/\//i.test(v)))]
+  if (!relays.length) {
+    const any = url.searchParams.getAll('relay')
+    return { error: any.length
+      ? `the URI's relay(s) are not wss:// — NIP-46 traffic is not sent over ${any[0].split('://')[0]}://`
+      : 'the URI names no relay, so there is no transport to reach the signer over. Copy the whole string.' }
+  }
+  return {
+    uri: raw,
+    pubkey: match[1].toLowerCase(),
+    relays,
+    // Not an error. A secret is single-use on first pairing, and a URI re-copied after the client
+    // key is already authorized legitimately has none. Refusing here would break the second run.
+    hasSecret: !!url.searchParams.get('secret'),
+  }
+}
+
+/**
+ * Did a bunker URI arrive by a route that records it?
+ *
+ * Checked BEFORE the prompt, so the answer is a refusal rather than a prompt the operator fills in
+ * while the same secret already sits in their history.
+ *
+ * @returns {{error: string}|{}}
+ */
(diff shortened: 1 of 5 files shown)


JAFairweather:
The pty driver, so the echo claim can be re-run rather than taken on trust. Save as `ptydrive.py`:

```python
import os, pty, select, sys, time
cmd = sys.argv[1:]
typed = os.environ['TYPE_THIS']
pid, fd = pty.fork()
if pid == 0:
    os.execvp(cmd[0], cmd)
time.sleep(1.2)                      # let the prompt come up and raw mode take effect
os.write(fd, (typed + "\r").encode())
out = b''
deadline = time.time() + 6
while time.time() < deadline:
    r, _, _ = select.select([fd], [], [], 0.3)
    if r:
        try: chunk = os.read(fd, 65536)
        except OSError: break
        if not chunk: break
        out += chunk
os.close(fd)
try: os.waitpid(pid, os.WNOHANG)
except ChildProcessError: pass
sys.stdout.buffer.write(out)
```

Then:

```
export TYPE_THIS="bunker://deadbeef?relay=wss://r.example&secret=ZZMARKERZZ"
WAGGLE_HOME=$(mktemp -d) BUZZ_RELAY_URL=http://127.0.0.1:1 \
  python3 ptydrive.py node tools/relay-invite.mjs mint --bunker | grep -c ZZMARKERZZ   # 0

# positive control 1 — the driver CAN see an echo
python3 ptydrive.py cat | grep -c ZZMARKERZZ                                           # 2

# positive control 2 — remove the mute from THIS prompt and the marker returns
perl -0pi -e 's/\Qif (!muted) output.write(s)\E/output.write(s)/' tools/relay-invite.mjs
```

The truncated pubkey is deliberate: it refuses before any network, so what is under test is
whether the typed characters come back, not whether the connect succeeds. The refusal message
naming the length is also what proves the line was read at all — an earlier attempt with
`script` and a pipe reported a marker hit that was `script` echoing its own input, on a run
where the prompt had closed before reading anything. That result was discarded.

JAFairweather:
## Approve — both fixes from the last round are in and verified

Re-reviewed at `69828666` (was `970fd1d7`). `npm test` **rc=0, 0 failures**.

**Fix 1 — the env scan reads values, not key names.** This was the sharp one: `findBunkerUriExposure`
scanned argv by value but env by *key name*, so any variable not on the name list carried a pasted
bunker URI straight through. Driven:

```
SIGNER=bunker://…?secret=…        CAUGHT  "SIGNER holds a bunker:// URI, and this tool will not take on…"
NOSTR_CONNECT=bunker://…          CAUGHT
BUNKER=bunker://…                 CAUGHT
FOO_BAR_BAZ=bunker://…            CAUGHT   <- the case that mattered; no name list can enumerate this
argv carries it                   CAUGHT   (control)
nothing anywhere                  {}       (control — no .error)
```

Closed, and closed at the right layer: by value, so the key name is irrelevant.

**Fix 2 — the refusal no longer contradicts itself.**

```
64 non-hex   "the pubkey in the URI is the right length but is not hex — it contains "z". A pubkey is 64 characters of 0-9 and a-f…"
63 hex       "…is 63 characters, not 64 — the paste looks truncated. Copy the whole string."
65 hex       "…is 65 characters, not 64 — the paste looks to have picked up extra characters."
```

`is 64 characters, not 64` is gone, and naming the offending character is more than I asked for —
that is the difference between a refusal and a diagnosis.

The pty evidence from the last round still stands (real prompt 0 markers, `cat` control 2,
mute-removed control 1), and the refusal naming "8 characters" is what proves the 0 is a reading
rather than an absence.

### One nit, not blocking

```
bunker://npub1p28qwgx…?secret=…  ->  "is 63 characters, not 64 — the paste looks truncated. Copy the whole string."
```

An `npub1…` is 63 characters, so the most likely paste mistake is diagnosed as the one thing it is
not, and the remedy offered ("copy the whole string") is the one that cannot work — they did copy the
whole string, of the wrong encoding. The prefix is trivially detectable and you already have the
string. Same one-line shape as fix 2: `that is an npub, not a hex pubkey — a bunker:// URI carries
the 64-hex form`.

PR #483, opened 2026-08-15
Attempt the community leg without an auth tag, and report what the relay says (#482)

### LINKED ISSUE #482: publish_profile skips the community leg without BUZZ_AUTH_TAG, so it cannot test whether membership is now enough
`publish_profile.mjs` gates the community leg on **both** env vars:

```js
const community = !!(process.env.BUZZ_RELAY_URL && process.env.BUZZ_AUTH_TAG)
```

With `BUZZ_AUTH_TAG` unset it does not attempt the community push — it **skips** it, exits 0, and
prints "skipped". The community copy is the only leg that writes the `users` row an at-word
resolves against, so a run with the relay URL right, the signer right and custody proven reports
success while moving the name not at all.

That gate made sense when an owner-minted NIP-OA auth tag was the only way past
`enforce_relay_membership`. It is not any more. Since #357/#477 a key can claim a relay invite and
be inserted into `relay_members`, which is what the AUTH gate reads — live-proven tonight:
`200 — joined` for `ad05b00e`.

So whether the auth tag is still required **on top of** membership is now an open, answerable
question, and the tool cannot ask it. Refusing to attempt is not the cautious option here; it is
the one that cannot learn anything, and it hides the answer behind a green run.

Fix: attempt the community leg whenever `BUZZ_RELAY_URL` is set, send `x-auth-tag` only when there
is one, and report the relay's answer as the answer.

Two things fall out that are worth getting right rather than papering over:

- **With no tag the two copies are the same event.** No tag means no `tags[]` entry, so the
  community copy is byte-identical to the trio copy — same pubkey, same `created_at`, same content,
  therefore the same id. Signing it twice would cost a second bunker approval and print two ids
  that are one id. The tool's standing claim is "one profile, not one event"; in this case it
  inverts, and it should say so rather than let a reader notice two equal ids.
- **`x-auth-tag: undefined` is not "no tag".** A `fetch` header set from an unset env var puts the
  literal string `undefined` on the wire, which the relay reads as a malformed tag. The run would
  then answer a question nobody asked.

JAFairweather:
Closes #482. Branches from `main` — independent of #478 and #481.

## Why

`publish_profile.mjs` gated the community push on `BUZZ_AUTH_TAG` as well as the relay URL. Unset,
it **skipped** the leg and exited 0. That leg is the only one that writes the `users` row an at-word
resolves against, so a run with everything else right reported success and moved the name not at all.

The gate made sense when an owner-minted NIP-OA auth tag was the only way past
`enforce_relay_membership`. Since #357/#477 a claimed key is in `relay_members`, which is what the
AUTH gate reads — live-proven tonight: `200 — joined` for `ad05b00e`. Whether the tag is still
needed *on top of* membership is now an answerable question, and the tool could not ask it.
Refusing to attempt is not the cautious option; it is the one that cannot learn anything.

## Two consequences, handled

**With no tag the two copies are one event.** No tag means no `tags[]`, so the community copy is
byte-identical to the trio copy — same id. It is reused rather than signed again, which would cost a
second bunker approval and print two ids that are one id. This tool's standing claim is "one profile,
not one event"; here it inverts, and it says so.

**`x-auth-tag: undefined` is not "no tag".** A `fetch` header from an unset env var puts the literal
string `undefined` on the wire, which the relay reads as malformed. Omitted entirely now.

## Evidence

92 suites `ALL PASS`, exit 0. Lint 0 errors, 1 known warning (#438).

| mutation | result |
|---|---|
| re-gate the leg on the tag | 3 FAIL |
| sign a second identical kind:0 instead of reusing | 3 FAIL |
| treat a malformed tag as absent | 5 FAIL |
| revert the signature-count formula | 1 FAIL |

Both directions: with a tag the copies are still two events with different ids, and a malformed tag
is still refused rather than read as absent.

**One thing is not asserted, and the test says so.** Removing the announced-vs-actual signature
count guard alone survives a green suite — no input this suite can construct makes the counts
disagree. What proves the guard is the second-identical-kind:0 mutation, which without it survives
silently because the two events share an id. An assertion that the guard's message is absent would
pass whether or not the guard existed, which is worse than none.

Live dry run against the real Bunker pairing: custody proven for `ad05b00e`, one event built for
both sides, nothing published.

## Review

The interesting question is whether reusing the trio event for the community leg is right, or
whether the community copy should carry something that distinguishes it even without a tag.

### DIFF
--- a/tests/publish_profile.mjs
+++ b/tests/publish_profile.mjs
@@ -82,13 +82,64 @@ const run = (env, argv) => {
   ok('  ...and nothing was published on the refusal', /Nothing published/.test(r.stderr))
 }
 
-// 4. The community leg is skipped, loudly, when its env pair is absent — never silently.
+// 4. The community leg is skipped, loudly, when there is no relay URL — never silently.
 {
   const r = run({ BUZZ_PRIVATE_KEY: hex }, ['--dry-run', '--content-file', contentFile])
   ok('skipping the community leg is stated, not silent',
-    r.status === 0 && /BUZZ_RELAY_URL \/ BUZZ_AUTH_TAG unset/.test(r.stderr))
+    r.status === 0 && /BUZZ_RELAY_URL unset/.test(r.stderr))
   ok('  ...and the signature count drops to 1 with it, rather than staying wrong',
     /this run makes 1 signature\b/.test(r.stderr))
+}
+
+// 4b. A RELAY URL AND NO AUTH TAG (#482). This combination used to skip the community leg — the
+//     one that writes the `users` row an at-word resolves against — so a run with everything else
+//     right reported success and moved nothing. Since #357/#477 a claimed key is in relay_members
+//     and passes the gate on its own, so the tool must ATTEMPT it and report what the relay says.
+{
+  const r = run({ BUZZ_PRIVATE_KEY: hex, EXPECT_PUBKEY: pk, BUZZ_RELAY_URL: 'wss://relay.invalid' },
+    ['--dry-run', '--content-file', contentFile])
+  const e = r.stderr
+  ok('a relay URL with no auth tag ATTEMPTS the community leg rather than skipping it',
+    r.status === 0 && !/skipped/.test(e), `exit ${r.status}`)
+  ok('  ...and the two copies are ONE event with one id, because there is no tag to differ by', (() => {
+    const t = e.match(/trio copy\s+([0-9a-f]{64})/), c = e.match(/community copy\s+([0-9a-f]{64})/)
+    return t && c && t[1] === c[1]
+  })())
+  ok('  ...and that inversion is said out loud, not left for the reader to notice two equal ids',
+    /the SAME id: with no auth tag the two copies are one event/.test(e))
+  ok('  ...and the signature count is 1, not 2 — a bunker operator is not told to wait for a prompt that never comes',
+    /this run makes 1 signature\b/.test(e))
+
+  // BOTH DIRECTIONS. With a tag it must still build two distinct events — otherwise this change
+  // reads as "works" while having quietly dropped the auth-tag path.
+  const withTag = run({ BUZZ_PRIVATE_KEY: hex, EXPECT_PUBKEY: pk,
+    BUZZ_RELAY_URL: 'wss://relay.invalid', BUZZ_AUTH_TAG: '["auth","probe"]' },
+    ['--dry-run', '--content-file', contentFile])
+  ok('  NEGATIVE CONTROL — WITH a tag the copies are still two events with different ids', (() => {
+    const t = withTag.stderr.match(/trio copy\s+([0-9a-f]{64})/)
+    const c = withTag.stderr.match(/community copy\s+([0-9a-f]{64})/)
+    return t && c && t[1] !== c[1] && /this run makes 2 signatures/.test(withTag.stderr)
+  })())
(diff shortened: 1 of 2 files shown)


JAFairweather:
**Live answer, from a real run: the auth tag is not required.** The community relay accepted the
kind:0 over NIP-98 alone, no `x-auth-tag` header at all:

```
community relay        200 {"accepted":true,"event_id":"f41f036e1840…","message":""}
```

That closes the open question in #482. `relay_members` membership — inserted by the `claim` in
#478's path, `200 — joined` for `ad05b00e` earlier tonight — is sufficient to write. The
owner-minted NIP-OA tag is not a second requirement on top of it.

Which also means the gate this PR removes was not conservative. It was the only thing standing
between a correct configuration and a successful publish, and it failed by reporting success.

**The read half is still refused, and the tool says so correctly.** Community read-back is
`403 RBAC: access denied` — the same refusal an unadmitted key gets, so it distinguishes nothing
and the tool reports `INCONCLUSIVE` rather than a pass. That matches the standing claim in
CLAUDE.md: membership buys write, not read. Nothing here should be read as native read arriving.

Trio: 3/4 CONFIRMED on the run, 4/4 on a re-read a minute later — `relay.ditto.pub` had not served
the new event yet when the tool looked. Propagation at read time, not a comparison fault.

JAFairweather:
## Approve

First read, at `f1946512`. 1 ahead / 2 behind `origin/main`, no conflict. `npm test` **rc=0, 0
failures**.

The gate change is right and the reasoning behind it is the part I checked hardest. Dropping the
community leg from `!!(URL && TAG)` to `!!URL` means a run with no tag now *attempts* the push and
reports what the relay says, instead of skipping and reporting success while leaving the `users` row
— the only thing an at-word resolves against — unwritten. Refusing to try is not the safe option
when the open question is whether the tag is still required; it is the option that cannot learn
anything. That is the correct read of #357/#477.

`authHeader()` omitting the header entirely rather than sending the string `"undefined"` is the
right mechanic: `undefined` on the wire is a *malformed* tag, not an absent one, so the old shape
would have answered a question nobody asked.

### The mutation you named — I ran it

Your test block declines to assert on the count guard and says why, then names the mutation that
proves it. That is the right way to leave a hole open, so I closed the loop rather than taking it:

```
BASELINE                                                        ALL PASS

MUTATE  communityEvent = trioEvent
     -> communityEvent = await sign({ kind: 0, created_at, content })

FAIL a relay URL with no auth tag ATTEMPTS the community leg rather than skipping it  [exit 2]
FAIL   …and the two copies are ONE event with one id, because there is no tag to differ by
FAIL   …and that inversion is said out loud, not left for the reader to notice two equal ids
FAIL   a whitespace-only tag counts as absent, not as malformed  [exit 2]
```

The guard fires (`exit 2`) and four assertions go red, exactly as documented. Without it that
mutation is invisible — the two events have the same id, so nothing in the output differs — and it
costs a bunker operator a real approval at a prompt they were told not to expect. The comment is
accurate and the guard is load-bearing.

The rest of the suite change is unusually careful: a negative control proving that *with* a tag the
copies are still two distinct events (so the change cannot read as "works" while having quietly
dropped the auth-tag path), and the malformed-vs-absent pair — `BUZZ_AUTH_TAG='not json'` still
refused, `'   '` treated as absent — which is precisely the distinction that making a required field
optional tends to erase.

No blockers.

Issue #486, opened 2026-08-15
One paste, then a working Pi: the acceptance test for remote-agent onboarding


JAFairweather:
Umbrella for #309's close condition, extended to the agent's own machine. Sub-issues follow.

## The flow

An operator opens the console, mints a key, enrols it in their bunker, and pastes **one**
`bunker://` URI. Nothing else is typed. The console does the rest, and hands back a prompt. The
operator pastes that prompt into a fresh Pi session, and the Pi takes on the identity, has its MCP
and its shell, has every credential in the right place, and starts collaborating under its own name.

That single paste is the only manual step that survives. Everything else is a defect, not a caveat.

## Where it stands, measured

| # | Step | State |
|---|---|---|
| 1 | Console mints the key, shows it once | built — `console/mint-agent-key.mjs`, take-once secret |
| 2 | Operator enrols it, pastes the URI, custody proved | built — `console/index.html`, `proveCustody` |
| 3 | Grants issued, correctly directed | built — `console/connect-plan.mjs`, `connect.html` steps 1–4 |
| 4 | Relay invite minted and claimed | tool only — `tools/relay-invite.mjs` (#357, #477) |
| 5 | `kind:0` published both sides, `@Name` resolves | tool only — `tools/publish_profile.mjs` (#459) |
| 6 | Broker seat: the channel key in `authorized_keys` | **nothing** — no tool anywhere writes it |
| 7 | The Pi gets a NIP-46 pairing to the identity | **nothing**, and the obvious route is forbidden |
| 8 | Pi machine: manifest, dirs, channel keypair, MCP | #465 / #467 / #471 / #473 / #475, unmerged |
| 9 | A copyable first prompt | partial — `startupDoc` (#467) is the content; step 5 emits a CLI |

Rows 1–3 are the step the operator asked to keep. Rows 4–7 and 9 are the work.

## Two things that shape every sub-issue

**The console has no backend.** `tools/serve-console.mjs` is a 92-line static server on
`127.0.0.1`, and `deploy/README.md` is explicit that the console is not served from the bridge
host. Rows 4–5 may be reachable from the browser directly; that is a measurement, not an
assumption. Row 6 is not, and belongs on the actuator — `DESIGN_AGENT_LIFECYCLE_PLANE.md` already
says host-touching work is console-signed intent with runner execution. No new HTTP surface.

**The prompt may not carry a credential.** A `bunker://` URI is a secret. `assertNoSecrets` refuses
[...]

PR #491, opened 2026-08-15
The pairing token, so an approved session can act without a pasted secret (#489)

### LINKED ISSUE #489: The first prompt cannot carry the pairing, so the R/A join in DESIGN_JOIN has to be built (#486)
Part of #486. Row 7, and the hardest one.

## Problem

A fresh Pi session has to end up holding a NIP-46 pairing to the agent identity. The obvious route
is to put the `bunker://` URI in the prompt the operator pastes. That is forbidden, and correctly:
a pairing URI is a credential, `CLAUDE.md` forbids printing one anywhere, and `assertNoSecrets`
already refuses to emit a startup file containing one — `src/agent_startup.mjs` lists
`bunker://` and `nostrconnect://` as the first two shapes it fails closed on.

So the pasted prompt must carry **no secret at all**, and the pairing has to reach the Pi by another
route.

**Correction to this issue as first written.** It said row 7 "has no implementation" and that
"nothing was built from" the design. That is wrong and understates what is on `main`:
`src/join_request.mjs`, `src/join_approval.mjs`, `tools/join.mjs` and their suites all exist, and
`tools/join.mjs` already mints R, publishes the request with a cold read-back by id, and burns R on
every exit path. The precise gap is **one artifact** — the sealed pairing token that turns an
approval into a working session, and the custody proof that has to sit between opening it and
writing it. Everything else in the ceremony was already here.

## The design already written

`docs/DESIGN_JOIN.md`, two keys doing two different jobs:

- **R, the request key** — minted in the Pi session, signs the join request, receives the pairing
  token **sealed to R**, then is burned. Never granted anything, never in the roster. Its only job
  is to be an address the owner can reply to, once.
- **A, the agent identity** — lives in the owner's bunker and never in the session at all. The
  session holds a pairing, not a key.

The pasted prompt then contains only public material: name, pubkey, channel id, and where to look.
The secret never appears in a chat window, a scrollback, or a file.

⚠ Do not shortcut this to "mint an ephemeral key, act, burn it". Burning R is the design; burning
the *identity* is the shape #141 weighed and set aside — a fresh npub per session is not admitted,
is not in `return_lane`, and has no continuity.

## What has to be built

- Pi side: mint R, emit the join request, wait for the sealed token, pair, burn R.
- Console side: show the pending request, let the owner approve, seal the single-use token to R.
- The approval is signed by the owner's own key, not a tap on a button in someone else's app
  (`DESIGN_JOIN.md`, the approval plane).

## Acceptance

- A Pi session that was handed nothing but public text ends up signing as A.
- The token is single-use: a replay is *seen* to be refused, not silently ignored.
- R never appears in the roster, and the pairing never appears in any file the prompt touches.
- Negative control that fires: a join request signed by a key the owner did not approve is refused,
  and a legitimate one still gets through — both directions asserted, because a guard proved only to
  reject cannot be told apart from one that rejects everything.

JAFairweather:
Closes #489. Stage 4 of the #486 sequence.

`docs/DESIGN_JOIN.md` ended with a session that had been approved and still could not act.
`tools/join.mjs` published the request, slept, and burned the request key; its own closing line
said pairing "is not built yet". The obvious finish — paste the `bunker://` URI into the prompt
handed to the new session — is the one route this project forbids. A pairing URI is a credential,
and `src/agent_startup.mjs` fails closed on exactly that shape.

So the pairing arrives sealed to the request key instead. R receives it and is burned; A never
leaves the owner's Bunker.

## What `src/pairing_token.mjs` guarantees

Each of these prevents a failure that would not announce itself.

| | |
|---|---|
| take-once | the plaintext is on no plain field of the result. A second read yields `null`, not a copy that reaches a log |
| bound to the request | checked against the id **this session sent**, never one read out of the token. Without it, one owner approval releases a pairing any listener holding any request key can spend |
| custody unproven | opening proves who sealed it, not who controls the identity. The result says `custodyUnproven: true` and promises nothing further |
| an nsec is refused **by name** | on build and on read. That slip *works* if uncaught, so nothing would report it — it silently converts the design into the one it exists to prevent |

## What `tools/join.mjs` now does

Listens for a token from the hive it asked to join, opens it, then — because the module
deliberately refuses to claim custody — mints a NIP-46 client key, signs one challenge through
`withPinnedCustody` pinned to the approved identity, and never publishes it. Only after that
signature verifies does anything reach disk, and only under an explicit `--seat <dir>`. Without
one the ceremony still runs, still proves custody, and discards the pairing rather than picking a
location of its own. The request key burns on every path, including the no-token timeout — which
exits 4 and says plainly that a token sealed to it now can never be opened.

## Verification

95 suites, rc=0. `tests/pairing_token.mjs` asserts **reasons**, not just refusals, and pairs every
refusal with a legitimate value still getting through — including that a token refused for one
session **does** open for the session that actually sent that request.

Seven mutations, all caught, no anchor misses, green on restore:

```
caught  <- 1: the token stops being bound to the request — any approval pairs any session
caught  <- 2: take-once becomes take-many — the URI can be read twice
caught  <- 3: the read side stops screening the URI — an nsec arrives and is handed on
caught  <- 4: expiry is off by one — a token is honoured at the instant it expires
caught  <- 5: unknown fields ride along unexamined
caught  <- 6: an opened token claims custody is proven
caught  <- 7: the build side stops refusing an nsec
```

## Not claimed

The responder is still not built. The owner seals the token by hand from the console
(`docs/JOIN_RUNBOOK.md` §4–§5). The live leg — a real bunker on the other end of a real URI — has
not been walked; that is stage 6.

## Review

Substantive: it touches a key path, so it needs a read from someone who did not write it.
**@My Dude** — the two things worth attacking are (a) whether pinning custody *before* the write
is actually sufficient, or whether a signer can pass one challenge and serve a different key after,
and (b) whether refusing a token but continuing to listen is the right shape, or whether a refused
token should abort the ceremony.


JAFairweather:
## First read of `05a37ef1` — changes requested. One must-fix, one that bites on the happy path.

Read at `05a37ef1`, worktree `waggle-491`, `node tests/pairing_token.mjs` → ALL PASS. The module is the strongest thing in this PR: take-once, unknown-key refusal, the nsec named rather than generically rejected, and the rid checked against what *this* session sent rather than what the token says. Rule 2's comment — the refusal deliberately not echoing the token's own rid — is the kind of care this path needs. Nothing below is about `src/pairing_token.mjs`.

---

### Your question (a): can a signer pass one challenge and serve a different key afterwards?

**In this process, no — and the repo already proves it.** `withPinnedCustody` re-verifies *every* signature against `pinned` (`src/nostr_signer.mjs:156-168`), and `tests/publish_profile.mjs:104-125` drives exactly that shape: a `twoFaced` signer that returns A for signature 1 and B for signature 2, refused on the second. Your challenge is sound as far as the process goes.

**Across the seat boundary, yes — and that is the must-fix.**

### MUST-FIX — the ceremony proves `A`, then throws `A` away

`tools/join.mjs:234-241` writes exactly two files:

```js
for (const [name, value] of [['bunker-uri', pairingUri], ['bunker-client', clientNsec]]) {
```

Those two are precisely what `WAGGLE_BUNKER_URI_FILE` / `WAGGLE_NIP46_CLIENT_NSEC_FILE` feed to `loadNostrSigner`. **Nothing in the seat records `opened.identityPubkey`.** It is proved at line 217, printed at 222, printed again at 249 — and never written.

So the next session that loads that seat has nothing to pin to. Downstream pinning is `EXPECT_PUBKEY`, an env var a human has to type, and `src/relay_invite_signer.mjs:75-78` says what happens when it is unset — in its own words:

> Pinning to the RESOLVED identity when EXPECT_PUBKEY is unset is not a no-op: a bunker holding several identities can answer `sign_event` as one it did not report here…

It pins to whatever the bunker *reports*. Consistently, verifiably — to the wrong key. The `bunker://<hex>` in the seat is the remote **signer** address and NIP-46 permits it to differ from the identity it signs with, so nothing recoverable from the seat files names `A`. The failure is silent and reads clean: the operator ran the ceremony, saw `custody proved — the signer signs as <A>`, and a later session signs as `B` with every check green.

**Fix:** write a third file in the same loop — `identity` (or `expect-pubkey`), `opened.identityPubkey`, mode 600 — and print the `EXPECT_PUBKEY=<hex>` line the operator needs, so the proof this ceremony exists to obtain is carried rather than spent. Without it, the challenge protects one signature in a process that then exits.

**Same block, same theme:** `mkdirSync` + `writeFileSync` overwrites an existing `bunker-uri` with no check. `tools/connect-agent.mjs:11` states the repo's rule as *"It NEVER overwrites and never deletes. An existing credential is left exactly as found"*, and `--seat` pointed at a live agent's directory clobbers a working pairing unrecoverably. It compounds with the above: with no identity recorded, nothing afterwards can even tell you which identity you overwrote. `existsSync` → refuse and name the path.

---

### SHOULD-FIX — the happy path waits the full timeout after the token has already arrived

`tools/join.mjs:212`:

```js
const opened = (await Promise.all(RELAYS.map(u => listen(u, deadline)))).find(Boolean)
```

Each `listen` resolves independently: on the relay that delivers, `clearTimeout(t); fin(opened)`. The other three keep their own `setTimeout(…, deadline - now)` and resolve `null` **at the deadline**. `Promise.all` waits for all four. `relaySet` defaults to four relays (`src/relays.mjs:39-44`) and `--wait` defaults to 900.

So: owner approves, token lands in seconds, and the custody challenge does not start for another ~15 minutes. Structural repro of the exact shape, WebSocket replaced by a timer, relay `a` delivering at 200ms against a 3s deadline:

```
  token opened on relay a at t=201ms
  custody challenge would start at t=3001ms (opened=true)
```

Not a live run of `join.mjs` — a repro of its control flow, which is where the claim lives. It also puts the whole `--wait` window between the token's mint and the challenge, so a short-expiry pairing can go stale between "opened" and "used" while the process sits there. `Promise.race` on a found-signal, or an `AbortController` the winner trips.

### Your question (b): should a refused token abort the ceremony?

**No — keep listening. But fix what the timeout then says.** `ev.pubkey !== hive` already drops strangers before decryption, so the only party who can produce a refused token is the hive itself; that is a typo or a stale mint, and letting the owner reseal without restarting the whole ceremony is the useful behaviour. Logging the reason is right.

What is wrong is the report on the way out. Tokens arrive, get refused, get logged to stderr — and then `tools/join.mjs:206`:

```
join: no pairing token arrived before the deadline.
```

which is false, and false in the direction that sends the operator to look at the relay instead of at the token they mis-sealed. Count refusals; when there was at least one, say so and exit differently. This is the same shape as the contradictory banners on #467.

### Nit

The unknown-key refusal echoes attacker-supplied strings — `fields this build does not understand: ${extra.sort().join(', ')}` — twenty lines below the comment explaining that the rid is deliberately *not* echoed for exactly that reason. Same input, same reader. Cap the count and the length, or list only known-key names.

### Coverage

No test in the tree references `tools/join.mjs` (`git grep -l join.mjs -- tests/` → empty). The module is asserted seven ways; the tool that uses it — the author pin, the listen loop, the custody challenge, the seat write, the timeout report — has none. Every finding above is in the tool, not the module, and that is not a coincidence. Same note I left on #465: extracting the decision into `src/` is right, and it moves the untested surface to the call site rather than removing it.

### Scope

"The responder is not built" and "the live leg has not been walked" are correctly stated and I am not treating either as a gap. The must-fix is not about the missing responder — it is that the artifact this PR *does* ship does not carry the fact it proves.

@MC Claude

JAFairweather:
All four addressed at `ac55113d`, rebased onto `main` (now 5 commits ahead of what you read: #465, #467, #471, #473, #475 all merged).

**MUST-FIX — the seat did not carry the identity.** You were right about the mechanism, and the part I had not seen is that `bunker://<hex>` is the *signer's* address, so nothing in the two files could name `A` even in principle. `identity` is now a third file at mode 600, and the tool prints `EXPECT_PUBKEY=<hex>`.

**MUST-FIX — the overwrite.** `--seat` refuses if *any* of the three names is already present, and names which. A partial seat is worse than none because it reads as complete.

**SHOULD-FIX — `Promise.all`.** Replaced with `firstTruthy`: first truthy answer wins, losers are cancelled. Measured in the suite — 7ms against a 60s loser, where the old shape waited the loser out.

**Your (b) — a refused token.** Kept listening, as you said. Fixed the report instead: `timeoutReport` counts refusals, says the relay carried them, and exits **5** rather than 4 so the two cases are distinguishable by something other than reading stderr.

**Nit.** Bounded to four names, each clipped at 24 chars, remainder counted.

**Coverage — your real point.** The decisions moved into `src/pairing_seat.mjs` and `tests/pairing_seat.mjs` is 34 assertions over them. I take the caveat you attached to the same note on #465: this moves the untested surface to the call site rather than removing it. What is left in `join.mjs` is wiring.

**Six mutations, one per fix, each reverted in turn:**

```
CAUGHT   — drop the identity file from the seat plan (the must-fix)
CAUGHT   — stop refusing an existing seat (the overwrite must-fix)
CAUGHT   — always report the quiet timeout, never the refused one
CAUGHT   — wait for every task instead of the first truthy (the Promise.all bug)
CAUGHT   — stop cancelling the losers
CAUGHT   — echo every unknown field unbounded again (the nit)
```

The last one **SURVIVED the first run** — I had bounded the echo and written no test for the bound. That is why `tests/pairing_token.mjs` now asserts the cap, the clip, the count, and a negative control that one mistyped field is still named in full. A bound that hid the useful case would be a worse bug than the one it fixed.

`npm test` rc=0, 0 failures, 100 suites.

Still true and still stated: the responder is not built, and the live leg has not been walked against a real bunker.

@My Dude

JAFairweather:
CI on `ac55113d` was red for **lint, not tests** — `B` declared for an assertion I did not write. `npm test` is the suite chain only; CI also runs `eslint src tools tests`, so a green local run said nothing about it.

Fixed at `882d9132` by writing the assertion rather than deleting the constant. It was worth having: a `seatPlan` that always wrote the same identity would satisfy every other check in that file, so two different keys now have to produce two different seats.

`npm run lint` rc=0, 36 assertions, six mutations still caught, full suite rc=0.

One thing worth flagging from the same few minutes, since it is the kind of failure this repo keeps a list of: the mutation script hit a command timeout and was killed **between mutation and restore**. Its restore is in a `finally`, which SIGKILL does not run — so the tree was left holding the reverted `firstTruthy`, the exact Promise.all bug the commit fixes, staged and one command from being committed as real code. Caught by reading `git status` before committing. A mutation run that is killed prints nothing, and a chain that continues past it looks identical to one that passed.

PR #493, opened 2026-08-15
Step 05 hands over the agent's first prompt, not a command line (#490)

### LINKED ISSUE #490: Console step 5 hands over a command line, not the agent's first prompt (#486)
Part of #486. Row 9.

## Problem

`console/connect.html` finishes the connect flow by rendering this:

    node tools/connect-agent.mjs \
      --name <name> \
      --pubkey <hex> \
      --owner <hex>

That is a command for the operator to run somewhere, which is the thing #486 says must not survive.
What the flow should end with is the text the operator pastes into the new session — the agent's
first prompt, telling it who it is, what it may do, which lane carries what, and what has *not*
been proved.

#467 already builds exactly that body. `startupDoc` in `src/agent_startup.mjs` renders it from the
install report, under two absolute rules: nothing secret, and nothing unproven stated as fact. An
agent whose kind:10050 was never checked is told it was never checked, rather than told it is
reachable — that specific confusion has cost this project a day more than once.

## Fix

Replace `renderHandoff` with the pasteable prompt, built from the **same** `startupDoc` body the
runtime reads on disk, so the session prompt and the startup file cannot drift into two different
accounts of the same agent. Gate the emitted text on `assertNoSecrets`, which already fails closed.

Depends on #467 landing.

## Acceptance

- The flow ends with a copy button and text a human can paste, and no command anywhere in it.
- The emitted prompt and the on-disk startup file are byte-identical in body, asserted, not assumed.
- Negative control that fires: feed the emitter a report carrying a `bunker://` and confirm it
  refuses to render rather than rendering it — the sweep's bound is credential *shapes*, and a raw
  64-hex private key is structurally undetectable there, which the header already says and this
  test must not pretend otherwise.

JAFairweather:
Closes #490. Plan stage S4.

## What was wrong

The connect flow ended by rendering a command:

    node tools/connect-agent.mjs --name … --pubkey … --owner …

That is a command for the operator to go and run somewhere else, which is the thing #486
says must not survive. What the flow has to end with is the text they paste into the new
session — the agent's own first prompt.

## The fix

Step 05 now emits that prompt, built from the same `startupDoc` body `tools/connect-agent.mjs
--startup` writes to disk, with a copy button. The paste and the on-disk startup file cannot
become two different accounts of the same agent.

**Why there are two copies.** Both directions are closed, on purpose:

- `tools/serve-console.mjs` pins `DOCROOT` to `console/` and refuses anything above it
  (`serve-console.mjs:72`) — an earlier version served the repo root and exposed `.env`. So the
  page cannot import `../src/`. Verified live: `GET /console/../src/agent_startup.mjs` → 404.
- `console/` is not in the deploy ship list (`deploy-runner.sh:63`), so shipped code importing
  the page's copy would not load on the box.

That leaves a twin under the same bind as `console/scope-hash.mjs`, and the same obligation:
prove they agree. `tests/console_first_prompt.mjs` renders both over a matrix of install reports
covering every branch of the document — the all-confirmed tail, the N-not-confirmed tail, the
singular/plural boundary, no report, and an empty rows array — and asserts byte-identical output.
The refusal path is compared too, message included: `!ok` cannot tell a correct refusal from a
correct refusal with a misleading reason, and the operator acts on the reason.

## Two behaviour changes that fall out of it

- **The prompt renders whether or not the grants landed.** It is built from install state, so a
  failed run produces a prompt saying admission is MISSING and that the agent does not post in
  at all — the thing its first session most needs to know. The old code hid step 05 on failure
  and left the operator at a blank step.
- **`startupDoc` takes `writtenBy`, default unchanged.** The line is a claim about provenance
  inside a document whose rule 2 is that nothing unproven is stated as fact. A pasted prompt
  opening "Written by `tools/connect-agent.mjs --startup`" is wrong on its first line, and an
  agent acting on "regenerate it" would produce a different document from a machine this page
  never observed. `connect-agent`'s own output is byte-for-byte what it was, asserted.

Only one row is reported confirmed: the approvals the page read back **by id from a fresh
connection**. Everything else lives on the agent's machine and is reported as never checked —
never as fine.

## Evidence

- 99 suites green; `npm run lint` and `eslint console` both 0 errors.
- **Eight mutation controls, one per property, all caught.** The eighth SURVIVED its first run:
  the suite proved `writtenBy` existed without proving the page passed it. The assertion that
  closes it is in §6.
- **Loaded live** under `tools/serve-console.mjs`: module graph resolves, the document renders
  at 3749 bytes, and a report carrying a `bunker://` refuses rather than rendering it.
- **A negative control that did not fire, reported as such.** `read_console_messages` showed a
  clean console for a page whose module had 404'd and never executed — so "no console errors"
  proves nothing here. The detector that works is whether the module script ran to completion
  (`#planes` populated), and that is what the live check asserts.
- The sweep's bound is asserted rather than implied: a raw 64-hex private key is structurally
  undetectable and reaches the document. The test says so instead of pretending otherwise.

## Review

**@My Dude** — substantive: it changes what the operator hands a new agent, and it adds a
duplicated module whose only safety is the binding assertion. The two things worth attacking are
whether the matrix in §3 really covers every branch of the document, and whether reporting eight
rows as UNKNOWN produces a prompt an agent reads usefully rather than ignores.


JAFairweather:
Reviewed at `4cdaa567`, worktree provisioned from `waggle-main`. 99 suites green locally, exit 0.

## Changes requested — one must-fix

**`console/connect.html:563` passes the owner's public key as the agent's channel.**

```js
channel: toHex($('owner-key').value) || undefined,
```

`startupDoc`'s `channel` is a channel UUID — `tools/connect-agent.mjs:85` refuses anything else (`--channel must be a channel uuid (8-4-4-4-12 hex)`). The page holds that value in `$('channel-id')` and uses it correctly at lines 362, 372 and 413. `renderHandoff` reaches for `owner-key` instead, so the pasted prompt renders:

```
- **Your channel:** `<64-hex owner pubkey>`
```

Wrong value, wrong type, under a label that is not what it is — in the "Who you are" section, which is the first thing the agent reads.

This is the exact failure the PR is built to prevent. For one agent, `connect-agent --startup` writes the UUID and the console pastes the owner's key: two different accounts of the same agent, one at connect time and one every session after. The twin binding cannot see it, because the suite renders `startupDoc` against `startupDoc` and never asserts what `connect.html` hands it — §6's only argument assertion is the `writtenBy` regex.

Fix is `$('channel-id').value.trim().toLowerCase()`, or omit `channel` entirely and let the `if (channel)` branch drop the line. If omitting is deliberate because step 02 says the id never rides publicly, then say so — but do not substitute a different identifier under the same label.

Add an assertion that the page passes the channel field it collects. The `writtenBy` mutation that survived your first run and this defect are one gap: nothing drives the call site.

## Q1 — does §3 really cover every branch, or is there a tail it never renders?

Yes, on branches. I checked it with V8 coverage over the matrix alone rather than by reading it. Every conditional that affects output executes. The only unexecuted ranges in the twin are:

- `SAYS[r.state] || r.state` at lines 85 and 90 — unreachable; `agent_install_state.mjs` exports exactly four states and §1 pins all four.
- the leak throw at 189 — §4 drives it.
- the `!agent` throw at 82 — nothing drives it, in this suite or any other.

Mutation, 7 applied with the replacement verified present before each run: 5 caught, 2 survived. The survivals are in the parameter dimension, not the branch dimension — the same class as your mutation 8, not the same instance.

| mutation | result |
|---|---|
| twin: UNKNOWN phrase drift | caught, 5 |
| src: wall paragraph drift | caught, 8 |
| twin: singular/plural inverted | caught, 3 |
| twin: all-confirmed tail | caught, 2 |
| twin: UNVERIFIED phrase | caught, 1 |
| src: `briefPath` default diverges | caught, 7 |
| src: `writtenBy` default changed | caught, 8 |
| **twin: ignores the `briefPath` param** | **survived** |
| **twin: `!agent` guard removed** | **survived** |

Neither is live: no caller passes `briefPath` on either side, and `connect.html:538` defaults the name to `'agent'`. But a fixture that passes every non-default parameter closes the class instead of the instance, and `writtenBy` is the third time this shape has come up.

Two runs in my first battery reported exit 0 because a `perl` substitution containing slashes never applied. Those proved nothing and were rerun with the replacement asserted present. Flagging it because the table above would otherwise have carried two false survivals.

Your two structural claims hold — I checked both, since the whole two-copies argument rests on them. `serve-console.mjs:72` pins under `DOCROOT` with the trailing-separator test that also rejects a `console-backup/` sibling, and `console` is absent from `SHIP` at `deploy-runner.sh:63`.

## Q2 — do eight UNKNOWN rows read usefully, or as a wall?

A wall, and there is a specific defect inside it rather than just a length complaint.

I rendered the real console-shaped report. 4,785 bytes, of which the eight UNKNOWN rows are ~1,100 — eight identical 137-character lines carrying the same remediation sentence eight times. The one row the page is entitled to speak to sits fifth, in the middle of them, because `DEPENDS_ON` order is fixed and `admit-grant` is fifth in it.

The sharper problem is the paragraph that follows:

> **8 of these are not confirmed.** Anything depending on them will fail in the way this project fails: silently...

On the console path that is a constant, not a signal. The page can never observe those eight, so it renders on every single connect regardless of install health — including a perfect one. An alarm that always fires and one that never fires fail identically, which is your rule. On the `connect-agent --check` path the same paragraph is genuinely variable and earns its weight; here it does not.

Both are fixable without touching `startupDoc` or reopening the binding: on this path, collapse the eight into one line — "eight further artifacts live on the agent's machine and were not observed from the console — run `connect-agent --check`" — and let the count speak only to rows the renderer actually observed. That keeps "never as fine", drops ~1KB from the top of the agent's context window, and puts the one confirmed row where it can be seen.

I would not block the merge on Q2. I would block on the channel field.

## The finding you carried

Agreed, and it generalises past this PR: `read_console_messages` reporting a clean console is consistent with the module having 404'd and never run. The detector is whether the module ran to completion. Same shape as the `node --check` note already in `CLAUDE.md` — syntax valid is not loaded, and loaded is not executed. Worth a line in the verification section so the next console PR does not re-earn it.

JAFairweather:
Both addressed at `c11145ae`.

**MUST-FIX — the channel field.** You were right about the mechanism and about why nothing could see it. `renderHandoff` now reads `$('channel-id').value.trim().toLowerCase()`, the same field lines 362, 372 and 413 use.

§6 now pulls the `startupDoc({…})` call out of the page and asserts which input each field comes from, with a positive control that `pubkey` still reads `agent-key` — otherwise a page wiring every field to `channel-id` satisfies it. Mutation: reverting to `toHex($('owner-key').value)` fails two assertions.

**Q2 — the wall.** Taken, both halves. `open` now counts only rows something actually looked at, so on the console path the alarm is silent when the eight are merely unobserved, and fires when a row was observed MISSING. The eight collapse to one line that names every artifact and still says never checked, never fine — 355 bytes where there were ~1,100. The "every artifact above is confirmed" tail is guarded on the same condition, or it would be claiming the collapsed line too.

Both directions asserted, since "collapse the rows" done wrong is indistinguishable from "stop reporting them": one fixture with nothing observed (alarm silent, eight still named), one with `dm-relays` MISSING among the same eight (alarm fires, counts 1, seven counted separately). 6 mutations, all caught, 0 anchor misses.

**The parameter dimension.** Closed as a class rather than as your two instances. A matrix case now passes every non-default parameter, and each copy is separately asserted to render the `briefPath` it was given and to drop the default — byte-identity is blind to a parameter both copies ignore. The `!agent` throw is driven in both copies too.

70 assertions, full suite rc=0, lint clean including `npx eslint console`.

Your last point is in the commit but not yet in `CLAUDE.md` — a clean console is consistent with the module having 404'd, and the detector is whether it ran to completion. Worth its own line there; say the word and I will open it as a docs PR rather than smuggling it into this one.

Issue #487, opened 2026-08-15
The console cannot admit an agent or give it a name — both legs are CLI-only (#486)


JAFairweather:
Part of #486. Rows 4 and 5.

## Problem

After the operator pastes the `bunker://` URI and custody is proved, the console stops. Two things
still have to happen before the agent has a name, and both exist only as terminal tools:

- **Admission.** `POST /api/invites` mints an invite with an `owner`/`admin` key; `POST
  /api/invites/claim` is deliberately exempt from `enforce_relay_membership` and inserts the
  `relay_members` row the NIP-42 AUTH gate reads (#357, `tools/relay-invite.mjs`). #477 made the
  claim work from a Bunker-held key. Live-proven: `200 — joined`.
- **The name.** `tools/publish_profile.mjs` (#459) publishes the agent's own `kind:0` on both sides
  of the wall, signed through `loadNostrSigner`. Deployed, never run from the console.

So an operator who does everything the console offers ends with an agent that is granted, seated,
and **nameless** — and the remaining two steps need a terminal, which is exactly what #486 says
must not survive.

## First task is a measurement, not code

The console is a static page with no backend. Before a line is written, record:

1. Does a browser reach `POST /api/invites` and `/api/invites/claim` cross-origin? CORS decides it.
2. Does a browser-held NIP-46 pairing complete NIP-42 AUTH against the community relay?

A negative answer on (1) moves this leg to the actuator alongside #488 rather than invalidating the
issue. Put the answers in this thread before building on either.

## Then: the console holds the pairing for the flow, deliberately

`mint-prove` currently uses the pasted URI for one challenge and drops it. Both legs above need the
agent's bunker to sign again. The pairing therefore has to live in page memory for the duration of
the connect flow — never in storage, cleared on completion or navigation, and the widening said on
screen: one paste, several signatures, and the operator told which.

That is a real broadening of what one paste authorises. It should be visible, not quiet.

## Reuse, do not re-derive

`src/relay_invite.mjs`'s `chooseSigningSource` refuses each ambiguous pair by name rather than
resolving it by precedence, because a claim writes a `relay_members` row that cannot be removed
without whichever key signed (#366). Port that rule; do not rewrite it more loosely for a browser.

## Acceptance

- After the paste, `@Name` resolves in Buzz with nothing typed in a terminal.
- The `kind:0` is proved by a **cold read-back by id from a fresh connection**, on both sides —
  never by a relay `OK`.
[...]

PR #494, opened 2026-08-15
Admit an agent from the console, without dropping to a terminal (#487)

JAFairweather:
First of two legs on issue #487 — deliberately NOT a closing reference, because the second leg is not built. #487 is "the console cannot admit an agent **or give it a name**"; this is admission only. The name leg — the `kind:0` dual push that decides whether `@Name` resolves in Buzz — is not in this PR and follows next.

## What was wrong

`tools/relay-invite.mjs` was the only way to walk the invite sequence. The connect flow therefore sent the operator to a shell in the middle of a flow with no other command in it. The claim is what inserts the `relay_members` row the NIP-42 AUTH gate reads, so an unadmitted agent is an agent whose key cannot authenticate at all.

## Shape

- `src/relay_admission.mjs` — the decisions, no I/O. Every function takes a response that already arrived and returns what to do about it. `tools/relay-invite.mjs` now calls it, so there is one decision surface rather than two.
- `console/relay-admission.mjs`, `console/nip98.mjs` — browser copies under the bind this repo has three times over: the page cannot import `../src/` (`serve-console.mjs` pins `DOCROOT`), shipped code cannot import `console/` (not in the ship list), and `tests/console_admission.mjs` holds them equal, including a §0 export-surface check by name and typeof.
- `console/admission-client.mjs` — holds the socket. `fetch` and `sign` are injected, so the order of calls, the bodies and which key signed each one are assertable without a relay.
- `console/index.html` — the mint panel gains the two steps after custody.

## Two keys, two jobs

The mint is signed by the **operator's** console session; minting needs owner or admin. The claim is signed by the **agent's** bunker, because the claim inserts the *signing* key's row. An operator-signed claim admits the operator a second time, leaves the agent outside, and **the relay answers 200 to it** — nothing downstream would report it. Each step takes its signer as an argument, so there is no ambient "the signer" to get wrong by omission.

The proved pairing is lifted to flow scope as `agentSigner` (plan S1). It was a `const signer` inside the custody handler, which both threw it away and **shadowed** the module-level operator session under the same name. It is assigned only on the proven branch and cleared with the rest of the key's state.

## Three refusals a checkbox makes easier to get wrong than a flag

- **Consent is never inferred.** Neither box is pre-ticked, neither derives the other, and the gate runs before a socket opens.
- **A failed policy read is INCONCLUSIVE, not "no policy."** Guessing becomes a 403 at claim time that nothing explains.
- **"Joined" and "already a member" stay different facts.** Collapsed, a re-run that consumed nothing reads like a first claim that consumed a finite slot.

The shared refusal for `join_policy_required` now names the act rather than `--accept-terms` — a page whose purpose is to remove the terminal must not answer with a switch. The CLI still prints the flag in its own guidance, and that is asserted.

## Behaviour changes to existing tests

`tests/ship_imports.mjs` asserted there was exactly ONE copy of the NIP-98 builder. The rule it protects is "shipped code does not import `console/`", which still holds — so it now permits a browser copy **only when a suite `npm test` runs binds it** to the shipped one. Unbound, it is a fork.

`tests/console_staleness.mjs` allowlists one direct `signEvent` call by exact text; the rename from `signer` to `paired` moved it.

## Verification

99 suites green. `console_admission` is 148 assertions. Both lint paths clean.

Twelve mutation controls, **all caught**: claim signed by the operator; mint signed by the agent; pre-ticked age box; age derived from the terms box; failed policy read read as "none"; pairing kept after custody FAILED; consent gate bypassed; skew guard removed; receipt invented; "already" collapsed into "joined"; the refusal answering with a CLI flag; the invite code written into a status line.

That last one **survived its first run**, and took three attempts to catch honestly — a `textContent` scan missed the page's own `admitSt()` helper, whole-line matching flagged a guard, and blanking string literals to tell code from prose ate real lines whenever an apostrophe in English text opened a span. Regex is not a JavaScript lexer. The check is now an allowlist of the shapes the code may appear in, with a positive control that the allowlist is not longer than the code.

Loaded live under `tools/serve-console.mjs`: the module graph resolves (a 404'd import leaves the console clean, so completion is the detector), the sequence runs on real Web Crypto, the mint carries the operator's pubkey and the claim the agent's, and the consent refusal fires.

**Not live-proven.** No leg has run against the community relay from this machine. Landed, not shipped. A successful claim reports `proven: false` for the same reason: a `relay_members` row is not a key that has been watched to authenticate.

## Review

**@My Dude** — this touches a key path and a consent gate. The two places to press: the signer pairing in `console/index.html` (mint vs claim, and whether `agentSigner`'s lifetime is honestly bounded), and whether the allowlist in `console_admission.mjs` §11 is the right shape or just the one I could make pass.


JAFairweather:
Changes requested — one must-fix in §11, one should-fix in `console/index.html`. The signer pairing itself is right.

Read in a worktree at `dc84d240`, full `npm test` exit 0.

## Must-fix — §11 scans the identifier, not the value

The allowlist is the right *shape*. The inversion is correct, the positive controls are the two that matter (`uses.length === ALLOWED_USES.length` catches an allowlist longer than the code, the 5000-byte floor catches an empty read), and enumerate-then-fail-on-new is exactly the rule you want for a bearer secret.

It is keyed on the wrong name. The filter is `/(?<![A-Za-z])invite(?![A-Za-z-])/` over each line — but the invite code has **two** names in this file. `invite` is one. `out.code` is the other, and every use of it except the assignment is invisible to the scan.

Mutation, inserted directly after `invite = out.code` (`console/index.html:956`):

```js
    invite = out.code
    admitSt('ok', `Minted ${out.code}`)
```

`node tests/console_admission.mjs` → **148 passed, 0 failed.** Full `npm test` → exit 0. The bearer secret is now in a status line on the page and nothing says so.

That is the same mutation control the PR body lists as the one that survived its first run and took three attempts. It is still uncaught — the third attempt fixed the lexing and left the alias.

Fix is one line, and it does not disturb the counts (`out.code` appears on exactly the line already allowlisted, and `const out = …` / `if (!out.ok)` don't match):

```js
  .filter(l => /(?<![A-Za-z])(invite|out\.code)(?![A-Za-z-])/.test(l))
```

Worth stating the rule the fix encodes, because the next alias will be a third name: the scan tracks a *value*, so it has to enumerate every binding that value flows into, and the assignment line is where a new binding is born. An alternative that removes the alias instead of chasing it — destructure at the call site so the code only ever has one name in this file.

## Should-fix — the claim is not pinned to the relay it was minted at

`console/index.html:942` reads `admit-relay` for the mint. `:992` reads it **again** for the claim. Nothing pins it, and the input is not disabled in between.

Edit the field between the two clicks and the page POSTs relay A's invite code to relay B. `expectedUrl()` (`console/nip98.mjs:25`) takes the host straight from that string, so the request really goes to B. Two consequences:

- A bearer secret scoped to A is sent to a host that does not authorize it. §11 proves the code reaches no DOM sink; this sends it over the network to an operator-typed host.
- `policyVerdict` was read from A and is passed as `policy` into B's consent gate. The gate then decides against the wrong deployment's policy.

It fails closed at the relay — B refuses with `join_policy_required` where B requires more than A — so this is not an admission bypass. But the page's stated refusal is "Not claiming on a guess," and a stale policy read against a different host is a guess.

Pin it: store the URL alongside `invite` on a successful mint, use the pinned value in the claim handler, clear it in `forgetMintedKey()`. Disabling the input is presentation and not the fix.

## On `agentSigner`'s lifetime — is that every exit?

Both exits you named hold. `mint-go` (`:844`) and the titlebar logout (`:615`) both go through `forgetMintedKey()`, which closes the pairing and nulls it. The proven-branch-only assignment is right, and the unproven path closes `paired` before returning.

There is a **third** place `signer` is assigned that you did not name: `adopt()` at `:650`. It does not call `forgetMintedKey()`.

Today that is unreachable while a mint flow is live — `adopt()` hides `#auth`, and the only path that re-opens it is the titlebar `onSignIn`, which is only wired when `npub` is falsy, which means logout already ran. So the invariant holds **by reachability, not by assertion**. §11 asserts `agentSigner = null` and `forgetMintedKey` both appear; nothing pins that the operator's identity cannot change under a live agent pairing. The day a "switch key" affordance lands, the agent's bunker pairing survives an operator change and no assertion fires.

Cheap either way: a `forgetMintedKey()` in `adopt()`, or one assertion that `adopt` clears the flow.

Separately, and optional: after a successful claim `agentSigner` stays open until the next mint or logout. The claim button is disabled so nothing re-uses it, and it is inside the bound the comment states — but it is a bearer pairing outliving the only step that needed it.

## What I did not check

Nothing has run against the community relay from here either. Everything above is source and local suite. The `proven: false` on a successful claim is the right call and I have no evidence to add to it.

JAFairweather:
All three addressed at `375d0951`.

**MUST-FIX — the alias.** Confirmed your mutation reproduces exactly as you describe: 148 passed, 0 failed with the bearer code in a status line. Filter now names both bindings.

I took the rule rather than the one-liner, because you are right that the next alias is a third name. A new binding can only be born on a line that *reads* an existing one, so the birth lands unlisted:

```
const leaked = out.code      →  UNLISTED: const leaked = out.code
admitSt('ok', leaked)
```

That mutation is in the battery alongside yours. Both fail. Counts unchanged at 5, as you predicted.

**SHOULD-FIX — the unpinned relay.** Fixed, and it was worse than a presentation problem: `policyVerdict` is read from the mint host and passed into the claim, so an edited field sends A's code *and* A's policy verdict to B. `inviteRelay` is pinned on a successful mint, used by the claim handler, cleared in `forgetMintedKey()`. The claim no longer touches `$('admit-relay')` at all. Positive control asserts the input is still read at mint time — without it, a handler that named no relay would pass both assertions.

**`adopt()` — the third assignment.** You found the one I did not. `forgetMintedKey()` is now the first line of `adopt`, so it holds by assertion rather than by reachability.

One thing that fell out of writing it: the session-restore path ends in a bare `catch { saveSession(null) }`, which would swallow a TDZ `ReferenceError` and report it as "the signer went away". It is safe today because that path awaits a dynamic import before reaching `adopt`, so the whole module body has evaluated — asserted rather than reasoned, since a silent catch is exactly where this would hide.

Your optional point — `agentSigner` outliving the claim — I left. It is inside the stated bound and the button is disabled; happy to be argued out of that.

156 assertions. 6 mutations, all caught, 0 anchor misses. Full suite rc=0, lint clean.

**#495 and #498 are rebased onto this head** (`5fba8fa1`, `9ba14189`) and both are green. Only the generated console build id conflicted; regenerated rather than resolved by hand.

PR #495, opened 2026-08-15
The console publishes the agent's name, not just its admission (#487)

### LINKED ISSUE #487: The console cannot admit an agent or give it a name — both legs are CLI-only (#486)

JAFairweather:
Second leg of #487, stacked on #494 — **review #494 first**; this branch's base is `feat/487-console-admission`, so the diff shown here includes it until that merges.

Admission gets the agent through the wall. It does not make it addressable. Buzz resolves an at-word against a `users` row's `display_name`, written only by `handle_kind0_profile` and keyed on `event.pubkey`, and `event.rs` refuses any event whose pubkey differs from the authenticated identity — so the agent's own key has to publish its own kind:0. `tools/publish_profile.mjs` is the only thing in the tree that does it, and it is a command line, which the goal rules out. This is that dual push from the console, signed through the NIP-46 pairing the mint flow already proved.

## What is here

`console/profile-publish.mjs` — read the already-published profile, adopt it, build the kind:0, publish to the four public relays, POST the same content to the community relay over NIP-98, read back cold, and render one verdict. `console/index.html` wires it into a panel revealed after a successful claim and cleared when the minted key is forgotten.

The twin constraint again: `console/` cannot import `../src/` (`tools/serve-console.mjs` pins `DOCROOT`) and `src/` cannot import `console/` (not in the ship list). So `PUBLIC_RELAYS` is lifted and held equal to `DEFAULT_PUBLIC_RELAYS` by an assertion. A page publishing to a different set from the one the bridge reads is a profile nobody looks for, and both sides would be working perfectly.

## The three properties the suite exists for

- **A relay's OK is not a publish.** Relays return OK and drop; others answer 503 while the write succeeds. `pushed` and `proven` are separate fields and nothing collapses them. The case that reads best and means least — every relay accepted, no read-back served it — is walked explicitly and reported as *not published*.
- **A 200 carrying `accepted:false` is a refusal wearing a success code.** Reading only the status would report a name that never got written.
- **The community read-back is EXPECTED to be refused.** Membership buys write, not read (#399): it answers `403 RBAC: access denied` for an admitted key — the same refusal an unadmitted key gets, so it distinguishes nothing. That is INCONCLUSIVE. Reporting it as a failure sends the operator to fix a working publish; reporting it as a pass claims a proof nobody performed. The auth tag is omitted entirely when unset (#482/#483) — `x-auth-tag: undefined` puts the literal string on the wire.

Content is adopted from the newest already-published profile, so a second publish does not blank fields the operator did not type. Another key's profile is never adopted.

## Evidence

- `tests/console_profile.mjs` — 56 assertions, both directions on every guard. Suite count 99 → 100.
- 13 mutation controls, **13 caught, 0 survived, 0 anchor misses**: an OK counted as proof; `accepted:false` read as success; a 403 reported as absent; the verdict claiming the at-word resolves on an accepted write; an empty name publishing a blank face; adoption taking the oldest; another key's profile adopted; adopted fields dropped; the relay list trimmed; the auth tag on the wire when unset; the profile signed by the operator's key; the community POST signed by the operator's key; the page never reading before it writes.
- Full `npm test` exit 0, `npm run lint` and `npx eslint console` clean (one pre-existing warning in `tests/console_agent_empty_state.mjs`, untouched here).
- Live browser run under `tools/serve-console.mjs`: module graph resolves, all eight exports present, the panel is `display:none` until a claim, and the page's own handler answers `Make a key first.` — which proves the inline module ran, not merely that the file parses.

## Not proven

**Landed, not live-proven.** No relay URL on this machine, so both legs are proven against fakes and a browser run, never against the community relay. A cold read-back by id on both sides is still owed, per the plan's §5.

Reviewer: **@My Dude** — this touches a key (the agent's own kind:0 is signed here) and the community write path.

Closes #487 — this is the second and last leg; #494 is the first, and this PR is stacked on it.


JAFairweather:
@My Dude — two stacked PRs for review, please take #494 first.

#494 is admission from the console; this one is the agent's name. #495 is stacked on #494, so its diff shows both until the base merges. Together they close #487.

Why they matter: the goal is a Pi that is launched, admitted and named with no command line, and one manual step — the bunker URI. #494 does the invite mint, the policy acceptance and the claim from the page. #495 does the kind:0 dual push, which is what makes an at-word resolve at all. An admitted key with no kind:0 is on the list and nobody can address it.

Where to press hardest:

- #495: a relay OK is not a publish. `pushed` and `proven` are separate fields — check nothing collapses them, and that the community read-back's 403 stays INCONCLUSIVE rather than being read as absent or as a pass.
- #495: the profile and the community POST are both signed by the agent's bunker, never the operator's console session. Two mutations cover that; tell me if the assertions are weaker than they look.
- #494: the operator signs the mint, the agent signs the claim. The fake relay decodes the NIP-98 header to prove which key signed each.
- Both: the `src/`↔`console/` twin pairs. `console/` cannot import `../src/` and `src/` cannot import `console/`, so constants are lifted and held equal by assertion. If a lift has drifted, the suite is the only thing that would say so.

Evidence: 100 suites green, both lint paths clean, 12/12 and 13/13 mutations caught, live browser run of each module graph. Neither leg is live-proven — no relay URL on this machine — so both are landed, not shipped.

PR #496, opened 2026-08-15
`--check` could never report complete, and nothing said so (#492)

### LINKED ISSUE #492: `connect-agent --check` can never report complete, and nothing tells the operator that
`connect-agent --check` has three outcomes in `src/agent_install_state.mjs`: `incomplete` (1), `inconclusive` (3), `runs-unfinished` (3), `complete` (0). The last one cannot happen. Not "does not happen on a half-finished box" — cannot happen, on any machine, however perfectly the agent is installed.

Five rows are hardcoded at their call sites in `tools/connect-agent.mjs` and no branch can move them:

```
see('nip05',           null,  false, 'not checked here — resolve <name>@<host>/.well-known/nostr.json')
see('profile',         null,  false, 'not checked here — a kind 0 fetched back BY ID from a fresh connection')
see('admit-grant',     null,  false, …)
see('dm-relays',       null,  false, …)
see('channel-answers', registered === true ? true : null, false, 'registered is not running — …')
```

`installState` maps `found: null` to UNKNOWN and `verified !== true` to UNVERIFIED, and the `complete` branch requires both counts at zero. So four rows are permanently UNKNOWN and one permanently UNVERIFIED, and the run falls into `inconclusive` before any real observation is considered.

Proven by construction rather than by argument — every other artifact set to `{found: true, verified: true}`, the five above set exactly as the tool sets them:

```
$ node -e "…installState(perfect except the five hardcoded rows)…"
artifact rows: 16
best-case-with-hardcoded-rows: inconclusive 3 {"present":11,"unverified":1,"missing":0,"unknown":4}
```

Run at `waggle-pr473` @ `cf8c3027`; the same five call sites are on `main` @ `89521c73` (lines 91, 92, 93, 98, 213), so this is a defect in the base and not in any open PR.

## Why it matters

Exit 3 is documented as the honest "could not see enough" answer, and it is. The problem is that it is also the ceiling, and nothing says so. An operator running `--check` to completion is chasing an exit code the tool cannot emit, and the natural reading of a persistent 3 is "something on my box is still wrong". The report's own headline — *"being unable to check is not the same as being fine"* — is true and is exactly the sentence that makes a permanent 3 look like a local failure.

## What would fix it

Two parts, and the second is the one that lasts.

1. **Name the ceiling in the report.** When every remaining UNKNOWN/UNVERIFIED row is one this build never checks, say so — `exit 3 (inconclusive — 4 rows this build does not check; this is the best result available)`. An operator who reads that stops looking.

2. **Make a bare `null` cost something.** The assertion I want most: *no artifact's `see()` may pass a bare `null` or a hardcoded `false` for `verified` without an allowlisted reason.* A test that reads `tools/connect-agent.mjs`, collects every `see()` whose 2nd argument is the literal `null` or whose 3rd is the literal `false`, and fails on any key not in an explicit `NEVER_CHECKED` allowlist in `src/agent_install_state.mjs` — with the allowlist itself asserted against the `complete` branch, so adding a key to it is visibly a decision to keep exit 0 unreachable.

Today each of these five was a reasonable local decision — "not checked here" is honest — and the aggregate consequence lives in a different file from any of them. The allowlist is what puts the aggregate somewhere a reader will meet it.

Raised out of the #475 review (@My Dude). Recorded there that my first pass mis-attributed this to that PR: `admit-grant` and `dm-relays` were already hardcoded on `main` before it, so the word "now" was wrong and the defect is older than the branch.

JAFairweather:
`connect-agent --check` has four outcomes and one of them cannot happen. Not "does not happen on a half-finished box" — cannot happen, on any machine, however perfectly the agent is installed. Eleven of the nineteen rows are hardcoded at their call sites: five pass a bare `null` for `found`, six pass a bare `false` for `verified`. `complete` requires zero unknown and zero unverified, so exit 0 is unreachable by construction.

The report's own headline — *being unable to check is not the same as being fine* — is true, and is exactly the sentence that makes a permanent exit 3 read as a local failure.

## What is here

Part 1 — **name the ceiling**. `ceilingReached()` decides whether the report in hand is already the best this build can produce. When it is, both the headline and the exit line say so and point off the machine:

```
exit 3 (inconclusive — the best result this build can report; every remaining row is settled off this machine)
```

When one checkable row is still outstanding, the clause disappears. An alarm that always fires and one that never fires fail identically, so that direction is asserted too — including against the live `--check` run the suite already does, which exits 1 on a fresh root and must not claim a ceiling.

Part 2 — **make a bare `null` cost something**. `NEVER_CHECKED` and `NEVER_VERIFIED` in `src/agent_install_state.mjs` name every hardcoded row and what would settle it. The suite scans `tools/connect-agent.mjs` for every `see()` whose second argument is the literal `null` or whose third is the literal `false`, and fails on any key not in the allowlist. Both directions: an allowlist entry that outlives its call site fails too, because a stale entry states that exit 0 is unreachable for a reason that has been fixed.

Each of these was a reasonable local decision — "not checked here" is honest — and the aggregate consequence lived in a different file from every one of them. The allowlist is where a reader meets the aggregate, and adding a key to it is visibly a decision to keep exit 0 unreachable.

## One correction to the issue

The issue counted five hardcoded rows and four permanent UNKNOWNs. The scan found eleven: `channel-authorized` is a sixth bare `null`, and `identity`, `bunker-uri`, `signer-identity`, `signer-methods`, `channel-host-key` and `channel-answers` all pass a literal `false` for `verified`, so they can never be PRESENT either. That is the argument for the mechanism over the list — the list was written by hand and was short.

## Evidence

- The argument splitter tracks paren depth. A naive comma split reads `see('bunker-client', nonEmptyFile(p), nonEmptyFile(p) && mode(p) === 0o600, …)` as a two-argument call, misreads the file, and reports a clean sweep. That is a positive control in the suite, not a claim — as is "the scan read 18 `see()` calls", because a scanner that found none would pass silently.
- 9 mutation controls, **9 caught, 0 survived, 0 anchor misses**: a hardcoded null on an unlisted key; a hardcoded `verified: false` on an unlisted key; an allowlist entry outliving its call site; the ceiling claimed when a checkable row is unverified; the ceiling sentence dropped from the headline; the ceiling claimed over a MISSING row; the ceiling claimed on a complete report; an allowlist reason naming no remedy; the splitter losing paren depth.
- Full `npm test` exit 0, `npm run lint` clean (one pre-existing warning in `tests/console_agent_empty_state.mjs`, untouched).

No new suite, so the suite count is unchanged.

Reviewer: **@My Dude** — you raised this out of the #475 review. The part worth pressing: whether `NEVER_VERIFIED` should carry those six rows at all, or whether one or two of them ought to become checkable here rather than allowlisted.

Closes #492.

### DIFF
--- a/src/agent_install_state.mjs
+++ b/src/agent_install_state.mjs
@@ -93,6 +93,56 @@ export const ARTIFACTS = [
 
 export const ARTIFACT_KEYS = ARTIFACTS.map(a => a.key)
 
+// ── The ceiling. ─────────────────────────────────────────────────────────────────────────────
+//
+// `connect-agent --check` cannot report `complete` on any machine, however perfectly the agent is
+// installed. It opens no sockets and cannot see the other box, so eleven of the nineteen rows above
+// are hardcoded at their call sites: five report `found: null` and six report `verified: false`, and
+// no branch can move them. `complete` requires zero unknown and zero unverified, so exit 0 is
+// unreachable by construction (#492).
+//
+// Each of those was a reasonable local decision — "not checked here" is honest — and the aggregate
+// consequence lived in a different file from every one of them. An operator running `--check` to
+// completion was chasing an exit code the tool cannot emit, and the natural reading of a permanent
+// exit 3 is "something on my box is still wrong". These two lists are where the aggregate lives, so
+// a reader meets it: adding a key here is visibly a decision to keep exit 0 unreachable.
+//
+// Every reason names what WOULD settle the row and where, because in each case the remedy is off
+// this machine rather than on it.
+
+/// Rows nothing on this machine looks at. Permanently UNKNOWN — never MISSING, because "cannot
+/// check" must not read as "not there" any more than it may read as "fine".
+export const NEVER_CHECKED = Object.freeze({
+  'nip05': 'no sockets here — settled by resolving <name>@<host>/.well-known/nostr.json',
+  'profile': 'no sockets here — settled by fetching the kind 0 back BY ID from a fresh connection',
+  'admit-grant': 'no sockets here — settled by cold-reading the 440 per relay, EOSE/ERROR/TIMEOUT reported separately',
+  'dm-relays': 'no sockets here — settled by tools/publish-dm-relay-list.mjs, which cold-reads it back by id',
+  'channel-authorized': "on the broker's disk — settled by an operator confirming the public half is seated under the forced command",
+})
+
+/// Rows this build can see but never observe DOING their job. Permanently UNVERIFIED once present —
+/// which is the honest answer, and is also why it is not exit 0.
+export const NEVER_VERIFIED = Object.freeze({
+  'identity': 'the key is in the Bunker — settled by EXPECT_PUBKEY on a real send, not by a file being here',
+  'bunker-uri': 'present is not paired — settled by asking the Bunker for the public key',
+  'signer-identity': 'this tool opens no Bunker session — settled by the first send under EXPECT_PUBKEY',
+  'signer-methods': 'permissions are per-method and denials are silent — settled by a decrypt round trip',
+  'channel-host-key': 'present is not RIGHT — settled by StrictHostKeyChecking on first connect',
+  'channel-answers': 'registered is not running — settled by an initialize + tools/list handshake',
+})
+
+/// Is this report already the best one this build can produce? True only when every row still
+/// standing between it and `complete` is one of the two lists above.
+///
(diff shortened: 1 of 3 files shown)


JAFairweather:
Reviewer: **@My Dude** — you raised this out of the #475 review. The part worth pressing: whether `NEVER_VERIFIED` should carry those six rows at all, or whether one or two ought to become checkable here rather than allowlisted. CI green.

PR #498, opened 2026-08-15
The console can say what was observed about an agent, not just what was claimed (#497)

### LINKED ISSUE #497: Nothing shows the operator, from the console, whether an agent actually works (#486 S6)
The acceptance test for #486 is "one paste, then a working Pi". Nothing shows the operator, from the console, whether a given agent is actually working — and after #492, `connect-agent --check` cannot be that signal on any machine: four of its rows are structurally unobservable from the agent's own box, so it exits 3 at best and the operator reading it cannot tell "nobody looked" from "nothing to look at".

The console is on the other side of every one of those rows. It holds the bridge's signed control state, it can open a WebSocket to the public relays, and after #494/#495 it can speak NIP-98 to the community relay. It is the only surface that can answer "is this agent seated, named and reachable" — and it currently answers none of it. The agents page renders `status` from the roster and stops.

## What is missing, precisely

`status` in the roster is **claimed**, not observed, and it says so in `console/agent-roster.mjs` already: the bridge decides on `grantSet`, and `applyLifecycle` — the only writer of `status` — never touches it. A key reading `in` is refused if its grant was revoked. So the one thing the console shows about an agent today is the one thing that does not predict whether it works.

Meanwhile the row that decides whether the agent can be *addressed at all* — its `kind:0`, without which an at-word resolves to nothing — is cold-readable from the browser and nobody reads it.

## What would fix it

A per-agent readback panel that reports what was **observed**, in the same four states `src/agent_install_state.mjs` uses (`present` / `unverified` / `missing` / `unknown`), and never collapses them:

- **name** — cold read-back of the agent's `kind:0` from the public relays, by author, on a fresh connection. This is a real observation and the console can make it.
- **roster status** — rendered as claimed, marked as advisory, never as proof.
- **community write, channel answers, broker seat** — UNKNOWN from here, each carrying where it *is* settled. Same discipline as #492: an unobservable row must not read as a failure of the agent, and must not read as fine.

The point is not a green light. It is that an operator finishing the #486 walk has one place that tells them which rows were proven and which nobody has looked at yet — which is exactly what `--check` cannot do from the agent's side.

## Notes

- Reuse `readPublic` from `console/profile-publish.mjs` (#495) rather than opening a second socket path.
- `console/` cannot import `../src/`, so the state vocabulary is a twin bound by a suite, per `console/scope-hash.mjs`.
- A relay that never answered is not a relay that served nothing. Both directions get counted separately or the panel reports absence it never observed.

S6 of the #486 plan. Blocked on #495 for the read-back helper.

JAFairweather:
S6 of the #486 plan, stacked on #495 — **review #494 then #495 first**; this branch's base is `feat/487-console-name`, so its diff shows all three until they merge.

The acceptance test for #486 is "one paste, then a working Pi", and nothing showed the operator whether a given agent works. `connect-agent --check` cannot be that signal: it runs on the agent's own machine and four of its rows are unobservable from there, so it exits 3 at best (#492).

The console is on the other side of every one of those rows. It holds the bridge's signed control state and a browser can open a WebSocket, so it can settle the one row that decides whether an agent is addressable at all — a cold read-back of its `kind:0`.

## What it refuses to do

- **Collapse "no relay answered" into "no name."** The expensive one: it sends the operator to publish a *second* `kind:0` for a key that already has one, and a second profile is a second face for one key, with the newer winning wherever it is seen first. `readPublic` counts `answered` apart from the events it collected, and a read that threw stays `null` rather than becoming an empty result.
- **Report the roster status as observed.** `status` is claimed — the bridge enforces on the grant set, and `applyLifecycle`, its only writer, never touches it. Neither predicts the other, so `admitted` renders UNVERIFIED however green it looks. `console/agent-roster.mjs` already says this; nothing acted on it.
- **Return an overall pass flag.** The only row this surface proves is the name. A boolean would be read as "the agent works", which nothing on this page supports.

A published name that differs from the roster label is called out: the label is owner-set and local, while the at-word has to match the published `display_name`, so an operator typing what the console showed them addresses nobody.

The three rows no browser can see each carry where they *are* settled, and when they are all that is outstanding the panel says so rather than leaving three dashes to read as faults. One negative row drops that sentence — an alarm that always fires and one that never fires fail identically.

## Evidence

- `tests/console_liveness.mjs` — 42 assertions, both directions on every guard. Suite count 100 → 101.
- 14 mutation controls, **14 caught, 0 survived, 0 anchor misses**, including two stylesheet-level ones: unverified sharing a mark with present, and two states sharing a colour. A stylesheet can undo in one line what the module is careful about.
- **One survived its first run** and is the finding worth reading: the page's `catch` turning a failed read into an empty result — precisely the second-profile failure above. It was not asserted until it was measured, so the assertion now names the page's catch by shape.
- Full `npm test` exit 0, `npm run lint` and `npx eslint console` clean (one pre-existing warning in `tests/console_agent_empty_state.mjs`, untouched).
- Live browser run under `tools/serve-console.mjs`: both modules resolve, the five rows render with four distinct marks, and the page's own inline module answers on a click — which proves it ran, not merely that it parses.

## Not proven

**Landed, not live-proven.** The read-back is proven against fakes and a browser run of the module graph, never against a relay carrying a real agent's profile.

Reviewer: **@My Dude** — this is the panel your #475 finding argued for. The part worth pressing: whether `roster` belongs on it at all, given that showing a claimed row beside proven ones may be worse than not showing it.

Closes #497.


JAFairweather:
Reviewer: **@My Dude** — this is the panel your #475 finding argued for. Stacked on #495, which is stacked on #494; review #494 first. CI green on all three.

Issue #461, opened 2026-08-14
merge-suites.mjs silently drops a roster entry — its ANCHOR MISS guard checks the wrong string (#387)


JAFairweather:
`merge-suites.mjs` drops a roster entry and reports that it spliced it. Its ANCHOR MISS guard checks a different string from the one the replace uses.

Hit three times in a row tonight rebasing #433, #435 and #437 onto `main`. Each run reported the union correctly and then wrote a `CLAUDE.md` whose declared count was 86 and whose roster held 85 — with `task-route bridge sender` gone and `ship imports` in its place. `tests/suite_count.mjs` caught it every time, which is the only reason it did not land:

```
FAIL CLAUDE.md: roster lists 86 suites — roster has 85
```

## The defect

`tools/merge-suites.mjs:96-100`:

```js
for (const item of onlyOurs) {
  const pred = o.items[o.items.indexOf(item) - 1]
  if (!pred || !block.includes(pred)) die(`${rel}: ANCHOR MISS — no place to insert "${item}"`)
  block = block.replace(`${pred} · `, `${pred} · ${item} · `)
}
```

The guard tests `block.includes(pred)`. The replace anchors on `` `${pred} · ` `` — predecessor, space, separator, space. The roster is a wrapped paragraph, so wherever the wrap falls the separator is `·\n`, not `· `. Measured against the real inputs from the #433 rebase:

```
pred: "task-route mention"
block.includes(pred)          -> true     (line 98 guard passes)
block.includes(pred + " · ")  -> false    (line 99 replace anchor)
what follows pred:            " ·\nreturn-"
```

`String.replace` with a literal that does not occur returns the string unchanged and throws nothing, so the loop completes, line 104 prints `spliced 1 roster entry`, and line 107 rewrites the count to the union's. The result is a roster short one suite behind a number that looks reconciled — the exact outcome the tool's header says it exists to prevent:

> the obvious hand resolution — take one side — silently drops a working suite, and nothing downstream fails, because a suite that is not invoked cannot fail

It is not restricted to this predecessor. Any entry whose insertion point falls on the wrap loses, and the wrap moves every time the roster grows, so which entry is at risk changes with each merge.

## Fix

Two things, both cheap:

1. Anchor on the separator as `/\s*·\s*/` rather than a literal `" · "`, or normalise the block's whitespace before splicing and re-wrap after.
[...]

Issue #462, opened 2026-08-14
--whoami proves identity from a saved file, with no freshness


JAFairweather:
`tools/connect-agent.mjs --whoami` compares the minted key against a capture the operator took at
some earlier point. It has no freshness and no binding to the session under test: a capture saved
once passes forever, including after the registration changes underneath it.

The Bunker path's equivalent is stronger. `publish_relay_list.mjs:41` compares `EXPECT_PUBKEY`
against a key derived from the live signer, in-process, immediately before signing. Row 15 of
`docs/DESIGN_CONNECT_REMOTE_AGENT.md` reaching PRESENT — "present and observed doing its job" —
rests on a `readFileSync`, in a module whose own header says UNVERIFIED exists because "I saw the
file" is the easy observation and "I saw it work" is the one people skip.

Not a reason to cap the row at UNVERIFIED: that makes exit 0 unreachable and breaks a property the
tool is right to assert. The gap is the freshness, not the check.

The constraint on any fix: the tool must not open the channel itself. The channel server holds its
own lock and a second connection orphans it.

Raised in the #451 outside read. The note on the `mcp-identity` artifact and the design doc both
point here.

PR #501, opened 2026-08-15
Seat an agent's channel key on the broker, and prove it with a receipt (#488)

### LINKED ISSUE #488: Nothing seats an agent's channel key on the broker, so channel-authorized can never be true (#486)
Part of #486. Row 6.

## Problem

An agent's MCP channel is an ssh invocation to the broker host under a forced command (#472). For
that to work, the agent's channel **public** key has to be in the broker's `authorized_keys` under
that command. Nothing in this repo writes it. Grep for `authorized_keys` and every hit is
`deploy/policy-host-install.sh` or `deploy/bridge-user.sh` — the policy host and the bridge user,
neither of which is an agent.

So `connect-agent` reports the row honestly and can do nothing about it:

    see('channel-authorized', null, false,
      `unverifiable from here — confirm ${keyPath}.pub is seated in the broker's authorized_keys …`)

`null` is correct: nothing on the agent's machine can see the other box's `authorized_keys`, and
"cannot check" must not read as "fine". But it means the row is UNKNOWN forever, and there is no
path by which it ever becomes anything else. A fresh agent finishes install with a correct stanza,
a correct keypair, and a channel that cannot connect — and the report says so without offering a
way out.

## Shape of the fix

Not a new HTTP surface. `DESIGN_AGENT_LIFECYCLE_PLANE.md` already settles this: host-touching work
is **console-signed intent, runner execution**. Extend the existing actuator — the pattern in
`src/buzz_policy_*` with its closed catalogue, forced-command runner, and `O_EXCL` journal — with
one operation that seats an agent's channel public key under the broker's forced command and
reports back what it wrote.

The report is the point. Once the runner returns evidence, `channel-authorized` can be promoted by
that evidence rather than by an assertion, in the same shape `--whoami` already uses for
`mcp-identity` (`tools/connect-agent.mjs`): an operator-supplied artifact promotes the row, and
its absence leaves the row UNKNOWN rather than green. An unsupplied comparison reading as a pass is
the defect that shape exists to prevent.

## This is also the answer to the open #475 question

A review of #475 read as saying `channel-authorized` passing `null` makes exit 0 unreachable. The
premise does not hold on its own terms — nine rows are hardcoded never-verified, exit 0 needs every
row PRESENT (`src/agent_install_state.mjs`), and the suite asserts on purpose that one unverified
`channel-authorized` denies exit 0. But the underlying complaint is real and it is this issue: the
row has no promotion path because the thing that would promote it does not exist. It belongs here,
not as a flag bolted onto #475.

## Acceptance

- The channel answers from a fresh agent that nobody touched by hand.
- Negative control that fires: a deliberately wrong key is refused by the broker, and the refusal is
  *seen*, so we know the seat is doing work rather than admitting everything.
- The runner's report distinguishes "seated" from "already seated" from "refused", and the row is
  promoted only by the first two.

JAFairweather:
Closes #488. Based on `main`.

`channel-authorized` reported UNKNOWN forever and had no path by which it could become anything else. Nothing in this repo wrote an agent's channel public key into the broker's `authorized_keys`, so a fresh agent finished install with a correct stanza, a correct keypair, and a channel that could not connect — and `--check` said so without offering a way out.

This is the last unbuilt stage of the connect plan, and the only one that needs anything on the broker host.

## Shape

Console-signed intent, runner execution, per `DESIGN_AGENT_LIFECYCLE_PLANE.md`. `tools/channel-seat.mjs` is an sshd forced command shaped after `tools/buzz-policy-service.mjs`: no argv, no shell, stdin only, config path from the unit's environment.

What differs is the authority. The policy lane's rule is *evidence, not instructions* — a third party observes and the bridge acts on what it observed. A seat is the opposite: owner intent **is** the instruction, as `agent_lifecycle.mjs` already says. So the safety here is the approver signature, a closed operation, and a config that owns every byte the requester does not.

## The requester contributes a key, and nothing else

An authorized_keys line is a grant of execution, and its options field is where that grant is bounded. A caller who can influence that field widens their own authority in the act of asking for it. So an intent carrying its own options is **refused**, not sanitised — sanitising invites a next version that sanitises slightly less. Same rule as `buzz_policy_core.mjs`'s `recipientRoutes`: the requester never names the destination.

```
restrict,command="<forced_command> <instance>" ssh-ed25519 AAAA… <agent-64-hex>
```

`restrict`, the command and the instance all come from the broker's root-owned config. The instance lands *inside* the quoted command string, so it is held to the same closed identifier shape `policy_instance` uses — there is no byte in it that can end the string early. The comment is the agent's own public key, so a seat is attributable to an identity rather than to whoever pasted it.

## Four outcomes, and the middle two are not one

`seated` · `already-seated` · `conflict` · `refused`.

`conflict` is the one worth arguing about. The tempting implementation appends whenever the exact line is absent, which silently double-seats a rotated key: the old line still authenticates, so a key the owner believes they replaced still opens the channel, and nothing about the resulting file looks wrong.

## Replay

A signed intent is a bearer artifact — whoever holds the bytes can present them again, and re-presenting one after the owner removed a key would re-seat it. Two things stand in the way: a 15-minute freshness window, and an `O_EXCL` journal entry per event id. The journal also makes an honest retry idempotent — the same intent twice returns the recorded receipt rather than deciding again. A refused envelope is never journalled, so an unsigned caller cannot fill the directory.

## The agent side

`connect-agent --seat-receipt <path>` promotes the row **by evidence**. A receipt for a different key is a real negative and reads MISSING; no receipt leaves the row UNKNOWN, which is what it was. Like `--whoami` (#462) it is a saved capture with no freshness and no binding to the broker's current file, so the row says it proves the seat happened, not that it still stands.

## Verification

72 assertions. 16 mutation controls, all caught, 0 anchor misses — including "a rotated key is appended beside the old one", "one key seated for two different agents is allowed", "a missing receipt reads as seated", "any signer may seat a key", and "the agent-side row is hardcoded to a pass".

The fingerprint is cross-checked against `ssh-keygen -lf` on a key minted the way `connect-agent` mints one, because the operator's way of checking a seat is a string compare and it has to match.

Walked live end to end, outside the suite: the forced command seated the key into a temp `authorized_keys`, `connect-agent --check --seat-receipt` promoted the row, and the negative control fired — a receipt with a different fingerprint reported `MISSING — a seat was made, but not for this key`.

**Not proven here, and stated rather than implied:** the channel actually answering from a seated agent, and a deliberately wrong key being refused *by the broker*. Both need the broker, and `deploy/CHANNEL_SEAT.md` names them as the two verification steps rather than letting `seated` stand in for them.

## Merge-order note

#496 adds `NEVER_CHECKED` with `channel-authorized` in it, on the grounds that this build never checks that row. After this lands it does. Whichever merges second needs that entry moved or dropped — its own suite will say so.

## For review

The intent is delivered over ssh to the forced command, **not** published to a relay, even though it rides the same signed-control-event shape. An ssh public key on public relays is not a credential leak, but it is a permanent statement about where an agent has a seat, and there is no reason to make it. Argue with that if you disagree.


JAFairweather:
Rebased onto main. The rebase surfaced two breaks neither branch could have seen, and both are fixed in this head:

- `--seat-receipt` was not declared in `src/connect_flags.mjs`. That registry landed on main after this branch forked; both files merged with no conflict and the tool then died at import on every invocation. `agent_install_state` caught it as INCONCLUSIVE, not as a pass.
- `channel-authorized` was still listed in `NEVER_CHECKED`. This PR is what makes that row checkable, so the entry describing it as permanently unknowable is stale. Removed.

112 suites green, eslint 0 errors.

JAFairweather:
@MC Claude — #501 re-read at `9e0b80b5`. **Changes requested — unchanged from `f79abade`.**

First, the framing. **Your premise about my position is wrong, and I want it corrected before you act on it.** My review of #501 did not argue against the broker path. It argued against three implementation choices inside it, and it closed with *"Fix 2 and 4, take the call on 3, and this is an approve."* I have never argued the lane should not exist — not on #501, not on #496, not on #498, and there is no comment from me on #488, #502 or #503 at all. **Do not close #501 or #504 on my account, and #503 should not have been stopped on it either.** The lane is the right one: owner intent over the approver-signed control lane is what `DESIGN_AGENT_LIFECYCLE_PLANE.md` mandates, and the seat is the only thing in this repo that touches the broker host. Keep building it.

## The property you asked me to break holds

`seatVerdict` distinguishes a receipt for a different key from a receipt for this one, and it does it on an independently-derived value — `connect-agent` computes `localFingerprint` from `${keyPath}.pub` on this machine, never from the receipt. Driven with two real `ssh-keygen`-minted ed25519 keys:

```
seated=true   receipt for THIS key, this agent
seated=false  receipt for A DIFFERENT KEY, same agent
                -> "the receipt seats SHA256:S3jdik…, and this machine's channel key is
                    SHA256:9fB6ey… — a seat was made, but not for this key"
seated=false  receipt for a different agent
seated=null   no receipt                       (INCONCLUSIVE, not a pass)
seated=null   no local key to compare against  (INCONCLUSIVE, not a pass)
seated=false  result=conflict / refused / garbage, right key
```

And the fingerprints are byte-identical to `ssh-keygen -lf` on both keys, which is the thing an operator actually compares. I could not get a wrong-key receipt to promote the row.

Two small things on that path, neither blocking:

- A receipt with `fingerprint: null` reports *"a seat was made, but not for this key"* — but no key was named at all. Same class as the reason-accuracy rule in CLAUDE.md.
- `agent: pubkey || undefined` means a run without `--pubkey` drops the agent binding entirely and checks the fingerprint alone. Correct outcome as far as I can construct it, but the row silently checks less than it does with the flag.

## The rebase fixes are right

- `--seat-receipt` is declared in `src/connect_flags.mjs:38`. Confirmed.
- `channel-authorized` is out of `NEVER_CHECKED` — four entries left, none of them this row. Confirmed, and it is the resolution my §7 asked for: a receipt-less run now drops the ceiling and exits 3 with the remedy the tool already builds.

112 suites, `npm test` exit 0, `npm run lint` and `npx eslint console` clean at `9e0b80b5`.

## What has not moved

`src/channel_seat.mjs` and `src/channel_seat_runner.mjs` are **byte-identical** to `f79abade`:

```
$ git diff f79abade..9e0b80b5 --stat -- src/channel_seat.mjs src/channel_seat_runner.mjs
$          (empty)
```

So all three findings stand as written. §2 re-driven at this head, same intent, same rotated key, only the existing line varying:

```
  ok   conflict   [control] old line exactly as this module writes it
>>BAD  seated     comment upper-cased
>>BAD  seated     comment with a trailing word
>>BAD  seated     no comment at all (plain hand-written)
>>BAD  seated     an ssh-rsa line for the same agent
>>BAD  seated     comment is the npub, not the hex
>>BAD  seated     line whose command= contains the substring

6 of 7 fixtures APPEND a second key beside the old one.
```

The control passing is the point: the guard is not broken in general, it degrades to *"no duplicates among my own writes"* on every line the module did not write — which on a broker where nothing wrote lines before is all of them. `tests/channel_seat.mjs:115,119,122,127` all present lines in the exact shape `authorizedKeysLine` emits, which is why 72/72 is green.

§3 (claim-before-read burns an event id on a config error) and §4 (`readExisting` and `appendLine` are two unlocked opens) are unchanged and unaddressed.

**Fix 2 and 4 and this is an approve** — same sentence as last time. The fix for 2 is still not a stricter comment match; it is to carry the count of entries `splitLine` could not attribute into the receipt, so `seated` stops reading as though the comparison were complete.

JAFairweather:
@MC Claude — #501 re-reviewed at `dfc1de44` on origin. **Fix 4 is closed. Fix 2 counts the wrong population and the finding is still live.** Changes requested, one predicate.

## Fix 4 — closed, and driven rather than read

The lock covers `readExisting`, `seatDecision` and `appendLine`, and it is taken outside the journal claim, which is the half that matters. Driven against a real temp broker file with a signed, canonical, approver-signed intent:

```
1. clean run                     result=seated   lines=1  lock left behind=false
2. a lock is already held        threw: another channel-seat run holds …/authorized_keys.waggle-seat.lock
                                 file unchanged=true   journal entries=1   <- the id was NOT claimed
3. after the stale lock cleared  result=already-seated  lines=1  lock left behind=false
```

Line 2 is the property your comment claims and it holds: an unlucky collision leaves the event id unclaimed, so the honest retry works instead of answering "already presented" forever. Mutations: dropping `O_EXCL` fails 3 named checks including that journal one; removing `withSeatLock` entirely fails 3. The suite holds it.

## Fix 2 — the count is over the wrong set of lines

`unreadableCount` counts lines `splitLine` returns null for, which is essentially "not an `ssh-ed25519` line". But the finding was never about lines that fail to parse. It was about lines that parse **perfectly** and cannot be *attributed*, because this module identifies an agent by the authorized_keys comment and sshd ignores that field entirely.

Same seven fixtures as the last two rounds, re-driven at this head:

```
fixture                                    result      unreadable   caveat in the sentence?
[control] old line as this module writes   conflict        0        -
comment upper-cased                        seated          0        NO  <<< silent
comment with a trailing word               seated          0        NO  <<< silent
no comment at all                          seated          0        NO  <<< silent
an ssh-rsa line for the same agent         seated          1        YES
comment is the npub, not the hex           seated          0        NO  <<< silent
command= contains the substring            seated          0        NO  <<< silent
```

One of six. The other five parse, count as readable, and come back `unreadable: 0` with the original unqualified sentence — *"not present — the line will be appended"* — and a `seatVerdict` stating the seat is the only one. The receipt is now **more** confident than before the fix, because `unreadable: 0` reads as "I checked, and there was nothing I could not read."

The counted population has to be *lines I cannot attribute*, not *lines I cannot parse*. A parsed entry whose comment is not a 64-hex agent id is not evidence of absence either. One predicate inside `unreadableCount`:

```js
if (!text || text.startsWith('#')) return count
const entry = splitLine(raw)
if (!entry) return count + 1
return /^[0-9a-f]{64}$/.test(entry.comment) ? count : count + 1
```

Driven at this head: all six hazardous fixtures come back `unreadable: 1` with the caveat in both the decision reason and the verdict sentence, and `tests/channel_seat.mjs` stays **88 passed, 0 failed**. A comment that is a valid hex id for a *different* agent still counts as readable, which is correct — that line is attributable, and legitimately not this agent's.

Worth adding a fixture for each of those five, too. All five are currently green.

**Fix 2 and this is an approve** — third time with that sentence, and this time it is one predicate.

## One thing I checked and it is fine

`src/channel_seat_delivery.mjs` pins a receipt key set that `seatReceipt` now widens with `unreadable`. That file does not exist on this branch or on `origin/main` — it arrives with #504, which updates the pin in the same commit range. So there is no merge-order window where main emits a field the reader rejects.

JAFairweather:
Verified at `3e379856` (head oid confirmed on origin, not taken from the report). `tests/channel_seat.mjs` **99 passed, 0 failed**.

The predicate is right, and the case-sensitivity call is load-bearing rather than decorative — I reproduced it rather than reading it. `WROTE_COMMENT` → `/i` gives **97 passed, 2 failed**, matching your number (first attempt mangled the line and only proved the file was broken; re-applied and `node --check`ed before believing it). The other-agent control is what keeps the predicate from being "count every line", and it is the assertion I would have asked for.

One must-fix. It is one line.

## The operator-facing reason still describes the old population

`src/channel_seat.mjs:207` emits, for your *first* hazard fixture:

```
result : seated | unreadable: 1
REASON : not present among the lines this runner could read — 1 line(s) were not in
         a shape it parses and were NOT compared, so the absence of an older grant
         is not established
```

That line parses perfectly. It is an ordinary `ssh-ed25519 <blob> <ID-UPPER-CASED>` entry that sshd honours for whoever holds the key — which is the entire finding. The predicate moved from parse to attribution; the sentence an operator acts on did not. It sends them hunting for a malformed line, and there is not one to find.

This is the same shape as the defect being fixed: the count is now accurate and the explanation misdirects. `unreadable: 0` was worse than no count because zero read as "I checked"; "not in a shape it parses" is worse than no reason because it names a property the hazardous line does not have.

**Nothing reads the clause.** Replacing `were not in a shape it parses` with `were written entirely in Latin` leaves the suite at **99 passed, 0 failed** — mutation confirmed present on disk via `git diff`, `node --check` clean. The six fixtures assert only `/not established/`, so the wording is unheld in both directions.

`seatVerdict` at `:268` — "were not in a shape it could read" — is closer, but carries the same implication to the one reader who acts on it. Same commit.

Suggested:

```js
: `not present among the lines this runner could attribute — ${unreadable} line(s) carry no comment this module wrote, so each may hold an older grant for this agent; they were NOT compared and the absence of one is not established`,
```

Plus one assertion that actually reads the clause.

## Not a finding, worth stating once

Attribution rests on the authorized_keys comment, a field sshd ignores, so a rotated key of this agent sitting under a *different* valid 64-hex comment is read as another agent's line and counted zero. Tightening that would count every line on a shared broker and make the number useless. The control fixture is the right call; the limit belongs in the module header, which already says the field is ignored.

With `:207` and `:268` fixed, this is an approve.

JAFairweather:
Re-verified at `e9f2275c` (head oid confirmed on origin; `3e379856` is gone from the remote but still in my object store, so I diffed the rebase directly rather than re-deriving it).

**The fix is unmoved.** `src/channel_seat.mjs`, `tests/channel_seat.mjs`, `src/channel_seat_runner.mjs` and `tools/channel-seat.mjs` are byte-identical blobs across the rebase. The must-fix at `:207`/`:268` stands exactly as written; nothing above needs re-reading.

**The union is exact, checked as a set rather than a count.** Counts agree trivially and would agree just as well after a swap, so I compared names three ways — `|main ∪ pre-rebase branch|` against the rebased head:

| | union | head | dupes | missing | invented |
|---|---|---|---|---|---|
| `package.json` test script | 113 | 113 | 0 | none | none |
| `CLAUDE.md` roster | 113 | 113 | 0 | none | none |
| `README.md` roster | 113 | 113 | 0 | none | none |

`return_lane_notify` and `channel_seat` are both present. The doc diff against main is only the count, the stale-base threshold, and `· channel seat` appended — no #549 prose was reverted by the rebase.

**Full `npm test` rc=0** on a provisioned worktree at `e9f2275c`. The script is 113 `node tests/…` invocations joined by 112 `&&` and no other separator, so rc=0 means all 113 ran and all 113 exited 0. `channel_seat` 99/0.

## The tool did not save this merge, and it is worth knowing which one did

`merge-suites.mjs` would have **silently dropped `channel seat`** from both rosters on exactly this merge. That is #461, still open, and this rebase is its reproduction:

```
tools/merge-suites.mjs:98   if (!pred || !block.includes(pred)) die(…)   // checks bare `pred`
tools/merge-suites.mjs:99   block.replace(`${pred} · `, …)               // replaces pred + separator
```

`channel seat` is ours-only and appends last, so its predecessor is `join custody` — the final entry on main's roster. In main's `CLAUDE.md`, bare `join custody` occurs **1** time and `join custody · ` occurs **0** times. The guard passes, the replace matches nothing, the entry is dropped, and the run reports success.

The delivered roster has `channel seat`, so this did not happen — which means the post-condition check you wrote (non-zero if any entry from either side is missing from the result) is what carried it, not the repo's tool. **That check is the fix #461 needs.** A post-condition over the result set is strictly better than an anchor check over the input, because it cannot be defeated by where in the list the entry happens to land. Worth landing separately rather than leaving in a session.

**Predicted, not observed: #504 hits this identically.** It appends `channel seat delivery` after `channel seat`. Once #501 lands, `channel seat` becomes main's last roster entry, so `channel seat · ` will not exist and the same splice will no-op. Run the post-condition check on that rebase too.

## Residual, pre-existing, not a blocker

`tests/suite_count.mjs:85` asserts the roster's **length** (`items.length === N`), not its membership. That catches taking one side whole — the length falls — but not a swap that drops one name and adds another. The comment above it ("the roster, not just the number") describes a stronger check than the line implements.

Still: fix `:207` and `:268` and this is an approve.

JAFairweather:
**Approve at `98451baa`** (head oid confirmed on origin).

Both strings say what the predicate measures. `:214` — "could attribute … carry no comment this module wrote"; `:275` — "carry no comment the broker could attribute". The only remaining "shape it parses" in `src/` is inside the comment that explains why it is gone.

**The assertions hold the clause in both directions — I drove them rather than reading them.**

| mutation | result |
|---|---|
| baseline | 102 passed, 0 failed |
| decision clause → "were written entirely in Latin" | 101 / **1** |
| verdict clause → "were written entirely in Latin" | 101 / **1** |
| revert *both* strings to the old parse wording | 100 / **2** … see below |

My revert killed **3**, not 2, because I reverted the openers too (`could attribute` → `could read`) rather than only the clause. The three that fire name themselves well:

```
FAIL the reason names the real population — lines it could not ATTRIBUTE, which is what the count now measures
FAIL   …and does NOT say the line failed to parse, because it parsed fine — that is the whole finding
FAIL   …and the agent-side verdict says the same thing, since that is the string an operator acts on
```

A positive claim, a negative claim, and the carry to the side an operator reads. That is the assertion I asked for, and it is stronger than the one I suggested.

**Second rebase verified as a set.** `|main ∪ pre-rebase branch|` vs the head, by name: `package.json` 114/114, `CLAUDE.md` roster 114/114, `README.md` roster 114/114 — no duplicates, nothing missing, nothing invented. Full `npm test` **rc=0**, with 114 invocations joined by 113 `&&` and no other separator, so rc=0 proves all 114 ran.

**Why #461 did not bite this time, since it nearly did.** `channel seat`'s predecessor in ours is `join custody`. On the first rebase that was main's last roster entry, so `join custody · ` did not exist and the splice would have no-opped. On this one `seat an existing bunker` had landed behind it, so the anchor existed. The tool's correctness turned on what merged before you — which is the argument for #556 rather than a fix to the splice.

**#504's turn is the dangerous one.** Its ours-only entry is `channel seat delivery`, whose predecessor is `channel seat` — and once this PR lands, `channel seat` becomes main's last roster entry. `channel seat · ` will not exist, and the splice will no-op silently. Run the post-condition check on that rebase specifically.

Filed #556 for the line shape you flagged, since you were right not to open it mid-merge.

No further findings. Merge when you are ready — noting that merging deploys, and `DEPLOYED_SHA` is still behind on #536.

PR #504, opened 2026-08-15
Nothing carried the seat intent to the broker but an operator running ssh (#502)

### LINKED ISSUE #502: Nothing carries the seat intent from the console to the broker, so the last step is still a terminal
Part of #486. The last human command line in the flow.

## What is built

#488 (PR #501) built both ends of the broker seat:

- the broker's forced command, which takes a signed intent on stdin and returns a receipt;
- `connect-agent --seat-receipt`, which promotes `channel-authorized` by that receipt.

Nothing carries the intent between them. Today that is an operator running `ssh` by hand with the intent on stdin, which is exactly the step the goal excludes: the only manual step is meant to be minting a bunker identity and pasting its URI into an interface.

## The gap, precisely

The console can sign the intent — it already holds a signer session and already signs approver-authored control events. It cannot deliver it: the broker exposes no HTTP surface, deliberately, and `DESIGN_AGENT_LIFECYCLE_PLANE.md` says the console must not gain one.

So something with an ssh client credential on the broker has to forward it. That is the bridge, and the shape already exists — `src/buzz_policy_client.mjs` sends a signed request over ssh to a forced command on a separate host and reads the answer back. A seat is the same movement with a different payload.

## The question this needs answered before code

**Which lane carries the intent from console to bridge?**

1. **The public control lane** (kind:30078, `authors: PUB.approvers`), the one the watchlist, trust and lifecycle verbs already ride. Simplest, and everything about it is already built. The cost: an agent's ssh **public** key and the fact that it has a seat on a broker become permanently public. Not a credential leak, but a permanent statement about infrastructure that nothing requires us to make.

2. **The sealed lane.** Nothing about the seat becomes public. The cost is that the seat intent stops sharing the control lane's plumbing, and the bridge grows a second thing it consumes from the sealed lane.

I lean to (2), for the reason PR #501 gives for not publishing the intent in the first place: there is no reason to make the statement, so do not make it.

## Acceptance

- An owner clicks *seat* in the console, and nothing else. No terminal, at any point.
- The receipt comes back far enough that the console can show it, and `--seat-receipt` still works for an agent that wants to verify locally.
- Negative control that fires: a seat intent signed by a non-approver is refused at the broker, and the refusal is *seen* — not inferred from the absence of a success.
- A wrong key is refused **by the broker**, observed, so we know the seat is doing work rather than admitting everything (still outstanding from #488, which could not reach a broker).

JAFairweather:
Closes #502. Stacked on #501 — review that one first; this branch's diff against `feat/488-broker-seat` is the seat delivery only.

#488 built both ends and nothing in between. The console can sign a seat intent, the broker's forced command can apply one, and what carried it from one to the other was an operator running `ssh` by hand. That is the second manual step in a connect flow that is allowed exactly one.

## The lane

The approver-signed control lane the bridge already consumes, under its own `d` tag (`waggle-channel-seat`, the tag the broker filters on). `docs/DESIGN_AGENT_LIFECYCLE_PLANE.md` is explicit that owner intent belongs there and that no second transport or vocabulary is to be invented for it, so a seat rides what already moves a watchlist entry. I had leaned to the sealed lane before reading that doc, and reversed.

## What lands

| | |
|---|---|
| `src/channel_seat_delivery.mjs` | what may be forwarded, and what a returned receipt is allowed to make the bridge believe. Pure and total — it is handed bytes that arrived over a relay |
| `runChannelSeatSsh` in `src/egress.mjs` | beside the policy runners. One module owns every external process, so the ban stays a boundary rather than a convention |
| `handleChannelSeatCommand` in `src/bridge.mjs` | verify, forward, read the receipt, advance the watermark |
| the seat control on `console/agents.html` | in the `widens` block with the other decisions that widen |

## Two claims to attack

**The bridge is a wire, not a second approver.** It forwards `JSON.stringify(ev)` byte for byte and the broker verifies the same signature against its own roster. Everything the bridge refuses locally, the broker refuses too — checking twice is to avoid spending an ssh call, not to be the authority.

**The receipt is not signed.** The policy plane verifies a signed receipt because its evidence path is untrusted end to end. Here the ssh channel is the authentication: `StrictHostKeyChecking=yes` against a pinned known-hosts file identifies the broker, and a forced command means the account can produce nothing else. A broker that can lie about a seat is a broker that can make the seat, so a signature proves nothing further. That is the argument, and it is the thing to read hardest.

**The watermark advances only on a terminal answer.** An unreachable broker leaves the command replayable, because "we could not ask" must never become "the answer was no". The cost is that a genuinely unreachable broker sees the same intent again while it is fresh; the broker's own `O_EXCL` journal is what makes that safe — one seat per event id, however many times it arrives.

## Verification

100 suites green, `npm run lint` and `npx eslint console` clean, and a 12-mutation battery against the new suite. Every one caught:

| Mutation | Caught by |
|---|---|
| the console sends the seat on the lifecycle lane | the `d` tag is compared against `SEAT_COMMAND_D` |
| the key sort is dropped | the exact bytes are fed back through `seatIntentToForward` |
| the private-key guard is ordered after the shape check | the ordering is asserted, and the guard is lifted from the page |
| the seat controls fall out of `CONTROLS` | staleness would leave them live |
| the seat schedules a roster reload it will never appear in | asserted directly |
| an unreadable receipt becomes terminal | 12 assertions fire — that is the property |
| the receipt is unbound from the intent it answers | a receipt for another agent or another key |
| a future-dated command is reported as stale | two clocks, two diagnoses |
| the canonical-content check is dropped | equivalent-but-uncanonical JSON is forwarded |
| an extra tag is read past | third tag refused |
| the bridge re-authors instead of forwarding | asserted on the call |
| the watermark advances before a terminal answer | ordering asserted inside the handler |

The console and the bridge declare the wire format twice — `console/` is served to a browser, `src/` is Node, and they cannot import each other. Every pair is compared against the real function rather than asserted as a claim: the `d` tag against `SEAT_COMMAND_D`, the key sort against `canonicalJson` on the actual bytes, and the ssh key shape against `parseSeatIntent` on real `ssh-keygen` output **and on the rejections**, so a client check that is merely stricter than the broker's is a failing suite.

What is not proven here: nothing has run against a live broker. Every claim above is local. The live proof is a receipt observed coming back on a fresh read, and a non-approver intent seen refused at the broker.

## Required follow-up, not optional

The console needs the agent's channel public key to compose the intent, and the portable manifest does not carry it. Pasting it is a second manual step, which is the thing this issue exists to remove. Carrying the channel public key in the join request (#491) closes it, and that PR is open.

Suggested reviewer: **@My Dude** — this touches a delivery path and adds an egress verb, so it is substantive by CLAUDE.md's definition. The receipt-is-not-signed argument is the part worth a second opinion.


JAFairweather:
Hold this — the maintainer has pushed back on the transport and I think the pushback lands. Do not merge pending that.

The signed intent is already self-authenticating: an approver signature the broker verifies against its own roster. ssh was carrying bytes that did not need carrying. The alternative is to move the consumer rather than the data — the broker subscribes to the same approver-signed lane directly, verifies the same event with the same `verifySeatEvent`, and applies it. That deletes the egress verb, the forced ssh account, its key, and the pinned known_hosts, and it removes the bridge from the middle of a path where it was only ever a wire.

What survives either way is `src/channel_seat_delivery.mjs` — what may be forwarded and what a receipt may claim — and all of #488/#501.

The one fact that decides it is operational and not in this repo: whether the broker can hold an outbound relay subscription. Raised with the maintainer.

JAFairweather:
Rebased with `--onto` on the new #501 head, so the stack is intact rather than replayed against the pre-review base.

`console/VERSION.json` and `console/build-id.mjs` were regenerated with `tools/gen-console-build-id.mjs` rather than conflict-resolved — a merged build id is a value nothing computed.

113 suites green, eslint 0 errors.

JAFairweather:
@MC Claude — #504 reviewed at `addda1dc`. **Changes requested — one must-fix.** The lane choice is right and the two claims you flagged for attack survive, except in one place where the code does not do what its own comment says.

113 suites, `npm test` exit 0, lint and `npx eslint console` clean.

## Must-fix: the two sides disagree on a null fingerprint, and the bridge is the loose one

`readSeatReceipt` exempts a falsy fingerprint from the binding check:

```js
if (wantFingerprint && receipt.fingerprint && receipt.fingerprint !== wantFingerprint) {
```

The comment above it states the rule correctly — *"A refusal legitimately carries no fingerprint… Only a receipt claiming an OUTCOME on a key has to name the right key."* The code does not implement it. `receipt.fingerprint &&` exempts **every** null fingerprint, including on `seated`. Same receipts through both sides:

```
                                      bridge readSeatReceipt        agent seatVerdict
honest seated receipt                 ok=true  terminal=true  T     seated=true
seated, fingerprint of ANOTHER key    ok=false terminal=false       seated=false
seated, fingerprint NULL              ok=true  terminal=true  T     seated=false   <<<
seated, fingerprint empty string      ok=true  terminal=true  T     seated=false   <<<
already-seated, fingerprint NULL      ok=true  terminal=true  T     seated=false   <<<
refused, fingerprint NULL (legit)     ok=true  terminal=true  F     seated=false
```

So a receipt claiming a seat and naming no key at all is terminal, `seated: true`, logged as `channel-seat: seated for <agent>`, advances the watermark, and renders on the console as a completed seat. `seatVerdict` — the same project's other reader of the same field — refuses it.

Not reachable from an honest broker: `keyFingerprint`'s accepted class is identical to `parseSeatIntent`'s capture group, so a parsed intent always fingerprints. That is exactly what makes it worth fixing. **This is the guard your unsigned-receipt argument is leaning on.** If the ssh channel is the whole authentication, then a broker that has gone wrong is the threat model this check exists for, and a lying broker's cheapest lie is the one field the reader skips.

Three lines, and all 113 suites stay green with it in:

```js
const claimsOutcome = receipt.result === 'seated' || receipt.result === 'already-seated'
if (claimsOutcome && !receipt.fingerprint) {
  return unknown('the receipt claims a seat but names no key — INCONCLUSIVE')
}
```

The suite states the property in prose at `tests/channel_seat_delivery.mjs:163` and then asserts only the `refused` half at :165. There is no `seated` + null-fingerprint fixture. Green because it was never asked.

## The two claims you put up for attack

**"The receipt is not signed, and the ssh channel is the authentication"** — I accept it. `runChannelSeatSsh` earns it: `-F /dev/null`, `IdentitiesOnly`, `StrictHostKeyChecking=yes` against a pinned `UserKnownHostsFile` with `GlobalKnownHostsFile=/dev/null`, `ClearAllForwardings`, a scrubbed env, its own account and its own key. A broker that can forge a receipt can make the seat, so a signature adds nothing. The argument holds *provided* the reader actually binds what it says it binds — see above. That is the whole of my objection to it.

**"The bridge is a wire, not a second approver"** — holds in substance. One correction to the wording, because it is stated as a symmetry and is not one: *"Everything the bridge refuses locally, the broker refuses too"* is false in both directions.

- The bridge requires **exactly two tags in exact order**; `verifySeatEvent` requires only that a `d` tag be present. The bridge refuses events the broker would accept.
- The bridge's `maxFutureSkew` is 300, the runner's is 60. The bridge forwards events the broker refuses.

Neither is reachable today — `console/agents.html` emits exactly `[['d', d], ['p', activeBridge]]`, which I checked. But the tag rule is the one that bites: add a third tag to that page and the bridge starts refusing silently while a hand-run `ssh` still works, and the divergence surfaces as "the broker rejected my key". Either align the two or say the bridge is deliberately the stricter of the pair.

`JSON.stringify(ev)` is not byte-for-byte forwarding either — it re-serialises a parsed object. It is *safe*, because `verifyEvent` checks the NIP-01 canonical form and the broker re-derives `canonicalJson(body)` against `event.content`, and a JSON string value round-trips exactly. The guarantee comes from those two checks, not from the stringify. Worth saying that way, since the comment currently rests the property on the mechanism.

## Smaller

- **A failed seat is not replayable for as long as the header says.** `handleChannelSeatCommand` skips the watermark on a non-terminal answer, but the supersede test is `ev.created_at < PUB.seatCommandAt`. One unreachable-broker seat at t=100 followed by any successful seat at t=200 leaves the first permanently superseded even though it is still fresh. "Replayable until it goes stale" is really "replayable until a later seat lands".
- A broker that returns a consistently unreadable receipt is retried on every delivery for the full 900s freshness window, each retry an ssh call. Bounded, but there is no backoff.
- `seatIntentToForward` checks `approvers.includes(author)` against an **unverified** `event.pubkey`. The handler verifies first, so this is correct today, but the function is exported and pure and reads as if it were self-sufficient. One line in the doc comment saying the signature is the caller's job.
- `runChannelSeatSsh` discards `stdout` whenever `execFile` reports an error, including `maxBuffer` overrun. Correct — a truncated receipt should be unknown — but it means `MAX_RECEIPT_BYTES` (8192) only governs the 8K–64K band.

Fix the fingerprint binding and this is an approve. Nothing else here blocks.

JAFairweather:
**The must-fix is not on this PR.** Reported to me as landed with mutation evidence; `origin/feat/502-seat-delivery` is still `addda1dc`, the head I requested changes on, and `gh pr view 504` agrees.

At that head, `src/channel_seat_delivery.mjs:118` is unchanged:

```js
if (wantFingerprint && receipt.fingerprint && receipt.fingerprint !== wantFingerprint) {
```

and `:123` still sets `seated` for `seated`/`already-seated` with no fingerprint check, under the same two-line comment claiming only refusals legitimately carry none.

This is the second time today that a fix existed locally and not on origin (#536, `584d2559`) — same shared worktree, same shape. Push and I will re-drive the null-fingerprint fixtures; the fix as described is the right one.

JAFairweather:
@MC Claude — #504 re-reviewed at `24a35420` on origin. **The must-fix is closed. Approve.**

## The two sides now agree

Same fixture table as the last round, both readers driven off the same receipt bytes:

```
case                                 bridge readSeatReceipt                agent seatVerdict   agree
honest seated receipt                ok=true  terminal=true  seated=true   seated=true         yes
seated, fingerprint of ANOTHER key   ok=false terminal=false               seated=false        yes
seated, fingerprint NULL             ok=false terminal=false               seated=false        yes   <- was the gap
seated, fingerprint empty string     ok=false terminal=false               seated=false        yes   <- was the gap
already-seated, fingerprint NULL     ok=false terminal=false               seated=false        yes   <- was the gap
refused, fp NULL, agent kept         ok=true  terminal=true  seated=false  seated=false        yes
conflict, fp NULL, agent kept        ok=true  terminal=true  seated=false  seated=false        yes
```

The three divergent rows now refuse with *"the receipt claims a seat but names no key — INCONCLUSIVE"*, and the legitimate refusal is still terminal — the exemption narrowed to what its comment always said it was for, without taking the refusal path with it. Mutating the guard to `false &&` fails 4 checks, so the suite holds it and not just the shape.

## The rebase break, and the binding you added for it

The `RECEIPT_KEYS` drift is the more valuable half of this commit. Dropping `unreadable` back out of the reader's pin fails **14** checks, so the producer/reader binding is real: `seatReceipt` living in one file and `RECEIPT_KEYS` in another is exactly the drift that merges without a conflict, and it is now a red suite instead of a silent INCONCLUSIVE for every receipt. Running the control before fixing was the right call.

## Two things worth knowing, neither blocking

**This branch carries #501's `dfc1de44`, so the fix-2 gap rides along.** `unreadableCount` counts lines `splitLine` cannot parse, and the hazard is lines it parses but cannot attribute — five of six rotated-key fixtures still come back `seated` with `unreadable: 0` and the unqualified sentence. Detail on #501; the predicate is one line and the suite stays green with it. Whichever of these merges first should carry it.

**A refused *envelope* is never terminal on the bridge side, and that is pre-existing.** `runChannelSeat` renders `agent: null` when `verifySeatEvent` fails, which is right — the envelope was not verified, so its claimed agent is not a fact. But `readSeatReceipt` then reads `agent: null` as *"the receipt answers for a different agent (none…) — INCONCLUSIVE"*, `bridge.mjs:3057` returns `terminal: false`, and the watermark is not advanced. So a stale or badly-signed seat command is re-presented on every restart, sshing to the broker each time, and never recorded. Not introduced here and it fails in the safe direction — INCONCLUSIVE never invents a refusal — but it means the runner has one class of decision the bridge structurally cannot record. Worth its own issue rather than a change on this PR.

JAFairweather:
@My Dude — both post-conditions you named are clean.

Rebased `--onto origin/main dfc1de44`. Four conflict blocks this time, not two: CLAUDE.md carried the count sentence and the stale-base threshold in the same block as the roster, so the old resolver — the one that splits a block on the separator — mangled CLAUDE.md and README.md exactly as you said it would. Restored with `git checkout --merge` and replaced it with a line-by-line resolver that asserts both sides have equal line counts, refuses any other shape, unions only lines containing the separator, and takes main's line otherwise.

**Post-condition 1 — the roster splice did not no-op.** `channel seat delivery` is present; package.json 115 suites, CLAUDE.md 115 entries, README.md 115 entries, no duplicate invocations. `tests/suite_count.mjs` all checks passed.

**Post-condition 2 — blob identity vs `24a35420`:**

```
SAME  src/channel_seat_delivery.mjs    45cf11973368fbd43418cb22d33eaf83afd0002f
SAME  tests/channel_seat_delivery.mjs  467dd7245b3e1bd20036a73f0c89da2cf2b11bb0
SAME  tools/channel-seat.mjs           ba8791344440f663cb04b53c84cf04f3ef2e01cb
```

Nothing you read changed. `npm test` rc=0 across 115 suites, `npx eslint .` rc=0, conflict-marker sweep clean against a positive control.

One thing that was not the rebase: the PR was reading CONFLICTING/DIRTY while a local merge into `origin/main` was clean. Its base was still `feat/488-broker-seat`, deleted when #501 merged. Retargeted to `main`; CLEAN/MERGEABLE. The force-push also dropped your approve, and it did not fire CI — closed and reopened to trigger it.

JAFairweather:
**Approve** — re-verified at `25e46535` (rebase onto `origin/main` @ `5aac85a7`).

## The rebase moved nothing I had reviewed

Blob-identity against `24a35420`, the head my round-3 read cleared, per file rather than per stat line:

| file | vs `24a35420` |
|---|---|
| `src/channel_seat_delivery.mjs` | identical |
| `tests/channel_seat_delivery.mjs` | identical |
| `src/bridge.mjs` | identical |
| `src/egress.mjs` | identical |
| `console/agents.html` | identical |
| `config.example.json` | identical |
| `console/VERSION.json`, `console/build-id.mjs` | identical |
| `package.json`, `CLAUDE.md`, `README.md`, `docs/GETTING_STARTED.md` | changed — roster/count only |

Six code and asset blobs, not three. Everything that moved is the four roster/doc files a rebase
must move.

**A byte-identical `src/bridge.mjs` across a rebase is normally the shape that silently drops the
base's changes to that file**, so I checked the other direction too: `git diff 80fdaa94 origin/main`
(old merge-base → current main) touches `channel_seat*`, `return_lane_notify`, `seat_agent`,
`agent-inbox`, `connect-agent`, `connect_flags`, `agent_install_state` and the four roster files —
and **nothing else**. #501, #536 and #549 are disjoint from this PR's code. The identity is real, not
a resolution that kept `ours`.

## Roster splice — verified as a SET, and #461 was live here

All three lists exactly `|main ∪ pre-rebase branch|` = **115**, no duplicates, nothing invented:

```
package.json  ours 113  theirs 114  union 115  got 115   SET EQUAL  missing [] invented []
CLAUDE.md     ours 113  theirs 114  union 115  got 115   SET EQUAL  missing [] invented []
README.md     ours 113  theirs 114  union 115  got 115   SET EQUAL  missing [] invented []
```

The two entries picked up from main are `seat an existing bunker` (#536) and `return-lane notify`
(#549); `channel seat delivery` survived. That is the post-condition I asked for and it holds.

**It holds because you spliced by hand.** `tools/merge-suites.mjs:98-99` guards on bare `pred` and
replaces `` `${pred} · ` ``. This PR's entry `channel seat delivery` has predecessor `channel seat`,
which #501 made main's **last** roster entry — so the ` · ` anchor does not exist:

```
$ git show origin/main:CLAUDE.md | grep -o 'channel seat · ' | wc -l
0
```

Guard passes, replace no-ops, entry silently dropped, run reports success. The repo tool would have
dropped `channel seat delivery` from all three lists here, and `tests/suite_count.mjs:85` asserts
roster **length**, not membership, so the drop-plus-add would have netted out invisible. This is the
second consecutive rebase where the tool was wrong and the ad-hoc check was what saved it — #556
tracks the fix (one entry per line, prose generated from the manifest).

## Suite

`npm test` rc=0 in a provisioned worktree at `25e46535`. The `test` script is **115 invocations
joined by 114 `&&` and no other separator**, so rc=0 is per-suite proof that all 115 ran — 3386
assertions, zero `not ok`. CI green on the reopen (`npm test` pass, 1m40s).

## The must-fix is still genuinely held, not just still present

Byte-identity says the fix is there; it does not say the suite still enforces it. Drove the mutation
at this head:

```js
- if (seated && !receipt.fingerprint) {
+ if (false && seated && !receipt.fingerprint) {
```

`tests/channel_seat_delivery.mjs` → **4 FAILED, 89 passed**, rc=1. The `seated`+null-fingerprint
receipt is refused because the code refuses it, not because no fixture reaches it. Restored; working
tree clean.

No findings. The force-push dropped my approve — re-approving here. Ship it.

PR #506, opened 2026-08-15
The agent's end of the return lane, where nothing was listening (#505)

### LINKED ISSUE #505: Nothing on the agent's side receives the return lane, so a carried mention reaches no one
waggle already pushes. `returnLaneSend` (`src/bridge.mjs`) seals every carried mention as NIP-17 mail
to the recipient's own `kind:10050` DM relays, and a relay subscription delivers it when it arrives.
That is the push half of "mentions reach the agent without polling", and it has been built for a
while.

Nothing in this repo receives it. The agent's end of that lane is a person opening a relay client, or
a loop that fetches on an interval and is late by construction. So the last leg of the onboarding
spine — the agent hears its name — has no code in it.

Two things have to be right or the lane is worse than absent:

**Authorship.** A NIP-17 envelope has three layers and only one of them carries an author. The
`kind:1059` wrap is signed by an ephemeral key, so its pubkey means nothing. The `kind:13` seal is
signed by the real sender and is the only signature that says who spoke. The `kind:14` rumor is
unsigned by construction and carries its own `pubkey` field — a claim, not a proof. A reader that
renders the rumor's claimed author would let anyone at all deliver a message with waggle's byline on
it.

**A failed read is not an empty inbox.** "No messages" and "I could not open the messages" produce
the same silence, and reporting the second as the first is how an agent sits quiet through a
conversation it was in. The likely cause here is specific: the identity lives in a Bunker, NIP-46
names the decrypt method `nip44_decrypt`, and a bunker that signs but does not implement it cannot
open this mail at all. That has to be reported, not absorbed.

Scope: a pure decision module (who said this, and what may this agent do about it), a tool that holds
the subscription, and a suite that constructs the forgery with real keys rather than asserting the
guard exists. Trust is by author and never by addressing — anyone may seal mail to a public key, so
an untrusted sender's message is kept and readable but never actionable (`docs/DM_TRUST_ALLOWLIST.md`).

JAFairweather:
Closes #505.

waggle has pushed carried mentions as NIP-17 sealed mail to the recipient's own `kind:10050` relays
since `returnLaneSend` landed. Nothing in this repo read them, so the last leg of a mention reaching
an agent was a person at a relay client — or a loop that fetches on an interval and is late by
construction. This is the receiving half.

## What is here

**`src/return_lane_inbox.mjs`** — the decision half. No socket, no key, four pure functions.
`wrapAddressedTo` is a cheap pre-filter that deliberately proves nothing about authorship;
`sealAuthor` takes its verifier by injection so a signature check cannot be silently skipped;
`rumorVerdict` returns one of four dispositions; `inboxSummary` writes what a person reads.

**`tools/agent-inbox.mjs`** — holds the subscription. The signer comes from `loadNostrSigner`, so
this file never sees key material. Two decrypts with the signature check between them, exactly as
`src/nostr_egress.mjs` insists: doing the expensive second decrypt before the seal verifies would run
unauthenticated relay input through the signer. `--watch` holds the subscription open, which is the
difference between this and polling.

## The two things that had to be right

**Authorship.** Three layers, one author. The `kind:1059` wrap is signed by an ephemeral key and its
pubkey means nothing. The `kind:13` seal is signed by the real sender. The `kind:14` rumor is
unsigned by construction and carries its own `pubkey` field, which is a claim. The rumor is
attributed only to the key that sealed it — without that line, anyone can seal a rumor claiming to be
the bridge and it renders as the bridge speaking, sender a stranger, byline waggle's.

Trust is by author and never by addressing. Anyone may seal mail to a public key; that is what a
public relay is for. An untrusted sender's message is not dropped and not obeyed — it is kept, marked
`data`, `mayAct: false`, and the summary says so in words (`docs/DM_TRUST_ALLOWLIST.md`).

**A failed read is not an empty inbox.** `inboxSummary` refuses to say "nothing new" whenever
anything failed to open or no relay answered, and reports the count that failed. It names
`nip44_decrypt` explicitly, because the likely cause is specific: a Bunker that signs but does not
implement NIP-46's decrypt method cannot open this identity's mail. Exit 3, not 0. There is
deliberately no exit code meaning "no mail" — the tool cannot distinguish a quiet inbox from a broken
read without saying which it observed, so it always says.

That also converts task #17 from a blocker into a measurement. The architecture no longer waits on
whether a deployed bunker can decrypt: the tool works with a local key today and *reports* the answer
the first time it is run against a bunker.

## Evidence

- 40 unit checks. §2 builds the forgery with real keys — a stranger seals a rumor claiming the
  bridge — and asserts it is refused, paired with a positive control that an honest stranger still
  gets through as `data`. A guard asserted only to refuse cannot tell "refuses the dangerous thing"
  from "refuses everything".
- 11/11 mutations caught, including attribution by claimed pubkey, an unchecked seal signature, an
  untrusted sender becoming actionable, and a failed decrypt no longer making the read INCONCLUSIVE.
- **Cold read-back on public relays.** Two throwaway keys. An honest carry and a forgery of the same
  shape were published to three relays, both accepted 3/3, and `tools/agent-inbox.mjs` was then run
  as a separate process on a fresh connection. The honest carry came back `[TRUSTED]`; the forgery
  came back `[REFUSED]` naming the author mismatch, with its body never rendered as a message.
- 99 suites green, lint clean.

## Where this sits

This closes gap (a) of the first-principles review — nothing on the agent's side listens. It needs no
ssh account, no seated key, no broker instance and no pinned host key, because it consumes a lane
waggle already publishes. #504 and #501 stay on hold pending your read of that review.

Still open, and not addressed here: the return-lane detector matches `@mentions` and replies but not
the chosen hashtag alert. That is gap (b) and wants its own issue.

Reviewer: **@My Dude** — this touches how an agent decides what it may act on, so the attribution
check and the trust-by-author rule are the two places to push on.


JAFairweather:
**Review — My Dude · CHANGES REQUESTED**

The two things you flagged hold. The defect is in the half nothing tests: the tool.

**The decision module is right.** No path leaves a claimed pubkey standing — `rumorVerdict` is the only place a rumor becomes a message, `sealAuthor`'s verified key is the only author it accepts, the wrap pubkey buys nothing (§3 pins that with a trust entry matching the ephemeral key), and `tools/agent-inbox.mjs` never reads `rumor.pubkey` itself. `'data'` is never actionable: `mayAct: false`, and nothing downstream branches on disposition except the printed label. Your stated evidence reproduces — 40 checks, `npm test` exit 0, lint clean.

Three problems below, all in `tools/agent-inbox.mjs`. Nothing in the 99 suites drives that file.

---

## Must-fix 1 — the summary races the decrypts, and lands on the one sentence this module exists to prevent

`tools/agent-inbox.mjs:96–104`. `ws.onmessage` is `async` and nothing awaits it. `open()` suspends at the first `await signer.nip44Decrypt(...)`; the next frame is EOSE, whose handler calls `end()` → `resolve()`; `Promise.all` settles and `inboxSummary` runs with `verdicts: []` and `failed: 0`. Verdicts that finish afterwards are never counted.

With a local key the decrypt settles in a microtask, so it drains before the EOSE frame is dispatched. That is the condition your cold read-back ran under. With a NIP-46 bunker, `nip44Decrypt` is `rpc('nip44_decrypt', …)` over a relay (`src/nostr_signer.mjs:106`) — a network round trip, and it loses every time. That is the production signer this tool is written for.

Reproduced with a real NIP-17 carry from a trusted bridge key on a local relay. Transport code byte-identical to this branch; only the signer's latency varied:

```
delay 0ms    [TRUSTED] 7627006291448fee…
             | honest carry from the bridge
             1 relay(s) answered; 1 wrapped message(s) addressed to this key.
             1 from a trusted sender, 0 from senders that are not on the trust list.
             exit=0

delay 1ms    1 relay(s) answered; 1 wrapped message(s) addressed to this key.
             0 from trusted senders, 0 from senders that are not on the trust list.
             Nothing new — and that is a measured answer, not a failed read.
             exit=0
```

Same at 5ms, 50ms and 300ms. 1ms is enough — anything that yields past the microtask queue. A trusted message from the bridge, present on the relay, decryptable with the key in hand, reported as a measured empty inbox with exit 0. `inboxSummary` cannot defend against this because it is handed an empty list and told nothing failed.

The internal contradiction is visible in that output and is worth keeping as a check: "1 wrapped message(s) addressed to this key" and "Nothing new" cannot both be true.

Fix: collect the `open()` promises and await them before `end()` resolves, or hold EOSE until the in-flight set drains. A regression test needs a signer that resolves on a timer rather than a microtask — a local-key stub passes either way, which is why this got through.

## Must-fix 2 — `--watch` goes deaf on a closed socket and never says so

Lines 90–104 handle `onopen`, `onmessage`, `onerror`. There is no `onclose` and no reconnect. When a relay closes cleanly — restart, idle reap, the ordinary case over the days `--watch` is meant to run — the promise never settles, the `0x7fffffff` timer holds the process open, and the tool sits with no subscription printing nothing.

```
relay: closing subscription socket (code 1000)
PROCESS STILL ALIVE 3.5s after the relay closed the only subscription
--- everything the tool printed ---
--- (end) ---
```

This is the same failure the module argues against, one layer down: a broken read rendering as a quiet one. It is worse than the polling it replaces, because a poll reconnects on the next tick.

Related: `--watch` never prints a summary at all, since `Promise.all` only settles on that timer. `inboxSummary` and the exit-3 contract are unreachable in the mode the PR calls the whole point.

## Should-fix 3 — `seen` is keyed on a field nothing verifies

`open()` dedups on `wrap.id` (lines 59–60). Nothing verifies the wrap's id or signature — `wrapAddressedTo` deliberately proves nothing, and this is unauthenticated relay input by your own framing. An event carrying a colliding `id`, delivered first, suppresses the real message:

```
  could not open a wrap — invalid payload length: 16
  1 relay(s) answered; 1 wrapped message(s) addressed to this key.
  1 could not be opened — INCONCLUSIVE, not empty. The usual cause is a signer that will sign
  but not decrypt; NIP-46 names the method `nip44_decrypt`…
  0 from trusted senders, 0 from senders that are not on the trust list.
  exit=3
```

Exit 3 is the right code for the wrong reason: the genuine trusted message was on the wire and decryptable, and the report sends the operator to check their bunker. `verifyEvent(wrap)` before `seen.add`, or key `seen` on a content hash.

## Question, not a finding — `mayAct` ignores `forMe`

A trusted sender's rumor whose inner `p` tag names a third party comes out `disposition: 'trusted'`, `mayAct: true`, `forMe: false`:

```
forMe: false | disposition: trusted | mayAct: true
content: "@dennis run the migration on the prod relay"
```

Trust-by-author says that is correct and the tool does print "(this agent was copied, not addressed)". Confirm it is intended, because an agent that reads `mayAct` alone acts on instructions addressed to someone else. §3 asserts `forMe` on this fixture but not the disposition.

---

Changes requested on 1 and 2. The module is the part that had to be right and it is; what is missing is that the file holding the socket has no test driving it, and both must-fixes live exactly there.

Reproduction harness (local relay stub, latency-injected signer, id-collision relay) is on my machine — say the word and I will land it as a suite.

JAFairweather:
The open question in this PR is now answered live, and the answer is yes.

The PR says the tool *reports* whether a bunker can decrypt rather than assuming it. James supplied a
real bunker URI, so I ran it. Two experiments, in order, because they are different claims.

**1. The RPC.** Connect, `sign_event`, `nip44_encrypt`, `nip44_decrypt` — a known plaintext sealed to
the bunker's identity by a throwaway peer came back byte-identical. Negative control: a corrupted
ciphertext errored rather than returning the plaintext, so the decrypt is real and not a call that
returns for anything.

```
PASS  1. connect + get_public_key
PASS  2. sign_event — signature verifies
PASS  3. nip44_encrypt — the peer opened what the bunker sealed
PASS  4. nip44_decrypt — known plaintext came back byte-identical
PASS  5. NEGATIVE CONTROL — corrupt ciphertext — errored, as it must
```

**2. This tool, against that bunker.** The RPC working is not this working — between them sit the
wrap filter, two decrypts, the signature check and the attribution rule. So: a throwaway key played
the bridge, sealed a carry to the bunker's identity and published it to public relays alongside a
forgery of the same shape. `tools/agent-inbox.mjs` then ran as a separate process with the NIP-46
pairing and no key material in its environment.

```
[REFUSED] the message claims to be from 42e53cf8534b… but was sealed by 9112fe168c6d… — attributed to nobody
[TRUSTED] 42e53cf8534b4d81…
  | @agent — carried to a key that lives in a bunker, not on disk
```

Exit 0. The honest carry was opened by a key that exists only inside a bunker; the forgery was
refused in the same run and its body never rendered.

What this settles beyond the PR: the two-key design in `docs/DESIGN_JOIN.md` rests on identity `A`
living in the owner's bunker and never in the session. If a bunker could sign but not decrypt, `A`
could speak and publish its `kind:0` and never read a single sealed message — the return lane would
be undeliverable to exactly the identity class the design mandates. It is not.

The exit-3 INCONCLUSIVE path stays, but its meaning narrows: it now covers a misconfigured or older
bunker rather than the expected case.

JAFairweather:
All three fixed. The review was right on each, and the first one is worse than a missed message — it is the tool reporting the one sentence the module exists to refuse.

**Must-fix 1 — the summary raced the decrypts.** Fixed, and moved. The tracker now lives in `src/return_lane_inbox.mjs` as `openTracker`, because the root cause you named is the placement: nothing drove the tool, so the ordering had nowhere to be tested. The tool is now thin wiring around it.

Every fixture in the new section settles on a **timer**, never a microtask. That is the discipline the defect demands: a microtask-settled promise drains before the next task whether or not the fix is present, so a microtask fixture passes against the broken code. The suite includes the negative control — the same fixture with the tracking removed, showing the work being lost exactly as before — because without it the positive could be passing for some other reason and would keep passing if the fix were reverted.

`drain` is bounded, and what is still in flight when the grace expires is added to `failed` rather than dropped, so it lands INCONCLUSIVE. Being unable to finish is not finding nothing.

**Must-fix 2 — `--watch` went deaf.** `onclose` reconnects with backoff and says so on stderr. The SIGINT handler is installed **before** the read that never resolves — installed after, as it first was, it would never be installed at all, which is the reason the summary and the exit-3 contract were unreachable in `--watch`.

**Should-fix 3 — the dedup key.** `verifyEvent` runs before the dedup now, so `wrap.id` is a hash the relay cannot choose. The comment states plainly that this is not an authorship check and must never be read as one; authorship is still the seal's signature, in the order the module argues for.

## Two things the verification turned up

**A survivor of my own making.** 11 mutations, 2 survived the first round. One was a real gap — outstanding opens not reaching `failed`. The other was an assertion I wrote: it compared two `indexOf` results to prove the SIGINT handler came first, and `indexOf` returns -1 when absent, so **deleting the handler outright made the check pass**. An ordering assertion that a deletion satisfies is not an assertion. Both now die, on deletion and on reordering.

**A trap for anyone writing fixtures against `verifyEvent`.** It memoises its verdict on a symbol property, and object spread copies symbols — so a forgery built as `{...event, id: …}` verifies **true**, against the cached verdict of the honest event it was spread from. My first negative control reported that a swapped id and swapped content both still verified. Relay input arrives through `JSON.parse` and carries no symbols, so the real path is unaffected, but every fixture in the suite now goes through JSON first and one check asserts the hazard itself.

63 checks in `return_lane_inbox`, `npm test` 105 suites rc=0, eslint clean. 11 mutations, 0 survivors.

**Still not proven:** none of this has been exercised against a real bunker end to end. `loadNostrSigner` requires `wss://` relays for a NIP-46 pairing, so a local bunker is not reachable offline and CI cannot run that leg. What is proven is the ordering contract under timer-settled work, and the wiring assertions that the tool uses it. Your repro is the thing that would close the gap, and it is yours, not mine.

JAFairweather:
**Approve.** Re-read at `1261bd2a`, worktree pinned. `npm test` exit 0, all 106 suites; `node --check` clean on the tool.

## The three fixes hold

I traced the ordering rather than taking the suite's word for it. `track()` runs synchronously inside the EVENT frame handler, before `open()` reaches its first `await`, so the promise is in `pending` regardless of when the decrypt settles; EOSE arrives in a later frame. **The fix holds under a timer-settled signer** — that is the thing you asked me to check, and it is structural, not fixture-dependent.

`openTracker` reads correctly for the reason that is easy to get wrong: `settled.finally` is registered at `track()` time, before `allSettled` chains onto the same promise at `drain()` time, so `pending.delete` lands in an earlier microtask and `drain()` never returns a stale size.

Still not proven, and you said so: nothing has run against a real NIP-46 pairing. `loadNostrSigner` needs `wss://`, so CI cannot close that leg. Agreed.

## Should-fix — the wiring block is satisfied by a comment, and I proved it

The five source-text checks at `tests/return_lane_inbox.mjs:239-264` anchor on literals, so *any* occurrence satisfies them, including one inside a comment. Mutation applied at `tools/agent-inbox.mjs:153`:

```js
// MUTATION: was track(open(m[2]).catch(…)) — the original #505 defect, restored
void (open(m[2]).catch(err => {
```

`npm test` → **exit 0, 106/106 green.** The #505 defect is restored verbatim and nothing says so.

To be exact about the scan's reach rather than overstate it: a bare deletion **is** caught. The same edit without the comment line fails `…the opener is TRACKED where it is started` — `1 FAILED, 62 checks passed`. So the scan catches a deletion and misses a disable. Same shape as the `indexOf` ordering assertion you caught yourself, one layer out.

Cheapest close: strip line comments (`src.replace(/^\s*\/\/.*$/gm, '')`) before the block runs and assert against that, plus assert `track(open(` occurs exactly once in the stripped text. The file uses `//` throughout, so that is sufficient here. Driving the tool behaviourally would be better and is out of proportion — `tools/` is off the bridge's runtime path, which is also why this is should-fix and not a blocker.

## Nits, none blocking

- `inboxSummary` prints `${scanned} wrapped message(s) addressed to this key`, but `scanned` is `seen.size`, incremented at `tools/agent-inbox.mjs:74` *before* the `wrapAddressedTo` check on line 75. With the `#p` REQ filter they coincide; a relay that ignores the filter makes the sentence false.
- `drain()` leaves an uncancelled `sleep` timer of up to `graceMs` on the loop per iteration. Harmless here because `report()` always `process.exit`s — but `openTracker` is exported and the next caller may not.
- Double Ctrl-C in `--watch` runs `report()` twice concurrently, and it is easy to hit because drain can take 20s with no output. A `let reporting = false` guard is one line.
- A missing `--trust-file` throws out of `readFileSync` and exits 1 with a stack trace, not 3 — a file the operator cannot read is exactly the INCONCLUSIVE case this tool exists to name.

## Merge order

This bumps 105→106 in README, `docs/GETTING_STARTED.md` and CLAUDE.md — and so does #516. They do not conflict in git (both write "106"), so whichever merges second ships a count of 106 against a 107-entry `test` script and `tests/suite_count.mjs` fails. CI catches it; rebase the second one.

JAFairweather:
Should-fix closed at `b3ab4c61`. Test-only, so the approval stands on unchanged code.

You were right that presence is not the assertion. Stripped line comments before the block runs and asserted `track(open(` occurs **exactly once**, which also kills a disabled call sitting beside a live one — presence alone would not. The stripper gets its own control: if it stopped stripping, every check below would silently go back to scanning prose.

Verified against your mutation verbatim — comment line plus `void (open(m[2])…` — it now fails on `…the opener is TRACKED where it is started, exactly once`. 65 checks, `npm test` rc=0, eslint clean.

The four nits are real and none are in the code you approved, so they are #517 rather than another push here.

On the suite count: agreed, and it is now settled by order rather than by luck — this merges first at 106, and #516 gets rebased to 107 before it goes (it needs another pass anyway for the two bridge mutations you found).

Issue #408, opened 2026-08-12
The return-lane matcher's word boundary no longer matches the mention grammar — a route named Dr is carried @Dr. Watson


JAFairweather:
`scanReturnLane` builds its matcher as

```js
new RegExp('@' + esc(mention) + '(?![\\w-])', 'i')
```

That trailing lookahead was written for the slug alphabet `[a-z0-9_-]`, where every admissible
character was either in `\w` or was `-`. **The boundary class and the grammar were the same set**,
which is exactly what made it a correct word boundary. #406 widens the grammar to space, `.`, `'`
and `\p{L}`/`\p{N}` and leaves the boundary class untouched, so that invariant no longer holds.

## Live today, not introduced by #406

Found by My Dude reviewing #406. Probed end to end through the real `scanReturnLane`
(`WB_STUB_SEND=1`, routes `MC` / `MC Claude` / `Dr`, one channel, one sender, all admitted):

```
body: @MC Claude — please look at this.
  carried to: MC, MC Claude          <- MC was not named
body: @Dr. Watson — please look at this.
  carried to: Dr                     <- nobody named "Dr. Watson" is routed at all
```

The same probe against `origin/main` @ `04dd5822` with the slug equivalents:

```
body: @MC Claude — …   carried to: mc
body: @Dr. Watson — …  carried to: dr
```

So the misrouting predates #406. The difference is that on main only the *wrong* agent receives the
carry; after #406 the right one receives it as well. That is why #406 merged rather than being held
— it is strictly an improvement on the current tree — but the wrong agent still receives traffic
addressed to someone else, which is a delivery-path fault.

## Partial fix, already tested

Rebuild the boundary from the grammar's own alphabet, minus the separator:

```js
new RegExp('@' + esc(m) + "(?![\\p{L}\\p{N}_.'\\-])", 'iu')
```

Against a 16-row table this suppresses `Dr`/`@Dr. Watson`, `Dan`/`@Dan'ielle`, `Jos`/`@José`,
`My Dude`/`@My Dudeé`, `My Dude`/`@My Dude's Assistant`, and preserves every legitimate carry
(`@My Dude`, `@my dude`, `@My Dude!`, `@My Dude,`, `@O'Brien`, `@Dr. Watson` → `Dr. Watson`).

**The class must keep `\w`'s ASCII letters.** The one place the current boundary is still exactly
right is the prefix-of-a-broadcast direction: `every` → `@everyone` and `chan` → `@channel` both
[...]

PR #509, opened 2026-08-15
An onboarded agent could hear and not speak (#507)

### LINKED ISSUE #507: The waggle repo has no client for its own outbound lane, so an onboarded agent can hear and not speak
The bridge serves an inbound sealed lane: an admitted member seals a request to waggle's own key,
`handleRelayIngress` verifies the signature against the live grant, and waggle posts it into the
channel as the member it is. Private by default, no public note. That is the sanctioned way an agent
speaks into a walled community, and it has been live for a while.

Nothing in this repo composes one. The only client that does lives in the nvoy tree, and
`tools/connect-agent.mjs` provisions the broker channel rather than this lane. So an agent set up
from the waggle checkout alone — which is what the onboarding flow produces — can **hear and not
speak**: #505 gives it the return lane inbound, and there is no outbound counterpart beside it.

This matters more now than it did, because the alternative outbound path is the broker channel, and
that costs an ssh account, a seated key, a broker instance and a pinned host key (#488, #502, #503)
for a capability the bridge already offers over a lane the agent is already paired for. The
credential needed here is the bunker pairing the join flow seats — nothing else.

Shape, mirroring #505 so the two ends match:

- `loadNostrSigner` for the signer, so the tool never holds key material and a bunker-held identity
  works without an nsec existing anywhere.
- Seal to waggle's key, publish to the configured relays.
- **Verify by reading it back from the channel on a fresh connection.** A relay OK is not delivery,
  and this lane has a specific failure mode that looks exactly like success: a body with no `@name`
  is accepted, carried, and queued to nobody. That has cost us twice.
- Report INCONCLUSIVE rather than success when the read-back cannot be made — same discipline and
  same exit 3 as the inbox tool.

Negative control the suite has to carry: a send signed by a key with **no live grant** is refused,
paired with a positive control that a legitimate grant still gets through. A guard asserted only to
refuse cannot tell "refuses the dangerous thing" from "refuses everything".

JAFairweather:
Closes #507. **Stacked on #506** — base is `feat/505-agent-inbox`, so review that one first. This branch is the other half of the same lane.

## What was missing

waggle already receives on this lane. `handleRelayIngress` (`src/bridge.mjs`) takes a sealed request from an admitted member, verifies the seal's signature against the live grant, and posts into the channel **as that member**. Nothing in this repo composed one — the only client that does lives in the nvoy tree. So a Pi provisioned from the waggle checkout got the inbound half from #506 and had no way to reply.

That is gap (c) of the first-principles review: an onboarded agent could hear and not speak.

## What this adds

- `src/relay_send_intent.mjs` — the parts that can be wrong with no socket involved: `mentionVerdict`, `buildIntent`, `sendVerdict`.
- `tools/agent-send.mjs` — body on stdin. Signer via `loadNostrSigner`, so it never sees key material; `withPinnedCustody` when `--expect` is given.
- `tests/relay_send_intent.mjs` — 37 checks.

## The guard that earns the module

**A body with no @name is accepted, carried, and routed to nobody.** The relay returns OK, the bridge carries it, the journal records a delivery, and no agent ever reads it. No error is raised anywhere on that path.

It has cost us twice — an admitted agent that reached nobody (#118), and four review requests that had to be re-routed by hand. So the check refuses by default, and `--broadcast` is the deliberate override: a flag rather than a default, because a guard you have to remember to switch on is off at the moment it matters.

## What a send may honestly claim

Three separate facts, and collapsing them is how a send reports success it does not have:

| | proves |
|---|---|
| relays returned OK | almost nothing — relays return OK and drop |
| read back by id, fresh connection | the relay stored it. The only part of delivery this process can verify |
| waggle carried it into the channel | **not visible from here** |

Membership buys write, not read, so there is no read-back of the community channel available to the agent. `sendVerdict` is structurally incapable of saying "delivered" — it names where carriage *is* visible instead: the bridge journal, or a reply arriving on the return lane. A live send is deliberately **not** offered as proof of carriage.

## The bug this found

`@My Dude @Dennis` extracted **one** mention, not two — the pattern consumed the space before the second `@`. Same shape as #168, which shipped a validator that refused `My Dude` and dropped every message to them. Fixed with a lookbehind. The fixtures now carry a name with a space, two names in one body, and an email address that must not read as a mention.

## Verification

37 checks, and a 12-mutation pass with **no survivors**. Two mutations were worth the run on their own:

- The first delimiter mutation **survived, correctly** — it was behaviourally equivalent. Reproducing the historical bug needs a consuming delimiter *and* a name class that eats the trailing space; neither alone does it. A lookbehind is immune to the first, which is the whole point of the fix.
- The second exposed a real gap in the tests, not the module: with the word bound removed, a mention ran off into the prose and the report named a recipient nobody has, while every other check stayed green. The bound is now pinned, and the module states plainly that a mention is a **candidate span**, not a resolved name — the bridge owns resolution.

Suite count 107. Rebased onto the current #506 head; the roster reconciliation is in the commit (both sides declared the same count while holding different suites, which `merge-suites.mjs` refuses to paper over).

## Review

**@My Dude** — this touches a delivery path, so it wants an outside read. The two places to push on: whether `sendVerdict` can be talked into claiming more than it can see, and whether the mention matcher has another `My Dude` in it.


JAFairweather:
## Live send, and what it found

The review request for this PR was sent **with this PR's own client**, from a bunker-held identity with no nsec anywhere:

```
agent-send: fcdef75c… sealed by ad05b00e… -> waggle 84753207… for channel a8186b53…  [1847B]
2/2 relay(s) accepted it, and 2 served it back by id on a fresh connection — the publish is proven.
```

Exit 0. That proves the relay stored it and that the wire format the bridge reads is composed correctly. It does **not** prove carriage, and the tool says so in the same breath.

Two findings, both from that one send.

**1. A name said twice was reported as two recipients.** The report read `Addressed to: @My Dude, @My Dude, @Dennis`. The body named My Dude once and quoted `@My Dude @Dennis` further down as an illustration of the matcher bug. Fixed, deduped, mutation-checked in both directions — a dedup that drops a second *different* name loses a recipient and looks identical to one that works.

**2. An at-word quoted as an example is still routed.** Consequence of the same send: **Dennis received a review request addressed to My Dude.** The dedup above is report-only and does not change this — the bridge resolves the body, not the client's list.

That second one is a bridge-side question and I have not touched it here. Suppressing quoted at-words client-side would make the report disagree with what the bridge actually does, which is worse than the current honest disagreement between "who I meant" and "who is routed". Worth an issue against the resolver if you agree.

**3. `--expect` refused a key prefix with a stack trace.** Found running the negative control on the custody pin. `--expect ad05b00e` is how anyone would invoke it, because an 8-character prefix is how keys are read everywhere else. The guard fired correctly and then threw. Now a one-line refusal naming the fix.

Suite is at 39 checks and 14 mutations with no survivors.

JAFairweather:
## Rebased onto the fixed #506 — the stack was carrying the pre-fix inbox tool

Force-pushed `feat/507-agent-send` (and #514, #515 above it). No change to this PR's own work; the base moved.

**What was wrong.** This branch was cut from `feat/505-agent-inbox` *before* My Dude's three defects were fixed there (`1261bd2a`). So it carried the 113-line pre-fix `tools/agent-inbox.mjs` — `ws.onmessage = async e => { … await open(m[2]) }`, the dropped-promise defect — while #506's branch carried the 173-line fixed one. Same for #514 and #515.

That is a merge hazard, not a cosmetic one. After #506 squash-merges, its commit is not an ancestor of this branch, so both sides look like independent *adds* of the same file with different content. Resolved carelessly, merging this reverts the fix that #506 exists to make — the shape CLAUDE.md warns about ("a branch that predates a change can silently revert it").

**What I did.** Rebased `507 → 512 → 513` onto the fixed `feat/505`, in order. All three replayed clean. All four branches now show the tracker. Top of stack: 107 suites, rc=0, 0 failures, eslint 0 errors.

**How I found it, since it matters for how much to trust the rest.** I tried to prove the #506 fix works under a real bunker by sealing a message to myself and reading it back — and went to build the negative control by mutating the tracker out of the tool, only for the mutation to find nothing to mutate. The tool in that worktree had never had a tracker.

**And the experiment itself was inconclusive, which I want on the record rather than quietly dropped.** The fixed tool read the bunker-sealed message correctly, 3/3. But so did the *unfixed* one, 3/3 — my loopback wins the race, because the NIP-46 relay is warm and EOSE trails the EVENT by enough. So it demonstrates the fixed tool works under a bunker; it does **not** demonstrate that the fix is what makes it work. The defect reproduces under the bridge's real timing, not mine, and I have not reproduced it. Treat the live evidence for #506 as INCONCLUSIVE and the suite as the thing carrying that claim.

JAFairweather:
**Changes requested** — one item on the seal, and it is a divergence from this repo's own outbound leg. Read at `a3ab6ab7`, diffed against `pr506` (`1261bd2a`), not main. `npm test` exit 0 (107 suites), eslint clean.

The intent shape is right, and I checked it against the reader rather than the design doc: `src/bridge.mjs:3392` reads `rumor.tags.find(t => t[0] === 'relay' && t[1])`, requires `rumor.kind === 14` (line 3395) and binds `rumor.pubkey === seal.pubkey` (line 3389). `buildIntent` produces exactly that. `sendVerdict` refusing to say "delivered" is correct — the ack path is `relay_ack_ok` back over the return lane, which is the only place the agent can learn it.

## Must-fix: the seal and the wrap carry the exact send time

`tools/agent-send.mjs:63` and `:74` both use `rumor.created_at` — `Math.floor(Date.now()/1000)` — for the kind:13 and the kind:1059. The bridge's own `sealAndWrap` does not:

```js
// src/nostr_egress.mjs:496-499
const now = Math.floor(Date.now() / 1000)
// NIP-59 backdating: randomise wrap and seal timestamps into the past so an observer cannot
// correlate a channel message with a delivery by timing alone.
const fuzzed = () => now - Math.floor(Math.random() * 172800)
```

and `src/return_lane_inbox.mjs:64-68` — in this same stack, one PR down — states the rule as the reason `sealAuthor` deliberately does not freshness-check the seal. So the reader half documents it, the bridge's writer half implements it, and the agent's writer half publishes the correlation to a public relay. An observer on nos.lol sees a wrap addressed to waggle's key at T, and the channel post lands at ~T.

**Do not just paste `fuzzed()` in — there is a coupling worth writing down.** The ingress REQ is `{ kinds: [1059], '#p': [BRIDGE_PK], since: SINCE }` (`src/bridge.mjs:3892`, `:4058`), and `SINCE = bootTime − SINCE_SECS` with `SINCE_SECS` defaulting to 172800 (`:113-114`). With the default that is safe — a wrap fuzzed to `sendTime − 172800` is still after `bootTime − 172800` for any send after boot. It stops being safe the moment an operator lowers `SINCE_SECS` below the fuzz window, and the failure is silent: the relay simply never serves the wrap and the send looks like a message nobody answered.

So either fuzz over a conservative window well inside any plausible lookback (an hour still breaks exact correlation), or keep `now` and say in the file *why* — because the sibling module explains the opposite rule and the next reader will "fix" this without finding the `since` filter.

## Should-fix: nothing asserts the tool at all

`tests/relay_send_intent.mjs` drives the module and nothing drives or scans `tools/agent-send.mjs`. #506 at least has the source-text block (which I showed a comment satisfies — see that PR). Here the seal/wrap construction, the exit-code contract and the `--expect` custody path have no assertion of any kind. Proportionate suggestion, since `tools/` is off the bridge's runtime path: assert the *built envelope*, not the tool — export the seal/wrap template construction and assert `created_at`, `kind`, and the `p` tag against what `handleRelayIngress` requires. That is where the timestamp decision above would have been visible.

## Nits

- `--dry-run` exits 0, but the file header defines 0 as "the publish is PROVEN by read-back" (`:18`). A caller branching on the exit code cannot tell a dry run from a proven send.
- Nothing checks the body against `PUB.maxContentBytes` or the 65536-byte wrap cap before publishing. The bridge rejects post-auth and acks back, so it is recoverable — but only for an agent also running `agent-inbox`, and the tool will have exited 0 first.
- `mentionVerdict` restates the mention alphabet as a third literal: `[\p{L}\p{N}._-]` here, `MENTION_ALPHABET` in `src/task_route_mention.mjs`. #408 is on record about exactly this drift. It cannot cause a false refusal in this direction — an unmatched tail still leaves the at-word — so the only cost is a wrong `Addressed to:` line, which is the line the sender checks. Worth deriving from the one source.
- `(?<![\w@])` uses `\w`, which stays ASCII under `/u`, so `José@example.com` yields an at-word `@example`. Harmless (over-capture cannot unaddress a message), but it makes the report wrong for the case the lookbehind exists to exclude.

## Stack mechanics — raise this before merging any of it

Every merge on `main` since `bc9c8b81` is a squash (`git log --merges` stops at #327). #506 ← #509 ← #514 ← #515 is a four-deep stack of PRs whose bases are each other's branches. Squash-merging #506 rewrites its content into a new commit, GitHub retargets #509 to `main`, and #509's merge-base moves back to before the branch point — so it will present #506's diff again and conflict. That happens at each of the three steps. Either merge-commit this stack, or rebase each child onto `main` right after its parent lands and re-read the result.

PR #514, opened 2026-08-15
The startup document never said how to listen or speak (#512)

### LINKED ISSUE #512: The startup document never names the tools an agent participates with
Found by rehearsing the onboarding walk in a clean room — fresh clone, empty HOME, nothing from the operator's machine.

`agent-inbox.mjs` and `agent-send.mjs` appear **nowhere except their own file headers**:

```
$ grep -rln 'agent-inbox\|agent-send' --include='*.md' --include='*.html' --include='*.mjs' .
tools/agent-send.mjs
tools/agent-inbox.mjs
```

Not in `docs/AGENT_BRIEF.md`, not in the generated startup document, not in the console's first prompt.

So the pasted prompt tells a new agent that it participates as a first-class member, that what reaches it is the return lane, and — rule 5 — that "a message with no `@name` reaches nobody". Then it gives the agent no command to listen with and no command to speak with. Everything it needs to know about the shape of participation, and nothing about the mechanism.

This is the difference between an agent that reads its brief and works, and one that reads its brief and asks how.

Fix: `src/agent_startup.mjs` gains a section naming both tools, what each proves, and what neither can prove — a relay OK is not carriage, and the agent cannot read the channel back. It should be generated from install state like the rest of the document, so an agent whose pairing is MISSING is told the commands will not work yet rather than being handed commands that fail.

Verification wants both directions: an agent with a working pairing is given the commands, and an agent without one is told plainly they will not work — a document that always prints the same section proves nothing about whether it read the state.

Blocked on #509, which adds the send tool.

JAFairweather:
Closes #512. **Stacked on #509**, which is stacked on #506. Review those first.

## How this was found

I rehearsed the onboarding walk in a clean room — fresh clone, empty `HOME`, nothing from the operator's machine — because the end-to-end Pi walk needs a Pi but the *procedure* does not. Any step that fails there fails on the Pi too.

```
$ grep -rln 'agent-inbox\|agent-send' --include='*.md' --include='*.html' --include='*.mjs' .
tools/agent-send.mjs
tools/agent-inbox.mjs
```

The two tools that **are** participation appeared nowhere outside their own file headers. Not in `docs/AGENT_BRIEF.md`, not in the generated startup document, not in the console's first prompt.

So the pasted prompt told a new agent that it posts in as a first-class member, that what reaches it is the return lane, and — rule 5 — that a body with no `@name` reaches nobody. Then it gave the agent no command to listen with and no command to speak with. The entire shape of participation, and none of the mechanism.

## What changed

A `## How you listen, and how you speak` section, rendered from install state like the rest of the document. An agent whose pairing is not in place is told the commands will not work **and which piece is missing**, instead of being handed commands that fail in a way that looks like the lane being down.

The console twin carries the same section. It cannot import the artifact table, so it inlines the three titles the way it already inlines the four state constants.

## Two assertions that could not fail

Both found by mutation, both pre-existing in kind:

- **`console_first_prompt` measured the never-checked block by keyword across the whole document.** Any other section naming `connect-agent --check` broke it — and what it reported was a regression in a collapse that had not happened. Now scoped to the section it is measuring, with a non-empty guard so an empty slice cannot pass every filter beneath it.
- **The new warning's own test split on a marker and tested the tail**, which matched "Bunker pairing" forty lines below in the install-state section. Bounded to the warning line.

## Verification

7 mutations, no survivors. The first round had **four** survivors and each was a real weakness in my tests, not noise: an alternation that passed with either half deleted, the unbounded split above, and a case the fixtures never covered — a document built from no install state at all, which is exactly the bug this commit fixes. One mutant survived correctly and was replaced as equivalent: `laneOpen` already holds every non-PRESENT piece and unknown is a subset of non-present, so filtering over either yields the same set.

Suite unchanged at 107 — this touches existing suites rather than adding one.

## Review

**@My Dude** — docs-and-tests by the letter, but it changes what every future agent reads at session start, so it is worth a read. The thing to push on: whether the warning can be reached in a state where it says "works" and does not.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "fcaaa79527c28ee075c8e8917f7edcec"
+  "build": "9dfedec4a8a22e5fdccf5a7cbb038c40"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "68f5a100c2eac8d5944d85b9b7fec0aa"
+  "build": "fcaaa79527c28ee075c8e8917f7edcec"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -38,6 +38,8 @@ const LANE_TITLES = {
 
 // Anything that looks like a credential. Checked against the rendered text, not against the inputs,
 // because the failure that matters is what reaches the file.
+const HEX64 = /^[0-9a-f]{64}$/
+
 const SECRET_SHAPES = [
   [/bunker:\/\//i, 'a bunker:// pairing URI'],
   // The OTHER NIP-46 pairing URI, and the same credential class. Missing it in a tool whose whole
@@ -85,7 +87,7 @@ const SAYS = {
  * works. `facts` carries the public identifiers. Everything else here is invariant role text, and
  * every line of it is a rule an agent has broken confidently at least once.
  */
-export function startupDoc({ agent, pubkey, channel, runtimeLabel, briefPath = 'docs/AGENT_BRIEF.md', report,
+export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, briefPath = 'docs/AGENT_BRIEF.md', report,
   writtenBy = '`tools/connect-agent.mjs --startup`' }) {
   if (!agent) throw new Error('startupDoc needs the agent name')
   const rows = report?.rows || []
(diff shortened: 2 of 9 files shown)


JAFairweather:
Routed to @My Dude over the sealed lane, 2026-08-15 — carried, `RELAY[buzz] ok -> a8186b53… (wrap a4a93bc51da4…)`. Recording it here because the review request goes to the channel, not to GitHub: a PR with zero GitHub comments looks unrouted, and that misread is what left this one unrouted for days.

JAFairweather:
Force-pushed: rebased onto the fixed #506. This branch was carrying the pre-fix `tools/agent-inbox.mjs` inherited from an older `feat/505`, which would have reverted #506's fix on merge. Own work unchanged; see #509 for the detail. Top of stack: 107 suites, rc=0, eslint 0 errors.

JAFairweather:
**Changes requested** — one live defect in the section this PR exists to add. Read at `e3a87e89`. `npm test` rc=0.

## Must-fix: the documented "To speak" command cannot run

```
**To speak:** `echo "@Name — your message" | node tools/agent-send.mjs --channel <uuid>`
```

Run exactly as printed:

```
$ echo "@My Dude — test" | node tools/agent-send.mjs --channel a8186b53-…-b6486c58970e
agent-send: --bridge <64-hex> (or WAGGLE_BRIDGE_PUBKEY) is required — this seals to waggle's own key and will not guess it
exit=3
```

`tools/agent-send.mjs:33-34` requires `--bridge` or `WAGGLE_BRIDGE_PUBKEY`, and it is the first check in the file — ahead of the signer, ahead of the body. Nothing in the onboarding path supplies it: `WAGGLE_BRIDGE_PUBKEY` appears in two files in this repo (`tools/agent-send.mjs`, `tools/mint-consent.mjs`), and neither `connect-agent` nor the manifest writes it. The rendered document never names `--bridge`, never names the variable, and never gives the agent waggle's pubkey — that is the full render, not the source.

The warning block below is where a reader would learn something is missing, and it does not fire here: it renders only when a lane piece is not `PRESENT`. The fully-installed agent — the one this section is written for — gets the command and no caveat.

`agent-inbox --pubkey <64-hex> --watch` is complete as printed; `--relays` and `--trust` both default. Only the speak half is short.

## Should-fix: the "never checked" caveat is all-or-nothing

`src/agent_startup.mjs:206` gates the caveat on `laneUnknown.length === laneOpen.length`. One genuinely `MISSING` piece among the three silences it for the other two, which are then stated flatly:

```
bunker-uri MISSING, the other two never supplied
  ⚠ **Neither command works yet.** Bunker pairing, Client transport key, Signer resolves to the
    right key are not in place.
```

Two of those three were never examined. That is the defect `src/agent_install_state.mjs:27-31` records as the reason the fourth state exists — "missing sends an operator to create a thing that may already exist, and for a key or a published profile, creating a second one is not a harmless no-op."

Latent today, and I want to be exact about that: neither producer reaches it. `console/connect.html:546-548` marks all three `UNKNOWN`; `connect-agent` calls `see()` on all three with booleans (`tools/connect-agent.mjs:102`, `:103`, `:113`), so none of them is ever `UNKNOWN` beside a `MISSING` sibling. Your new test covers the all-`UNKNOWN` case only. Per-piece is a one-line change and holds when a fifth state starts flowing through this predicate — which is #515, one PR up: `laneState(k) !== PRESENT` will read `NOT_APPLICABLE` as "not in place" and `laneUnknown` will not count it.

## Nit

"…are not in place — or **was** never checked" — plural verb and singular verb in one sentence. It fires on the all-`UNKNOWN` case, which is the console's every render. `tests/agent_startup.mjs:228` matches the substring, so the wording is pinned.

JAFairweather:
## Must-fix, should-fix and the nit are all closed — `fed94982`

**The speak command runs now.** `startupDoc` takes `bridge` and renders it into the command; `tools/connect-agent.mjs` passes `--bridge` or `WAGGLE_BRIDGE_PUBKEY`, and `console/connect.html` passes `loadBridgeKey()`. When neither knows it, the flag is still printed — with the placeholder and a caveat saying the writer did not know waggle's key, that the tool refuses to guess it and exits before signing anything, and that it is a public key so nobody withholds it as a credential. An omitted flag is what made the command look complete.

Both directions asserted in new block `5c` of `tests/agent_startup.mjs`: with the key known, the rendered line carries the actual key and no caveat; with it unknown, the flag is still there and the caveat fires. The negative control is the second half — a caveat that always fires and one that never fires fail identically.

**The all-or-nothing caveat is per piece now.** Each open lane row carries its own verb: never-checked reads "was never checked, which is not the same as absent", missing reads "is not in place". One `MISSING` piece no longer states two unexamined ones flatly as absent. That also removes the plural/singular disagreement — the nit was the same sentence.

`console/connect.html` was reading no bridge key at all, so `tests/console_bridge_key.mjs` would have failed the new consumer. Rather than exempting it, the control now asserts the rule it actually protects: a consumer may read the key, and must never prefill a field with it.

Rebased onto #518 (which replaces #509 — that one merged into a branch, not into `main`). 107 suites rc=0, eslint 0 errors.

JAFairweather:

JAFairweather:
**Approved** — all three closed, verified in rendered output rather than in the diff. Read at `fed94982` from the top of the stack (`c6c45cda`); `npm test` rc=0, 107 suites.

I rendered the document from a clean instance root with no bridge key and read the section:

```
**To speak:** `echo "@Name — your message" | node tools/agent-send.mjs --channel <uuid> --bridge <waggle's 64-hex>`
⚠ **`--bridge` is not filled in above** — whatever wrote this did not know waggle's own
public key. The tool refuses to guess it and exits before signing anything. Ask for it, or
set `WAGGLE_BRIDGE_PUBKEY`; it is a public key, so it is not a secret and not a credential.
```

The flag is present as a flag, not as prose, which is the part that matters — an agent that copies the line and fills the placeholder gets a command that runs, and one that copies it blind gets a refusal it was warned about rather than an exit 3 it has no account of. Printing it always and caveating when unknown is the right shape; omitting it was what made the command look complete.

Per-piece verbs close the should-fix. The same render:

```
⚠ **Neither command works yet.** Bunker pairing is not in place; Client transport key is not
in place; Signer resolves to the right key is not in place.
```

Each row carries its own verb, so a `MISSING` sibling can no longer state an unexamined one flatly as absent. That was latent when you wrote it and stops being latent one PR up, which is the reason to have fixed it here rather than tracked it.

Three things I checked that you did not claim:

- `laneTitle` falls back through `ARTIFACTS` rather than a local copy, so the titles cannot drift out of step with the artifact table. Right call.
- The rendered `--bridge` value is a 64-hex public key and does not trip `SECRET_SHAPES`. Confirmed against the rendered text, which is where that sweep runs.
- `loadBridgeKey()` in `console/connect.html:574` is the one producer that legitimately has this value, and `tests/console_bridge_key.mjs` now asserts the rule it actually protects (may read, must never prefill) rather than exempting the new consumer. That is the correct way to widen a control.

One nit, no action needed unless it bothers you: this section opens "Both lanes run on the signer above", meaning inbox and send. One PR up, `--lane` starts meaning `sealed|broker`, and the same document then prints `connect-agent --check --lane sealed` about twenty lines below that sentence. Two meanings of "lane" on one page.

JAFairweather:
Rebased onto `main` at `87e9c5fb` and retargeted from `feat/507-agent-send`, which merged as `07654127`.

The rebase was `--onto origin/main 699b0c8c`, not a plain `git rebase origin/main`. The plain form replays #507's four commits, which are already on `main` inside the squash — every one of those files conflicts add/add, and resolving with `--theirs` restores the pre-review `\w` lookbehind that #518 was approved for fixing. Same shape as the stale `agent-inbox.mjs` this stack carried last time; checked for rather than discovered.

Verified after: `src/relay_send_intent.mjs` on this branch still has the Unicode lookbehind. 108 suites rc=0, eslint 0 errors.

PR #515, opened 2026-08-15
A correctly-onboarded agent was told it could not run (#513)

### LINKED ISSUE #513: Install state has no notion of the sealed lane, so a broker-less agent reports "cannot run"
Found in the same clean-room rehearsal as the issue above.

`connect-agent.mjs --check` on a fresh agent reports **10 required pieces missing — this agent cannot run**, exit 1. Two of those ten are the broker:

- `channel-host-key` — "run the broker doctor on the broker and write its host key… This tool will never mint one"
- `mcp-registration` — "no channel host — pass `--channel-host <host>`"

Both are `blocking: true` in `src/agent_install_state.mjs` with no way to say otherwise.

But the first-principles review concluded the broker is an **alternative** transport, not a requirement. waggle already runs the sealed lane in both directions: `returnLaneSend` pushes to the agent, `handleRelayIngress` accepts from it. An agent participating that way needs a bunker pairing and nothing else — no ssh account, no broker instance, no seated host key. That was the whole finding.

So the install-state model and the architecture disagree. An agent onboarded exactly as the plan describes is told it cannot run, and the tool exits 1 saying so. That is the wrong answer given confidently, which is the failure mode this repo cares most about.

What it needs is a lane concept: an agent declares how it participates, and the required set follows. Sealed lane → pairing, manifest, directories, relay list; broker → those plus the host key and the MCP stanza.

Two things not to get wrong:

1. **The broker items must become `not applicable`, never `ok`.** An agent that skips them has not satisfied them. Reporting them as satisfied is how a missing artifact reads as a present one, which is the exact class of defect the four-state model (`present / unverified / missing / unknown`) exists to prevent.
2. **The default must not silently become the permissive one.** If an agent does not declare a lane, the answer is `unknown`, not "assume sealed and pass".

Verification, both directions: a sealed-lane agent with a pairing reaches exit 0 without a broker; a **broker-lane** agent still fails without its host key. A gate that passes everything and one that passes the right things fail identically.

Blocked on #509 and the startup-document issue.

JAFairweather:
Closes #513. **Fourth in the stack** — #506 → #509 → #514 → this. Review those first.

This is the one that was blocking the Pi walk.

## The failure

`connect-agent --check` on a fresh sealed-lane agent:

```
10 required pieces missing — this agent cannot run.
exit 1 (incomplete)
```

Two of the ten were the broker's — its host key, and the MCP registration that ssh's to it. But the first-principles review settled that the broker is an **alternative** transport, not a requirement: waggle already runs the sealed lane in both directions, and an agent on it authenticates by signature. The install-state model and the architecture disagreed, and the model was the one giving a wrong answer with confidence.

The startup document in #514 says, three sections up, that neither lane "needs an ssh account, a broker, or a key held on this machine". Then the checker it points the agent at demanded all three.

## What changed

A declared participation lane. Seven rows scope to `broker`; a `sealed` agent reports them `not-applicable`.

```
[ok ] Declared participation lane          — sealed — seal to the bridge, and the bridge seals back
[ ? ] Bunker pairing                     UNVERIFIED  — …
[n/a] The broker's host key              NOT-APPLICABLE  — not part of the sealed lane
[n/a] Registered as an MCP server        NOT-APPLICABLE  — not part of the sealed lane
…
4 present but unchecked, 4 not looked at — being unable to check is not the same as being fine.
Every remaining row is one this build never checks. 7 rows did not apply to the sealed lane and
were not checked — that is not the same as satisfied.

exit 3 (inconclusive — the best result this build can report)
```

**Exit 0 is still unreachable by construction (#492), and this does not change that.** It changes exit **1**. "The check now passes" is the summary a reader reaches for and it is not what happens — a sealed-lane agent reaches the ceiling, which is the honest answer.

## The two things the issue said not to get wrong

**`not-applicable` is never `present`.** Its own state, its own glyph, and the seven rows are still rendered rather than dropped — a scope the reader cannot see is a scope nobody can audit. The headline says how many were skipped and that skipped is not satisfied.

**Undeclared is not sealed.** With no `--lane`, every row stays required and the lane row itself reads UNKNOWN, so the report is inconclusive rather than permissive. An unrecognised name is not a declaration either — the tool refuses it and names the lanes it takes, before rendering anything.

Also: the sealed lane no longer mints a broker ssh key. Key material created just in case is key material nobody is tracking.

And the startup document's remedy line now reads `--check --lane sealed`, in both the runtime copy and the console twin, or it hands the agent the command that produced the wrong answer.

## Found while measuring

`String(['sealed'])` is `'sealed'`. Without an explicit `typeof === 'string'`, a single-element array declared a lane and scoped seven rows out. That was a test assertion failing, not reasoning — the same sweep is why the lookup is `hasOwnProperty` and not `in`, so `'toString'` cannot name a lane.

## Verification

Both directions on every claim that matters, because a scope that excuses a row is one line from a scope that excuses every row and both look like a green suite:

- a sealed-lane agent with its artifacts in place is not told it cannot run — **and** a broker-lane agent with no broker artifacts still is, naming all seven
- the *same* observations pass sealed and fail broker, so the lane is what differs, not the box
- live: `--lane sealed` renders exactly seven `n/a` rows — **and** the same run with no `--lane` renders none and says why
- the negative control on the box: sealed lane with the pairing removed still exits 1 and names it

11 mutations, no survivors. The two that would have mattered most are L2 (`applies` reduced to `!declared`, excusing every row) and L3 (undeclared defaulting to `sealed`).

Suite unchanged at 107 — this extends `agent_install_state`, which is where the model's tests already live.

## Review

**@My Dude** — this one changes what a gate requires, so it is substantive by the letter. The thing to push on: whether there is a path where `applies()` returns true for a row that should have been demanded. That is the failure that would ship an agent reading as installed while a credential it does need was never asked for.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "9dfedec4a8a22e5fdccf5a7cbb038c40"
+  "build": "dd13e8336b81871b9ec9fce92b365187"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "dd13e8336b81871b9ec9fce92b365187"
+  "build": "c478d5ef6ff8287cba196c78088e7342"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -255,7 +255,7 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
     out.push(`- **${neverChecked.length} further artifact${neverChecked.length === 1 ? ' was' : 's were'} never checked` +
       ` — do not assume either way:** ${titles}.`)
     out.push(`  Whatever wrote this could not observe ${neverChecked.length === 1 ? 'it' : 'them'};` +
-      ` run \`connect-agent --check\` on the agent's own machine to settle ${neverChecked.length === 1 ? 'it' : 'them'}.`)
+      ` run \`connect-agent --check --lane sealed\` on the agent's own machine to settle ${neverChecked.length === 1 ? 'it' : 'them'}.`)
   }
   out.push('')
   if (open.length) {
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -36,6 +36,12 @@ const LANE_TITLES = {
   'signer-identity': 'Signer resolves to the right key',
 }
 
+// The lane names, inlined for the same reason and exported so `tests/console_first_prompt.mjs` can
+// pin them against `LANES` in `src/agent_install_state.mjs`. A stale list here does not render a
+// wrong sentence — it renders a `--lane` flag the node side would reject, which is a command an
+// agent is told to run and cannot.
+export const LANES = Object.freeze({ sealed: true, broker: true })
+
 // Anything that looks like a credential. Checked against the rendered text, not against the inputs,
 // because the failure that matters is what reaches the file.
 const HEX64 = /^[0-9a-f]{64}$/
@@ -98,6 +104,16 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
   // true of the system whatever this install looks like, and mixing the two is how a document tells
   // an agent it holds a pairing it does not have. Absent row reads as never-checked, never as fine.
   const stateOf = key => { const r = row(key); return r ? (SAYS[r.state] || r.state) : SAYS[UNKNOWN] }
+
+  // The check command this agent should actually run, rendered from the lane it was checked under.
+  // `--lane` is not cosmetic: declaring `sealed` scopes six broker rows out (#513), so a document
+  // that hardcodes one lane tells an agent on the other one to scope out the rows it depends on —
+  // `applies` refusing to assume the cheaper lane, arriving through the document instead of the
+  // flag. Undeclared renders no flag at all, for the same reason `installState` treats undeclared
+  // as "every row applies": absent is not sealed. An unknown name is dropped rather than echoed,
+  // because a command printed with a lane `installState` would reject is a command that fails.
+  const lane = Object.prototype.hasOwnProperty.call(LANES, String(report?.lane)) ? String(report.lane) : null
+  const checkCmd = `connect-agent --check${lane ? ` --lane ${lane}` : ''}`
 
   // Only the rows an agent's own behaviour depends on. A wall of install state is a wall nobody
   // reads, and this file competes for the top of a context window.
@@ -242,7 +258,7 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
       : `${laneTitle(k)} is not in place`
     out.push('')
     out.push(`⚠ **Neither command works yet.** ${laneOpen.map(laneSays).join('; ')}.`)
-    out.push(`Settle that first with \`connect-agent --check\`; running either tool before then fails in a`)
+    out.push(`Settle that first with \`${checkCmd}\`; running either tool before then fails in a`)
     out.push(`way that looks like the lane being down.`)
   }
   out.push('')
@@ -255,7 +271,7 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
     out.push(`- **${neverChecked.length} further artifact${neverChecked.length === 1 ? ' was' : 's were'} never checked` +
       ` — do not assume either way:** ${titles}.`)
     out.push(`  Whatever wrote this could not observe ${neverChecked.length === 1 ? 'it' : 'them'};` +
-      ` run \`connect-agent --check --lane sealed\` on the agent's own machine to settle ${neverChecked.length === 1 ? 'it' : 'them'}.`)
+      ` run \`${checkCmd}\` on the agent's own machine to settle ${neverChecked.length === 1 ? 'it' : 'them'}.`)
   }
   out.push('')
   if (open.length) {
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = 'dd13e8336b81871b9ec9fce92b365187'
+export const CONSOLE_BUILD_ID = 'c478d5ef6ff8287cba196c78088e7342'
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = '9dfedec4a8a22e5fdccf5a7cbb038c40'
+export const CONSOLE_BUILD_ID = 'dd13e8336b81871b9ec9fce92b365187'
--- a/src/agent_install_state.mjs
+++ b/src/agent_install_state.mjs
@@ -19,7 +19,7 @@
 //   not-applicable — this agent's declared lane never needed it. NOT the same as satisfied.
 //
 // The lane (#513) exists because there are two ways to participate and this model could describe
-// only one of them. Seven of the rows below reach an ssh channel on the broker box; an agent on
+// only one of them. Six of the rows below reach an ssh channel on the broker box; an agent on
 // the sealed lane authenticates to the bridge by signature and needs none of them. Until it could
 // say so, `--check` told a correctly-onboarded agent it could not run — a wrong answer given
 // confidently, which is the failure this file was written to prevent.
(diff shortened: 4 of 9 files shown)


JAFairweather:
Re-walked the clean room this came from — fresh clone, empty `HOME`, nothing from the maintainer's machine:

```
no --lane        10 required pieces missing — this agent cannot run.
--lane sealed     7 required pieces missing — this agent cannot run. 7 rows did not apply to
                  the sealed lane and were not checked — that is not the same as satisfied.
```

Worth being exact about what that shows and what it does not. The clean room has no bunker pairing and no manifest, so exit 1 is the **correct** answer there — the agent genuinely is not installed. What changed is that three of the ten are gone: the ones it was never going to have.

The at-the-ceiling case needs an agent that actually has its own artifacts, and that ran separately: a sealed-lane agent with a pairing and a manifest reaches `exit 3 (inconclusive — the best result this build can report)`. The same box on `--lane broker` exits 1 with four missing, and the same box on `--lane sealed` with the pairing removed exits 1 and names it.

JAFairweather:
Force-pushed: rebased onto the fixed #506. This branch was carrying the pre-fix `tools/agent-inbox.mjs` inherited from an older `feat/505`, which would have reverted #506's fix on merge. Own work unchanged; see #509 for the detail. Top of stack: 107 suites, rc=0, eslint 0 errors.

JAFairweather:
**Changes requested** — two scoping problems, both reproduced live at `d7ddfb6b`. `npm test` rc=0, 107 suites, working tree clean.

## Must-fix: the startup document hardcodes `--lane sealed`

`src/agent_startup.mjs:207` and `:220`, and both console copies:

```
Settle that first with `connect-agent --check --lane sealed`; …
… run `connect-agent --check --lane sealed` on the agent's own machine to settle them.
```

Rendered with the lane declared the other way:

```
$ node tools/connect-agent.mjs --name probe --root … --lane broker --check --startup --print
  | Settle that first with `connect-agent --check --lane sealed`; running either tool before then …
  |   Whatever wrote this could not observe them; run `connect-agent --check --lane sealed` on the …
```

A broker-lane agent that follows its own onboarding document scopes out the seven rows it needs. That is what `applies` refuses — "assuming the lane with fewer requirements is how a missing credential reads as satisfied" — arriving through the document instead of the flag.

`report.lane` is already in the object `startupDoc` receives: `installState` returns it at `:275`, and `agent_startup.mjs` never reads it. This PR touches neither `tests/agent_startup.mjs` nor `tests/console_first_prompt.mjs`, and nothing renders the doc under `--lane broker`. The twin binding holds the two copies byte-identical, so it will hold them identically wrong.

## Must-fix: `mcp-exclusive` is scoped by list position, not by what it reads

Six of the seven `lanes: ['broker']` rows reach an ssh channel on the other box. `mcp-exclusive` does not — it reads **this machine's** MCP runtime configs, through `foreignNvoyServers` → `foreignServers`. Same box, same commit, one flag apart:

```
--check                 [ x ] No other nvoy server registered   MISSING — also registered:
                              <a foreign nvoy server> (Codex CLI) — these carry the tools that
                              SIGN, and not as probe. Remove with `codex mcp remove …` before
                              acting. (Claude Code went unread, so there may be more.)

--check --lane sealed   [n/a] No other nvoy server registered   NOT-APPLICABLE — not part of the
                              sealed lane
```

The server name, the remove command and the "went unread, so there may be more" caveat go with it. #338's hazard is a session carrying signing tools bound to another identity; that belongs to the session, not to the transport. The row's own `why` names no broker and no ssh — it travelled with the block it sat in.

It is worse on the sealed lane specifically. `foreignServers(names, name)` excludes `nvoy-<name>`, and a sealed-lane agent has no such server — so *every* nvoy server in that session counts as foreign. The one lane where this row has nothing of its own to compare against is the lane it was scoped out of.

`mcp-registration` and `mcp-identity` are correctly scoped: both ask about `nvoy-<name>`, which a sealed-lane agent has no reason to own.

## The arithmetic checks out

`:141` says declaring `sealed` scopes three of the eleven ceiling rows out. I checked it because `:22` and `:104` both say seven and a reader will trip on that: the eleven are 5 `NEVER_CHECKED` + 6 `NEVER_VERIFIED`, and exactly `channel-authorized`, `channel-host-key` and `channel-answers` among them carry `lanes: ['broker']`. Three out, eight left. Correct as written.

## The lane suite is behavioural, and it holds

One mutation, `const applies = artifact => true`:

```
FAIL — a sealed-lane agent reports all seven broker rows NOT-APPLICABLE
FAIL —   …and not one of them as PRESENT — "did not apply" is never "satisfied"
FAIL —   …and the present count drops by exactly seven — the rows left the tally, they were not renamed into it
FAIL —   …with their own glyph, never a tick
FAIL —   …and the headline says how many were skipped, and that skipped is not satisfied
FAIL —   …and renders exactly seven n/a rows
```

Six checks, one of them the live subprocess run. The `String(['sealed'])` and `--lane brokr` cases are real refusals with the reason asserted, not source scans. This is a different standard from the wiring in #506 and #516, where a realistic disable survived.

JAFairweather:
## Both must-fixes closed — `c6c45cda`

**The hardcoded lane.** `report.lane` is read now, in both copies. `startupDoc` builds one `checkCmd` and both remedy lines render it. Live, same box, one flag apart:

```
--lane broker   | Settle that first with `connect-agent --check --lane broker`; …
                |   …run `connect-agent --check --lane broker` on the agent's own machine…
--lane sealed   | Settle that first with `connect-agent --check --lane sealed`; …
no --lane       | Settle that first with `connect-agent --check`; …
```

Undeclared prints no flag, for the same reason `installState` treats silence as "every row applies" — a document that supplied the flag anyway would talk an undeclared agent into the permissive reading. An unrecognised name is dropped rather than echoed: a command carrying a lane `installState` would refuse is a command that fails.

New block `5d` in `tests/agent_startup.mjs`, six checks, all three branches plus the bogus lane. Negative control: restoring the literal `--lane sealed` fails four of them, and the restore is clean.

The twin binding was blind to this — `report.lane` is a field on `report`, not a parameter, so nothing in the matrix walked it, and both copies could have hardcoded the same lane and stayed byte-identical. `tests/console_first_prompt.mjs` gains all three lane branches in `MATRIX`, a per-copy assertion that each actually renders the lane, and a pin that the browser copy's inlined lane list matches `LANES`.

**`mcp-exclusive`.** `lanes: ['broker']` removed and the row moved out of the block, with a comment saying why it is not there. Broker rows are six now, not seven — counts updated in the source comments and the suite. The ceiling arithmetic is untouched: `channel-authorized`, `channel-host-key` and `channel-answers` are the three of the eleven, and none of them moved.

Live on the lane it was scoped out of:

```
$ node tools/connect-agent.mjs --name probe --root … --lane sealed --check
[ x ] No other nvoy server registered   MISSING — also registered: <a foreign nvoy server> …
```

Two new assertions in `tests/agent_install_state.mjs`, both directions: the row is scoped to no lane, it is still `PRESENT` (checked, not skipped) on a sealed-lane report, and a foreign server still fails a sealed-lane agent.

**Not fixed, and flagging it rather than deciding it:** `console/connect.html` builds its report with no `lane`, so the pasted prompt still says the undeclared `connect-agent --check`. Declaring `sealed` from the page would be the console asserting which flow the operator ran, and if they also seated a broker it scopes out six rows they need — "assume the lane with fewer requirements", from the producer this time. Left undeclared deliberately. Worth an issue if you disagree.

Rebased onto the fixed #514. 107 suites rc=0, eslint 0 errors.

JAFairweather:

JAFairweather:
**Approved** — both must-fixes closed. Read at `c6c45cda`; `npm test` rc=0, 107 suites, and I drove the tool rather than the suite.

## Live, on a clean instance root, four ways

```
--check --lane sealed    exit 1 · 6 rows n/a · "8 required pieces missing — this agent cannot run.
                                 6 rows did not apply to the sealed lane and were not checked —
                                 that is not the same as satisfied."
--check --lane broker    exit 1 · 0 rows n/a — all six required again  (positive control)
--check                  exit 1 · 0 rows n/a · [ - ] Declared participation lane UNKNOWN
--check --lane sealled   exit 1 · "connect-agent: --lane must be one of: sealed, broker"
                                                                       (negative control)
```

The broker run is the one I care about most: a lane flag that scopes rows out and a lane flag that scopes nothing out are indistinguishable from the sealed run alone. Six rows come back. The `n/a` glyph is its own mark and no scoped-out row renders a tick, so the left column cannot be scanned into a false green.

`mcp-exclusive` is now unscoped and required on both lanes — confirmed in the table, not just the diff: 19 artifacts, 6 broker-only (`channel-key`, `channel-host-key`, `channel-authorized`, `mcp-registration`, `mcp-identity`, `channel-answers`), 13 that apply everywhere. Your reasoning for why it was worst on the lane it was excluded from is right, and the comment now says so at the row.

## The arithmetic, re-derived independently

`NEVER_CHECKED` is 5 and `NEVER_VERIFIED` is 6, so eleven of nineteen. Broker-only among those eleven: `channel-authorized`, `channel-host-key`, `channel-answers`. Three out, eight left. Matches.

## One precision point, no change requested

"exit 0 is unreachable by construction" is a property of **`connect-agent`**, not of `installState`. Handed synthetic best-case observations, the module returns exit 0 on every lane:

```
sealed      outcome=complete exit=0  {"present":13,…,"notApplicable":6}
broker      outcome=complete exit=0  {"present":19,…,"notApplicable":0}
undeclared  outcome=complete exit=0  {"present":19,…,"notApplicable":0}
```

What actually makes it unreachable is that the producer hardcodes the eight remaining ceiling rows: `verified: false` at `tools/connect-agent.mjs:123`, `:125`, `:136`, `:138`, and `found: null` at `:144`, `:145`, `:146`, `:151`. So the claim is true today and true for the reason the comment gives, but it is enforced in a different file from the one asserting it — and a second producer supplying `verified: true` would get exit 0 without tripping anything. Worth a line at `agent_install_state.mjs:145` naming where the guarantee actually lives, since that comment is what a future producer will read.

## On `connect.html` and the undeclared lane — leave it as it is

You made the right call, and I would not open an issue. Three reasons, in order of weight:

1. Declaring `sealed` from a page that never asked which flow the operator ran is exactly the move `applies` exists to refuse. The undeclared branch is the demanding one on purpose, and the console is the one producer with no evidence either way.
2. The page's own report already contradicts a sealed declaration. `connect.html:551-553` enumerates `mcp-registration` and `mcp-identity` as unseen — both broker-lane rows. The console's model is already "this might be a broker agent", and declaring sealed would scope out rows the same function just listed.
3. The undeclared path is not a dead end. I ran it: the no-lane check renders `[ - ] Declared participation lane UNKNOWN — no --lane given, so every row stays required — an undeclared lane is not a sealed one`. The agent gets the flag named, at the moment it matters, from the machine that can actually answer. The console handing off undeclared routes to the right answer instead of guessing it.

One small thing worth taking while you are here: that note names `--lane` but not its values, so a reader learns them by typing one wrong and reading the refusal. `--lane <sealed|broker>` in the note closes the loop.

## Merge order

Stack is #518 → #514 → #515, and merging to `main` deploys. Squash-merging a parent leaves the child showing content already in `main`; land them in order and re-read each child after its base lands rather than trusting the green from before. Three merges is three production restarts — worth spacing them.

JAFairweather:
Rebased onto the rebased #514 at `777bb8af`; base stays `feat/512-startup-lane-commands`, which is now itself on `main`.

Same `--onto` treatment as #514 — replaying #507's commits would have restored the pre-review `\w` lookbehind #518 was approved for fixing. Checked after: the Unicode lookbehind is present on this branch. 108 suites rc=0, eslint 0 errors.

Merges after #514.

PR #516, opened 2026-08-15
The alert hashtag reached nobody (#508)

### LINKED ISSUE #508: A message carrying only the alert hashtag is carried to nobody — the detector matches at-words and replies only
The return-lane detector (`scanReturnLane`, `src/bridge.mjs`) decides what gets carried out to an
agent from three signals: a direct-reply marker (`e` tag with `reply`), a `p` tag naming the
recipient, and an at-word matched against the recipient's mention name with longest-name arbitration.

A hashtag matches none of them. So a channel message carrying only the agreed alert tag — the case
where somebody is flagging something for whichever agent is watching, rather than naming one — is
carried to nobody, and nothing logs a refusal because nothing considered it a candidate.

This is the second half of "mentions, or anything with the hashtag we've chosen to be an alert, push
into the agent's context". The first half works; this half was never built.

Shape:

- The tag is configured per recipient, not global. An agent subscribes to the alerts it wants, and an
  agent that subscribes to none behaves exactly as today.
- Match the `t` tag as Nostr defines it, and the in-body `#word` form, because people write the
  second and clients normalise to the first inconsistently. Case-insensitive; a `#word` that is part
  of a longer word does not match, mirroring the at-word rule already in `task_route_mention.mjs`.
- Fan-out and dedup are unchanged: one message reaching an agent by both an at-word and a hashtag is
  one carry, deduped on the existing (source × recipient) key.
- Every non-carry keeps its logged reason, like every other refusal in this lane.

Negative control the suite has to carry, and it is the whole test: a message carrying a **different**
hashtag is seen **not** to be carried, and an agent with no configured tag is seen to be unaffected.
A matcher asserted only to fire cannot tell "carries the right thing" from "carries everything" — and
this one fans out to a live channel, so a matcher that is too broad is worse than one that is absent.

JAFairweather:
Closes #508.

`scanReturnLane` read three signals — a direct-reply `e` marker, a `p` tag, and an at-word arbitrated by `taskRouteMentionArbitrate`. A hashtag matched none of them, so a message flagging something for whichever agent is watching reached nobody, and nothing logged a refusal, because nothing ever considered it a candidate.

## Shape

**Per recipient, never global.** An agent subscribes to the alerts it wants via `alert_tags` on its `return_lane` row; an agent that subscribes to none behaves exactly as it does today. A global tag would make every configured agent a recipient of every alert in the channel — a fan-out dressed as a feature.

**Default-closed.** This matcher fans out to a live channel, so one that is too broad is worse than one that is absent: it puts messages nobody meant for an agent into that agent's context, and every one of them looks like a correct carry from the inside. Two deliberate tightenings:

- The alphabet is `\p{L}\p{N}_` — narrower than `MENTION_ALPHABET`, which also carries `.`, `'` and `-`. `#alert.` would otherwise swallow a full stop, and `#alert-2` vs `#alert` would be two tags whose relationship is decided by the boundary rather than by the operator.
- The matcher has a **leading** lookbehind the at-word matcher does not, because `#` is the URL fragment delimiter. Without it `https://example.com/runbook#alert` raises the alert lane, and a pasted link is not somebody raising an alarm.

Both `t` tags and body text are read, because clients disagree about which they write. Reading one and not the other makes the feature work in some clients and not others, invisibly from inside the channel.

## The near-miss, and why it moved

`why` reaches an allowlist in **both** return-lane carry templates, and `reject()` throws out of the fan-out loop — so a value outside the list does not fail one carry, it takes every **later** recipient in that scan with it. That is the #168 shape exactly.

A first draft here built `` `alert:${tag}` ``, which would have done precisely that on the first alert ever raised. I caught it by reading `src/nostr_egress.mjs`. The mutation run then showed **nothing in 106 suites would have caught it** — so the decision moved into `returnCarryReason`, which owns a closed set, and the suite asserts every reachable return value against the `whys` lists in the templates that enforce it. Reading found it once; the test makes it structural.

## What the mutation run found beyond that

- `alertHashtags` validated `String(value)` and then converted `value`. A number passed validation, failed conversion, and pushed a **null** into the subscription list with `problems` empty — an operator told nothing is wrong while holding a subscription that can never fire.
- A type guard in `alertHashtagHit` that no test could distinguish from its own absence. Removed rather than kept: a guard nothing can see is a second place for the rule to drift, not defence in depth.

13 mutations, 0 survivors. The ones that matter are the widenings — drop the leading lookbehind so a URL fragment fires, drop the trailing boundary so `#alerting` fires, make an empty subscription a wildcard, revert `why` to a value the allowlist rejects.

## Verification

`npm test` — 106 suites, rc=0. `eslint src tools tests console` clean. Every positive in the new suite is paired with a negative: `#alerting`, `#alert2`, `#alert_x`, `https://example.com/runbook#alert`, `page#alert`, `##alert`, a bare word, an at-word, and a different tag are each shown **not** to fire; a non-subscriber in the same fan-out is shown not to be woken while a subscriber in the same shape is.

**Not proven here:** no live channel message has raised an alert on the box. The issue's own verification — a message carrying only the alert hashtag seen to be carried, and the control of a *different* hashtag seen not to be — is a post-merge observation, not something this branch can claim.

Based on `main`, not on the #506 stack, so it can merge independently.

Review: **@My Dude** — it touches a delivery path and a fan-out that can abort mid-scan.


JAFairweather:
Routed to @My Dude over the sealed lane. Recording it here because the request goes to the channel, not to GitHub — zero GitHub comments does not mean nobody was asked.

JAFairweather:
**Changes requested** — the test gap, not the code. Read at `2130c833`. `npm test` exit 0, eslint clean on the four changed files.

## Must-fix: the feature can be entirely dead with 106/106 green

Two mutations, both survivors.

**M1** — `src/bridge.mjs:450`, rename the config key:

```js
alert_tagsMUTATED: (() => { … })(),
```

Every configured subscription is dropped at load. `r.alert_tags` is `undefined` on every route, `alertHashtagHit` returns `null` for every message, the lane is a no-op. `npm test` → **exit 0, ALL PASS.** That is #508's own failure shape one layer up: reaches nobody, logs no refusal.

**M2** — `src/bridge.mjs:3579`, neuter the tag half:

```js
const ttags = tags.filter(t => t[0] === 't' && t[1]).map(t => String(t[1])).slice(0, 0)
```

`npm test` → **exit 0, ALL PASS.** The assertion at `tests/alert_hashtag.mjs:214` still matches, so one of the two sources this module exists to read can be dead and nothing says so.

Both survive because every bridge-side assertion in the new suite is a source-text scan. The harness to close it already exists and is not a new build: `tests/return_lane_scan.mjs` writes a real `config.json`, imports the real `src/bridge.mjs`, and drives the real `scanReturnLane`. One case there kills both mutations — a recipient configured with `alert_tags` carries on a `#tag` message nobody named them in; a second recipient in the same fan-out subscribed to a *different* tag does not; the carry reaches `returnLaneSend` with `why === 'alert'`; and one case with the tag in a `t` tag rather than the body.

I am not claiming the code is wrong. I traced `activeReturnLane` (`src/bridge.mjs:925`) — it filters `PUB.returnLane` and preserves the field, and dynamic rows omit `alert_tags` so they default closed. The claim is narrower: nothing would tell us if it were.

## The two you asked me to push on

**Leading lookbehind — keep it, but it does not close the case it was written for.**

```
[see the runbook](#alert)   →  fires
```

The character before `#` is `(`, which is deliberately allowed by your own positive at `tests/alert_hashtag.mjs:57`. So a markdown link target raises the alert lane, same class as the URL fragment and just as ordinary to paste into a channel. JS lookbehind is variable-length, so this is precise and costs nothing:

```js
`(?<![${ALERT_ALPHABET}/#])(?<!\\]\\()#${…}(?![${ALERT_ALPHABET}])`
```

Verified: `[see the runbook](#alert)` → false, `the build is down (#alert)` → still true, `#alert at start` → true, `#alerting` → false.

**Alphabet — `.` is right, `-` I would widen.**

Your stated reason for refusing `-` is that `#alert-2` and `#alert` would then be "two tags whose relationship is decided by the boundary rather than by the operator". That is already true of `#alert` and `#alerting`, and the trailing lookahead settles both the same way. The cost is concrete: `-` is admissible in a Nostr `t` tag and common in practice (`p0-incident`), and the `t`-tag path is a whole-value comparison where no boundary applies at all — so the tightening buys nothing there and loses coverage. Your call; it is one string either way and default-closed today is defensible.

## Worth stating because it is the part that is easy to get wrong

The closed-set chain is sound and has no hand-maintained middle term: `tests/alert_hashtag.mjs:177` proves every reachable `returnCarryReason` output over the full signal matrix is in `RETURN_CARRY_REASONS`, and `:200` proves `RETURN_CARRY_REASONS` is admitted by both templates' `whys`. Composed, reachable ⊆ whys.

## Nit

`if (alerted && !mentioned && !repliedTo)` at `src/bridge.mjs:3649` restates the precedence `returnCarryReason` owns, so the two can drift. `if (why === 'alert')` cannot.

## Merge order

This bumps 105→106 in README, `docs/GETTING_STARTED.md` and CLAUDE.md — and so does #506. They do not conflict in git (both write "106"), so whichever merges second ships a count of 106 against a 107-entry `test` script and `tests/suite_count.mjs` fails. CI catches it; rebase the second one. "Off main, merges independently" is true for conflicts and not for the count.

JAFairweather:
## Both mutations closed, and both of your boundary calls taken — `70248958`

**Must-fix.** You were right that the whole lane could be dead with everything green, and right that the harness already existed. Added behavioural cases to `tests/return_lane_scan.mjs`, which writes a real `config.json` and drives the real `scanReturnLane`:

- two subscribers on **different** tags, so "carries to the subscriber" is distinguishable from "carries to everyone" — the single-subscriber version of this test cannot tell those apart
- a carry from the body; a carry from a `t` tag with nothing matching in the body; an unsubscribed tag carrying nothing
- precedence asserted as behaviour rather than read out of the source

Verified: **M1** (rename the config key) now fails four checks; **M2** (truncate the tag list) fails the t-tag case. Both restore clean.

One process note, because it nearly cost the result: my first run of M1 used the wrong indentation, the anchor missed, and it reported `rc=0` — a pass for a mutation that was never applied. The assert caught it. It is the failure mode the runbook names and it still very nearly landed.

**Markdown link — taken, and it found something I had not seen.** `](` as its own lookbehind rather than removing `(`, because "the build is down (#alert)" is a real alarm. Both directions asserted.

**Hyphen — taken, and your argument is better than my objection.** What settled it was writing the test: I configured `p0-incident` as a fixture and it was **rejected at config load**, so the subscription never existed at all. That is a step worse than the body not matching, and it is what an operator would actually hit. A dot is still refused, so the alphabet stays a decision rather than an absence of one.

**Suite count.** No new suite file, so this stays at 106 and does not collide with #506 — that one merges first at 106 and this needs no rebase for the count. Your warning was right about the general case; it just does not bite here.

@My Dude — one more pass when you have it, on the behavioural cases specifically.

JAFairweather:
**Approved.** Re-read at `70248958`. `npm test` rc=0, working tree clean.

Both mutations die now, on the real harness:

```
M1  alert_tags: → alert_tagsMUTATED:              tests/return_lane_scan.mjs → 6 FAIL
M2  …map(t => String(t[1])).slice(0, 0)           tests/return_lane_scan.mjs → 1 FAIL
      FAIL — a tag carried in a t tag rather than the body still fires the lane
```

M2 failing exactly one check, and that check, is the right shape — the `t`-tag path is the only thing it breaks and the body path is untouched.

Two more I ran against the new surface, because the cases you added are the ones I have not seen fail:

```
M3  swap returnCarryReason to alert-first     tests/alert_hashtag.mjs → 2 FAIL
                                              tests/return_lane_scan.mjs → 1 FAIL
      FAIL — a message that both names the agent and carries its tag is labelled mention
M4  delete the (?<!\]\() lookbehind           tests/alert_hashtag.mjs → 1 FAIL
      FAIL — a markdown link target does not raise the lane
```

M3 dying at both levels is the part worth having: the module owns the precedence and the bridge is proven to consult it, so neither can drift alone.

The widened alphabet, driven rather than reasoned about:

```
FIRE   #p0-incident   paging #p0-incident now  /  … now.  /  the build is down (#p0-incident)
quiet  #p0-incident   see [runbook](#p0-incident)
quiet  #alert         #alert-2 is different  /  x.com/a#alert
FIRE   #alert         "#alert"  /  #alert
config p0-incident → accepted;  -alert → refused, and named
```

## Two nits, both mine to own since the widening was my argument

`ALERT_ALPHABET` does three jobs — config validation, the leading lookbehind, the trailing lookahead — and `-` earns its place in two of them. In the lookahead it is doing real work: `#alert-2` no longer matches `#alert`. In the *lookbehind* it buys nothing and costs a case, `P1-#alert` → quiet. A hyphen before a `#` is not a word continuation the way a letter is.

`alert-` passes `alertHashtagProblem`. `STARTS_WELL` constrains the first character and nothing constrains the last, so a trailing hyphen is configurable. Harmless — it matches, and the `t`-tag path compares whole values — but it is a tag nobody meant to write.

Neither blocks. Nothing else outstanding from me on this one.

PR #518, opened 2026-08-15
The agent can speak, and the envelope does not carry the send time (#507)

JAFairweather:
Reopens #507. **#509 was merged into `feat/505-agent-inbox`, not `main`** — that branch had itself merged thirteen seconds earlier, so GitHub records #509 as MERGED and `tools/agent-send.mjs` never reached `main`. Nothing was lost; the branch survived. This is the same PR, rebased onto `main`, with My Dude's review addressed.

## Must-fix: the seal and the wrap carried the exact send time

Both used `rumor.created_at`. `src/nostr_egress.mjs:496` fuzzes both into the past so an observer cannot correlate a channel message with a delivery by timing, and `src/return_lane_inbox.mjs:64` documents that rule from the reading end — it is why `sealAuthor` does not freshness-check a seal. The agent's writer half was publishing to a public relay the correlation the bridge's writer half removes: a wrap addressed to waggle's key at T, channel post at ~T.

The window is **an hour, not the bridge's two days**, and that is a coupling rather than a preference. The ingress REQ is `{ kinds: [1059], '#p': [BRIDGE_PK], since: SINCE }` with `SINCE = bootTime − SINCE_SECS` (`src/bridge.mjs:113`, `:3892`). A wrap dated before that is never served, and the send then looks like a message nobody answered. Two days is safe only while `SINCE_SECS` keeps its default; an hour is safe for any lookback above an hour and already destroys exact correlation. The reasoning is in the module, because the sibling file states the opposite rule and the next reader will otherwise "fix" this without finding the `since` filter.

## Should-fix: nothing asserted the tool at all

Taken as suggested — the decision moved into `src/relay_send_intent.mjs` as `envelopeTemplates`, which is where a suite can drive it, the same move #506 made with its tracker. `rnd` is injectable so the window is asserted at its bounds instead of sampled, and the shape is asserted against what `handleRelayIngress` requires: kind 13, kind 1059, and the `p` tag.

Twelve checks, both directions. Mutation-checked — restoring `const back = () => at` fails three of them:

```
FAIL   …and does move — a window that never fires is the defect this replaced
FAIL over 200 draws the envelope almost never carries the exact send time (200/200)
FAIL   …and the seal and the wrap are dated INDEPENDENTLY (200/200 collide)
```

## Stack mechanics — the warning was right

Rebasing this onto `main` after #506 squash-merged produced exactly the predicted add/add conflict on all three #506 files, presenting them as independent adds. Resolved to `main`'s side and then verified rather than assumed: `tools/agent-inbox.mjs` is the 173-line fixed version with `track(open(` present exactly once, and diffs clean against `origin/main`. Rebased with `--onto origin/main 1261bd2a` so only this PR's four commits replay.

`npm test` rc=0, eslint 0 errors.

## Not addressed — filed rather than folded in

The four nits (dry-run exit code, no size pre-check, the third copy of the mention alphabet, `\w` in the lookbehind) are real and none is a delivery-path defect. Filing them so this PR stays the timestamp fix.


JAFairweather:
## This replaces #509, which never reached `main`

#509 shows MERGED, but its base was `feat/505-agent-inbox` — not `main` — and that branch had itself merged 13 seconds earlier. `tools/agent-send.mjs` is not on `main`. Nothing was lost; this is the same work rebased onto `main`, plus the privacy must-fix from that review.

**The seal and the wrap no longer carry the send time.** Both were stamped `rumor.created_at`, which publishes to a public relay exactly the correlation `src/nostr_egress.mjs:496` fuzzes away and `src/return_lane_inbox.mjs:64` documents from the reading end as the reason a seal is not freshness-checked. `envelopeTemplates` in `src/relay_send_intent.mjs` now backdates each event independently — two events stamped the same backdated second are themselves a correlator.

It lives in `src/`, not in the tool, for the reason #506 moved its tracker: the tool has no suite, so a decision made there is a decision nothing can assert. The reviewer found this one by reading; nothing in 107 suites could have.

**An hour, where the bridge uses two days, and that is a coupling rather than a preference.** Ingress subscribes `{ kinds: [1059], '#p': [BRIDGE_PK], since: SINCE }` with `SINCE = bootTime − SINCE_SECS` (`src/bridge.mjs:113`, `:3892`). A wrap dated before that is never served, and the send looks like a message nobody answered. Two days of fuzz is safe only while `SINCE_SECS` keeps its default; an hour is safe for any lookback above an hour and already destroys the exact correlation.

`rnd` is injectable, so section 4 of `tests/relay_send_intent.mjs` asserts the bounds rather than sampling them: never in the future, never more than an hour back, never equal to the send time, and the two stamps independent across 200 draws. Plus the kinds, the `p` tag, `FUZZ_SECS`, and a refusal whose reason names the p tag — a wrap with no `p` reaches no one.

Four nits from the #509 review are unaddressed and I will open an issue rather than widen this: the `--dry-run` exit code, no size pre-check before signing, a third copy of the mention alphabet, and `\w` in the lookbehind.

@JAFairweather — this needs a re-read of the envelope change before merge; it is a delivery path.

JAFairweather:

JAFairweather:
**Approved** — the envelope change is correct and does not introduce a silent drop. One should-fix I am raising out of your deferred-nit list, and one nit. Read at `699b0c8c` (reviewed from the top of the stack, `c6c45cda`); `npm test` rc=0, 107 suites.

## The envelope change holds — what I checked, and where

The privacy fix is right, and so is the hour. Backdating is the change that could have cost a delivery path, so I traced it to every reader rather than to the `since` filter alone:

- **The `since` coupling is accurate.** The relay lane subscribes `{ kinds: [1059], '#p': [BRIDGE_PK], since: SINCE }` at `src/bridge.mjs:3892`, and `SINCE = bootTime − 172800` is computed once at boot (`:113-114`). It only recedes further into the past as the process runs, so an hour of backdate is never the reason a wrap goes unserved — under the default, or under any `SINCE_SECS` above an hour. The comment's warning about widening this is the right thing to have written down.
- **`handleRelayIngress` applies no timestamp check at all.** I read `:3364-3411` end to end: dedup, size cap, decrypt budget, seal verify, rumor/seal pubkey bind, `relay` tag, allowlist, grant, body size, rate cap. Nothing reads `created_at` on the wrap, the seal or the rumor.
- **The A5 clamp never touches this lane.** All four `clampCreated` call sites (`:1709`, `:1797`, `:1857`, `:4006`) are in the public kind:1 read lane and `connectPublic`. A backdated wrap is not clamped and cannot be dropped out of range.
- **Latency tracing does not distort.** `markLatency(sourceId, stage, at = Date.now())` (`src/latency.mjs:65`) keys on the event id and stamps with wall-clock, so a backdated wrap does not produce a bogus trace.

Independent stamps are the right call — a shared backdated second is a correlator, and section 4's 200-draw check kills the mutation that reintroduces it. The injected `rnd` bounds check is the part that makes this assertable rather than sampled.

## Should-fix: the `\w` lookbehind lets an email defeat the mention guard

You listed this as a nit. It is not cosmetic — it defeats the guard's stated purpose and puts a false recipient in the report the sender checks.

`\w` is ASCII-only, so `@` preceded by a non-ASCII letter is not excluded. Run against `mentionVerdict` at `699b0c8c`:

```
refused  "mail me at someone@example.com"
PASSES   "mail me at café@example.com"           -> ["example.com"]
PASSES   "ping 김철수@example.com"                 -> ["example.com"]
PASSES   "x@example.com and наташа@mail.ru"      -> ["mail.ru"]
```

The last one is the sharp case: a body with no mention at all passes the guard, is sent, is routed to nobody — and `sendVerdict` reports `Addressed to: @mail.ru`. That is the #118 failure with a recipient line on top of it, which is worse than the bare version, because the sender now has a reason not to go looking.

One character class closes it, and every positive control survives:

```js
/(?<![\p{L}\p{N}_@])@([\p{L}\p{N}][\p{L}\p{N}._-]*(?: [\p{L}\p{N}][\p{L}\p{N}._-]*){0,3})/gu
```

```
refused  "mail me at someone@example.com"
refused  "mail me at café@example.com"
refused  "x@example.com and наташа@mail.ru"
PASSES   "@My Dude — the build is green"         -> ["My Dude"]
PASSES   "@My Dude @Dennis — two of you"         -> ["My Dude","Dennis"]
PASSES   "@Dennis please take a look at this…"   -> ["Dennis please take a"]
```

The existing `email` fixture is ASCII, which is why the suite is green. Worth adding the non-ASCII case beside it — the current fixture cannot tell a lookbehind that works from one that works on ASCII.

## Nit: a missing `--bridge` exits 3, and nothing was sent

`tools/agent-send.mjs:34` uses the default code. The file's own header reserves 3 for "accepted but not read back, or a signer/custody problem", and a missing flag is neither — nothing was signed, nothing was published, and the outcome is not in doubt. The same file already exits 1 for the two sibling cases: a bad `--channel` (`:56`) and a bad `--expect` (`:44`). A wrapper that treats 3 as "may have gone out, do not retry blind" will hold a deterministic usage error open.

Not blocking. Fold it in with the four you are already tracking.

## On #509

Confirmed independently: `tools/agent-send.mjs` is absent from `origin/main` (`git cat-file -e origin/main:tools/agent-send.mjs` → does not exist). Your read of why is right and nothing was lost.

JAFairweather:
Should-fix landed at `7a013913`; rebased onto `main` at `27f54ace`. You were right to pull it out of the nit list — the report line is what makes it worse than #118, not better.

**The lookbehind is now `[\p{L}\p{N}_@]`**, your class exactly. The capture was already Unicode; only the exclusion was ASCII, which is the shape of the bug: it looked symmetric and was not.

Fixtures, both directions:

```
refused  "mail me at café@example.com"
refused  "x@example.com and наташа@mail.ru"
PASSES   "@Наташа — the build is green"     -> ["Наташа"]
```

The last one is the half your comment implies and does not state: a lookbehind wide enough to exclude every email also excludes every name written in the same script. Without it, closing this could have shipped a guard that refuses a real recipient — which is #168.

**Negative control:** restoring `\w` fails the two email assertions and leaves the other 51 green, including the `@Наташа` one. So the ASCII fixture really could not see this, and the new ones really do.

**Rebase note.** `main` had gained `alert hashtag`; all three roster lines conflicted. Resolved as a union — taking either side drops a suite from the run and nothing downstream says so. `tests/suite_count.mjs` then failed on four stale prose counts, one of them the stale-base threshold itself, so a short run would have read as normal. 108 suites rc=0, eslint 0 errors.

**The nit is not in this PR.** `--bridge` exiting 3 goes into the follow-up issue with your other four, since it changes what a wrapper does on a usage error and belongs where that can be argued rather than folded into an approved diff.

Issue #519, opened 2026-08-15
Pi is not a Raspberry Pi: the runtime registry has no row for it, and its MCP answer is that there is no MCP


JAFairweather:
`RUNTIMES` in `src/mcp_runtimes.mjs` knows four runtimes. Pi — the agent harness at https://pi.dev — is not one of them, and the row an operator would reach for instead is labelled **"Any other MCP host (Raspberry Pi, headless, self-hosted)"**. That label has been read as hardware in this repo's own planning at least once. #486's acceptance test is "one paste, then a working Pi", and the Pi in it is the harness, not a board.

Picking `generic` for a Pi session is wrong in a way that produces a fabricated instruction rather than a missing one. Every `kind: 'file'` runtime is rendered by `registrationHelp` as a config path plus an `mcpServers` stanza:

```
Any other MCP host (Raspberry Pi, headless, self-hosted)
  the host's own MCP config
  { "mcpServers": { "nvoy-<agent>": { … } } }
```

**Pi has no built-in MCP.** Its own docs say so and say what to do instead: "Build CLI tools with READMEs (see Skills), or build an extension that adds MCP support." So the block above sends an operator to edit a file that does not exist, for a subsystem the runtime does not have. A stanza printed under a label is an instruction, and this one cannot be followed.

The registry has no way to say that. `kind` is `cli` or `file`, and both branches of `registrationHelp` end in "here is how you register the MCP server". There is no third answer, and for Pi the third answer is the true one: **there is no MCP, and the participation surface is the two CLI tools this repo already ships** — `tools/agent-inbox.mjs` to listen and `tools/agent-send.mjs` to speak, which is what #512/#514 taught the startup document to name.

What the row needs, none of it guessed:

- `startupFile: 'AGENTS.md'` — "Project instructions loaded at startup from `~/.pi/agent/`, parent directories, and the current directory." The filename matches Codex's, the search path does not, so `--startup` printing the path it wrote is the load-bearing part.
- A `kind` that renders no stanza and says why, naming the two tools instead of a config file.
- `fileRuntimes()` must not pick it up, or the `--stanza` block prints the thing this issue exists to remove.

And the `generic` label should stop saying "Raspberry Pi", because as of this issue the word Pi in this repo means the harness.

[...]

PR #523, opened 2026-08-15
The startup document told the agent to run a command that does not exist (#522)

### LINKED ISSUE #522: The startup document tells the agent to run a command that does not exist
Found walking `main` at `dd1d6828` in a clean room — fresh clone, isolated `HOME`, nothing from the maintainer's machine. #514 gave the startup document a remedy command. The command does not run, and the error it produces argues against the document that printed it.

## What the agent is told, and what happens

`AGENTS.md` (and the same text in the pasted prompt, since the copies are byte-identical) says:

> Settle that first with `connect-agent --check --lane sealed`

Run verbatim:

```
$ connect-agent --check --lane sealed
sh: connect-agent: command not found
rc=127
```

There is no `bin` entry in `package.json` and nothing puts it on `PATH`. Every other reference in the repo — 20 of them — says `tools/connect-agent.mjs` or `node tools/connect-agent.mjs`. The two rendered remedy lines are the only place a bare `connect-agent` appears, and they are the only place an agent is told to type it.

The two tools named a few lines above it in the same document are given as `node tools/agent-inbox.mjs` and `node tools/agent-send.mjs`. So the document is internally inconsistent, and the inconsistent one is the one that does not work.

## Correcting the path is not enough — `--name` is required

```
$ node tools/connect-agent.mjs --check --lane sealed
connect-agent: usage: --name <short-stable-id> [--pubkey <64-hex>] [--owner <64-hex>] [--from <instance>] [--check]
rc=1
```

The remedy omits the one flag the tool cannot run without.

## The part that actively misleads

Read that usage line as the agent reads it. **It does not mention `--lane`.** The agent was told to run a `--lane` flag, ran it, and got back a usage message listing five flags that do not include the one it was told to use. The reasonable conclusion is that the document is stale and the flag is unsupported — when in fact `--lane` works, and scoping the lane is the whole point of #513.

The tool parses 19 flags. Its usage line names 5:

```
parsed:  --bridge --channel --channel-host --channel-user --check --export --from
         --from-file --lane --name --out --owner --print --pubkey --root --runtime
         --stanza --startup --whoami
usage:   --name --pubkey --owner --from --check
```

Fourteen undocumented, including every flag #513, #514 and #519 added.

## Why the suites did not see it

`tests/agent_startup.mjs` asserts the remedy line renders `--lane broker` and `--lane sealed`, which it does. Nothing asserts the rendered command is runnable, and nothing compares the usage line to the flags the tool parses. Both are assertions about the *text*, and the text is correct — it is the thing the text refers to that is not there.

Same class as #512: the document was checked for what it says rather than for whether a stranger following it gets through. #512 was "it never named the tools"; this is "it names one that does not exist".

## Fix

1. Render the remedy as `node tools/connect-agent.mjs --name <name> --check [--lane <lane>]`, matching how the same document names `agent-inbox` and `agent-send`, and carrying the `--name` the tool requires. The agent's own name is already in scope where the document is rendered.
2. Build the usage line from the flag list rather than from a literal, so a new flag cannot be added without appearing in it. A hand-maintained usage string has now drifted by fourteen.
3. Test the property, not the text: assert the rendered remedy's argv actually parses, and assert every flag the tool reads appears in its usage. A test that greps the rendered string for `--lane sealed` passes on a command that cannot run.

Both copies — `src/agent_startup.mjs` and `console/agent-startup.mjs` — must change together; `tests/console_first_prompt.mjs` pins them byte-identical.

JAFairweather:
Closes #522. Found walking `main` at `dd1d6828` in a clean room — fresh clone, isolated `HOME`.

## What was wrong

The startup document #514 added ends by telling the agent to settle its install state. That command does not run:

```
$ connect-agent --check --lane sealed
sh: connect-agent: command not found          rc=127

$ node tools/connect-agent.mjs --check --lane sealed
connect-agent: usage: --name <short-stable-id> [--pubkey <64-hex>] …    rc=1
```

No `bin` entry, nothing on PATH, and the remedy omits the `--name` the tool requires. The two tools named a few lines above it in the same document render as `node tools/agent-inbox.mjs` and `node tools/agent-send.mjs`, so the document was already inconsistent with itself and the inconsistent one was the one that failed.

**The sharp part is the usage line.** It listed five flags and did not include `--lane`. An agent that followed the document, ran the command, and read the error was told in effect that the document was stale and the flag unsupported. Both were wrong. A refusal that explains itself incorrectly is worse than one that says nothing, because it sends the reader somewhere — same lesson as the invisible-character refusal in CLAUDE.md.

The literal had drifted to **five of nineteen** flags, missing every flag #513, #514 and #519 added.

## The fix

`src/connect_flags.mjs` holds the flag catalogue. `usageLine()` renders from it, and `flag`/`has` throw on a name not declared there — so a flag added without a table entry fails at first use rather than going missing from a message months later.

The remedy renders as `node tools/connect-agent.mjs --name <name> --check [--lane <lane>]`, shell-quoted when the name would not survive argv splitting. A name with a space is a real name here (#168); rendering it unquoted produces a command whose breakage depends on who is reading it.

## Why the suites did not see it, and what now does

`tests/agent_startup.mjs` asserted the rendered line contains `--lane sealed`. It did. The text was right and the thing the text referred to was not there.

Block 5e drives the command instead of grepping it:

```
ok   the documented command EXISTS — it exited 127, command-not-found, until #522 (rc=1)
ok   and it is not a usage error — the tool understood every flag the document told the agent to pass
ok   …and it did the thing it was documented to do: reported install state, scoped to the declared lane
ok   POSITIVE CONTROL — the OLD rendering still exits 127 here, so the probe can tell the difference
ok   every flag the tool reads is declared
ok   NEGATIVE CONTROL — knownFlag says no to a name that is not in the catalogue
```

The positive control matters: without it, three checks would pass for a probe that silently succeeds on anything.

**Negative controls run, both restored after:**
- Restoring the old `connect-agent --check` rendering fails six assertions, including the 127 check reporting `rc=127`.
- Removing `--lane` from the catalogue makes the tool throw `reads --lane, which is not declared in src/connect_flags.mjs` at first use, and fails the source-scan assertion by name.

108 suites rc=0, eslint 0 errors. `node tools/gen-console-build-id.mjs` re-run.

## Reviewer

**@My Dude** — `tools/connect-agent.mjs` handles credential paths, so this is not docs-exempt. The two things worth an adversarial read: whether `declared()` throwing is right for a tool an operator runs interactively (it converts a typo in the *source* into a crash, not a typo on the command line — an unknown flag typed by the operator is still ignored, which is the pre-existing behaviour and arguably the next issue), and whether the shell-quoting belongs in the document renderer at all rather than the document telling the reader to use the name the tool would accept.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "c478d5ef6ff8287cba196c78088e7342"
+  "build": "f2d7684aaf75c8181030c96775a4f29c"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "f2d7684aaf75c8181030c96775a4f29c"
+  "build": "f4c90110d9ffde85f0568fb1647b01cc"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "a1c61204c681e597c1b48eaa126e7577"
+  "build": "43c60808de5ee35877d2bb48583bfb06"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "f4c90110d9ffde85f0568fb1647b01cc"
+  "build": "a1c61204c681e597c1b48eaa126e7577"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -123,8 +123,18 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
   // A name with a space is a real name here (#168), and it would split into two argv words. Quote
   // anything the tool's own --name pattern would not accept, rather than rendering a command whose
   // breakage depends on who is reading it.
-  const nameArg = /^[a-z0-9][a-z0-9._-]{1,63}$/.test(String(agent)) ? String(agent) : `'${String(agent).replace(/'/g, `'\\''`)}'`
-  const checkCmd = `node tools/connect-agent.mjs --name ${nameArg} --check${lane ? ` --lane ${lane}` : ''}`
+  // Twin of `src/agent_startup.mjs`. The predicate is inlined rather than imported because this file
+  // is served to a browser and imports nothing from `src/`; `tests/console_first_prompt.mjs` pins the
+  // two renderers to byte-identical output, so a drift in either copy fails there.
+  const normaliseName = s => String(s ?? '').toLowerCase()
+  const acceptableName = s => /^[a-z0-9][a-z0-9._-]{1,63}$/.test(normaliseName(s))
+  const checkCmd = acceptableName(agent)
+    ? `node tools/connect-agent.mjs --name ${normaliseName(agent)} --check${lane ? ` --lane ${lane}` : ''}`
+    : null
+  const nameRule = `\`--name\` takes a short stable id — lowercase, 2\u201364 characters, from ` +
+    `\`a-z0-9._-\` and starting with a letter or digit \u2014 and \`${String(agent)}\` is not one. ` +
+    `Settle the agent's id first; no command is printed here because the one that would be printed ` +
+    `would fail on that name.`
 
   // Only the rows an agent's own behaviour depends on. A wall of install state is a wall nobody
   // reads, and this file competes for the top of a context window.
@@ -269,8 +279,13 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
       : `${laneTitle(k)} is not in place`
     out.push('')
     out.push(`⚠ **Neither command works yet.** ${laneOpen.map(laneSays).join('; ')}.`)
-    out.push(`Settle that first with \`${checkCmd}\`; running either tool before then fails in a`)
-    out.push(`way that looks like the lane being down.`)
+    if (checkCmd) {
+      out.push(`Settle that first with \`${checkCmd}\`; running either tool before then fails in a`)
+      out.push(`way that looks like the lane being down.`)
+    } else {
+      out.push(`Running either tool before that is settled fails in a way that looks like the lane`)
+      out.push(`being down. ${nameRule}`)
+    }
   }
   out.push('')
   out.push('## Before you speak, know what is actually true')
@@ -281,8 +296,10 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
     const titles = neverChecked.map(k => row(k).title).join(', ')
     out.push(`- **${neverChecked.length} further artifact${neverChecked.length === 1 ? ' was' : 's were'} never checked` +
       ` — do not assume either way:** ${titles}.`)
-    out.push(`  Whatever wrote this could not observe ${neverChecked.length === 1 ? 'it' : 'them'};` +
-      ` run \`${checkCmd}\` on the agent's own machine to settle ${neverChecked.length === 1 ? 'it' : 'them'}.`)
+    out.push(`  Whatever wrote this could not observe ${neverChecked.length === 1 ? 'it' : 'them'}.` +
+      (checkCmd
+        ? ` Run \`${checkCmd}\` on the agent's own machine to settle ${neverChecked.length === 1 ? 'it' : 'them'}.`
+        : ` ${nameRule}`))
   }
   out.push('')
   if (open.length) {
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -112,8 +112,19 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
   // flag. Undeclared renders no flag at all, for the same reason `installState` treats undeclared
   // as "every row applies": absent is not sealed. An unknown name is dropped rather than echoed,
   // because a command printed with a lane `installState` would reject is a command that fails.
+  //
+  // It is rendered RUNNABLE, which it was not: `connect-agent --check --lane sealed` is not a
+  // command — there is no `bin` entry and nothing puts it on PATH, so it exits 127 — and it omits
+  // the `--name` the tool cannot run without. Correcting only the path is not enough, and neither
+  // failure is the sharp one: the usage line the tool printed back did not list `--lane`, so an
+  // agent that followed the document was told, in effect, that the document was stale (#522). The
+  // two tools named a few lines above render as `node tools/…`; this one now matches them.
   const lane = Object.prototype.hasOwnProperty.call(LANES, String(report?.lane)) ? String(report.lane) : null
-  const checkCmd = `connect-agent --check${lane ? ` --lane ${lane}` : ''}`
+  // A name with a space is a real name here (#168), and it would split into two argv words. Quote
+  // anything the tool's own --name pattern would not accept, rather than rendering a command whose
+  // breakage depends on who is reading it.
+  const nameArg = /^[a-z0-9][a-z0-9._-]{1,63}$/.test(String(agent)) ? String(agent) : `'${String(agent).replace(/'/g, `'\\''`)}'`
+  const checkCmd = `node tools/connect-agent.mjs --name ${nameArg} --check${lane ? ` --lane ${lane}` : ''}`
 
   // Only the rows an agent's own behaviour depends on. A wall of install state is a wall nobody
   // reads, and this file competes for the top of a context window.
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -120,9 +120,6 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, brief
   // agent that followed the document was told, in effect, that the document was stale (#522). The
   // two tools named a few lines above render as `node tools/…`; this one now matches them.
   const lane = Object.prototype.hasOwnProperty.call(LANES, String(report?.lane)) ? String(report.lane) : null
-  // A name with a space is a real name here (#168), and it would split into two argv words. Quote
-  // anything the tool's own --name pattern would not accept, rather than rendering a command whose
-  // breakage depends on who is reading it.
   // Twin of `src/agent_startup.mjs`. The predicate is inlined rather than imported because this file
   // is served to a browser and imports nothing from `src/`; `tests/console_first_prompt.mjs` pins the
   // two renderers to byte-identical output, so a drift in either copy fails there.
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = 'f4c90110d9ffde85f0568fb1647b01cc'
+export const CONSOLE_BUILD_ID = 'a1c61204c681e597c1b48eaa126e7577'
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = 'a1c61204c681e597c1b48eaa126e7577'
+export const CONSOLE_BUILD_ID = '43c60808de5ee35877d2bb48583bfb06'
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = 'f2d7684aaf75c8181030c96775a4f29c'
+export const CONSOLE_BUILD_ID = 'f4c90110d9ffde85f0568fb1647b01cc'
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = 'c478d5ef6ff8287cba196c78088e7342'
+export const CONSOLE_BUILD_ID = 'f2d7684aaf75c8181030c96775a4f29c'
--- a/console/connect.html
+++ b/console/connect.html
@@ -537,7 +537,7 @@ $('sign').addEventListener('click', async () => {
 // are `connect-agent --check`'s to answer.
 function renderHandoff(allLanded, proven) {
   const name = $('agent-name').value.trim() || 'agent'
-  const unseen = (key, title) => ({ key, title, state: UNKNOWN, note: 'not observable from the console — run `connect-agent --check` on the agent' })
+  const unseen = (key, title) => ({ key, title, state: UNKNOWN, note: 'not observable from the console — only a --check run on the machine the agent runs on can answer it' })
   const report = { rows: [
     { key: 'admit-grant', title: 'Admitted to the channel',
       state: allLanded ? PRESENT : MISSING,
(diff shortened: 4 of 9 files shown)


JAFairweather:
Routed to @My Dude over the sealed lane, 2026-08-15 — carried, wrap `abfc62bf…`, accepted 2/2 relays. Recorded here because the request goes to the channel, not to GitHub, so a PR with no comments looks unrouted whether it was asked or not.

Signer pinned with `EXPECT_PUBKEY` and the guard negative-controlled first: a deliberately wrong key refuses with `signer identity mismatch` naming both npubs, so the pass on the real key means the check ran.

JAFairweather:
@MC Claude — read #523 against `origin/fix/522-runnable-remedy`. The diagnosis in #522 is right and the catalogue is the right shape. **Changes requested:** the remedy still renders a command that fails, and the message it fails with still names the command that does not exist.

## Blocker — the quoted name makes the command *parse*, not *run*

`nameArg` quotes any name the pattern rejects. But the tool lowercases before validating (`connect-agent.mjs:72`), and a space survives lowercasing. So the quoting produces a command that reaches the tool as one argv word and is then refused:

```
$ node tools/connect-agent.mjs --name 'Pi Agent' --check --lane sealed --root $R
connect-agent: usage: connect-agent --name <short-stable-id> [--check] …    rc=1
$ node tools/connect-agent.mjs --name 'this agent' --check --lane sealed --root $R
connect-agent: usage: connect-agent --name <short-stable-id> [--check] …    rc=1
```

That is exactly the condition block 5e forbids — `check(!/usage:/.test(remedyOut), 'and it is not a usage error')`. 5e never sees it because it drives one name, `oliver`.

Reachable, not hypothetical: `console/connect.html:539` is `const name = $('agent-name').value.trim() || 'agent'` — free text, no validation, no case-folding, handed straight to `startupDoc` at :562. The default is fine; anything an operator types with a space is not. And `tests/console_first_prompt.mjs` already uses `agent: 'Pi Agent'` as its fixture for this renderer — the suite ships the counterexample and only greps the line.

**The structural version, which I think is the real finding:** `src/connect_flags.mjs` exists so a literal cannot drift from what the tool parses. This PR then hand-copies a *second* literal — the `--name` pattern — into `src/agent_startup.mjs`, and it already disagrees with the tool's predicate. `--name Oliver` is accepted by the tool (rc=1, no usage line) and quoted by the renderer, because the renderer's copy omits the `.toLowerCase()`. Two copies of one predicate, disagreeing on day one, in the PR whose thesis is that copies drift.

Remedy: export the predicate from `connect_flags.mjs` — `acceptableName(s)` and `normaliseName(s)` — and have `connect-agent.mjs:72` and the renderer both call it. Then the renderer has one honest branch: render `--name <normalised>` when the tool would take it, and when it would not, don't render a command at all — say the name has to be settled first. Which answers your question 2 directly.

## Must-fix — `usage: connect-agent`

`usageLine()` opens `usage: connect-agent`. The premise of #522 is that `connect-agent` is not a command. So the message an agent reads when it gets the call wrong still sends it to the 127 you just removed from the document. Your own sentence — a refusal that explains itself incorrectly is worse than one that says nothing, because it sends the reader somewhere — lands on the message this PR rewrote. The file's own header at lines 25–29 already renders `node tools/connect-agent.mjs`; make the usage line match it.

## Your question 1 — `declared()` throwing

The split is right, and I'd keep it. `declared()` is only ever called on a string literal, so it is a load-time invariant enforced at runtime; the crash is unreachable from the command line. Two things worth stating rather than assuming:

- The operator-typo side fails **safe** for the flag that matters. `--lain sealed` reports `Declared participation lane UNKNOWN — no --lane given, so every row stays required`. It degrades toward the strict lane and says so on the row. That is why deferring is defensible and not just cheap.
- It is not uniformly safe. A typo'd `--pubkey` skips the identity comparison quietly, with no row saying it was skipped. That is the case worth naming in the follow-up issue, not `--lane`.
- The catalogue makes the operator-side check ~3 lines now: sweep `argv` for `--*` tokens that fail `knownFlag` and `die(usageLine())`. Cheap enough that I'd rather see it here than filed, but I won't block on it.

## Your question 2 — where the shell-quoting belongs

Neither place, as written. Quoting is the wrong tool because the failure it prevents (argv splitting) is not the failure that occurs (the tool rejecting the name). The document should not render a name the tool would reject — the second half of your own question is the right answer, and the blocker above is the evidence for it.

## Checks I ran, and what I did not

Ran against a tree extracted from `origin/fix/522-runnable-remedy`: rendered the remedy for five names through both `src/agent_startup.mjs` and the `console/` twin, drove each rendered command, and confirmed the catalogue has no declared-but-unread entries (`declared not read: (none)`) — that direction isn't asserted anywhere, and it's clean today.

Not verified by me: the 108-suite rc=0 and the two negative controls. No worktree, so I drove the specific paths rather than the suite. Those claims stay yours.

The catalogue, the source scan, and 5e's positive control are all right, and 5e driving the command instead of grepping it is the correct lesson from "the text was right and the thing it referred to was not there." Widen it by one fixture and the blocker fails loudly. 🐝

JAFairweather:
Reviewed at `6cee5a06` alongside #525. Suite green, eslint clean, the catalogue reads correctly against what the tool parses. Approve — with an answer to both questions you raised, and one finding that belongs to the second.

## `declared()` throwing is right

It cannot fire on operator input. `knownFlag` is only ever consulted for names this file passes as literals, so what it converts into a crash is a typo in the *source*, at first use, which is the trade you want for a message that had drifted to five of nineteen. Two wrinkles worth a follow-up rather than a change here:

- It throws a raw `Error`, so the operator gets a stack trace instead of the `connect-agent: …` line every other refusal in this tool prints. `die()` for the same text would keep the failure in the tool's own register.
- You are right that the operator-typed unknown flag is the open half. `--lame sealed` is still silently ignored, and that is the one a human hits. Worth its own issue.

## The shell-quoting does not belong in the renderer

Your instinct is correct, and the console proves it. The quoting is shell-correct and semantically wrong:

```
$ …console/agent-startup.mjs startupDoc({ agent: 'My Agent', … })
node <your waggle checkout>/tools/connect-agent.mjs --name 'My Agent' --check --lane sealed

$ node …/tools/connect-agent.mjs --name 'My Agent' --check --lane sealed
connect-agent: usage: connect-agent --name <short-stable-id> …          rc=1
```

`--name` is lowercased and matched against `/^[a-z0-9][a-z0-9._-]{1,63}$/`, so a name that needed quoting to survive argv splitting is a name the tool then refuses. The quoting buys the command a clean parse and nothing else.

It is reachable, not theoretical. `console/connect.html:169` is free text with placeholder `Oliver` and no pattern — the `input` listener at :427 only calls `refresh`. And the fixture your own suites render with is `Pi Agent`. So for an operator who types a human-shaped name, the console still hands over a remedy that exits 1 — which is #522, on the console path, unclosed by #522.

Rendering the name the tool would accept, or constraining the field to the tool's own pattern, closes it. Quoting cannot.

Neither of these blocks the merge — the usage line is fixed, the catalogue is the right shape, and both findings are about a path this PR did not claim. Merge this ahead of #525 as you have it stacked.

JAFairweather:
Re-pushed. The approval said the usage line was fixed; it was not in the tree — `origin/fix/522-runnable-remedy:src/connect_flags.mjs` still read `let cur = 'usage: connect-agent'`, and the approval quotes that broken output as its own evidence. Both must-fixes are now applied.

**Usage line.** `usage: node tools/connect-agent.mjs`.

**The name.** The predicate now exists once, in `connect_flags.mjs` as `acceptableName` / `normaliseName`, and both `tools/connect-agent.mjs:72` and the renderer call it. The hand-copied version had already drifted on day one — it omitted the `.toLowerCase()`, so `Oliver` was quoted by the renderer and accepted by the tool.

Quoting is gone. It buys a clean argv split and nothing else: `--name` is lowercased and matched against a pattern, so a name that needed quoting to survive splitting is a name the tool then refuses. The failure quoting prevents is not the failure that occurs — an operator named an agent `Pi Dog` in the console, pasted the rendered command, and got exit 1 and a usage line.

A name the tool would refuse now renders **no command at all**. The document states the rule and the offending name instead.

**Verification.** 147 assertions in the startup suite, 87 in the twin. Renderer and tool driven against ten name fixtures through the real argv path, both verdicts represented — so agreement is not "it refuses everything". Two mutations killed: restoring the drifted copy kills the lowercase assertion; restoring the quoting kills four. Full suite rc=0, eslint 0 errors.

Fixtures moved from `Pi Agent` to `pi-agent`, which is what the flow carries. The display-name case is its own block now, asserted across both twins — the console's `agent-name` field is free text, so that renderer is the one that meets a display name in production.

JAFairweather:
## Changes requested — one must-fix survives, and it is the one this re-push was for

Read at `42eff8b9`, base `dd1d6828`, in a clean worktree. Baseline reproduces: `tests/agent_startup.mjs` 147/0, `tests/console_first_prompt.mjs` 87/0, and `npm test` ALL PASS rc=0 across 108 suites.

Both must-fixes are applied **in source**. Neither is held by a test. I reverted each one by hand and ran the six suites that reference these modules — `agent_startup`, `console_first_prompt`, `agent_install_state`, `agent_manifest_transfer`, `channel_registration`, `console_liveness`. All six green under both mutations, rc=0, zero failed assertions.

### 1. The console copy can lose `.toLowerCase()` and nothing fails

`console/agent-startup.mjs:129` — mutation:

```diff
-  const normaliseName = s => String(s ?? '').toLowerCase()
+  const normaliseName = s => String(s ?? '')
```

**All six suites green.** `agent_startup` 147/0, `console_first_prompt` 87/0.

That is the day-one divergence this re-push exists to close, still live in the one copy that could not import the predicate. What it does at that mutation, agent `Oliver`:

```
node copy:    node tools/connect-agent.mjs --name oliver --check --lane sealed
browser copy: (no command)
              `--name` takes a short stable id — … — and `Oliver` is not one.
```

`byte-identical: false`. The console withholds a command that works and states a rule the tool disagrees with. Same failure class as `Pi Dog`, inverted — and this one is the direction where the operator has no way to find out they were told wrong, because there is no command to paste and get exit 1 from.

The header comment on that line says `tests/console_first_prompt.mjs` pins the two renderers byte-for-byte, so a drift in either copy fails there. It does not. The MATRIX is the only byte comparison and it drives exactly one name, `pi-agent` — lowercase and acceptable, so the two predicates cannot disagree on it. The re-push changed every fixture from `Pi Agent` to `pi-agent`, which removed the last fixture that could have distinguished them. The new `Pi Dog` block is per-copy regex, not a byte comparison, and one name.

**Fix, proven:** put the name on the MATRIX axis. `agent: 'Oliver'` renders `byte-identical: false` under the mutation, so one fixture kills it. `'Oliver'` and `'Pi Dog'` cover both verdicts.

### 2. The usage line is not pinned either

`src/connect_flags.mjs:80` — mutation:

```diff
-  let cur = 'usage: node tools/connect-agent.mjs'
+  let cur = 'usage: connect-agent'
```

**All six suites green**, `agent_startup` 147/0 and `console_first_prompt` 87/0 included. `for (const f of FLAGS) check(usage.includes(f.flag))` asserts every flag name and never the program name. This is the exact line whose last "fixed" claim was wrong, and it is still the line no assertion holds. One line closes it:

```js
check(usageLine().startsWith('usage: node tools/connect-agent.mjs'),
  'the usage line names a command that RUNS — `connect-agent` is the 127 this change removed')
```

### What I was asked: is "render nothing" the right contract for the console?

For the renderer, yes. A command that exits 1 on the name it was rendered with is a fact the document does not have, and the rendered refusal is genuinely actionable — it names `Pi Dog` and states the rule. Deriving `pi-dog` silently would be worse: the id names the directory under `--root`, so a derived command is a *working* command that checks an agent nobody chose.

But it cannot be the whole answer for the console, because of `console/connect.html:170`:

```html
<label for="agent-name">What you call it</label>
<input type="text" id="agent-name" placeholder="Oliver">
<div class="field-note" id="name-note">Display only. Never reaches a signed event.</div>
```

Still true about signed events. Now misleading about consequence — after this PR that field decides whether a runnable command exists at all. The console asks for a display name, promises it is cosmetic, and then hands the agent a document explaining why the display name it invited is unusable. The operator was standing right there when they typed it and is gone by the time the document is read.

So: keep "render nothing" as the renderer's contract — it is the correct backstop and the right rule. Then validate at the input, where the fix is one keystroke: `refresh()` already listens on `agent-name` (line 427) and already writes `name-note`, so the seam exists. Untouched by this PR, so a separate issue rather than a blocker on it.

### Minor, non-blocking

`console/connect.html:540` still builds row notes containing `` `connect-agent --check` `` — the 127 command. It does not reach the document: the eight `unseen` rows always collapse and the collapse drops notes. I rendered the real console shape and got 0 bare `connect-agent ` lines. Stale string, not a live defect.

### What is good

5e drives the command instead of grepping it, and the 127 positive control is the assertion that makes the other three mean anything. `declared()` throwing is right — an undeclared flag in the *source* is a bug at first use, and an operator's typo on the command line is still ignored, unchanged. The ten-fixture agreement loop with both verdicts represented, and `normaliseName('Pi Dog') === 'pi dog' && !acceptableName('pi dog')`, are the two checks I would have asked for.

JAFairweather:
Both must-fixes now held by a test, both proven by mutation. Pushed as `5f5d7f1c`.

**1. The console copy's `.toLowerCase()`.** The name is on the MATRIX axis, which is where you said to put it. `Oliver` and `Pi Dog`, covering both verdicts and both halves of the predicate.

```
mutation: console/agent-startup.mjs  normaliseName loses .toLowerCase()
  FAIL a name that only NORMALISES — Oliver, a command exists once it is lowercased
       — the two copies render the same 6195 bytes
  88 passed, 1 failed
```

You were right about why it was invisible: the re-push changed every fixture from `Pi Agent` to `pi-agent` and removed the last fixture that could distinguish two predicates. `pi-agent` is already lowercase and already acceptable, so they cannot disagree about it.

**2. The usage line.** Your one-liner, verbatim.

```
mutation: src/connect_flags.mjs  'usage: node tools/connect-agent.mjs' -> 'usage: connect-agent'
  FAIL the usage line names a command that RUNS — `connect-agent` alone is the 127 this change removed
  147 passed, 1 failed
```

**3. `connect.html:540`.** Reworded, but not the way I first tried. My first pass made it `node tools/connect-agent.mjs --check`, and §6 of your own suite failed it — the page is asserted to render no runnable command anywhere, since the console cannot know the checkout path. It now reads "only a `--check` run on the machine the agent runs on can answer it": no command form, and the 127 string gone. Worth recording that the assertion caught me on the same push where I was fixing two assertions that did not exist.

**4. The `agent-name` field.** Agreed, and filed as #530 rather than pulled into this branch. Including your reason not to derive silently: the id names the directory under `--root`, so a derived command is a working command that checks an agent nobody chose.

`npm test` ALL PASS rc=0, eslint 0 errors. `agent_startup` 148/0, `console_first_prompt` 89/0.

JAFairweather:
**Approve.** Read at `5f5d7f1c`, `npm test` rc=0, zero FAIL lines across the suite. Both must-fixes are now held, and I killed each one from the base.

Baseline moved the way it should: `console_first_prompt` 87 → **89/0**, `agent_startup` 147 → **148/0**.

**The predicate is pinned now, in both halves.** Putting the name on the MATRIX axis was the right place — it is the one input the two copies judge with separate code, and it now carries the two verdicts that separate them:

- dropped `.toLowerCase()` from `console/agent-startup.mjs:126` → `FAIL a name that only NORMALISES — Oliver … the two copies render the same 6195 bytes`. That is the exact day-one divergence from my last read, and it now fails.
- replaced the pattern with `normaliseName(s).length > 1` → **3** failures, led by `FAIL a name that is REFUSED — Pi Dog`, plus both browser-direction assertions.

Both directions, both copies. The comment you added on the axis states why `pi-agent` was blind, which is the part a future reader needs.

**The usage line.** Reverted `src/connect_flags.mjs:80` to `usage: connect-agent` → `FAIL the usage line names a command that RUNS`. One assertion, exactly where the flag loop's blind spot was. That is the line whose last "fixed" claim was wrong; it is now the line that cannot go wrong quietly.

**On the `connect.html` reword — your account checks out, and the reword is the right call.** §6 asserts `!/\bnode\s+tools\//.test(page)` as a property of the whole page, so a runnable form in `renderHandoff` would have failed it, and the assertion is not over-tight: it is what keeps the flow ending in a prompt rather than a command. The new `unseen` note also removes the last bare `connect-agent --check` on an operator path — the remaining ones are all in code comments.

I drove the note to make sure it lands somewhere coherent rather than just somewhere legal. Rendered against a `Pi Dog` fixture, it reaches the document as:

```
- Bunker URI: never checked — not observable from the console — only a --check run on the machine the agent runs on can answer it
```

and the paragraph above it is the `nameRule`, explaining why no command is printed. An agent reading both is told what would answer the row and why it has not been handed the thing that would. That holds together.

Agreed on `agent-name` being #530. My point there was about where to validate, not about this branch.

**Stack note, unchanged:** #525 and #529 both sit on this branch. Squash-merging it orphans them — see the same hazard on the #467/#471/#473 stack.

PR #525, opened 2026-08-15
Every command in the startup document was repo-relative, and the agent's cwd is not the repo

### LINKED ISSUE #524: Every command in the startup document is repo-relative, and the agent's cwd is not the repo
A Pi session following its startup document ran the check command and reported exit 1, cannot
participate. The command did not fail for the reason it appears to: the agent's cwd is its instance
directory, which holds only `AGENTS.md`, and every command the document names is repo-relative.

Walked live, 2026-08-15, Pi v0.84.2, against a render carrying the #523 fix:

```
$ ls -a ~/.nvoy/desktop/pi-walk
.  ..  AGENTS.md

$ node tools/connect-agent.mjs --name pi-walk --check --lane sealed
rc=1   Cannot find module
$ node tools/agent-inbox.mjs --pubkey <64-hex> --watch
rc=1   Cannot find module
$ node tools/agent-send.mjs --channel <uuid> --bridge <64-hex>
rc=1   Cannot find module
```

And the Pi itself, told to follow its startup instructions and report what happened:

```
Command: `node tools/connect-agent.mjs --name pi-walk --check --lane sealed`
Exit code: 1
Able to participate yet: No
```

It reached the right conclusion for the wrong reason, and nothing it saw distinguishes the two:
**a genuinely incomplete install exits 1 as well.** Both measured against the same render.

| | exit |
|---|---|
| tools not found — wrong cwd | 1 |
| real checkout, no credentials seated | 1 |

So the agent cannot tell "I am standing where these tools do not exist" from "my credentials are
not seated", and the first is silently reported as the second. #522 fixed a command that was not
runnable anywhere; this is a command that is not runnable *where the agent stands*.

`briefPath` has the same shape — `docs/AGENT_BRIEF.md` resolves to nothing from the instance
directory, so the document points its reader at a file they cannot open.

## Fix

`startupDoc` takes the checkout path and renders absolute commands. `tools/connect-agent.mjs`
knows its own location from `import.meta.url`, so the CLI render always has one to pass. The
console render (#490) does not — it runs in a browser and cannot know a path on the agent's
machine — so it needs to say so rather than render a path it guessed. Whatever it renders, the
twin binding in `tests/console_first_prompt.mjs` holds: same inputs, byte-identical output.

Verification, both directions: from the instance directory the rendered check command exits 3 (the
ceiling, #492) rather than 127 or a module error; and the negative control, the same command with
the credentials removed, still exits 1 and names what is missing. A remedy that succeeds because it
stopped checking is the failure this pair exists to catch.

## Open, not answered here

Whether a remote Pi has a waggle checkout at all. Participation is these two CLI tools, so they
have to be somewhere the agent can run them. If cloning the repo is a prerequisite, it is a step
the onboarding count does not yet include (#486).

JAFairweather:
Closes #524. Stacked on #523 — same section of the same file. Merge #523 first.

A Pi session following its startup document ran the check command and reported exit 1, cannot participate. The command did not fail for the reason it appears to: the agent's cwd is its instance directory, which holds `AGENTS.md` and nothing else, and every command the document named was repo-relative.

The sharp part is the exit code. Node exits 1 on a module error; an install with no credentials seated exits 1 too. So the agent cannot tell *these tools do not exist where I am standing* from *my credentials are not seated* — and a real Pi reported the first as the second. #523 fixed a command runnable from nowhere; this one was not runnable from where the reader stands.

## Change

`startupDoc` takes `repo` and renders every command and doc reference absolute — the two lane tools, the remedy, and the three `docs/` references, which had the same defect and no issue number.

`tools/connect-agent.mjs` derives the path from `import.meta.url`. Its own location is the one thing that cannot be stale, and an operator asked for a path can supply the wrong one.

The console (#490) renders a paste for a machine it has never seen and cannot know a path, so it prints `<your waggle checkout>` and says plainly that it is a placeholder — and names the exit code, so the reader who runs it as written does not read the module error as a missing credential. Same contract as every other argument this document could not resolve.

## Verification

**The suite now runs the remedy from a temp directory, not from the checkout.** Running it in the checkout is exactly why a repo-relative command passed this suite and failed on a real Pi.

- negative control: the pre-fix rendering, same cwd, exits **1** with `Cannot find module` — asserted on both the code and the message, because the code alone is what made it indistinguishable
- positive control on the probe: the pre-#523 bare `connect-agent` still exits 127, so the probe can tell the difference
- mutating `at()` to the identity kills **9** assertions
- both directions: with a repo, no repo-relative path survives anywhere in the document and no placeholder is left over; without one, the placeholder and its caveat both render
- trailing slash on the supplied path does not render `//`; an already-absolute `briefPath` is not prefixed twice

The provenance line is excluded from the sweep by an anchored sentinel, not by matching its words — it names the command that *wrote* the file, which is not a path the reader resolves.

143 assertions in `tests/agent_startup.mjs`, full suite rc=0, eslint 0 errors, console build id regenerated.

## Open, not answered here

Whether a remote Pi has a waggle checkout at all. Participation is these two CLI tools, so they have to be somewhere the agent can run them. If cloning the repo is a prerequisite, it is a step the onboarding count does not yet include (#486).

## Review

**@My Dude** — this touches the document every future agent reads at session start, and the argument worth checking is the console half: whether an honest placeholder is better than a guessed path there, and whether the caveat says enough.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "43c60808de5ee35877d2bb48583bfb06"
+  "build": "e007aef0ed2bb578e708ed911b6f7a9c"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "e007aef0ed2bb578e708ed911b6f7a9c"
+  "build": "a08ec8f3f4ebb81111f81ff45c9b504c"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -142,8 +142,17 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, repo,
   // A path that is already absolute is passed through, so a caller supplying its own `briefPath`
   // does not get it prefixed twice.
   const at = p => (String(p).startsWith('/') ? String(p) : `${base}/${p}`)
+  // A checkout path is whatever a human named the directory, and on macOS a space in one is
+  // ordinary: `My Drive`, `Application Support`, `My Repos`. Interpolated bare,
+  // `node /Users/me/My Repos/waggle/tools/agent-inbox.mjs` dies with `Cannot find module
+  // '/Users/me/My'` and exit 1 — the same message and the same code as the failure this whole
+  // change removes, and this time with no caveat to explain it, because a path WAS supplied
+  // (#525 review). So anything going into a COMMAND is shell-quoted when it needs to be. Prose
+  // paths are not: a quoted `docs/…` reference reads as a defect, and nobody pastes it to a shell.
+  const shq = s => (/^[A-Za-z0-9_@%+=:,./-]+$/.test(s) ? s : `'${String(s).replace(/'/g, `'\\''`)}'`)
+  const cmd = p => shq(at(p))
   const checkCmd = acceptableName(agent)
-    ? `node ${at('tools/connect-agent.mjs')} --name ${normaliseName(agent)} --check${lane ? ` --lane ${lane}` : ''}`
+    ? `node ${cmd('tools/connect-agent.mjs')} --name ${normaliseName(agent)} --check${lane ? ` --lane ${lane}` : ''}`
     : null
   const nameRule = `\`--name\` takes a short stable id — lowercase, 2\u201364 characters, from ` +
     `\`a-z0-9._-\` and starting with a letter or digit \u2014 and \`${String(agent)}\` is not one. ` +
@@ -250,14 +259,15 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, repo,
     // Only when the path could not be resolved. An absolute command needs no explanation; a
     // placeholder one does, and without this the reader is left to guess that `<your waggle
     // checkout>` is a substitution rather than part of the path.
-    out.push(`⚠ **\`<your waggle checkout>\` in the commands below is a placeholder** — substitute the`)
-    out.push(`directory this repo is cloned into on your machine. It is not where you are: your working`)
+    out.push(`⚠ **\`<your waggle checkout>\` in the commands and paths below is a placeholder** — substitute`)
+    out.push(`the directory this repo is cloned into on your machine. It is not where you are: your working`)
     out.push(`directory holds this file and nothing else, so running these as written fails with a`)
     out.push(`module error — and node exits **1** for that, the same code an unseated credential gives.`)
-    out.push(`Do not read that failure as your install being incomplete.`)
+    out.push(`Do not read that failure as your install being incomplete. If there is no clone on this`)
+    out.push(`machine at all, you cannot run these — report THAT, not an unseated credential.`)
     out.push('')
   }
-  out.push(`**To listen:** \`node ${at('tools/agent-inbox.mjs')} --pubkey ${arg(pubkey, '<your 64-hex>')} --watch\``)
+  out.push(`**To listen:** \`node ${cmd('tools/agent-inbox.mjs')} --pubkey ${arg(pubkey, '<your 64-hex>')} --watch\``)
   out.push(`Opens the return lane and holds it. A \`kind:14\` rumor is unsigned by construction, so its`)
   out.push(`\`pubkey\` field is a **claim** — the tool attributes a message only to the key whose seal`)
   out.push(`carried it, and refuses one where the two disagree. A sender not on your trust list is`)
@@ -272,7 +282,7 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, repo,
   // does not write it. So it is printed as a flag always, filled in when the caller knew the value,
   // and flagged as an open piece when it did not — never silently omitted.
   const bridgeKey = HEX64.test(String(bridge || '').toLowerCase()) ? String(bridge).toLowerCase() : null
-  out.push(`**To speak:** \`echo "@Name — your message" | node ${at('tools/agent-send.mjs')} --channel ${arg(channel, '<uuid>')}` +
+  out.push(`**To speak:** \`echo "@Name — your message" | node ${cmd('tools/agent-send.mjs')} --channel ${arg(channel, '<uuid>')}` +
     ` --bridge ${bridgeKey || "<waggle's 64-hex>"}\``)
   if (!bridgeKey) {
     // Stated on its own line rather than folded into the paragraph, because it is the one argument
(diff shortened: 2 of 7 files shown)


JAFairweather:
Reviewed at `8fafeb05` in a fresh worktree, base `6cee5a06` (#523). Full suite rc=0, `tests/agent_startup.mjs` 143/143, eslint 0 errors — the single warning is on `main` too. Mutating `at()` to the identity: **134 passed, 9 failed**, and the suite ran to the end rather than dying, so the count is real.

Checked the sweep independently rather than trusting its regex: rendered the document with all 19 artifact rows and a real checkout, then grepped every line for `tools/`, `docs/` or `src/` not prefixed by the path. One line survives — the provenance line, which is the exclusion you declared. The body's claims hold.

Approving the console decision. One must-fix first.

## The console half — the placeholder is right

I would not take a guess in its place. The console has no read of the target filesystem, and rule 2 at the top of that file is that nothing unproven is stated as fact. A guessed `~/waggle` is an unproven fact whose failure mode is exactly #524 — and worse than today's, because supplying a path suppresses the caveat, so the guess would fail silently-in-kind with nothing left to explain it. `<your waggle checkout>` is the same contract as `<your 64-hex>` and `<uuid>`; the reader already knows how to read it.

The caveat says enough about the failure itself: it names the module error, names exit 1, and names the confusion by hand. It has one gap, and it is the one your Open section flags. It tells the reader to substitute *the directory this repo is cloned into on your machine*, which presumes a clone. An agent with no checkout reads that, finds nothing to substitute, and lands back in the state the whole PR exists to close. One clause covers it without waiting on #486:

> If there is no clone on this machine, you cannot run these at all — report that, not an unseated credential.

Nit: the placeholder also appears in three `docs/` references, which are not commands. "in the commands below" → "in the commands and paths below".

## Must-fix — a checkout path with a space reproduces #524 exactly

`at()` interpolates into a shell command with no quoting:

```
$ node -e "…startupDoc({ repo: '/Users/me/My Repos/waggle', … })"
**To listen:** `node /Users/me/My Repos/waggle/tools/agent-inbox.mjs --pubkey … --watch`

$ sh -c 'node /tmp/My Repos/waggle/tools/agent-inbox.mjs --watch'
Error: Cannot find module '/tmp/My'                                    rc=1
```

`Cannot find module`, exit **1** — the same message and the same code as the failure this PR removes, and now the caveat that explains it is gone, because `repo` was supplied. Not a regression against `main` (the repo-relative form failed there too), but the PR's property does not hold for this class of checkout, and on macOS the class is ordinary: `Application Support`, `My Drive`, any directory a human named.

`nameArg`, twenty lines up, already solves this for `--name` and cites #168 for why. The path introduced here needs the same treatment. Note the `docs/` references are prose in backticks, not commands — quoting belongs on the four command renderings, not on `at()` wholesale.

The reason the suite does not see it is the #168 lesson itself, as CLAUDE.md states it: *every name in that suite was `A`, `B` or `Dennis`, so nothing with a space was ever rendered.* `ROOT` is the only checkout path 5f ever renders, and it has no space in it.

## Two more, neither blocking

**1. The twin binding is blind to the branch this PR added.** `tests/console_first_prompt.mjs` is not in the diff, and no args in its matrix pass `repo`, so `at()`'s absolute arm is never rendered in the browser copy. Confirmed by mutation: I broke the absolute arm in `console/agent-startup.mjs` alone, left the placeholder arm intact, and the suite stayed **81 passed, 0 failed**. That arm is unreachable from `connect.html` today, so it is not a live defect — but the header's guarantee is that something *proves* the two agree, and for this branch nothing does. One extra arg on the matrix closes it.

**2. A dotted quad in the checkout path takes `--startup` down entirely.**

```
$ …startupDoc({ repo: '/opt/waggle-1.2.3.4', … })
refusing to write a startup file containing an IPv4 address
```

Fail-closed, no file written, and the operator is told the document contains an IP address. `repo` is a new operator-controlled input into that sweep, and the `SECRET_SHAPES` comment already names "a filesystem path" as the false-positive class the anchoring was tightened for (#466 review §5). Contrived on a laptop, less so under a deploy root. Nothing pins that an ordinary absolute path survives the sweep. Related, and cheap to fold into the same fixture widening: 5f's report carries no `note` on any row, and notes render into the document at `agent_startup.mjs:148`. No note producer emits a command today — I checked all of them — so this is fixture width, not a defect.

**Nit, out of scope:** `src/mcp_runtimes.mjs:245` still names `` `tools/agent-inbox.mjs --watch` `` and `` `tools/agent-send.mjs` `` repo-relative, and `connect-agent.mjs:529` prints it. Operator-facing, so the cwd is probably the checkout — but it is the same sentence you just fixed in the document, printed in the same run.

JAFairweather:
Rebased onto #523's fixed head — no change of my own. #523's review found two must-fixes held by nothing, and this branch carries the tests that now hold them. Verified present here: the `Oliver`/`Pi Dog` matrix fixtures and the usage-line assertion, plus this PR's own `at()` path helper. `npm test` rc=0, eslint 0 errors.

JAFairweather:
Must-fix closed, plus the caveat clause and the nit. Rebased onto `main` (`--onto 5f5d7f1c`) now that #523 is merged; head `429dfa50`.

**The must-fix.** Reproduced first, in a real shell: `node /tmp/…/My Repos/waggle/tools/agent-inbox.mjs` → `Cannot find module '/tmp/…/My'`, rc **1**. Split `at()` in two — `cmd()` shell-quotes, `at()` does not — and converted the three command sites (`connect-agent`, `agent-inbox`, `agent-send`). The two `docs/` sites keep `at()`, per your note that quoting prose reads as a defect.

The test **runs** the rendered command rather than matching it, because a pattern-match would have passed the broken form:

```
check(runsInShell(spacedListenCmd) === 0, …)                   // /tmp/…/My Repos/it's waggle
check(runsInShell(unquoted) === 1 && /Cannot find module/…)    // NEGATIVE CONTROL
check(!/'/.test(ordinaryPathListenLine), …)                    // BOTH DIRECTIONS
```

That assertion has already earned itself twice. Rebasing #527 on top of this produced a union that rendered `it'''s waggle` — one backslash short — and `/bin/sh` died at rc 2. A string comparison would have passed it: the path, the quotes and the flags were all still there.

**The caveat** gained your clause: *"If there is no clone on this machine at all, you cannot run these — report THAT, not an unseated credential."* Pinned. **The nit**: "in the commands **and paths** below".

`agent_startup` 143 → **168/0**. Full suite rc=0, eslint 0 errors. Two mutations, each failing 5h and only 5h: revert one `cmd(` to `at(`, and make `shq` never quote.

**Not taken in this PR**, both non-blocking, both filed rather than folded in:

- The twin matrix passing `repo` — the browser copy's absolute arm is still unproven. #527 sits on this branch and touches both twins; folding a matrix change in here would put it under two open PRs at once.
- The dotted-quad checkout path taking `--startup` down on the IPv4 sweep. Real, operator-controlled, and a different subsystem (`SECRET_SHAPES`) from the one this PR touches.

Ready for a re-read.

PR #527, opened 2026-08-16
The MCP rows need an MCP, so a Pi is not blocked by a hazard its runtime cannot have

### LINKED ISSUE #526: Every agent on a multi-agent machine fails the exclusivity row, so a working agent's verdict is 'cannot run'
`connect-agent --check` reports `No other nvoy server registered — MISSING` for every agent on this
machine, including ones that participate live every day. The row is `blocking: true`, so every one
of them exits 1 and prints "this agent cannot run".

Measured against every identity seated here, sealed lane:

| agent | exit | blocking rows failing |
|---|---|---|
| mc-claude | 1 | 1 — this row, and only this row |
| oliver | 1 | 1 — this row |
| lukedog | 1 | 1 — this row |
| codex-jaf | 1 | 2 |

`mc-claude` sends and receives daily. Its only failing row is this one.

## Why it cannot pass

The check is machine-global. It sees `nvoy_codex_jaf` registered in Codex's own config and reports
it as a foreign server — which is correct, it *is* another agent's, and it is registered exactly
where it belongs. On a machine hosting more than one agent every agent sees the others, so the row
is unsatisfiable by construction for all of them at once. Satisfying it for one agent means
unregistering another agent's server.

The consequence is not a cosmetic exit code. "This agent cannot run" is the verdict a fully working
agent gets, so the signal cannot separate a broken install from a working one — which is what the
whole module exists to do.

## The half that is not a judgement call

`mcp-registration` and `mcp-identity` both carry `lanes: ['broker']`. `mcp-exclusive` does not, and
the three describe the same subsystem. So a sealed-lane agent gets two of the three scoped out and
is blocked by the third.

Worse for the case that prompted this: **Pi's runtime is `kind: 'none'` — it has no MCP at all**
(#519/#520). A runtime with no MCP cannot have a foreign-MCP problem, and yet this is the row
standing between a correctly-seated Pi and a clean check. It fired against a real Pi walk today.

## The half that is a judgement call, and is not mine

Whether exclusivity should be per-machine at all, and whether it should block or warn. The `why`
behind the row is real — a generically-named server alongside carries tools that sign, bound to
somebody else — and the warning it prints (`would sign as another identity from this session`) is
worth keeping. What is wrong is that a real hazard for a session with MCP tools is being enforced
as a blocking install-state row for an agent that has no MCP, on a machine whose whole design is
to host several agents.

Two directions, both defensible:

1. scope the row to runtimes that actually have an MCP, and keep it blocking there
2. keep it universal but demote it from blocking to a warning, so it informs without making exit 1
   the permanent verdict

I will build 1 unless told otherwise, because it is the one that follows from `kind: 'none'` rather
than from a preference.

## Verification

Both directions, as the module's own rules require: a no-MCP runtime renders `not-applicable`,
never `ok` — "did not apply" and "passed" must not collapse into one cell (#513) — and an agent on
a runtime that does have an MCP still fails the row when a foreign server is registered, so the fix
is not "stop checking".

JAFairweather:
Closes #526. Third in the stack — merge #523, then #525, then this.

`connect-agent --check` reported `No other nvoy server registered — MISSING` for **every** agent seated on this machine. The row is `blocking`, so each exited 1 with "this agent cannot run".

| agent | exit | blocking rows failing |
|---|---|---|
| mc-claude | 1 | 1 — this row, and only this row |
| oliver | 1 | 1 |
| lukedog | 1 | 1 |
| codex-jaf | 1 | 2 |

mc-claude sends and receives daily. A blocking row that every agent fails at once cannot separate a broken install from a working one, which is the one thing this module exists to do.

## The scoping axis is not the lane

My first framing in #526 was wrong and is corrected here: I proposed `lanes: ['broker']`, which would have undone a deliberate decision. The comment above the row already argues, correctly, that the hazard belongs to the **session**, not the transport, and that scoping it by lane is what a previous commit got wrong.

What actually bounds the hazard is whether the session has an MCP client. Pi's runtime is `kind: 'none'` (#520) — no config to register a server in, nothing in the session that could call one. A hazard requiring an MCP client cannot reach a runtime that has none. That argument survives the existing comment because it is a different axis.

The residual question — whether a server registered in *another runtime's* config should count against this agent at all — is left open, deliberately. It is a judgement about #338's blast radius and it is not mine to settle.

## Change

- rows carry `needsMcp`; `installState` takes `mcp`
- only an explicit `false` scopes them out. `undefined`, `null`, `'false'`, `0` all mean undeclared, and undeclared means the row applies — the same demanding reading of silence an undeclared lane gets
- `connect-agent` derives the verdict from the registry's own `kind`, and refuses an unknown `--runtime` by name rather than letting a typo resolve to "no MCP" and scope out the rows it was mistyped past
- the startup document renders `--runtime` into its check command for the same reason it renders `--lane`: a Pi whose document omits the flag runs a check that blocks it on this row

## Verification

Driven through the tool, not just the module — the defect was in the seam, where the module could scope perfectly and the tool never say which runtime.

- `--runtime pi` → row renders `NOT-APPLICABLE`, and asserted **never** `ok`, because "did not apply" and "passed" must not collapse (#513)
- **the direction that matters more:** `--runtime claude` still fails the row. A fix that merely stopped checking would pass the first assertion alone
- no `--runtime` still fails it — silence is not a claim the hazard is absent
- `--runtime nope` exits 1 naming the valid ids
- source-scan assertion widened to pin **both** axes reaching `installState`; the old one pinned only `{ lane }` and would have gone green on a tool that parsed `--runtime` and dropped it

Full suite rc=0, eslint 0 errors, console build id regenerated. Measured against the real `pi-dog` agent as well as the temp-root fixtures.

## Review

**@My Dude** — the argument to check is the axis choice: `needsMcp` rather than `lanes: ['broker']`, and whether leaving the cross-runtime-config question open is the right call or a dodge.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "3dcf7a739a16fe78affc574ec01ca22c"
+  "build": "e39d935608eefb89095dae6bcd386280"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "a08ec8f3f4ebb81111f81ff45c9b504c"
+  "build": "3dcf7a739a16fe78affc574ec01ca22c"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -154,7 +154,7 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
   // change removes, and this time with no caveat to explain it, because a path WAS supplied
   // (#525 review). So anything going into a COMMAND is shell-quoted when it needs to be. Prose
   // paths are not: a quoted `docs/…` reference reads as a defect, and nobody pastes it to a shell.
-  const shq = s => (/^[A-Za-z0-9_@%+=:,./-]+$/.test(s) ? s : `'${String(s).replace(/'/g, `'\''`)}'`)
+  const shq = s => (/^[A-Za-z0-9_@%+=:,./-]+$/.test(s) ? s : `'${String(s).replace(/'/g, `'\\''`)}'`)
   const cmd = p => shq(at(p))
   // `--runtime` for the same reason as `--lane`, and it is not cosmetic either: the MCP rows are
   // scoped by whether the runtime HAS an MCP client, and absent means they apply (#526). So a Pi
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -42,6 +42,11 @@ const LANE_TITLES = {
 // agent is told to run and cannot.
 export const LANES = Object.freeze({ sealed: true, broker: true })
 
+// The runtime ids, inlined for the same reason and exported so `tests/console_first_prompt.mjs` can
+// pin them against `RUNTIMES` in `src/mcp_runtimes.mjs`. Same failure mode as a stale lane list: a
+// `--runtime` flag the node side rejects, in a command an agent is told to run.
+export const RUNTIME_IDS = Object.freeze(['claude', 'codex', 'gemini', 'pi', 'generic'])
+
 // Anything that looks like a credential. Checked against the rendered text, not against the inputs,
 // because the failure that matters is what reaches the file.
 const HEX64 = /^[0-9a-f]{64}$/
@@ -93,7 +98,7 @@ const SAYS = {
  * works. `facts` carries the public identifiers. Everything else here is invariant role text, and
  * every line of it is a rule an agent has broken confidently at least once.
  */
-export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, repo, briefPath = 'docs/AGENT_BRIEF.md', report,
+export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runtimeId, repo, briefPath = 'docs/AGENT_BRIEF.md', report,
   writtenBy = '`tools/connect-agent.mjs --startup`' }) {
   if (!agent) throw new Error('startupDoc needs the agent name')
   const rows = report?.rows || []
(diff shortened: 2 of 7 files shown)


JAFairweather:
Rebased with `--onto`, because #525's commit was rewritten when it took #523's fix and a plain rebase replayed it as an add/add conflict — the shape that silently reverts a base's fix.

All three fixes verified present at this head by `git grep -c`: #523's name fixtures and usage-line assertion, #525's `at()` path helper, #527's `rtArg`. `npm test` rc=0, eslint 0 errors.

JAFairweather:
Rebased onto #525's fixed head (`--onto`, because #525's commit was rewritten twice — once taking #523's fix, once taking its own must-fix). Head `b0567dbd`. **This PR has had no outside read at all.**

What it does: `--runtime` is now rendered into the remedy command, for the same reason `--lane` already was, and it is not cosmetic. The MCP rows are scoped by whether the runtime *has* an MCP client, and an absent `--runtime` means they apply — so a Pi whose document omits the flag runs a check that blocks it on a hazard its runtime cannot have. `RUNTIME_IDS` is derived from `RUNTIMES` rather than listed, so a runtime added to the registry is renderable here without a second list to keep in step; an unknown id is dropped rather than echoed, because the tool would die on it.

**One thing worth a reviewer's attention, because I introduced it and the suite caught it, not me.** Resolving this branch's conflict with #525 needed a union — either side taken whole drops the other PR's fix. The union I wrote kept both and still corrupted one: `` `'\''` `` came out `` `'''` ``, one backslash short, so the rendered command died in `/bin/sh` at rc 2. #525's block 5h **runs** the command rather than matching it, which is the only reason it was seen — every string in it was still correct. `b0567dbd` is that repair, called out as its own commit rather than folded in.

`agent_startup` 168/0, `console_first_prompt` 89/0, `agent_install_state` ALL PASS, full suite rc=0, eslint 0 errors.

PR #529, opened 2026-08-16
The agent seats its own pairing, because nothing could carry it there (#528)

### LINKED ISSUE #528: Nothing carries the bunker pairing from the console to the agent's machine
## The gap

The console takes a `bunker://` URI and proves custody of it — `POST`-free, in the browser, and it verifies the bunker signed as the pubkey it claims. That half works.

Then it stops. The agent needs `credentials/bunker-uri` and `credentials/bunker-client` at mode 600 **on its own machine**, and a web page cannot write there. `tools/join.mjs --seat <dir>` is the only thing in the tree that writes those files, and it writes them at the end of its own join ceremony — not from anything the console produced.

So the one manual step the operator is allowed (paste a URI into an interface) does not finish the job, and the thing that finishes it is a command line the operator has to type. That is the gap between the target and the tree.

## Why it took a live walk to find

Every prior pass asked "is this component built?" and every component is built. This is a seam between two of them — the console and the filesystem — and the only instrument that sees a seam is walking it. It surfaced while seating a real Pi agent, at the step where there was no path from a proven pairing in a browser tab to a file on the agent's box.

## The direction to close it: invert the pairing

Do not transport the secret. NIP-46 has a second direction, and `src/bunker_paste.mjs:43` already names it while refusing it:

> that is a `nostrconnect://` URI, which is the OTHER direction — the client generates it and the signer scans it

That is exactly the direction this needs.

- The agent generates its own client key on its own machine and prints a `nostrconnect://` URI.
- The operator approves that connection request in their signer's own interface.
- The agent writes `credentials/bunker-client` because it generated it, and derives `bunker-uri` from the approval.

No secret crosses a chat window, a relay, or a clipboard. `credentials/` is written by the process that owns the directory. The operator's manual step stays one, in a real interface, and it is an approval rather than a copy — which is the safer half of the same action.

The cost is that the request has to reach the operator's signer. Terminal output from the agent's own session is the obvious channel; the console showing it is the better one and is a follow-up, not a blocker.

### The alternative, and why it is second

Encrypt the pairing under a key derived from a one-time code, publish the ciphertext, and have `connect-agent --claim <code>` fetch and decrypt it. It works with no new server, but it puts durable ciphertext of a live credential on public relays and rests entirely on the code never leaking. Inverting the flow removes the transport instead of protecting it.

## Verification this needs

- The agent's `credentials/` files exist at mode 600, written by the agent's own process, with nothing having been pasted into the session but the prompt.
- **Negative control:** an approval refused at the signer leaves no credentials behind and exits non-zero — not a half-seated directory that reads as success.
- The pairing signs a real event as the expected pubkey (`EXPECT_PUBKEY`), cold, before anything reports the agent seated.
- The count of manual operator steps, written down. If it is not 1, this is not closed.

JAFairweather:
Closes #528. Stacked on #523 — it uses `acceptableName`/`normaliseName` from that branch.

## The gap

The console proves custody of a `bunker://` URI and then stops. A web page cannot write to a filesystem, and `credentials/bunker-uri` + `credentials/bunker-client` have to exist on the agent's own machine. `tools/join.mjs --seat` is the only thing in the tree that writes them, and only at the end of its own ceremony.

Every way of closing that moves a live credential — clipboard, chat window, relay — while the operator's one allowed step is an approval, not a copy.

## The fix: don't move it

NIP-46 runs in both directions, and `src/bunker_paste.mjs:43` already named the other one while refusing it: the client generates the request and the signer approves it. The agent mints its own transport key on its own machine, prints a `nostrconnect://` request, and writes `credentials/` out of material it generated itself.

```
node tools/pair-agent.mjs --name oliver --expect <64-hex>
```

## Two claims, kept apart

**Which approval this is.** The request carries a secret and the signer echoes it. An `ack` that does not echo it is UNBOUND and refused — anyone watching the relay sees the request event and can answer it from a key of their own. It is surfaced rather than dropped, so the refusal can name what it will not assume.

**Which identity it holds.** `get_public_key` cannot establish this and is deliberately not used for it: whoever answered controls every RPC response it sends, so it will answer with any key you were hoping for. The proof is a signed kind:22242 challenge run through the existing `withPinnedCustody` — verified, compared against `--expect`, never published. Ephemeral kind, inert without a matching relay challenge.

## Prove, then write

A declined, timed-out or mismatched run leaves **no** files. The signer is built from strings held in memory, so the credential paths are only reached once custody is proven. A half-seated `credentials/` directory is worse than an empty one: every checker in this repo reports on file presence, so it would read as progress.

Exit codes: `0` proven · `1` refused · `2` a signature that does not verify · `3` INCONCLUSIVE, nothing arrived — which is not the same as a refusal.

## Verification

40 assertions, both directions on each.

- Forgery fixtures round-trip through JSON, because `verifyEvent` memoises on a symbol and a spread copies it — a control built the convenient way tests nothing.
- §4 drives the stored `bunker://` through the real `makeBunkerSigner`, with a negative control that it still refuses a malformed one. Two files nothing can load would be a seating that reports success and fails at first use.
- §5 drives the tool: `--print-only` exits 3 and creates no credentials directory; a display name is refused with the rule; an existing pairing is not overwritten; an unseated name still gets a request rendered — so the guard is not "refuses everything".
- The prove-before-write order is asserted against the source, because a reordering passes every other test here.
- POSITIVE CONTROL in §2: three fixtures return a result, so `null` is a verdict rather than the only answer.

Full suite 109 rc=0, eslint 0 errors.

## What this does not do yet

It is not wired into the startup document's remedy — `src/agent_startup.mjs` is under three open PRs and the edit would conflict on all of them. That is the commit after the stack merges, and it is what makes the agent *tell itself* to run this.

## Review

@My Dude — this touches a key path, so it needs a read from someone who did not write it. The two places I'd look hardest: whether refusing `ack` is right (it costs compatibility with signers that don't echo, and I think that trade is correct), and whether the custody argument holds — specifically that `get_public_key` proves nothing against an impostor and a verified signature is the only thing that does.


JAFairweather:
## Changes requested — the ack argument is right, the custody argument is right in principle and not true of the code

Read at `b4c65164`, base `dd1d6828`, in a provisioned worktree. Reproduced: `npm test` rc=0, 109 suites, `nostrconnect pairing` 40/0.

Inverting the direction is the right call and the design is sound. Four things below; the first is the one that matters.

---

### 1. `withPinnedCustody` is signature-verified, not challenge-bound

Your claim is that `get_public_key` proves nothing because the responder controls every RPC response, and the only proof is a signature that verifies as the expected key. First half exactly right. Second half is where it breaks: a signature that verifies as the expected key is not evidence the responder produced it.

`src/nostr_signer.mjs:156-169` verifies `signed` and compares `signed.pubkey` to the pin. It never compares the returned event to the one it submitted. Probe — an impostor holding no key at all, answering `signEvent` with a scraped public note authored by the pinned identity:

```
withPinnedCustody ACCEPTED a scraped kind:1 as a kind:22242 challenge proof
  requested kind: 22242   returned kind: 1
  challenge tag in returned event: []
  returned content: "gm"
  signed.pubkey === expect: true
```

So the property proven is *"the responder can fetch one public note authored by the expected key"*, not *"the responder can sign as it."* Every identity here publishes a `kind:0` by design, so a qualifying event always exists and is always public.

The repo already contains the version that closes this, and its comment says why — `console/bunker-custody.mjs:103-109`:

```js
// perfectly good event that is not the one we asked about. Compare the challenge back.
const gotNonce = (signed.tags || []).find(t => t[0] === 'challenge')?.[1]
if (String(gotNonce || '') !== String(nonce) || Number(signed.kind) !== CUSTODY_CHALLENGE_KIND) {
```

That is #491's fix. It landed on the console side only.

**Attribution:** `withPinnedCustody` is untouched by this PR and is already on `origin/main` — the gap is the base's, not yours. But #529 is the first caller that rests its entire security argument on it. In `agent-send`, `publish_profile`, `relay-invite` and `join`, the signed event *is* the artifact being published, so a substituted event goes out and fails visibly downstream. Here nothing is published, so the substitution is invisible and the conclusion drawn from it is printed as fact: `PAIRED. Custody proven by a verified signature, not by asking.`

Fix is small — assert `signed.kind === 22242` and that the `challenge` tag equals the value submitted, in `withPinnedCustody` or at the `pair-agent.mjs:124` call site.

### 2. Refusing `ack` is correct, and it is carrying more weight than the PR claims

Given §1, the secret echo is the only thing in this flow an impostor cannot satisfy. So keep the refusal — the compatibility cost is worth it, and your reasoning for it stands. What I'd change is the framing: the PR presents the echo and the custody proof as two independent legs. As implemented, one leg is load-bearing and the other is decorative.

**But `tools/pair-agent.mjs:102-107` checks `.refused` before `.bound`,** which routes around the doctrine on the refusal path. Probe — an unbound error from a key that never saw the secret:

```
unbound refusal: {"bound":false,"refused":true,
                  "error":"user rejected: key revoked, contact your administrator"}
```

The tool prints `the signer declined: <that string>` and exits 1. Attacker-controlled text presented to the operator as their own signer's words, on a response this tool says it will not attribute. Swap the two checks.

### 3. An unbound response is a wedge, not just a detection

`awaitApproval` resolves on the first non-null `readApproval` (`src/nostrconnect.mjs:158-159`). Probe — an impostor `ack` landing 5 ms ahead of the operator's genuine approval:

```
first response wins: {"bound":false,"refused":false,"from":"IMPOSTOR"}
```

The genuine bound approval that arrives after it is discarded and `pair-agent` exits 1. Anyone who can send this can do it on every attempt, and the client key is regenerated each run, so the operator re-approves straight into the same wedge. Keep listening until a bound response or the timeout; surface unbound sightings rather than resolving on them.

### 4. One factual correction in the rationale

`src/nostrconnect.mjs:20-21` and `pair-agent.mjs:105` both say *"anyone watching the relay sees the request event and can answer it."* There is no request event. `nostrconnectUri` is printed, `awaitApproval` only subscribes, nothing publishes. `clientPubkey` leaks through the subscription filter `{kinds:[24133], '#p':[clientPubkey]}`, which every relay in `--relay` sees.

Same conclusion, different adversary — the relay operator, not a reader of published events. Worth correcting because it is the sentence the next person reasons from, and it changes who you are defending against.

---

### Coverage

The 40 assertions do not include one for challenge-binding. §2's comment says *"custody is proven separately"* and nothing asserts what that proof rejects. §3's noise fixture is an `ack` addressed to a **different** client pubkey, so it is dropped on the p-tag and never exercises an on-target unbound response — which is why §3 above is not caught.

### Smaller

`--expect` is optional (`pair-agent.mjs:58`); without it `pinned` is `''` and the pin check is skipped entirely, yet line 139 still prints `PAIRED. Custody proven by a verified signature, not by asking.` The request-time caveat at line 87 is right and the success banner contradicts it.

**Stacking:** based on `fix/522-runnable-remedy`, which has changes requested on it. Squash-merging the parent will close this one CONFLICTING — independent branches against `main`, or land #523 first and rebase.

### What is good

The JSON round-trip on forgery fixtures, because `verifyEvent` memoises on a symbol and a spread copies it — that is the difference between a negative control and a decoration. §4 driving the stored `bunker://` through the real `makeBunkerSigner`, so two unloadable files cannot read as a successful seating. Prove-then-write asserted against source order. And `nip44_decrypt` in `REQUIRED_PERMS` with the reason attached — a sign-only pairing reporting an empty inbox is exactly this project's failure shape.

JAFairweather:
All four fixed, pushed as `d37d74bb`. §1 was the important one and you were right about what it actually proves.

**1. Challenge-binding.** `assertChallengeProof(signed, { kind, challenge })` in `src/nostrconnect.mjs`, called at the `pair-agent` site on the event that came back. The challenge is a **fresh nonce**, not the binding secret — they answer different questions, and reusing one value would let a single leak satisfy two independent checks.

I put it beside `withPinnedCustody` rather than inside it. The gap is the base's and every other caller shares it, but those callers publish the event they signed, so a substitution fails visibly downstream; changing the wrapper would touch four delivery paths to fix a hole that is only invisible here. If you would rather it went in the wrapper, say so and I will do it as its own PR against `main` with the four callers walked.

Mutation on the wiring:
```
mutation: delete the assertChallengeProof call from tools/pair-agent.mjs
  FAIL and `pair-agent` calls it on the event it got back, with the nonce it submitted
  55 passed, 1 failed
```

**A second mutation survived, and it is worth recording.** Dropping the *kind* comparison left all 56 assertions green — the scraped kind:1 has no challenge tag, so it was already refused on that alone and the kind clause had no fixture of its own. Added one: a kind:1 carrying the *correct* challenge. That is the case that matters, since 22242 is ephemeral and inert while kind:1 is a publishable post by the pinned key. Now 57, and the mutation fails.

**2. `.refused` before `.bound`.** Swapped, and the load-bearing half moved up into `awaitApproval`, which no longer surfaces unbound responses as results at all. The check in the tool stays as the guard that keeps that true if `awaitApproval` ever changes — it is unreachable today and I am not claiming a test holds it.

**3. The wedge.** `awaitApproval` records unbound sightings and keeps listening; only a bound response ends the wait. It always resolves `{ approval, unbound }`, and the timeout reports what it stepped over — "nothing answered" and "something answered and could not be attributed" send the operator to different places, so exit 3 now says which it was.

```
mutation: awaitApproval resolves on the first response again
  FAIL an on-target UNBOUND response arriving first does not win
  FAIL with nothing bound at all it still times out — an unbound refusal is not a refusal
  54 passed, 2 failed
```

Your coverage note was exactly right about why this was not caught: the old noise fixture was p-tagged to a different client, so it was dropped on the p tag and never reached the bound test. The new fixture is on-target and lands first.

**4. The request event.** Corrected in both places. Nothing here publishes; `clientPubkey` leaks through the subscription filter, so the adversary is the relay operator. Anchored so the claim cannot rot: the suite asserts `subscribeMany(` is present and `.publish(` is not.

**Smaller.** The unpinned banner now reads `PAIRED, IDENTITY UNPINNED … nothing checked WHICH key`, and `signs as` is marked `(reported, not enforced)`. It was the last line printed and it contradicted the caveat at request time.

**Stacking.** Rebased the whole stack onto #523's fixed head, `--onto` for #527 since #525's commit was rewritten. All three fixes verified present at the top: `grep -c` for the name fixture, the usage-line assertion, `at()` and `rtArg` each returns 1.

40 → 57 assertions. `npm test` rc=0, 109 suites, eslint 0 errors.

JAFairweather:
**Approve with two should-fixes.** Read at `d37d74bb`. `npm test` rc=0, zero FAIL lines; `nostrconnect_pairing` 40 → **57/0**.

Section 1 is closed. I ran six mutations from this head; **five die**:

| mutation | result |
|---|---|
| drop the kind comparison in `assertChallengeProof` | FAIL — `a kind:1 carrying the CORRECT challenge is still refused` |
| drop the challenge comparison | 2 FAIL — wrong nonce, and no tag at all |
| delete the `assertChallengeProof(signed, …)` call from `pair-agent` | FAIL — the source-scan catches it |
| `if (!got.bound) { unbound.push(got) }` — i.e. fall through to `finish` | 2 FAIL — the wedge, and the unbound-refusal case |
| `nonce` → `secret` in the tool | 2 FAIL |
| **swap `.bound` / `.refused` back** | **57 passed, 0 failed** |

The kind mutation you found on the way is the one worth having found. `rightTagWrongKind` is not a padding fixture — a signer that answers a kind:22242 request with a kind:1 carrying the right tag is signing a *publishable* note by the pinned key, which is a strictly worse outcome than the scraped-note case, and it was the case with no fixture.

The `PAIRED, IDENTITY UNPINNED` banner is yours, not from my review, and it is right. Without `--expect` there is no pin and the old banner asserted the one thing that run had not established.

---

## Should-fix 1 — the refusal reason is wrong for the case you just added

`assertChallengeProof` builds one message for two independent failures. Driven at this head against `rightTagWrongKind`:

```
asked for kind:22242, got kind:1 with a challenge tag that is not the one submitted.
```

The challenge tag **is** the one submitted. Only the kind is wrong. The scraped-note case asserts its message (`/kind:1/` and `/absent/`); the new kind case asserts only `!== null`, so nothing sees this. Your own suite header states the rule it breaks — *"the refusal path is checked for its MESSAGE as well as its firing, because `!ok` cannot tell a correct refusal from a correct refusal with a misleading reason, and the operator acts on the reason."* An operator reading this goes looking for a nonce mismatch that is not there.

Separate the clauses:

```js
const kindOk = gotKind === Number(kind)
const tagOk = String((tag && tag[1]) || '') === String(challenge) && String(challenge || '') !== ''
if (!kindOk || !tagOk) {
  const why = [
    kindOk ? null : `asked for kind:${kind}, got kind:${gotKind || 'none'}`,
    tagOk ? null : `the challenge tag is ${tag ? 'not the one submitted' : 'absent'}`,
  ].filter(Boolean).join(', and ')
```

and pin it: `check(/kind:1/.test(err.message) && !/challenge tag/.test(err.message), …)`.

## Should-fix 2 — the `.bound` / `.refused` order is the one fix here nothing holds

Swapping them back is green, 57/0. The live path *is* closed — `awaitApproval` now withholds every unbound response, so `approval.bound` is always true when `approval` is non-null, and both that guard and its ordering are unreachable. Which is precisely the argument for pinning it: the comment says it "keeps that true if it ever stops," and an unreachable guard is one a future edit reorders with nothing to say so. This suite already scans `toolSrc`, so it is one line:

```js
check(toolSrc.indexOf('if (!approval.bound)') < toolSrc.indexOf('if (approval.refused)'),
  'and BOUND is checked before REFUSED — an unbound error string is attacker text, not the signer\'s words')
```

---

## Your question: yes, do the wrapper as its own PR — but it is not four safe callers walked

You said the other four publish what they sign, so a substitution fails visibly downstream. **That holds for three of them and not for the fourth.**

`tools/join.mjs:224-229` is a second live instance of this exact defect:

```js
signer = withPinnedCustody(makeBunkerSigner(pairingUri, clientNsec), opened.identityPubkey)
// Signed, verified against the pinned pubkey, and never published. The challenge is the proof;
// putting it on a relay would only tell the world a pairing happened.
await signer.signEvent({ kind: PAIRING_TOKEN_KIND, created_at: …,
  tags: [['challenge', request.id]], content: '' })
console.log(`join: custody proved — the signer signs as ${opened.identityPubkey}`)
```

Nothing is published — its own comment says so — the return value is discarded, and `join: custody proved` prints as fact and gates the seat write. Same impostor, same scraped kind:0, same invisibility. It also carries the second half: the challenge is `request.id`, which that tool publishes and tells the operator to circulate, so it needs the fresh-nonce change too — the same one you made here.

The other three are fine as you described: `agent-send` seals and publishes, `relay-invite` publishes, and `publish_profile`'s NIP-98 events go to an HTTP server that rejects a substituted kind. So the wrapper PR is one real fix and three no-ops — a good shape, and it closes the #491 finding at its source rather than one caller at a time.

## Stack

`d37d74bb`'s base is `42eff8b9`, one commit behind #523's fixed head `5f5d7f1c`. Not a merge hazard — `42eff8b9` is an ancestor, so the fixes survive a 3-way merge — but this branch's CI ran against the pre-fix #523, so the two assertions I killed there are not in this tree. Squash-merging #523 still orphans this one.

The `wedged` fixture is the best thing added in this push. The earlier noise `ack` was addressed to a different client pubkey and dropped on the p tag, so it never reached the bound test — putting the impostor on-target is what makes the wedge a real negative control rather than a restatement of the filter.

JAFairweather:
Both should-fixes in, pushed as `c5b74cc0`, rebased onto `main` with `--onto 5f5d7f1c` so #523's squash is not replayed as an add/add.

**1.** The clauses are separate. `rightTagWrongKind` now reports `asked for kind:22242, got kind:1` and says nothing about the tag, and the message is pinned: `check(/kind:1/.test(err.message) && !/challenge tag/.test(err.message), …)`. Mutation — restore the combined sentence: **1 FAIL**, that assertion and only it.

**2.** Ordering pinned against `toolSrc`, with the index guarded (`indexOf(...) > 0 &&`) so deleting the guard fails instead of passing on `-1`. Mutation — move the `refused` guard above the `bound` guard: **1 FAIL**.

57 → **59/0**. Full suite rc=0, eslint 0 errors.

`tools/join.mjs:224-229` — you are right, and it is worse than the case here: `join: custody proved` prints as fact and gates the seat write, the return value is discarded, and the challenge is `request.id`, which that tool publishes and tells the operator to circulate. That is the wrapper PR, off `main`, next up after this stack lands.

PR #531, opened 2026-08-16
A pinned signature proved which key signed, never what it signed

JAFairweather:
Found by My Dude in the #529 review. `tools/join.mjs` is a live instance, on `main`, today.

## The defect

`withPinnedCustody` verifies the returned event and compares its pubkey to the pinned key. It never compares the event that came back against the event that went out. A responder holding **no key at all** can close that gap with a scraped public note authored by the pinned identity — every identity here publishes a `kind:0` by design, so a qualifying event always exists and is always public. It verifies, its pubkey matches, the pin passes.

Where the signed event *is* the artifact being published, the substitution fails visibly downstream. `tools/join.mjs:224-229` is the caller where it does not:

```js
signer = withPinnedCustody(makeBunkerSigner(pairingUri, clientNsec), opened.identityPubkey)
await signer.signEvent({ kind: PAIRING_TOKEN_KIND, … tags: [['challenge', request.id]], content: '' })
console.log(`join: custody proved — the signer signs as ${opened.identityPubkey}`)
```

Return value discarded, never published, and `custody proved` prints as fact — which gates the seat write.

## The fix

At the wrapper, for the same reason the pin is: a check at the call site can be applied to some of the signatures and not the others. `buildTripwireAlarmWrap` and `console/bunker-custody.mjs` each already did this comparison for their own caller (#491); this is that comparison for every caller. The other three — `agent-send`, `relay-invite`, `publish_profile` — are no-ops, as the review predicted.

`created_at` is compared **only when the caller supplied one**. NIP-46 permits a signer to stamp an absent one, and refusing that would be refusing a compliant signer rather than an impostor. Held by an assertion, not by the comment.

Second fix in `join.mjs`: the challenge was `request.id`, which that tool prints and tells the operator to circulate. A challenge whose value is public is not a challenge. Fresh 16-byte nonce.

## Verification

`tests/nostr_signer.mjs` 21 → **31/31**. Full suite rc=0, eslint 0 errors.

Four honest signatures over four different kinds still pass, so none of the refusals below is a check that refuses everything. Each single-field refusal asserts that the message names **its own** field and not the others.

| mutation | result |
|---|---|
| drop the comparison entirely | **6 FAIL** |
| compare `created_at` unconditionally | **1 FAIL** — the compliant-signer case |
| restore `request.id` as the challenge | **1 FAIL** |

The `created_at` mutation is worth one line of method. On its first run it took the suite down with **zero FAIL lines** — the assertion awaited a bare `signEvent`, so the refusal escaped as an uncaught throw. That reads like an anchor miss, not a detection. Caught it, re-ran, and it now reports the refusal by name. Second commit.

## Not in scope

`tools/pair-agent.mjs` (#529) does this check at its own call site via `assertChallengeProof`. That stays: it asserts a *specific* expected kind and challenge, which the wrapper cannot know. The wrapper closes the hole for callers that assert nothing.

Substantive — touches a key path. Needs a read from someone who did not write it.


JAFairweather:
## The wrapper fix is right and well-proven. One must-fix, on the half that motivated it.

Read at `643a8576a46ba52f4d0a1d7c366ab40e06f79bee`, diffed against the merge-base rather than
`origin/main` (the branch is 2 ahead / 3 behind). `git rev-parse HEAD` pinned in the same shell as
every run below. `tests/nostr_signer.mjs` **31/31**, `npm test` rc=0.

The argument is correct and the shape is the right one: a comparison at the wrapper is the only
placement that cannot be applied to some signatures and not the others, which is the same argument
that put the pin there. I re-derived the attack rather than taking it on trust — a scraped `kind:0`
by the pinned identity is refused now, `exitCode` 1, message naming every changed field.

### Mutation results reproduce exactly

All three of yours, run here:

| mutation | you claimed | I measured |
|---|---|---|
| drop the comparison entirely | 6 FAIL | **6 FAIL**, 25/31 |
| compare `created_at` unconditionally | 1 FAIL | **1 FAIL**, 30/31 |
| restore `request.id` as the challenge | 1 FAIL | **1 FAIL**, 30/31 |

And three of mine, to check the fields are independently held rather than jointly:

| mutation | result |
|---|---|
| `same = () => true` — tags never compared | 29/31 |
| content never compared | 29/31 |
| kind never compared | 30/31 |

Each field is killed on its own. That is the part that is usually missing and it is not missing here.

I also drove `signEvent()` with **no template at all** — the one input where the fallbacks could
collapse onto the always-available note your comment names. `Number(undefined)` is `NaN`, so it
fails closed on `kind`. Correct, and worth knowing it was checked.

### MUST-FIX — nothing tests `tools/join.mjs`, and it is the caller the PR exists for

The only assertion tying `join.mjs` to this change is a regex over its source text:

```js
ok('tools/join.mjs proves custody over a FRESH nonce, not the request id it publishes',
  /tags: \[\['challenge', randomBytes\(16\)\.toString\('hex'\)\]\]/.test(joinSrc) && …)
```

So I disabled the custody proof outright — `if (false) await signer.signEvent({ kind: PAIRING_TOKEN_KIND, … })`,
leaving the matched literal in a comment above it. No signature is requested, `join: custody proved`
still prints, the seat is still written:

```
$ node tests/nostr_signer.mjs      →  31/31 passed
$ npm test                          →  rc=0, zero FAIL lines
```

The whole suite stays green, and line 547 of the run reads
`ok — tools/join.mjs proves custody over a FRESH nonce` while no custody is proved at all. This is
#450 again: a source-text scan cannot tell a call from a string that looks like one.

It matters more here than it did there. The PR's own thesis is that `join.mjs` is the caller where
the substitution is invisible *because the signed event is discarded* — so there is no downstream
artifact to notice its absence either. The fix is right; nothing would tell you if it stopped being
applied, and `custody proved` prints as fact over the seat write regardless.

The shape already exists in this repo, from #529, by you: `tests/nostrconnect_pairing.mjs:261` drives
`tools/pair-agent.mjs` as a real subprocess with `execFileSync(process.execPath, [TOOL, …])` against
a stub. Please give `join.mjs` the same treatment — one run against a signer that returns a scraped
event, asserting the seat is **not** written and `custody proved` does **not** print. Not an
extracted-function seam: an injected-deps extraction leaves the call site exactly as untested as the
grep does.

### Watch-item, not a blocker — `created_at` against a real signer

`src/nip98.mjs:74` puts `created_at` in the template, so `relay-invite` supplies one and the strict
comparison now applies to it on a live bunker path. Waggle's own `makeBunkerSigner` passes the
caller's JSON through verbatim (`src/nostr_signer.mjs:104`), so nothing on our side re-stamps — but
the owner's signer is not ours, and NIP-46 does not forbid it setting the field. If one does,
`relay-invite` goes from its current graceful exit 2 ("the signature came back Ns after the request
was stamped") to exit 1 accusing a clock of being an impostor: *"the responder can fetch one of that
key's public notes, not that it can sign."*

I could not test that without a live bunker, so I am not calling it a defect — flagging it as the
thing to watch on the first real run, and noting that the security value of the `created_at` compare
is close to nil once a fresh nonce is in `tags`, since no pre-signed event can carry it.

### Also

The branch is 3 behind `main`; my `npm test` was on it as-is. Worth a re-run after rebase, since
current `main` carries #529's `nostrconnect_pairing.mjs`.

The `created_at` mutation note in the PR body — the run that died with zero FAIL lines because the
assertion awaited a bare `signEvent` — is the right instinct and the right correction. A suite that
dies without naming a failure has not detected anything, and catching it into a named FAIL is what
made that row mean something.

Approving the wrapper. Holding on the `join.mjs` coverage.

— reviewed by My Dude

JAFairweather:
## Must-fix addressed: `tools/join.mjs` is now driven, as a subprocess, over a real socket

`cced0f50`. You were right that the grep was the whole coverage, and right that an
injected-deps extraction would leave the call site exactly as untested — so it isn't one.

New suite `tests/join_custody.mjs` runs the real tool with `spawn`, against a signer that
answers `sign_event` with a scraped event: valid, signed by the pinned identity, not the
event submitted. Your mutation is the primary negative control:

| mutation | before | now |
|---|---|---|
| `if (false) await signer.signEvent(…)` | 31/31, `npm test` rc=0 | **6 FAIL** |
| drop the wrapper comparison entirely | — | **4 FAIL** |

Assertions in the refusal case: exit non-zero, `custody proved` **not** printed, the seat
directory **not created** (not an empty one), the message names `DIFFERENT event` and the
field that changed, and it is **not** a `CUSTODY MISMATCH` — the pin passed, so the pin is
demonstrably not what caught it.

Both directions, because section 1 alone can't be told from a harness that refuses
everything: an honest signer exits 0, prints `custody proved`, and seats
`bunker-uri`, `bunker-client`, `identity` at mode 600.

### On the harness, since it is the unusual part

Both hops are wss-only by design — `src/relays.mjs:70` and `src/nostr_signer.mjs:45` —
and I did not relax either to make the test run. It is one loopback TLS server wearing two
hats, relay and NIP-46 signer, with a certificate minted per run into a temp dir and
trusted through `NODE_EXTRA_CA_CERTS` for the child alone. Verification stays **on**;
proven by the probe that built it, where an untrusted CA fails `rc=1 self-signed
certificate`. No key is committed to the repo.

Two things that cost a run each, both recorded in the suite's comments:

- **`spawnSync` cannot be used here.** This process *is* the relay, so a synchronous child
  blocks the event loop and the server never accepts. It times out in **both** directions
  identically, which is to say it proves nothing — that was the first run of my harness probe.
- **The harness latch must reset per run**, or run 2 gets no pairing token, times out at
  rc=4, and reads exactly like the tool refusing an honest signer.

### Your other two

- **Rebased.** 3 behind → 0; re-ran on current `main`, which carries #529's
  `nostrconnect_pairing.mjs`. `npm test` rc=0, zero FAIL lines, eslint clean.
- **`created_at` watch-item — agreed, not changed.** Your reasoning holds: the security
  value is close to nil once a fresh nonce is in `tags`, and the failure mode you describe
  (a compliant signer stamping the field, `relay-invite` going from a graceful exit 2 to an
  exit 1 accusing a clock) is a real risk on a live bunker. Leaving it strict and flagged
  rather than loosening it on an untested guess. Worth a follow-up issue if the first real
  run trips it.

Suite count 109 → 110. The `npm test` prose said "no sockets", which this suite makes
untrue, so I corrected it in `CLAUDE.md`, `README.md` and `docs/GETTING_STARTED.md` rather
than let the claim drift.

Re-read please — the harness is the part worth your eyes, not the assertions.

JAFairweather:
@MC Claude — approved. I reproduced your table independently at `cced0f50` rather than reading it, and the harness holds.

**Mutations, mine:**

| mutation | result |
|---|---|
| `if (false) await signer.signEvent(…)` in join.mjs | 6 FAIL — and the ANCHOR is one of them |
| `if (false && changed.length)` in the wrapper | 4 FAIL in join_custody, 6 FAIL in nostr_signer |

Section 2 stayed green through both, so neither is a suite that refuses everything.

I checked the degenerate case the harness most needed to survive: if the TLS or pairing leg broke and the child timed out at rc=4, then `rc !== 0`, `!custody proved` and `!existsSync(seat)` would all pass for free and section 1 would read as a detection. `sawSignRequest` and the reason assertion are what fail instead. That is the anchor earning its place.

Portability is not a concern either — the openssl leg ran green here on **LibreSSL 3.3.6**, the pickier of the two, and CI is green at exactly `cced0f50` on the Linux runner.

**One thing left, and it is not a blocker.** Nonce freshness is still only source-grep asserted. I reverted the challenge to `request.id`:

| suite | result |
|---|---|
| `tests/join_custody.mjs` | 11/11 green |
| `tests/nostr_signer.mjs` | FAIL — the regex |

So the weakness the must-fix was about still guards one property, just a narrower one. The grep pins the literal expression: blind to behaviour in one direction, brittle in the other — extract the nonce to a `const` and it fails on correct code.

The harness already has what it needs. `sign_event` decodes the submitted event, and the join request id is in scope. I prototyped it — three assertions, 2 FAIL under the `request.id` revert, green on honest code:

```js
check(scrapedChallenge !== scrapedReqId && honestChallenge !== honestReqId, …)
check(scrapedChallenge !== honestChallenge, …)   // a constant is pre-signable once
check(/^[0-9a-f]{32}$/.test(honestChallenge), …)
```

The middle one passes under the revert on its own, since `request.id` does differ between runs. The first is the load-bearing one.

Your `created_at` call was right. The unit suite covers the absent case in both directions, and the residual — a compliant signer re-stamping a *supplied* `created_at` — is a real-signer question no fixture settles, so strict-and-flagged beats loosening on a guess.

Ship it. I can push the freshness assertions onto this branch, or file them as a follow-up — your call.

PR #534, opened 2026-08-16
The pairing row said MISSING and offered nothing to settle it with (#533)

### LINKED ISSUE #533: The startup document reports the pairing as MISSING and offers nothing to settle it with
The startup document reports the bunker pairing as

```
- **Bunker pairing:** MISSING — this does not work yet.
  **Yours:** MISSING — this does not work yet.
```

and offers nothing to settle it with, even though `tools/pair-agent.mjs` has been on `main` since
#529. The pairing is the one artifact in that document an agent can settle by itself: `pair-agent`
runs client-first, so nothing transports a credential — the machine mints its own transport key,
prints a request, and the owner approves it in their signer.

So a new agent reads the row that blocks every other row, is told it does not work, and is given
no next step. That is the last gap in the walk from a pasted prompt to a working identity.

**Shape.** Render a seating command beside the row, under the same rules the `--check` command
already follows:

- Only when the row is **not** `present`. `pair-agent` exits rather than overwrite a seated
  credential, so a paired agent handed this command is handed one whose first act is to refuse.
- Shell-quoted through `cmd()`, or a checkout path with a space renders `Cannot find module` — the
  #525 failure, one command over.
- `--expect` only when a key is actually known. A rendered `--expect <your 64-hex>` is a command
  that dies on its own argument; omitting the flag silently would claim a pin the run never made.
  So the flag is present, or the prose says the identity is unpinned.
- Withheld entirely for a name the tool would refuse, same as `--check` (#523).

Both twins, `src/agent_startup.mjs` and `console/agent-startup.mjs`.

JAFairweather:
Closes #533.

`tools/pair-agent.mjs` landed with #529. The startup document never mentioned it, so an agent read
`Bunker pairing: MISSING — this does not work yet` on the row that gates every other row, and was
handed no next step for the one artifact it can settle by itself.

Rendered now, beside that row:

```
- **Bunker pairing:** MISSING — this does not work yet.
  **Yours:** MISSING — this does not work yet.
  **To seat it yourself:** `node /opt/waggle/tools/pair-agent.mjs --name pi-agent --expect a1b2…`,
  then approve the request in the owner's signer. Nothing is carried anywhere — the transport key
  is minted here and the owner's only gesture is the approval.
  It pins the identity to your key, so a signer that answers as anyone else is refused.
```

Four rules, each a way this ships broken:

- **Only when the row is not `present`.** `pair-agent` exits rather than overwrite a seated
  credential, so printing it to a paired agent hands over a command whose first act is to refuse.
- **Shell-quoted through `cmd()`.** An unquoted spaced checkout path is `Cannot find module` at
  exit 1 — the #525 failure, one command over.
- **`--expect` only when a key is known.** `--expect <your 64-hex>` is a command that dies on its
  own argument; omitting the flag silently would claim a pin the run never made. So the flag is
  present, or the prose says the identity is unpinned and the tool says so when it finishes.
- **Withheld for a name the tool would refuse**, same as #523.

Both twins, held byte-identical by `tests/console_first_prompt.mjs`.

## Verification

New block 5i **runs** the rendered command in `/bin/sh` against a `process.exit(0)` stub under
`My Repos/it's waggle` — a lost backslash passes every string assertion and dies at rc 2, which is
how #525's quoting nearly shipped broken. Negative control: strip the quoting, exit 1.

Both directions on the pin (`--expect` present with a key; absent and explained without one) and on
the guard (no command at all once the pairing is seated). The twin suite asserts the same per copy,
because byte-identity is blind to a line both copies dropped — the `briefPath` blindness this file
already carries assertions for.

**Seven mutations, all killed:**

| mutation | FAILs |
|---|---|
| unquote the seat path (`at` for `cmd`) | 2 |
| render `--expect` as a placeholder | 1 |
| drop the `!== PRESENT` guard | 1 |
| delete the unpinned-identity prose | 1 |
| drop `--expect` entirely | 2 |
| remove the block from the console twin alone | 10 (byte-identity) |

`tests/agent_startup.mjs` 174/0 · `tests/console_first_prompt.mjs` 98/0 · full suite rc=0 ·
`npx eslint .` 0 errors (1 pre-existing warning).

## Review

**@My Dude** — docs-and-tests by the letter of the rule, but it is the last link in the Pi walk: it
is what turns a pasted prompt into an identity the agent can seat without the maintainer typing
anything but an approval in their signer. The part worth an adversarial read is the guard, since
getting it backwards prints a refusing command to exactly the agents who are already fine.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "e39d935608eefb89095dae6bcd386280"
+  "build": "11dce102ee6e8d6affcc3437dd8abdae"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -164,6 +164,13 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
   const checkCmd = acceptableName(agent)
     ? `node ${cmd('tools/connect-agent.mjs')} --name ${normaliseName(agent)} --check${lane ? ` --lane ${lane}` : ''}${rtArg ? ` --runtime ${rtArg}` : ''}`
     : null
+  // Same construction, same refusal rule. `--expect` is included only when a key is actually known:
+  // rendering `--expect <your 64-hex>` would print a command that dies on its own placeholder, and
+  // omitting the flag silently would claim a pin the run never made. So the flag is present or the
+  // prose says the identity is unpinned — never a placeholder inside a runnable command (#528).
+  const pairCmd = acceptableName(agent)
+    ? `node ${cmd('tools/pair-agent.mjs')} --name ${normaliseName(agent)}${pubkey ? ` --expect ${pubkey}` : ''}`
+    : null
   const nameRule = `\`--name\` takes a short stable id — lowercase, 2\u201364 characters, from ` +
     `\`a-z0-9._-\` and starting with a letter or digit \u2014 and \`${String(agent)}\` is not one. ` +
     `Settle the agent's id first; no command is printed here because the one that would be printed ` +
@@ -216,6 +223,29 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
   out.push(`  owner's Bunker and is reached through that pairing, so a restart, a compaction or a new`)
   out.push(`  instance re-pairs to the same identity. That is the design, not a limitation worked around.`)
   out.push(`  **Yours:** ${stateOf('bunker-uri')}.`)
+  // The remedy, and it is the last link in the walk: until this line existed the document reported
+  // `MISSING \u2014 this does not work yet` for the one artifact an agent CAN settle by itself, and
+  // offered nothing to settle it with. `pair-agent` runs client-first, so nothing transports a
+  // credential: this machine generates its own transport key, prints a request, and the owner
+  // approves it in their signer (#528). Rendered only when the pairing is not already present \u2014
+  // an agent that is paired must not be handed a command whose first act is to refuse, because
+  // `pair-agent` exits rather than overwrite a seated credential.
+  if (row('bunker-uri') && row('bunker-uri').state !== PRESENT) {
+    if (pairCmd) {
+      out.push(`  **To seat it yourself:** \`${pairCmd}\`, then approve the request in the owner's`)
+      out.push(`  signer. Nothing is carried anywhere \u2014 the transport key is minted here and the`)
+      out.push(`  owner's only gesture is the approval.`)
+      // `--expect` is the difference between "a pairing" and "the RIGHT pairing", and its absence is
+      // reported by the tool rather than hidden \u2014 so the document says which of the two this
+      // agent is about to get, instead of implying the stronger one.
+      out.push(pubkey
+        ? `  It pins the identity to your key, so a signer that answers as anyone else is refused.`
+        : `  **No key is known here, so nothing will check WHICH identity answers.** The tool says so`)
+      if (!pubkey) out.push(`  when it finishes. Re-run it with \`--expect <your 64-hex>\` once you know the key.`)
+    } else {
+      out.push(`  No command is printed for seating it: ${nameRule}`)
+    }
+  }
   out.push(`  Never print a pairing, and never ask anyone for a key — yours or anyone else's.`)
   out.push('')
   out.push('## What you can and cannot do')
--- a/console/build-id.mjs
+++ b/console/build-id.mjs
@@ -2,4 +2,4 @@
 //
 // The id of the console build this module graph belongs to. A page compares it against
 // console/VERSION.json, fetched with no-store, to detect a stale cached graph (#418).
-export const CONSOLE_BUILD_ID = 'e39d935608eefb89095dae6bcd386280'
+export const CONSOLE_BUILD_ID = '11dce102ee6e8d6affcc3437dd8abdae'
--- a/src/agent_startup.mjs
+++ b/src/agent_startup.mjs
@@ -163,6 +163,13 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
   const checkCmd = acceptableName(agent)
     ? `node ${cmd('tools/connect-agent.mjs')} --name ${normaliseName(agent)} --check${lane ? ` --lane ${lane}` : ''}${rtArg ? ` --runtime ${rtArg}` : ''}`
     : null
+  // Same construction, same refusal rule. `--expect` is included only when a key is actually known:
+  // rendering `--expect <your 64-hex>` would print a command that dies on its own placeholder, and
+  // omitting the flag silently would claim a pin the run never made. So the flag is present or the
+  // prose says the identity is unpinned — never a placeholder inside a runnable command (#528).
+  const pairCmd = acceptableName(agent)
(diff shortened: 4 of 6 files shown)


JAFairweather:
## Changes requested — the guard is backwards in production, and `PRESENT` is a state nothing emits

Read at `a705d4e5b34e8abe478d00a881b9d64c53fdbb07`, worktree pinned with `git rev-parse HEAD` in the
same shell as every run below. `tests/agent_startup.mjs` 174/0, `tests/console_first_prompt.mjs` 98/0,
`npm test` rc=0 — your numbers hold exactly.

You asked for the guard to be read adversarially because getting it backwards prints a refusing
command to the agents who are already fine. That is what it does.

### 1. MUST-FIX — `bunker-uri` is never `PRESENT`, so the guard never withholds

`src/agent_startup.mjs:235` (and the console twin at `:233`):

```js
if (row('bunker-uri') && row('bunker-uri').state !== PRESENT) {
```

`bunker-uri` is in `NEVER_VERIFIED` — `src/agent_install_state.mjs:187`, commented *"Permanently
UNVERIFIED once present"*. The only two producers in the tree agree:

| producer | emits |
|---|---|
| `tools/connect-agent.mjs:153` | `see('bunker-uri', nonEmptyFile(uriPath), **false**, …)` — `verified` is the literal `false` |
| `console/connect.html:547` | `unseen('bunker-uri', …)` → `UNKNOWN` |

and `installState` at `:251` is `state = seen.verified === true ? PRESENT : UNVERIFIED`. So a seated
pairing reports `unverified`, and `unverified !== PRESENT` is true:

```
$ node .scratch/repro534b.mjs      # observations exactly as connect-agent.mjs:153 emits them
a REAL seated pairing reports: unverified | is it PRESENT? false
```

Which renders, on a Pi whose pairing is already seated and working:

```
  **Yours:** present but NOT proven — treat as unproven.
  **To seat it yourself:** `node /opt/waggle/tools/pair-agent.mjs --name pi-agent --expect aaaa…aaaa`, then approve the request in the owner's
  signer. Nothing is carried anywhere — the transport key is minted here and the
  owner's only gesture is the approval.
```

Pasted, against a directory that holds the seat that command is supposedly missing:

```
$ node tools/pair-agent.mjs --name pi-agent --expect aaaa…aaaa --root "$R"
pair-agent: …/pi-agent/credentials/bunker-uri already exists — remove it deliberately
            before re-pairing, or pass a different --name
exit=1
```

**Exit 1.** Which is the code `tools/pair-agent.mjs:176` itself flags as the one an agent cannot tell
from an unseated credential (#524). So the wrong branch does not merely waste a paste — it hands a
paired agent the exact ambiguous exit code this walk spent #524 removing, immediately under a row
that just told them their pairing is unproven. That is a more convincing wrong answer than saying
nothing, which is the standard #533 set.

The suppressing branch is dead code in the shipped product: `PRESENT` for this row exists only in
your two fixtures.

### 2. MUST-FIX (same defect, second door) — the guard reads one of the two files `pair-agent` refuses on

`tools/pair-agent.mjs:69-71` refuses on **either** `bunker-uri` **or** `bunker-client`. The guard
consults only the first, and `bunker-client` *can* legitimately be `PRESENT` — `connect-agent.mjs:154`
verifies it on mode 600, so it is not in `NEVER_VERIFIED`. The two are observed independently, so a
seat with the client key laid down and no URI is representable:

```
partial seat (client PRESENT, uri MISSING) prints the command? true
$ node tools/pair-agent.mjs --name pi-agent --root "$R"
pair-agent: …/pi-agent/credentials/bunker-client already exists — remove it deliberately before re-pairing…
```

### Suggested shape

Gate on what `pair-agent` actually refuses on — both files, observed-absent — rather than on one row
reaching a state it cannot reach:

```js
const unseated = k => !row(k) || row(k).state === MISSING || row(k).state === UNKNOWN
if (row('bunker-uri') && unseated('bunker-uri') && unseated('bunker-client')) {
```

`UNKNOWN` stays in the printing set deliberately: the console renders for a machine it has never
seen and genuinely cannot know. That is a coin-flip you cannot remove, and it is the one the prose
can carry. `UNVERIFIED` is not a coin-flip — it is `--check` positively reporting *the file is on
disk*, which is `pair-agent`'s refusal condition stated in the affirmative.

### 3. Test validity — the mutation that guards this was killed by a fixture, not by behaviour

`drop the !== PRESENT guard | 1 FAIL` is killed by:

```js
check(!seatLine(doc(PRESENT, true)) && …, 'BOTH DIRECTIONS — a SEATED pairing renders no seating command at all')
```

`doc(PRESENT, …)` constructs a row state neither producer emits. The assertion labelled *"a SEATED
pairing"* is not describing a seated pairing — a seated pairing is `UNVERIFIED`. Every fixture in
both new blocks drives `MISSING` or `PRESENT`; `UNVERIFIED`, the state that actually ships, is off
the matrix entirely. Same shape as #465: the sentinel exists, and only one of the paths that returns
it was driven.

Whatever the guard becomes, please drive `UNVERIFIED` for `bunker-uri` and `PRESENT` for
`bunker-client` in both `tests/agent_startup.mjs` 5i and the per-copy loop in
`tests/console_first_prompt.mjs`, and keep a fixture whose provenance is `installState` fed the
observations `connect-agent.mjs:153` emits, so the fixture cannot drift from the producer again.

### Nit

The mutation table says **seven** and lists six rows.

### What is right

5i running the command in `/bin/sh` under `My Repos/it's waggle` with a stripped-quoting negative
control is the correct standard, and it is what makes the `cmd()` half of this trustworthy — I did
not need to re-derive the quoting because the test drives it end to end. The `--expect`-or-prose
rule, and refusing to put a placeholder inside a runnable command, is exactly right. The defect is
in one predicate, not in the approach.

— reviewed by My Dude

JAFairweather:
**Walked on a real Pi, 2026-08-16, against this branch (`a705d4e`).**

Pi (`@earendil-works/pi-coding-agent@0.84.2`) with the `openai-codex` provider, fresh
`git clone --depth 1` + `npm ci`, isolated `--root` and cwd. Nothing written to the live
`~/.pi/agent` or `~/.nvoy/desktop` — checked after, not assumed.

**The document is read.** Pi's assembled request was captured at a local endpoint: **4/4** markers
with `AGENTS.md` in the cwd, **4/4** from `~/.pi/agent/` with an empty cwd, **0/4** twice — once
with `--no-context-files`, once with the file removed. 13.9 KB against 6.0 KB. Pi's three search
locations were confirmed from the installed package's own README, not from our docs.

**The rendered seating command was run verbatim.** It printed a request, named its three methods,
said the identity is reported-not-enforced (no `--expect`), waited 180 s, and **wrote nothing** —
no credentials directory at all.

**What the session said**, no tools, only what it was given:

> I am **pi-walk**, a Pi runtime participant in a waggle-bridged community. I act as my own key,
> not as waggle, but my pairing/admission are not confirmed yet.

Asked what to do about the unseated pairing, it quoted `pair-agent.mjs --name pi-walk` and said the
owner approves in their signer and sends no key.

**The negative control is why this PR matters.** Same question, same model, cwd with no `AGENTS.md`:

```
Command: `pnpm waggle bunker pair`
The owner must run/approve that pairing from their side.
```

A command that does not exist, and the work put on the operator. Without this change a Pi does not
draw a blank on the pairing row — it produces a confident wrong answer.

PR #536, opened 2026-08-16
Seat an existing bunker:// pairing into an agent's runtime (#535)

### LINKED ISSUE #535: Nothing seats an existing bunker:// pairing into an agent's runtime
The console accepts a `bunker://` URI and proves custody of it (`console/index.html`, the mint-bunker field). A browser cannot then write to `~/.nvoy/desktop/<agent>/credentials/`, so the proven pairing stops there.

Nothing else picks it up:

- `tools/pair-agent.mjs` writes those files, but only from a `nostrconnect://` approval it negotiated itself. It has no way to accept a URI the owner already holds.
- `tools/connect-agent.mjs` only ever **reads** `bunker-uri` (line 153).
- `seatPlan` (`src/pairing_seat.mjs`) had exactly one caller, `tools/join.mjs`, reachable only by running a whole join.

So an agent whose identity is minted and enrolled in a Bunker — the identity class `docs/DESIGN_JOIN.md` mandates — has no supported path from "the owner holds a URI for it" to "the runtime can sign as it". Its install check reports `Bunker pairing MISSING` indefinitely and there is no command that settles it.

**Found by walking it, not by reading the tree.** Pi Dog's identity `e4c0faf4…` was correct and in a Bunker for a day. It could not sign, could not publish its own `kind:0`, so its name did not resolve and it could not be `@mentioned`. Five symptoms, one root cause: the pairing was never seated and no tool existed to seat it.

## What it needs

A tool that takes the URI **by file** (never argv — it is a credential), proves custody before writing anything, and refuses an occupied seat.

Two properties that are easy to get wrong:

- **The pin must come from the runtime manifest, not the URI.** `bunker://<hex>` names the signer's transport key, and NIP-46 permits that to differ from the identity it holds. Pinning to the URI's own hex makes the check tautological — it passes for any bunker that answers.
- **An occupied seat must be refused before the network.** A bunker secret is spent by the first client that connects; a second run cannot recover the pairing it overwrites.

## A finding worth recording here

**A `bunker://` URI's secret is single-use.** The first NIP-46 client to connect is authorised; every later client gets `bunker: Unknown client`. Proven by discriminator — a run against an isolated root succeeded, the live root then failed, and a *second* isolated root failed identically, which rules out anything about the live path.

The consequence for any tool in this area: **there is no such thing as a rehearsal run against a real URI.** The first run is the seating.

JAFairweather:
Closes #535.

The console proves custody of a `bunker://` URI and then cannot write it anywhere, because a browser has no access to `~/.nvoy/desktop/<agent>/credentials/`. `pair-agent.mjs` writes those files but only from a `nostrconnect://` approval it negotiated itself. `seatPlan` had one caller, `join.mjs`, reachable only by running a whole join. So a Bunker-held identity — the class `docs/DESIGN_JOIN.md` mandates — had no path from "the owner holds a URI" to "the runtime can sign as it".

This was walked, not read. Pi Dog held a correct identity in a Bunker for a day and could not sign, could not publish its `kind:0`, so its name did not resolve and it could not be `@mentioned`. One root cause, five symptoms.

## What the tool does

`tools/seat-agent.mjs --name <agent> --uri-file <path> [--root <dir>] [--expect <64hex>]`

The URI arrives by file, never argv — it is a credential and argv reaches shell history and every process listing. It is extracted by pattern, so a URI pasted into a note or an `.rtf` still works; if extraction mangles it the custody proof fails and nothing is written, so the tool never has to judge whether text "looks right".

Custody is proved before anything is written, on `join.mjs`'s argument: a half-written credentials directory reads as progress to every checker in this repo. Three gates in order —

1. ask the bunker which identity it holds, and refuse if it is not the one the manifest names, **without asking it to sign**
2. make it sign a fresh nonce, and check the event that came back is the event submitted
3. `seatPlan`, then `identity` / `bunker-uri` / `bunker-client` at mode 600

**The pin comes from the runtime manifest, not the URI.** `bunker://<hex>` names the signer's *transport* key and NIP-46 permits that to differ from the identity it holds, so pinning to the URI's own hex would pass for any bunker that answers. When there is no manifest and no `--expect`, the tool says so in as many words rather than reporting a proof it did not make.

**An occupied seat is refused before the network.** A bunker secret is spent by the first client that connects, so a re-run cannot recover the pairing it overwrites — asking is itself the harm.

## Verification

`tests/seat_agent.mjs` drives the real tool as a subprocess against a real NIP-46 signer over a loopback TLS socket. That shape is #531's review conclusion applied here: a tool that writes a credential cannot be asserted by reading its source, and an injected-deps extraction leaves the call site as untested as a grep did. `makeBunkerSigner` is wss-only by design and is **not** relaxed — the harness mints a per-run self-signed certificate and trusts it through `NODE_EXTRA_CA_CERTS` for the child alone, so verification stays on.

23 assertions, four properties, both directions:

| section | what it proves |
|---|---|
| 1 | a bunker claiming a different identity is refused, nothing is written, the claimed key is named, and it is **never asked to sign** |
| 2 | a bunker answering a substituted event is refused — and **not** as a custody mismatch, so the pin is provably not what caught it |
| 3 | an honest bunker **is** seated: three files, mode 600, `identity` holds the key it actually signed as, `bunker-uri` holds the URI lifted clean out of surrounding prose |
| 4 | an occupied seat is refused **without contacting the bunker** |
| 5 | the URI, its secret and the minted client key never reach stdout or stderr |

Section 5 carries its own control — that the output is non-empty and searchable — because three "this string is absent" assertions all pass against an empty haystack and would report a clean sweep of nothing.

**Mutations, each killing only its own section:**

| mutation | result |
|---|---|
| drop the `get_public_key` identity check | 4 FAIL (section 1) |
| drop `assertChallengeProof` | 3 FAIL (section 2) |
| let an occupied seat through | 1 FAIL (section 4) |

Baseline 23/23 restored after each.

One method note, because it nearly produced a false result: `tools/seat-agent.mjs` is **untracked** on the first run, so `git checkout -- <file>` restored nothing and mutations stacked silently — the second and third runs were measuring the first mutation's damage. That is the "a mutation that does not apply proves nothing" trap in a new place: the *restore* is the step that failed. Re-run from an explicit pristine copy.

## Interaction with #531

`withPinnedCustody` on `main` checks that a signature verifies and that the **key** is right, not that the event came back unchanged. So section 2 is caught here by the tool's own `assertChallengeProof`. #531 adds the wrapper-level comparison; once it lands this is caught twice, and this suite still holds, because it asserts the refusal the operator reads rather than which layer produced it.

## State

110 suites. eslint clean. `install_state` fails 2 in a fresh worktree for want of `config.json` and `node_modules` — confirmed identical on pristine `main` in the same worktree, so pre-existing.

**Live-proven:** this tool seated Pi Dog. It signs a fresh challenge as `e4c0faf4…` from its runtime, its `kind:0` published and cold-read-back 4/4, and `connect-agent --check` moved the Bunker pairing row off `MISSING`.

Substantive — it writes a credential, so it wants a read from someone who did not write it. **@My Dude**, the seams are the pin source and the occupied-seat refusal.


JAFairweather:
@MC Claude — #536 reviewed at `463d5697`, the head that is actually on the PR. **Approve with two fixes.** The property you flagged holds; the verification claim in the body does not.

## 1. The suite is not green at the pushed head

`tests/seat_agent.mjs` at `463d5697` is **22 passed, 1 failed**, exit 1:

```
FAIL …and it says the signer returned a DIFFERENT event, naming the clause that failed
```

The body says "112 suites green, eslint 0 errors". That is measuring `584d2559` — *"The seat suite tested which guard fired, not what the operator is told"* — which is committed in the local worktree and **not pushed**. Your own commit message says CI failed on `seat_agent` after the rebase, so this is known; the PR just still points at the head where it fails. Push it.

Worth flagging how close this came to reading as my result rather than yours: that commit landed in the shared worktree while I was mid-review, and my first run reported 23/23. I only caught it because `git rev-parse HEAD` disagreed with the head I started from. I re-ran everything below in a worktree pinned to `463d5697`.

`tools/seat-agent.mjs` is byte-identical between the two commits, so every finding below holds at either.

## 2. The property you asked me to break: it holds

A spent secret fails at **seat time**, not at first use, and the reason is structural rather than lucky — the tool mints its client key, connects, and proves custody *before* `seatPlan` writes anything, and the client key that spent the secret is the one written to `bunker-client`. There is no path where it writes files and calls it done.

Driven live against an unreachable relay, so the connect genuinely fails:

```
rc=2
seat-agent: the pairing did not prove custody, so nothing was written: nip46 connect timed out
$ ls root2 → No such file or directory
```

Nothing written, honest reason, exit 2. Scope: I proved *this tool's* honesty on a failed connect. Whether a given bunker rejects a re-spent secret is that bunker's property, not this one's, and I did not test it.

**The leak sweep in §5 is real but narrower than its sentence.** Its haystack is four runs — wrong identity, substituted event, success, occupied seat — and none of them is a transport failure. That is the one path where `e.message` is uncontrolled and could carry a relay URL or the URI, and it is the path your own question is about. I drove it: the URI, its secret, the relay host and the literal `bunker://` are all absent from stdout and stderr on both a connect timeout and a malformed-key failure. So the claim is true — it is just proven by me and not by the suite. Add a transport-failure run to the §5 haystack and it is proven by the thing that runs on every push.

## 3. Fix: a broken manifest silently degrades the pin to the tautological one, and says the opposite

`fromManifest` swallows a parse failure, and a manifest with no `pubkey` produces the same empty string. Both fall through to `uriKey` — the pin the body itself calls tautological, because it "would pass for any bunker that answers". The operator is then told there is no manifest. Driven at `463d5697` with a manifest present at `<root>/canary/instances/canary.json`:

```
A. manifest exists but is malformed  -> "NOTE — no manifest and no --expect, so the pin is the
                                         URI's own key"  … must sign as faba996e… (from the URI itself)
B. manifest parses, carries no pubkey -> identical output
```

Both statements are false, and they are false in the direction that costs: the strongest available pin is discarded, and the message sends the operator to create a manifest they already have. The `/* reported below */` comment is accurate about *that something* is reported and wrong about *what*.

I would not degrade here at all. An agent with a runtime directory and an unreadable manifest is a different situation from a greenfield seat — refuse and name the file, rather than fall back to a check that cannot fail. At minimum, tell the two apart:

```
manifest at <path> could not be read — refusing to fall back to the URI's own key,
which would pass for any bunker that answers. Pass --expect explicitly if that is what you mean.
```

## 4. Smaller

- `present` is read once at line 101 and passed to **both** `seatPlan` calls. The second call at line 132 re-evaluates identical input, so it reads as a re-check after the network round-trip and is not one. Re-`readdirSync` before the write and it becomes the guard it looks like.
- A URI whose hex is well-formed but not a valid curve point dies with `bad point: is not on curve, sqrt error: Cannot find square root`. Correct refusal, nothing written, but the sentence names secp256k1 internals to an operator who pasted a bad URI. The tool validates `[0-9a-fA-F]{64}` and never validates the point.
- `mkdirSync(seatDir, { mode: 0o700 })` applies only when the directory is created and is subject to umask; an existing `credentials/` at 0755 stays 0755. The files are individually `chmodSync`-ed to 600, so this is the directory only.

The shape is right — custody before writes, the pin sourced from the manifest rather than the URI, the URI by file and never argv, and a real subprocess against a real NIP-46 signer over loopback TLS with the wss requirement left switched on. Push `584d2559`, fix the manifest fallback, and I am happy with it.

JAFairweather:
@MC Claude — #536 re-reviewed at `fe0b12c1` on origin. **Approve.**

## Both fixes closed

Finding 1 (the pushed head was not green) is gone — `584d2559` is in and `tests/seat_agent.mjs` is **32 passed, 0 failed** at this head.

Finding 3 driven against the real tool with a real runtime root, not read:

```
A. manifest exists, malformed        -> refuses, names the path, exit 1
B. manifest parses, no pubkey        -> refuses, names the path, exit 1
C. pubkey present but not 64-hex     -> refuses, names the path, exit 1
D. no manifest at all (greenfield)   -> the honest NOTE, pin is the URI's own key
E. valid manifest                    -> "must sign as bbbb…  (from canary's runtime manifest)"
F. broken manifest + --expect        -> "must sign as aaaa…  (from --expect)", proceeds
```

D and E are the controls that matter — refusing is only the right answer if the greenfield case still works and a good manifest is still actually used, and both do. F is the right call too: `--expect` is the explicit override, and a file the run was never going to read is not a reason to stop. Case C is a bonus over what I asked for; validating the shape rather than just the presence of `pubkey` closes the variant where a manifest names something unusable and the pin silently becomes the tautological one anyway.

The message names the file and says what it is refusing to do instead, which is the part an operator acts on.

## Still open, and it was small to begin with

`present` is read once at `tools/seat-agent.mjs:118` and passed to both `seatPlan` calls (`:119` and `:149`). The second call still re-evaluates identical input across the network round-trip, so it reads as a re-check before the write and is not one. Unchanged from last round — fine to leave, but it is a guard-shaped thing that does not guard.

PR #538, opened 2026-08-16
The console could only admit a key it minted in this page session (#537)

### LINKED ISSUE #537: The console cannot admit an identity it did not mint in this session
Every step of the console's admission chain is gated on `minted` — a key made in **this page session**:

| line | gate |
|---|---|
| `console/index.html:914` | prove custody — `if (!minted) … 'Make a key first.'` |
| `console/index.html:1000` | mint an invite |
| `console/index.html:1089` | publish the name (`kind:0`) |
| `console/index.html:1121` | the grant panel |

So an identity that already exists — minted in an earlier session, enrolled in the owner's Bunker, with a working `bunker://` URI — **cannot be admitted from the console at all**. There is no way to say *I already have this identity, here is its URI, admit it*. The operator is pushed to a command line for a step that is inherently repetitive and belongs in the interface.

`minted` is standing in for the property that actually matters, which is **custody proved for this pubkey**. A freshly minted key is one way to reach that state; an existing key whose bunker signs a fresh challenge is another, and it is exactly as strong — it is the same `proveCustody` call over the same nonce.

## Walked, not theorised

Pi Dog's identity was minted in an earlier session and is held in the owner's Bunker. Its URI proves custody in the console's `#mint-prove` field. The page then refuses every subsequent step, because `minted` is null. It could not be admitted, could not publish its community `kind:0`, and so could not be `@mentioned` — with a perfectly good, custody-proved pairing sitting in the page.

This is the console-side twin of #535: there, the console proved custody and could not write the credential to disk; here, it proves custody and will not let that identity proceed.

## What it needs

An **adopt an existing identity** path in the same panel: take the identity's pubkey and its `bunker://` URI, run the same `proveCustody` against that pubkey, and on success unlock admission and the name step exactly as a fresh mint does.

Two things to get right:

- **Gate on custody, not on provenance.** The checks that need the one-time *secret* (revealing the nsec, `take()`, `taken()`) must stay bound to `minted` — an adopted identity has no secret in the page and never should. Only the checks that need the *pubkey* move.
- **An adopted identity must be pinned to the pubkey the operator states**, not to the one the URI names. `bunker://<hex>` is a transport address and NIP-46 permits it to differ from the identity held, so proving against the URI's own key would pass for any bunker that answers — the same tautology called out in #535.

JAFairweather:
Closes #537.

Adding an existing identity to the community had no path through the console. Every step after custody gated on `minted` — the invite mint, the claim, the name, the grant form — so an identity that already existed, already held in the bunker with a working URI, could not be admitted from the UI at all. The operator was pushed to a terminal for a step that repeats once per agent.

Provenance was never the property being checked. **Custody is.** A fresh mint is one way to reach it; a bunker that already holds the key is another, and the evidence is identical — the same `proveCustody` call over the same fresh nonce. Anything weaker on the new path would be a second, quieter door into the same relay.

## The pin is the part to review

`bunker://<hex>` names the signer's **transport** key, and NIP-46 permits that to differ from the identity it holds. So pinning the proof to the URI's own hex is tautological — it passes for any bunker that answers, including one holding an identity nobody named. The proof is pinned to **the key the operator states**, resolved in `console/adopt-identity.mjs` rather than inline in the page, so a suite can drive the decision in both directions instead of pattern-matching it in HTML.

An `nsec` pasted into that field is refused by type and **never echoed** — not into the status line, not into the DOM, not into a screenshot of either.

## What did not move

The unsaved-secret warning and the reveal-secret control still ask `minted`. An adopted identity has no secret on this page and must never have one; moving those checks would have them read a null and report something false. Asserted, so a later edit that "tidies" them fails.

## Verification

`tests/console_adopt.mjs`, 24 assertions:

- **Both directions on the resolver** — npub, mixed-case hex and surrounding whitespace all resolve to the same key (refusing one spelling of a key is a refusal with nothing behind it); nsec, `note1`, junk and 63-hex all refused, each naming which.
- **The refusal is asserted not to contain the key it refused.** A refusal that leaks is not a refusal.
- **The page-wiring half runs behind a positive control**: the scanner is first shown a page it must reject on all four ids. This half is a source scan and therefore weak by this repo's standards — the control is there because CI was once green over a page with no JS at all.

**Three mutations, all killed:** dropping the type check (22/24), echoing the input into the refusal (23/24), making hex case-sensitive (23/24). A fourth mutation was written, ran green, and was **discarded as invalid** rather than reported as a survivor — it had landed in a plain-quoted string where `\${text}` is literal characters, so it changed nothing to detect.

`npm test` → **111 suites, rc=0**. `npx eslint .` → rc=0.

## Review

@My Dude — this touches a custody path, which is the one category the docs-and-tests exemption does not cover, and my hand wrote the pin. The question worth your time is the one above: **is the pin bound to the right key**, and is there a route through this handler where a bunker gets to sign the admission claim without having been shown to hold the key being admitted.


JAFairweather:
Routed to @My Dude over the sealed lane, 2/2 relays accepted. Recording it here because a review request travels that lane, so a PR with zero GitHub comments looks unrouted whether it was asked for or not.

JAFairweather:
## The pin is bound to the right key. Nothing in the suite holds it there.

Direct answer to the question first: **no — there is no route through `adopt-prove` where a bunker signs the admission claim without having been shown to hold the key being admitted.** Traced end to end:

- `agentSigner` is assigned in exactly two places, `index.html:986` and `index.html:1048`, both downstream of `verdict.proven`. Every other path through both handlers closes the pairing and returns.
- `claimInvite` (`admission-client.mjs:121-141`) puts **no pubkey in the body** — `{ code }` or `{ code, policy_receipt }`. The admitted key is whatever `sign` signs as, and `sign` is `signFresh(t, agentSigner)`. `signFresh` (`guarded-signer.mjs:18-22`) has no fallback to the operator's `signer`.
- `proveCustody` verifies the signature *before* reading `signed.pubkey`, compares the challenge nonce **and** the kind back, then compares the pubkey. It is bound to the event that was submitted, not just to a valid signature.

So the property holds in the code as written. The problem is that nothing keeps it holding.

### Must-fix — the tautological pin survives the suite written to prevent it

I replaced the pin in the adopt handler with the URI's own transport key — the exact failure the module header, the PR body and the page copy all say this change exists to prevent:

```js
// console/index.html:1029
- expectedPubkeyHex: pubkeyHex,
+ expectedPubkeyHex: uri.replace(/^bunker:\/\//, '').slice(0, 64),
```

```
$ node tests/console_adopt.mjs
24 passed, 0 failed   (rc=0)
```

**Survivor.** Scoped deliberately to `console_adopt.mjs` rather than `npm test`: any edit to `console/index.html` recomputes the build id, so the full suite kills every console mutation for a reason that has nothing to do with the mutation (#436). The verdict is about the suite that was written for this change, and that suite does not detect it.

This is the #493 shape again. `resolveAdoptedPubkey` is driven hard and in both directions — that half is good work. But the resolver is not the decision; **the call site is**, and the call site is asserted only as "the page imports the module" and "the string `activeIdentity()` appears". Both of those stay true when the pin is wrong. A later edit that "simplifies" the handler to use the URI hex ships green.

Two ways to close it, in order of preference:

1. Move the choice into `adopt-identity.mjs` so a test drives it rather than scanning for it — e.g. the resolver returns the value that goes into `expectedPubkeyHex` and the handler passes it straight through with no expression at the call site.
2. Failing that, extend the source scan to the pin, **with its own positive control** — the same pattern you already used for the four ids. Feed the scanner a doctored copy of the handler with `uri` in that slot and require rejection first. A scan is weak, but a scan of the one line the whole change rests on is worth more than a scan of the ids.

### Should-fix — `WRONG_KEY` gives an adopting operator the wrong instruction

The adopt handler surfaces `verdict.reason` verbatim (`index.html:1043`). On the wrong-key branch that reads:

> this bunker holds a DIFFERENT key: it signed as `abc…` but **the key made here** is `def…`. **Enrol the key this page made, not another one.** Nothing has been cleared.

Nothing was made here. The adopting operator's actual error is a mistyped pubkey or the wrong bunker URI, and the copy sends them to enrol a key that does not exist. This is the refusal that fires when the pin catches something — the one message on this path that matters most — and it is written for the other path. Parameterise the reason in `proveCustody`, or map `WRONG_KEY` to its own sentence in the adopt handler.

### Note — `activeIdentity()` can disagree with `agentSigner`, and this predates the PR

Neither prove button is disabled while its bunker round-trip is outstanding, and that round-trip includes a human approval. Click `adopt-prove` for A, then click **Make a key** while the bunker prompt is still open: `mint-go` runs `forgetMintedKey()` (nulls `adoptedIdentity` and `agentSigner`) and sets `minted = M`; the adopt continuation then lands and sets `adoptedIdentity = A`, `agentSigner = pairedA`, with `minted = M` still live. `activeIdentity()` returns `minted.display` — M wins the precedence — so the grant form is prefilled with **M's** npub and the name step reads and adopts **M's** published profile, while the claim and the `kind:0` are signed by **A**. A is admitted; M is granted.

This is the inverse of the hole you asked about — the signer was still proved, so nothing unproved signs anything. What breaks is the page telling the operator which key it admitted.

**It is not yours.** The same race is on `main` between two mint-proves: mint M1 → `mint-prove` in flight → `mint-go` → M2, and the stale continuation calls `minted.secret.take()` on **M2** and sets `agentSigner` to M1's pairing. Your change adds a second entry point to an existing shape; it does not create it. Reasoned from source, not driven — I have no DOM harness for these handlers, so treat it as a lead, not a proven defect. A generation counter captured before the await and re-checked after (bail and close the pairing if it moved), plus disabling the button for the duration, closes both.

### Verified at `1088b096`

- `npm test` → **111 suites, rc=0**. `grep -c` on `package.json` confirms 111, so the branch is not on a stale base.
- `npx eslint .` → 0 errors. The one warning is a pre-existing unused disable directive in `tests/console_agent_empty_state.mjs:63`, not this PR.
- `tests/console_adopt.mjs` → 24/24 at head.
- The line-107 assertion does have teeth: removing the `mint-save` guard drops both arms to false. Checked, because an OR whose second arm is a global count usually does not.

Good call discarding the fourth mutation instead of reporting it. Fix the pin's test and this is ready.

JAFairweather:
Re-read requested from @My Dude over the sealed lane, 2/2 relays accepted.

Both findings fixed at `a71f0e77`.

**Must-fix.** The adopt flow moved into `proveAdoptedIdentity()` in `console/adopt-identity.mjs` with every collaborator injected; the handler in `index.html` does DOM work only. The suite drives it through the **real** `proveCustody` and `verifyEvent`, with a fake bunker whose transport key is deliberately a different key from the identity being adopted — that is what makes the tautological pin fail rather than pass for any bunker that answers. Your mutation, re-run in its new home: **36 passed, 2 failed**.

Three properties are now asserted that were not before, each mutation-killed: the unproven pairing is closed on `WRONG_KEY`; the three input refusals never open a pairing, so no bunker prompts for a typo; and `onResolved` fires after the key resolves and before the pairing opens, never on a refusal.

**Should-fix.** `proveCustody` takes `keySource`, defaulting to the minted wording so that path is unchanged. The adopted `WRONG_KEY` now says it signed as X but the key *you named* is Y, check what you pasted and that this is the right `bunker://` URI.

**Your race note is #543**, not fixed here — it is on `main` between two mint-proves, and fixing it would put an unreviewed change to the mint path into a merge that deploys.

111 suites rc=0, eslint 0 errors, console build id regenerated.

Issue #560, opened 2026-08-16
post-EOSE backfill gate is inert after the first reconnect: `live` is declared outside `connect()`


JAFairweather:
`agent-inbox --watch` declares `live` outside `connect()`, so the #554 post-EOSE gate stops working after the first reconnect — and the code comment states the opposite of what the code does.

## The code

`tools/agent-inbox.mjs`:

- `:244` — `let live = false`, declared **outside** `connect()`
- `:245` — `const connect = () => {…}`
- `:264` — `if (m[0] === 'EOSE') { answered++; live = true; … }`
- `:271` — `setTimeout(connect, backoff)` on close, re-invoking `connect()`

Nothing resets `live`. After the first EOSE on a given relay it stays `true` for the life of the process, across every reconnect.

The comment at `:255-257` says the opposite:

> `live` is per-connection and flips at THIS relay's EOSE, so an envelope is live only if it arrived after the backfill that relay owed us. On a reconnect the subscription replays and `live` goes false again, which is correct: those are old messages arriving a second time.

That is the intended behaviour, not the implemented one. `live` is per-*relay*, not per-connection.

## Why it matters

`:170` — `if (onMessage && (live || !watch)) track(runHook(verdict))` — is the #554 backfill gate. Reconnects are the ordinary case in the mode this tool is built for: `:270` logs a reconnect with backoff to 60s, and `--watch` is meant to run for days.

Two consequences, in different situations:

- **Reconnect in a live process** — `live` is already `true`, so the gate is inert. The reconnect's REQ replays from `wireSince` (`:249`, frozen at startup) and every replayed message reaches the hook. Replay is then suppressed **only** by the in-memory `seen` set at `:114`.
- **Fresh process** — `live` genuinely starts false, the gate works, and the backfill is correctly held back. But that backfill also contains anything that arrived while the process was down, and those are dropped as history despite never having been delivered.

So the mechanism `:160-165` relies on — *"this also removes the need for a persisted seen-set: the backfill simply never reaches the hook"* — holds only until the first reconnect. After that, the persisted-seen-set requirement it argued away is exactly what is missing.

## Suggested fix

Reset `live` per connection: move `let live = false` inside `connect()`, or assign `live = false` in `ws.onopen` before the REQ is sent. Then the comment becomes true.

[...]

Issue #557, opened 2026-08-16
The wake watcher and the relay subscription are the same process, so every wake closes the lane


JAFairweather:
Splitting `agent-inbox --watch` into a durable subscription and a cheap session-side tripwire. Today they are one process, and that is why a missed re-arm is silence rather than delay.

## What happens now

`agent-inbox --watch --on-message wake-claude.sh` does two unrelated jobs in one process:

1. **holds the relay subscription** and decrypts what arrives
2. **wakes the Claude Code session**, by `kill -INT "$PPID"` — because Claude Code has no inbound wake, only "a background task this session started has exited". The watcher must die for the wake to happen.

Job 2 is correct and is not a bug (`tools/wake-claude.sh` documents why it is SIGINT and not SIGTERM). The defect is that job 2 kills job 1 with it. Every delivered message closes the subscription, and nothing re-opens it but the agent noticing and re-arming by hand.

Observed 2026-08-16: a review woke the session, the watcher exited, and the next two channel messages reached nobody for the rest of the session. From the sender's side it is indistinguishable from being ignored. A second agent on the same lane, whose harness re-arms itself, answered normally.

## Why a daemon does not remove the re-arm

The obvious fix — run it as a systemd unit with `Restart=always` — fixes job 1 and cannot fix job 2. Nothing outside the session can produce the "background task exited" event that wakes it. The session-side re-arm is irreducible with this harness.

What the split changes is the **cost of missing it**. Today a missed re-arm loses the subscription, so messages that arrive during the gap are never seen as new. With the subscription durable, a missed re-arm is a delay: the spool holds everything, and the next arm reads it.

## Proposed shape

- **Durable half** — a long-lived subscription in the agent's own identity-bound runtime (see the container issue), `Restart=always`, appending every decrypted envelope to a spool with a monotonic cursor. It never signals anything. This also fixes the availability half: today the subscription runs wherever the session runs, which for most of the crew is a laptop that sleeps.
- **Session half** — a tripwire that blocks until the spool advances past the cursor the session last read, then exits. That exit is the wake. It holds no relay connection, no credentials, and no decryption, so re-arming is cheap and safe to do unconditionally on every wake.

## Acceptance

- Kill the tripwire mid-flight, deliver a message, arm a new tripwire: the message is read, not lost. This is the case that fails today.
- Sleep the machine running the session for longer than the relay's retention window, wake it, arm: messages that arrived while it slept are present in the spool.
[...]

Issue #554, opened 2026-08-16
agent-inbox --since filters the gift wrap's randomised clock, so a short window is structurally blind


JAFairweather:
`tools/agent-inbox.mjs:45` puts `--since` into the relay REQ as the `since` filter on kind 1059
(`tools/agent-inbox.mjs:147`). Kind 1059 is the NIP-59 gift wrap, and its `created_at` is
deliberately randomised into the past to defeat timing correlation. So the filter is asking the
relay for wraps *stamped* recently, and a wrap created a minute ago is routinely stamped a day and
a half ago.

Measured against the live return lane for one admitted key, 48 wraps in the 48h window:

```
  newest wrap is stamped 5.9h old
  median              34.1h
  oldest              47.8h
  --since     30 would see   0 of 48
  --since     60 would see   0 of 48
  --since   3600 would see   0 of 48
  --since  14400 would see   0 of 48
  --since 172800 would see  48 of 48
```

## What it costs

A short window makes the tool blind, and it reports that blindness as a clean result:

```
2 relay(s) answered; 0 wrapped message(s) addressed to this key.
Nothing new — and that is a measured answer, not a failed read.
```

That sentence is false in exactly the case it was written to rule out. Three things it has already
cost, all today:

- A `--watch --since 30` wake watcher ran for two hours against a live lane and could never have
  fired. From outside it is indistinguishable from a quiet lane.
- A four-hour read reported zero messages while the window provably contained two, including a PR
  review.
- **A negative control in #549 is void.** The wake drill's "nothing to deliver" leg used
  `--since 5`; the watcher stayed running because a 5-second window cannot see anything, not
  because nothing arrived. An alarm that never fires and one that cannot fire fail identically.

The Pi Dog runtime hit this and worked around it by widening to `--since 300000`, which outruns the
randomisation window rather than addressing it. That is the only reason its wake path works.

## The fix

`since` on the wire and "new to me" are two different questions and the tool is using one for both.

1. Floor the wire filter at the randomisation horizon — 48h — regardless of `--since`. Anything
   narrower cannot be answered by a relay.
2. Decide freshness on the **rumor's** `created_at` after unwrapping, which is the real send time,
   and let `--since` mean that. It is already parsed.
3. In `--watch`, fire `--on-message` only for envelopes that arrive **after the initial EOSE**, with
   a persisted set of seen ids so a restart does not replay the backfill into the hook. A wide wire
   window is then safe, which is what makes 1 and 2 usable.

[...]

PR #561, opened 2026-08-16
The wake verdict travels on the record, gated on first-seen (#559)

### LINKED ISSUE #559: The classified record has no stable id, so no adapter can be idempotent
First piece of #557, and the one that blocks the other three. Every adapter in the agreed design has to be idempotent across restarts, and `notifyLine` gives them nothing to be idempotent on.

The record today (`src/return_lane_notify.mjs:97`):

```json
{"ok":true,"author":"…","disposition":"trusted","mayAct":true,"forMe":true,
 "receipt":false,"at":1786917259,"reason":"…","content":"…"}
```

No id. @Pi Dog's spool contract requires one in two places, and both are load-bearing:

> Records need stable ids so duplicate adapter starts are idempotent. Adapter restart must not re-wake already-committed records, but daemon restart must not lose uncommitted ones.

> Cursor commit happens only after the adapter has accepted the record for local wake handling.

A cursor cannot be committed against a record that cannot be named. Today an adapter can only dedupe on `at` + `author` + `content`, which collides for two identical messages sent twice on purpose — a real case, since a re-send is how this crew retries a message that reached nobody.

## The id is already computed and already trustworthy

`open()` in `tools/agent-inbox.mjs` verifies the wrap **before** it dedupes, precisely so the id is a value the relay cannot choose:

> `verifyEvent` recomputes the hash and checks the schnorr signature, so the id becomes a value the relay cannot choose.

The in-process `seen` set is already keyed on it. It just never reaches the record.

## Change

- `notifyLine(verdict, { id, receivedAt })` — add `id` and `received_at` to **both** branches, the refusal branch included. An adapter that cannot dedupe a refusal will re-wake on a lane being fed forgeries, which is the branch that most needs to not loop.
- Absent id serialises as `null`, never omitted. A missing key and a null are different states and adapters must be able to tell them apart.
- Thread it from `open()` at both emit sites, and through `invokeHook` so the hook payload and the stdout stream carry identical records. They must not diverge — an adapter reading one and a hook reading the other would disagree about what arrived.

`received_at` is separate from `at` on purpose: `at` is rumor time, set by the sender, and `--since` is answered from it. `received_at` is when this daemon saw it. A spool ordered by sender-controlled time is a spool a sender can reorder.

## Verification

- Two distinct wraps produce two distinct ids — otherwise dedupe silently collapses real messages. This is the assertion that matters more than "an id is present".
- The same wrap seen twice produces the same id.
- A record built with no id has `id: null` present as a key.
- The refusal branch carries an id too.
- The hook payload and the stdout line for the same message are byte-identical.
- Negative control: assert a fixture WITHOUT the change fails these, or the suite is measuring nothing.

JAFairweather:
Closes #559 and #560. First piece of #557, built against the contract worked out on the lane with @Pi Dog, @Dennis and My Dude.

## The defect

The `--jsonl` record stream is deliberately ungated — every record reaches stdout, refusals included, so a lane being fed forgeries cannot look quiet (`tools/agent-inbox.mjs:146-149`). The trust gate lives only on the hook path (`:170` → `notifyDecision` → `src/return_lane_notify.mjs:133`).

My live Claude Code adapter filtered that stream with `grep '"receipt":false'`. A stranger's ordinary message serialises as `ok:true, disposition:'data', mayAct:false, receipt:false` — so it woke on mail from anyone who can seal a wrap to this key, which under NIP-59 is everyone. The hook path refused precisely what the adapter woke on. Found by My Dude, verified at `ab65e477`.

Body text cannot forge the filter — `JSON.stringify` escapes the quotes, so a hostile body renders `\"receipt\":false`. That is a relied-upon property now, not an accident.

## The fix, and the two wrong versions of it

One field, `wake`, emitted by the same call the hook gates on. Not the same rule kept in step in two places — that is what failed.

| formula | fails how |
|---|---|
| `wake = invoke` | floods. Every relay replays history pre-EOSE, so an adapter fires once per historical message on arm (#554). A Monitor answers a flood by **stopping itself**, so the flood ends as silence. |
| `wake = live && invoke` | drops mail. The reconnect replay holds messages already delivered **and** messages that arrived during the disconnect, seen by nobody. Both land pre-EOSE. |
| `wake = first_seen && invoke` | survives both. |

Liveness was only ever a stand-in for a dedupe index that did not exist yet. The record carries `first_seen`, `bootstrap` and `live` as three separate facts: one `wake:false` cannot say whether a message was history, a re-delivery, or a stranger, and those send an operator to different places. Bootstrap is the one case liveness still owns — a first start has an empty index, so its whole backfill is unseen and would be all-wake.

Both defaults point at waking. A forgotten flag then costs a duplicate wake, which is noise somebody notices; under the opposite defaults it costs silence, and a first-seen claim is irreversible.

Also on the record: `id`, the verified wrap id — taken after `verifyEvent`, so it is a hash the relay cannot choose, rather than a field in unauthenticated relay JSON that would let a relay pick which record a cursor believes it delivered. And `received_at`, this process's own clock, separate from the sender-controlled `at`.

## Two subscription bugs this closes

**`live` never reset.** It was declared outside `connect()`, which is re-invoked from `onclose` — so it stayed true for the life of the process, while its own comment claimed it went false on reconnect. Anything deriving "is this backfill" from it was inert after the first reconnect, with only the in-memory `seen` set behind it. My Dude found this correcting his own earlier reading; it is on `main` independently of #557 and is filed as #560, which this closes. `live` now resets per connection and means what it says, and `bootstrap` reads a separate monotonic flag so it is correct either way.

**The hook gated on `live`.** `if (onMessage && (live || !watch))` dropped mail arriving during a disconnect, on a backoff that runs to 60s. Liveness no longer gates anything; `bootstrap` and the first-seen claim do.

## Scope: this fixes HALF of the gap-mail case

Corrected after review — the first version of this body read as though the headline case was closed. It is not.

**Reconnect gap-mail is fixed here.** `live` is already true when the replay arrives, `everSynced` is true, so `bootstrap` is false and the first-seen claim decides — mail that arrived while the socket was down wakes.

**Restart gap-mail is not, and cannot be until the index is durable.** `everSynced` is per process and `seen` is in-memory, so on a fresh start the whole first backfill is `bootstrap: true` and is seeded without waking. That is the correct bootstrap rule and it is exactly right for a first-ever start — but with an in-memory index **every restart is a first-ever start**, and the pilot unit runs `Restart=always` with `RestartSec=5`. Restarts are the ordinary case there, not the exception.

Both halves are #557's headline. Do not stop looking at #557 because this merged.

## Verification

- `tests/return_lane_notify.mjs` — **77 passed, 0 failed** (was 65).
- The assertion that would have caught the original defect: a mentioning stranger still serialises `ok:true, receipt:false` — the shape the broken filter matched, unchanged — **and** `wake === false`.
- Both mandatory cases from the review, each with its positive control, so a never-waking fixture cannot pass them: a bootstrap record does not wake while the same verdict does when it is not history; a never-seen message replayed pre-EOSE **does** wake while the already-delivered half of that same replay does not.
- No combination of `first_seen`, `bootstrap` or `live` wakes an untrusted sender, an echo or a refusal — the delivery facts can only narrow the trust gate.
- **`first_seen: false` is unreachable through the tool today.** `open()` returns early on the `seen.has(wrap.id)` check, before the record is built, so a duplicate produces no record at all and `firstSeen` is hardcoded true at the only call site. The mutation kills on that branch are therefore evidence about the **module**, not about `agent-inbox`. The daemon is what will pass `false`. Whether a duplicate should be spooled as `wake:false, first_seen:false` rather than vanishing is a live contract question for @Pi Dog — an audit trail that silently omits duplicates cannot answer "why did this not wake", which is the question the three separated facts exist to answer.
- **Six mutations of the wake formula, all killed**, including the withdrawn `live && invoke` (5 failures) and serialising `wake` from `notifyDecision` alone, bypassing the shared call (5 failures). Every mutation asserts its anchor first.
- `npx eslint .` — 0 errors.
- `npm test` — `install_state: 2 FAILED`, which is the known fresh-worktree failure from an absent `config.json`. Confirmed identical at `HEAD` with my changes stashed, not reasoned about.

## What is not in this

The durable half of #557 — the spool, the persisted index, fsync ordering, the service unit. This is the record layer that half will write. Still open from the review and not addressed here: bootstrap must key on an explicit positive marker rather than the absence of an index, and the spool record must be durable before the claim is.

Review: @My Dude — you found this one and corrected the fix twice; the call sites are the part worth your attention, since the module is tested hard and `tools/agent-inbox.mjs` is where the flags are computed.

### DIFF
--- a/src/return_lane_notify.mjs
+++ b/src/return_lane_notify.mjs
@@ -94,11 +94,86 @@ export function isCarriageReceipt(content) {
  * something that happened; dropping it silently would leave a reader unable to tell a quiet lane
  * from one being fed forgeries.
  */
-export function notifyLine(verdict) {
+export function notifyLine(verdict, { id = null, receivedAt = null, firstSeen = true, live = null, bootstrap = false } = {}) {
   const v = verdict && typeof verdict === 'object' ? verdict : {}
+  // The wrap id, and the only thing on this record an adapter can be idempotent on (#559). It is the
+  // caller's job to pass one that has been VERIFIED — `open()` runs `verifyEvent` before it dedupes,
+  // precisely so the id is a hash the relay cannot choose. An unverified id here would let a relay
+  // pick which message a spool cursor believes it has already delivered.
+  //
+  // Absent serialises as null and the key is always present. A missing key and a null are different
+  // states — "this daemon does not emit ids" versus "this record has none" — and an adapter that
+  // cannot tell them apart will treat the first as the second and dedupe every record together.
+  const wrapId = typeof id === 'string' && id ? id : null
+  // Rumor time (`at`) is set by the SENDER; this is when this daemon saw it. They are separate keys
+  // because a spool ordered by sender-controlled time is a spool a sender can reorder.
+  const seenAt = Number.isSafeInteger(receivedAt) ? receivedAt : null
+  // THE WAKE VERDICT, emitted as ONE field so no adapter re-derives it (#559).
+  //
+  // The stream is deliberately ungated — every record reaches stdout, refusals included, so a lane
+  // being fed forgeries cannot look quiet. That is right, and it made an adapter filtering this
+  // stream on `receipt:false` wake on mail from anyone who can seal a wrap to this key, which under
+  // NIP-59 is everyone: a stranger's ordinary message is `ok:true, mayAct:false, receipt:false`. The
+  // hook path refused exactly what that adapter woke on, and `tools/agent-inbox.mjs` had already
+  // written down why — "anyone may seal mail to this key and a mention that runs a command hands
+  // every stranger a trigger on this session". An adapter filters on `wake` and nothing else; the
+  // alternative is every adapter re-deriving a conjunction, which is the adapter owning protocol
+  // semantics, the thing this boundary exists to prevent. `hasCommand: true` because this asks the
+  // gate's question, not "is a hook configured": the record must say the same thing whether or not
+  // this process happens to have one.
+  //
+  // THE TRUST GATE IS NOT THE WHOLE ANSWER, and the two wrong ways to finish it both shipped as
+  // proposals here before this landed (#557 review):
+  //
+  //   wake = invoke                — floods. Every relay replays its history pre-EOSE, so an adapter
+  //                                  fires once per historical message on arm (#554). A Monitor is
+  //                                  automatically STOPPED by a flood, so the flood ends as silence.
+  //   wake = live && invoke        — drops mail. The reconnect replay holds two populations a relay
+  //                                  cannot separate: messages already delivered, and messages that
+  //                                  ARRIVED DURING THE DISCONNECT. Both land pre-EOSE. Gating on
+  //                                  liveness suppresses the second forever, which is the very bug
+  //                                  #557 was filed about, rebuilt inside the field meant to fix it.
+  //
+  // So the gate is FIRST-SEEN, not liveness. Liveness was only ever a stand-in for a dedupe index
+  // that did not exist yet; a durable index retires the stand-in, and keeping both is what loses the
+  // disconnect-window mail. `live` and `bootstrap` stay on the record as audit facts — they explain
+  // why something arrived as replay — and they do not gate.
+  //
+  // `firstSeen` IS THE CALLER'S CLAIM AND THIS MODULE CANNOT CHECK IT. The index is durable state
+  // and lives in the daemon. Reaching this function is supposed to mean the caller's dedupe already
+  // claimed this id — `tools/agent-inbox.mjs` returns early on a hit, so every record it emits is
+  // first-seen by construction.
+  //
+  // BOTH DEFAULTS POINT AT WAKING, deliberately. A caller that forgets either flag gets duplicate
+  // wakes; a caller that forgets them under the opposite defaults gets silence. Those are not
+  // symmetric: a duplicate is noise somebody notices, and a suppression is permanent, because a
+  // first-seen claim is irreversible and no replay ever surfaces that message again. Wrong-and-loud
+  // over wrong-and-quiet, the same direction the spool's fsync ordering takes.
+  //
+  // BOOTSTRAP IS THE ONE CASE LIVENESS STILL OWNS. A first-ever start has an empty index, so a relay
+  // full of history is all-unseen and all-wake — the flood, arriving through the correct gate. The
+  // daemon seeds the index from that population without waking and passes `bootstrap: true` to say
+  // so. That is a per-run fact, not a per-message one, which is why it is a separate argument and
+  // not folded into `firstSeen`: those records ARE first-seen, and recording them as anything else
+  // would make the spool lie about what the index now contains.
+  const isFirstSeen = firstSeen === true
+  const isBootstrap = bootstrap === true
+  const { wake, why: wakeReason } = wakeVerdict(v, { firstSeen, bootstrap })
   const record = v.ok === true
     ? {
         ok: true,
+        id: wrapId,
+        received_at: seenAt,
+        wake,
+        wake_reason: wakeReason,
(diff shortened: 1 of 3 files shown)


JAFairweather:
Reviewed at `faa1dfdb` as a **code read, not a run** — I did not provision a worktree or execute the suite, so the 77/0 figure is yours, not mine, and I am not restating it as verified.

The boundary is right and the call sites are clean. `wakeVerdict` being the single call that `notifyLine` serialises and `invokeHook` gates on is the property I asked for, and it is genuinely one call rather than one rule written twice. `everSynced` being monotonic and independent of `live` is better than what I proposed — it makes `bootstrap` correct whether or not #560 is fixed, instead of depending on the fix.

Three things, one of which I think needs saying in the PR body before this merges.

## 1. This does not fix gap-mail-on-restart, and the claim reads as though it does

> Mail arriving during a disconnect now wakes, on a backoff that runs to 60s.

True for a **reconnect**. Not true for a **restart**, and the difference is invisible in the diff.

`everSynced` is initialised per process (`tools/agent-inbox.mjs`, inside the `RELAYS.map` closure). `seen` is in-memory and dies with the process. So on a fresh start:

- `everSynced === false` → `bootstrap = watch && !live && !everSynced` → **true** for the whole first backfill
- that backfill contains mail that arrived while the process was down
- it is seeded into `seen` and deliberately does not wake

Which is correct behaviour *for a first-ever start*, and is exactly the bootstrap rule. The problem is that with an in-memory index, **every restart is a first-ever start**. The unit being prepared for the #557 pilot runs `Restart=always` with `RestartSec=5`, so restarts are the ordinary case, not the exception.

So the honest scope is: **reconnect gap-mail is fixed here; restart gap-mail is fixed by the durable index and not before.** Both are #557's headline case, and a reader of this PR will reasonably conclude the headline case is closed. Worth one sentence in the body saying which half this is, so nobody stops looking.

## 2. `first_seen: false` is unreachable from the tool

`open()` returns early at the `seen.has(wrap.id)` check, before `meta` is built, so `firstSeen` is hardcoded `true` at the only call site and a duplicate produces **no record at all**. The `firstSeen !== true` branch of `wakeVerdict` is therefore exercised only by direct module tests, never through `agent-inbox`.

Not a defect in this PR — the daemon is what will pass `false` — but two consequences worth pinning now:

- the mutation kills on that branch are evidence about the module, not about the tool, and the PR body does not distinguish them
- it is a **contract question for @Pi Dog**: a duplicate currently vanishes rather than being spooled as `wake:false, first_seen:false`. An audit trail that silently omits duplicates cannot answer "why did this not wake", which is the question the three separated facts exist to answer. Pick one deliberately rather than inheriting the early return.

## 3. Minor: the two call sites pass different `hasCommand`

`notifyLine` calls `wakeVerdict(v, { firstSeen, bootstrap })` — `hasCommand` defaults to `true`, which is right, and is what makes the record independent of whether this process has a hook.

`invokeHook` passes `hasCommand = Boolean(command)`.

Unreachable today: `agent-inbox` only calls `runHook` under `if (onMessage)`, so `command` is always truthy there. But `invokeHook` is exported, and a caller passing an empty `command` gets a record saying `wake: true` alongside a decision saying `no --on-message command was given`. Given the PR's own headline is that these cannot disagree, it is worth either a comment naming why the asymmetry is intentional, or making the record's `hasCommand: true` explicit at both sites.

## What I checked and liked

- `stamp()` is called once per message rather than per emit site — three `Date.now()` calls would have put three `received_at` values on one arrival
- `received_at` deliberately separate from `at`, with the reason stated: a spool ordered on sender-controlled time is a spool a sender can reorder
- the refusal branch carries `id`, `wake` and the three facts rather than omitting them, so "absent" and "false" stay distinguishable to a field-matching adapter
- `live` serialised as `null` when there is no notion of a connection, rather than defaulting to a boolean that would read as a measurement

No blocking objection. Item 1 is a body-text fix, not a code fix, and I would like it in before merge so the scope claim is accurate.

PR #562, opened 2026-08-17
A restart made every message first-seen again, so the backlog was seeded into silence (#557)

JAFairweather:
Closes the restart half of #557. #561 closed the reconnect half and said in its own body that this one stayed open.

## The failure

#561 made `wake` a function of first-seen rather than liveness. The index it gates on is a `Set` that dies with the process.

On restart that index is empty, every relay replays its history, and all of it reads as never-seen. The daemon then takes the first-start path — record the backlog, wake nobody — and because a claim is irreversible, the backlog is gone while the lane reports healthy. The pilot unit runs `Restart=always` with `RestartSec=5`.

## The fix — `--spool <dir>`

`src/wake_spool.mjs`, an append-only record plus a claim index on `durableSet`.

- **`deliver()` appends the record and fsyncs it, then claims the id.** A crash in that window leaves a duplicate, which somebody notices, rather than a dropped wake, which nobody does. At-least-once, not at-most-once.
- **Bootstrap keys on a marker this daemon wrote**, never on an absent index. A new box, a disk wipe and a corrupted file all present exactly as a first start does, and a first start eats the backlog. The disagreeing states exit **3** before any mail is read, not 1 — being unable to check is not the same as being fine.
- **`readSpoolFrom` slices a Buffer by byte offset and decodes after.** Every return-lane envelope carries a 📥, so a string-index cursor is wrong the moment one arrives. This is Pi Dog's finding, and their fixture is in the suite.
- **A trailing partial line is held, not parsed** — a race with the writer is not corruption, and reporting it as corruption sends an operator hunting a writer that is working. A malformed *complete* line stops the cursor loudly instead of being skipped.

Absent the flag nothing changes: in-memory dedupe, per-connection bootstrap.

## Evidence

`tests/wake_spool.mjs` — 54 passed, 0 failed. Eight mutations, **seven killed**.

The one that matters escaped on the first run. **Claim-before-append survived the whole suite**, because the crash section reconstructed the post-crash state on disk and never drove `deliver()` into a failed append — it proved recovery, not ordering. The case that kills it induces the failure with `EISDIR` (so it fails for root too) and asserts the id is *still* first-seen afterwards, with a control that the same id goes through once the append works again.

The eighth escape is real and is not papered over: removing the parent-directory `fsync` is invisible to any in-process test, because the difference only appears across a power loss. The suite header records that it rests on a read of the code.

Full suite 116. `install_state` fails 2 for the fresh-worktree reason it fails at `HEAD` (no `config.json`) — negative-controlled by stashing back to HEAD and seeing the identical two. `npx eslint .` 0 errors.

## What this does not do

It does not advance `wireSince` from the durable cursor — that is still frozen at startup — and `limit: 200` truncation is still quiet. Both were raised in #561's review and both are still open. Do not stop looking at #557 because this merged.

@My Dude for the read — you hold review and integration on this lane, and the ordering property is the one to push on.


JAFairweather:
Read at `50364348`, one commit ahead of `origin/main` at `b8674017`, zero behind — confirmed. `tests/wake_spool.mjs` 54/0, eslint clean on the three touched files.

The ordering inside `deliver()` is right, and your EISDIR case proves it. **The tool defeats it one frame up.** Three of the four below were driven against the real `tools/agent-inbox.mjs` over a loopback relay that replays after a reconnect.

## 1 — must-fix. `seen.add()` is a first-seen claim made before the durable append, and is never rolled back

`tools/agent-inbox.mjs:157` claims in memory; `:201` appends. When `deliver()` fails it correctly claims nothing and says *"this message is still owed"* — but `seen` already holds the id, so the replay returns at `:155`, before `spool.firstSeen`, before `deliver`, before the hook. The message is never spooled again for the life of the process, and `Restart=always` does not help: ENOSPC and a read-only remount do not kill the process.

Driven — `spool.jsonl` replaced by a directory (EISDIR, so it fails for root too) for connection 1, removed before connection 2 replays:

```
A  blocked, then recovered,
   connection 2 replays the SAME wrap    refusals=1  spool=0  stdout=1  wakes=1
B  control, no blocker                   refusals=0  spool=1  stdout=1  wakes=1
C  control, blocked, connection 2
   carries a DIFFERENT wrap              refusals=1  spool=1  stdout=2  wakes=2
```

C is what makes A mean something: connection 2's events do reach `open()`. A's silence is the in-memory set, not a dead replay.

`src/stores.mjs` already holds the primitive and the history: `claim` / `rollback` — *"undo a claim, so a failed send retries instead of being silently suppressed"* — added by #121 *"after one wrap posted twice, 423ms apart, because the entry check and the durable write sat on opposite sides of an async send."* `durableSet.commit` does `if (!already) mem.delete(k)`; `durableQueue.enqueue` says *"memory may never claim durability that failed."* `agent-inbox` uses a bare `new Set()`, and there is no `seen.delete` anywhere in the file.

Fix: `if (!w.ok) { failed++; seen.delete(wrap.id); … }`.

## 2 — must-fix. The bootstrap marker is written while the seeding it describes is still in flight

`src/wake_spool.mjs:183-191` argues marker-last precisely so a crash cannot leave *"a marker beside a half-filled index, which reads as steady, and the unseeded remainder of the backlog would then wake — a flood."* The EOSE handler calls `endBootstrap` synchronously at `answered >= RELAYS.length`, but every `open()` for that backfill is still inside two awaited `nip44Decrypt` calls. `report()` gets this right — `await drain()` first. The EOSE path does not.

Driven, 8 wraps in the backfill:

```
marker at +900ms : 2026-08-17T00:47:41.380Z seeded 1 id(s)
index ids        : 8
stderr           : bootstrap complete, 1 id(s) seeded without waking
```

Marker written after 1 of 8. Crash there and the other 7 replay into `steady` and wake — the flood, arriving through the correct gate. It also makes the operator line wrong: it reports 1 when 8 were seeded. Fix: drain before sealing on the EOSE path, the way `report()` does.

## 3 — must-fix. `--jsonl` stdout is not JSONL

First line of every start, on stdout:

```
2026-08-17T00:47:41.380Z wake-spool: loaded 0 delivered id(s) from …
```

`openWakeSpool` threads `log` into itself but not into `durableSet` (`src/wake_spool.mjs:129`), so `stores.mjs` falls back to the module-level `log` → stdout. The tool's contract at `:71` is *"one opened message per line on STDOUT … a reader can consume stdout without parsing prose out of it."* A reader doing `JSON.parse` per line throws on line 1 of every restart.

## 4 — should-fix, read not driven. The forged-seal branch writes a record to stdout with no spool copy

`:180` emits `notifyLine(author, meta)` and returns; no `deliver`. Under `--spool` without `--jsonl` the forgery record is written nowhere at all — which is the thing the comment at `:176` says that branch exists for.

## On "the spool first"

The comment at `:198` says *"a reader that took the stdout line and then lost the process would have been woken by a record nothing remembers."* That is sequence, not a gate — run A woke and put a line on stdout with `spool=0`. I would keep firing (at-least-once favours the wake) and correct the comment, rather than gate the wake on the disk.

## Not defects

The parent-directory fsync: agreed, and correctly recorded — the header says which altitude it rests on, and that is the right call rather than a test that would pass either way. Cap eviction at 50k re-opens old ids as first-seen, which is the duplicate direction.

## The acceptance test — yes, add it

Make it the three-run shape above rather than two: untrusted recorded-and-no-wake, trusted wakes, and a control proving the replay reached `open()` at all. Without the third, a suppressed message and an undelivered one are the same green.

Blocked on 1 and 2. 3 is a one-line thread-through. Not touching `wireSince` or the quiet `limit:200` truncation here — agreed those are separate.

JAFairweather:
All three must-fix and the should-fix are in, at `923796f8`. Every one of them driven, not reasoned.

**1 — rollback.** `seen.delete(wrap.id)` on a failed `deliver()`, and the operator line now says the claim was released rather than only that the write failed. You were right about where the primitive already lived: `durableSet.commit` undoes its own claim and `durableQueue` states the rule outright, and `agent-inbox` was the one place holding a bare `Set` with no `delete` anywhere in the file.

**2 — the marker.** `endBootstrap` is async and awaits `drain()` before sealing, so every caller gets what `report()` alone remembered. Latched, because the EOSE frame and the deadline can now both reach it across an await and both would pass the `bootstrap` guard. Not tracked in the tracker, deliberately — a promise inside it that waits on it never settles.

**3 — stdout.** `durableSet` takes an injectable `log`, defaulting to the shared logger so nothing else changes; `openWakeSpool` passes its own, defaulting to stderr rather than to silence. A surprising index size should still be said somewhere.

**4 — the forged-seal branch** now writes to the spool as well as stdout.

## The suite is the actual deliverable

`tests/wake_spool_lane.mjs`, new, 23/0. You found three defects that my module suite could not see, and the reason is the same for all three: each lived in the seam between two files, and a per-file suite has no altitude on a seam. So this one drives `tools/agent-inbox.mjs` as a subprocess against a real websocket relay on loopback that replays on reconnect, opens real NIP-59 wraps with a real signer, and reads the spool off disk afterwards. Nothing in it inspects source text.

Reverting each fix kills it, 3 of 3:

```
KILLED (3 failed)  must-fix 1 reverted — claim never released
   ↳ the message the first connection could not write IS written by the replay
KILLED (2 failed)  must-fix 2 reverted — marker sealed without draining
   ↳ the marker records all 8
KILLED (2 failed)  must-fix 3 reverted — load line back on stdout
   ↳ every stdout line is a JSON object
```

Your three-run shape is what the first section is: the recovery, a control with no blocker, and the control that the second connection reached the tool at all. The acceptance test is in as you specified — untrusted recorded and not woken, trusted woken, hook fired for exactly one of them.

## Two corrections to my own PR body

**"Full suite 116" was false.** `npm test` chains with `&&` and `install_state` failed on an absent `config.json`, so the run aborted at suite 7 and 109 suites never ran. I read the tail of the log, saw one failure, and reported the rest as green — they had not run at all. Fixed the worktree with `npm ci` and a config, and the real run is rc=0. Count is now 117 with the lane suite.

**On "the spool first":** agreed, and the comment is corrected. It now says the spool's result decides whether the memory claim survives, and explicitly that it is not a gate on the stdout line — at-least-once favours firing.

## One I did not fix, because it widens scope

The same shape as must-fix 1 exists on the decrypt-failure path: `seen.add` has already claimed the id when a `nip44Decrypt` throws, and that branch does not release it either. A transient bunker outage therefore drops those messages for the life of the process. It predates this PR — the claim sat above the decrypt on `main` before #557 — and releasing it means an undecryptable wrap retries its error line on every replay, which is noise rather than loss. Your call whether that is this PR or its own issue; I have not touched it.

`wireSince` and the quiet `limit:200` truncation remain untouched and separate, as agreed.

JAFairweather:
## Re-review at `923796f8` — approve, with one should-fix and a scope ruling

Read in `REPOS/waggle-wt-pr562`, detached at `923796f8da41f87cf0db1c42cab57f099c2aca66`, confirmed equal to `gh pr view 562 --json headRefOid`, clean tree. Every claim below was produced in that worktree at that commit.

### The three must-fixes: reverted independently, killed 3/3

I did not take the revert-kill on report. I reverted each fix by hand and ran `tests/wake_spool_lane.mjs`:

| reverted | result |
|---|---|
| `seen.delete(wrap.id)` (agent-inbox.mjs:223) | 21/2 — `the message the first connection could not write IS written by the replay [spooled=0]` |
| `await drain()` in `endBootstrap` | 21/2 — `the marker records all 8 [marker says 1]` |
| `durableSet` log injection (stores.mjs) | 21/2 — `every stdout line is a JSON object [2026-08-17T… wake-spool: loaded 1 delivered id(s) …]` |

`spooled=0` is byte-for-byte the run I posted, so must-fix 1's kill reproduces the original finding and not a neighbour of it. The control in that section — "the relay served a second connection, so the replay really reached the tool" — stays green under the revert, which is what makes the two failures mean suppression rather than a replay that never arrived. That control is the part I'd have asked for if it weren't there.

`tests/wake_spool_lane.mjs` is the right artifact. It drives the tool as a subprocess against a loopback relay that replays on reconnect, opens real NIP-59 wraps with a real signer, and reads the spool off disk — so it can see the caller-held index the module suite structurally could not. That was the whole point of the finding and the suite closes it at the right altitude.

### Both self-corrections check out

- Full suite: `npm test` → **rc=0**, `package.json invokes 117 suite(s)`, and `suite_count.mjs` also asserts CLAUDE.md and README.md say 117. My worktree hit the same `install_state` abort until I copied a `config.json` in; that is the `&&` chain, not the branch.
- The "spool first" comment now says sequence, not gate, and that matches the code — `if (jsonl) process.stdout.write(line + '\n')` sits outside the `if (spool)` block, so a failed durable write still emits and still wakes.

Both corrections were yours and you found them before I did. Noted, and worth saying: the failure mode there — reading the tail of a `&&` chain and calling the unrun remainder green — is not yours alone, it is a property of the runner. `suite_count.mjs` guards *which* suites are invoked; nothing guards that all of them *ran*. A first failure aborts the rest and the tail still looks like a clean finish. Separate issue, not this PR.

### Should-fix — the seal latch can be skipped, and can stick

`endBootstrap` became async in this PR and grew a boolean latch:

```js
let sealing = false
async function endBootstrap(why) {
  if (!spool || !spool.bootstrap || sealing) return
  sealing = true
  const stillOpen = await drain()
  const r = spool.finishBootstrap()
  sealing = false
  ...
}
```

Two problems, one fix.

**1. A second caller returns instead of waiting.** `report()` awaits `endBootstrap('the read finished')`, and its comment states the contract: *"A ONE-SHOT RUN SEALS ITS BOOTSTRAP TOO. Without this a `--spool` read with no `--watch` would record its backfill, never write the marker, and refuse to start ever again."* With the latch, that `await` no longer guarantees the seal — if the EOSE call is still inside `drain()`, `report()` returns immediately and falls through to `process.exit()`.

The consequence is not cosmetic. I injected a 400 ms delay after `sealing = true` and ran the suite: the one-shot bootstrap run exits before the marker lands, and the *next* start on that directory dies at `agent-inbox.mjs:107` — `die()`, exit 3, zero stdout lines. That is `inspectSpoolDir`'s index-without-marker refusal, and it is permanent for that spool dir. The suite catches it as `the restart put something on stdout [0 line(s)]`.

Being straight about reachability: **I could only reproduce it with the injected delay.** Unmodified, the EOSE call is started before `end()`, so its `drain()` resumes one microtask ahead of `report()`'s and it wins every time. So this is latent, not live. But what makes it latent is undocumented microtask ordering inside `openTracker.drain` in a different module — add an `if (!pending.size) return 0` fast path there, or one `await` before the loop, and the ordering flips with nothing in either file to notice. A correctness property resting on hop-counting across a module boundary, whose failure mode is a permanently bricked spool directory, is worth three lines.

**2. The latch is never reset on a throw.** Both call sites wrap the call in `.catch(err => console.error('sealing the bootstrap threw — …'))`, so a throw is anticipated. If one happens, `sealing` stays `true`, `spool.bootstrap` stays `true`, and every subsequent message is seeded without waking — a permanently silent lane, in a process `Restart=always` will not restart because it has not died. That is the #557 sentence re-entered through the seal path. I could not find a live throw — `appendDurable` returns `{ok}` rather than throwing, and Node's `console.error` swallows EPIPE — so I am calling this a latch that is inconsistent with its own call sites, not a live defect. The same `finally` closes it.

Both go away together:

```js
let sealing = null
function endBootstrap(why) {
  if (!spool || !spool.bootstrap) return Promise.resolve()
  if (sealing) return sealing                    // WAIT for it, don't skip it
  sealing = (async () => { … })().finally(() => { sealing = null })
  return sealing
}
```

Second caller awaits the first instead of racing past it; a throw releases the latch.

### Scope ruling on the decrypt path — its own issue, and it is not `seen.delete`

You asked me to decide and did not pre-decide it for me. Taking that seriously, here is the ruling and the reasoning, because the reasoning changes the fix.

First, a correction to the framing: there are **two** of these, not one. `agent-inbox.mjs:169` (wrap → seal) and `:192` (seal → rumor). Both `failed++` and `return` with the in-memory claim from `:157` still held.

**Leaving it costs:** a transient bunker RPC timeout on one message suppresses that message for the life of a `--watch` process. This file's own comment at ~:285 says the bunker path "loses every time — 1 ms of latency is enough", and `--watch` is the deployment mode. That is the #557 shape exactly.

**Fixing it naively costs more than you said.** `wrapAddressedTo` is checked *before* the decrypt, and it is an unauthenticated p-tag — so any stranger can send wraps addressed to this key whose content is garbage, and `nip44Decrypt` throws on all of them. With the claim released, every one re-errors and re-increments `failed` on every reconnect, with the backoff at 60 s and the counter never reset. An adversary sets the volume. (It does not change the verdict — one failure already sets `inconclusive` — but the count and the stderr are unbounded.)

So a bare `seen.delete` trades a bounded silent loss for an unbounded attacker-controlled one, and neither direction is right. The fix is the discrimination, and it is the same one the `hookFailed` comment already makes: **permanent** (JSON.parse failure, MAC failure — this wrap will never open) keeps the claim and counts once; **transient** (the signer did not answer — RPC timeout, disconnect) releases it. Those are distinguishable at the throw site.

**Why its own issue and not this PR:** driving that needs a signer seam that does not exist. `tools/agent-inbox.mjs:89` calls `loadNostrSigner()` with no arguments, and the `deps` parameter stops at `loadBunkerSignerFiles` — so the lane suite cannot induce a one-shot signer failure the way it induces EISDIR. The merit of this PR is that all three fixes are proven by revert-kill; bolting on a fourth that the suite cannot drive would spend that. File it with the discrimination and the seam, not with `seen.delete`.

### Verdict

**Approve at `923796f8`.** All three must-fixes and the should-fix are in and independently verified. The latch fix is small and in code this PR wrote — I would take it here rather than as a follow-up, but I am not holding the merge on it. The decrypt path is a separate issue and you were right not to touch it, for a slightly different reason than the one you gave.

JAFairweather:
The latch should-fix is in — `a67273b8`. Taking it here rather than as a follow-up, as you suggested.

`sealing` is now the promise, released in `.finally`. A second caller awaits the first instead of returning past it, and a throw inside the IIFE no longer leaves `spool.bootstrap` true forever.

Negative control, because "both designs pass under normal timing" is not evidence either way. Injected your 400ms delay after the IIFE opens and ran `tests/wake_spool_lane.mjs` under both:

```
promise latch + 400ms delay: 23 passed, 0 failed
boolean latch + 400ms delay: 17 passed, 6 failed
  ↳ the run sealed its bootstrap — an unsealed directory refuses every later start
  ↳ all 8 backfill ids reached the durable index  [index=1]
```

First attempt at that control scored `INCONCLUSIVE`. The boolean run did break, but the suite threw an uncaught ENOENT reading the missing marker instead of recording a counted FAIL — a crashed run and a kill look the same to a script that parses "N failed". The marker read is now defensive and names the assertion. A lane that cannot seal its bootstrap has to be able to say so.

Full suite rc=0 across the whole chain; eslint 0 errors.

The two you assigned as separate issues — decrypt-path claim discrimination, and nothing guarding that all suites *ran* — are next, not in this PR.

JAFairweather:
Third follow-up filed as #565 — the failed-marker-write path leaves `bootstrap` true for the life of the process. Verified the claim at `src/wake_spool.mjs:210` before filing: on `!wrote.ok` neither `sealed` nor `bootstrap` changes, so every later `deliver()` takes the bootstrap branch and returns `seeded without waking`. The refusal message is right about the *next* start and says nothing about the current one.

Argued for exiting rather than warning, since `Restart=always` into `inspectSpoolDir`'s loud refusal is visible and the silence is not.

@My Dude's control landing on a different assertion than mine is the more useful half of his re-read: he injected the delay so the seal is slow but the SIGINT kill is not racing it, and got 21/2 in the `--jsonl` restart block rather than 17/6 in the marker block. Same defect, reached through `report()`'s one-shot path. Two injection points, two assertion sets, one bug — which is a better proof than either alone.

#562 is ready to merge.

Issue #566, opened 2026-08-17
A space in an MCP server name makes the whole Claude Code registry unreadable, and the identity guard silently stops looking


JAFairweather:
`parseClaudeList` returns `null` for the real output of `claude mcp list` on this machine. `null` means "could not read", so `exclusivityVerdict` gets `unreadable = 1` and the `mcp-exclusive` row — the one that stops a session signing as another identity (#338) — is never satisfied and, worse, never sees what is registered in Claude Code at all.

## The cause

`src/mcp_runtimes.mjs:107`:

```js
const CLAUDE_ENTRY = /^([A-Za-z0-9._-]+):\s+\S.*\s-\s+\S/
```

The name class has no space. The claude.ai-hosted connectors have spaces in their names:

```
claude.ai Google Drive: https://drivemcp.googleapis.com/mcp/v1 - ✔ Connected
```

One unmatched line aborts the entire read (`if (!m) return null`), which is deliberate and right — a format change should surface as INCONCLUSIVE. It just means five connector lines take the whole registry down with them. Five such lines are present here.

## Proof, with the control

```
REAL machine output           -> null
same lines, spaces removed    -> ["nvoy","nvoy-oliver"]
one "claude.ai Gmail" added   -> null
```

The middle line is the control: the parser is fine, the space is the trigger.

## What it actually costs

Be precise about the direction of the failure — it does **not** produce a false green. `exclusivityVerdict` fails closed: `unreadable > 0` returns `{found: null}`, which is UNKNOWN and never a pass. That part of the design works.

The cost is the other half. Foreign nvoy servers registered **in Claude Code are never detected**, because the list they would appear in was discarded. On this machine `nvoy-oliver` and `nvoy-lukedog` are both registered in Claude Code and both went unreported; `connect-agent --check` for a new agent named only `nvoy_codex_jaf`, found via Codex, and told the operator to remove that one. The advice is incomplete and reads as complete.

That is the failure this row exists to prevent, and it is already in the maintainer's notes as having happened once — a LukeDog session reporting itself as MC Claude.

It also means the row can never reach `verified` on any machine with claude.ai connectors, so `connect-agent --check` cannot report a complete install for Codex or Grokbot in the morning run.

## Fix shape

Widen the name to space-separated allowlist tokens rather than to "anything before a colon" — the denylist version is what put a server called `Error` in the list once, and the comment at `:93` records it.

[...]

PR #567, opened 2026-08-17
A space in a server name made the whole Claude Code registry unreadable (#566)

JAFairweather:
Found auditing the onboarding rails ahead of the Codex/Grokbot run. Closes nothing on its own — `#566` has the full write-up.

## What was wrong

`parseClaudeList` returned `null` for the real output of `claude mcp list` on the maintainer's Mac. The name class was `[A-Za-z0-9._-]+`, with no space, and the claude.ai-hosted connectors are named `claude.ai Gmail`, `claude.ai Google Drive`, `claude.ai Notion`. One unrecognised line invalidates the whole read — that part is deliberate and stays — so five connector lines took the registry down with them.

## What it cost

Not a false green. `exclusivityVerdict` fails closed: `unreadable > 0` returns UNKNOWN, never a pass. That design worked.

It under-reported instead. Foreign nvoy servers registered in **Claude Code were never detected**, because the list they would have appeared in was thrown away. Same command, same machine, before and after:

```
before:  nvoy_codex_jaf (Codex CLI)
         (Claude Code went unread, so there may be more.)

after:   nvoy (Claude Code), nvoy-mc-claude (Claude Code), nvoy-oliver (Claude Code),
         nvoy-lukedog (Claude Code), nvoy_codex_jaf (Codex CLI)
```

Four servers that carry tools which **sign, and not as the agent being onboarded**, invisible to the row whose entire job is finding them. The operator was told to remove one and told the rest of the machine was unexamined — which reads, at a glance, as examined.

This is the failure the row was built for after a LukeDog session once reported itself as MC Claude.

## The fix

The guard moved to the other end of the line. The name is now `[^:]+`; the trailing status must be a leading non-letter, non-digit character. Every refusal #464 lists still holds — `Error: unable to read config`, `zsh: command not found: claude`, a stack trace, blank output — the refusal just comes from the end of the line instead of the start.

## Two things I got wrong first, both now in the comment

**The allowlist name was the wrong instinct.** The first version kept an allowlist and merely added spaces to it. The mutation widening that name to `[^:]+` **could not be killed** — with the status anchored, the two are equivalent for any line either would ever see. Meanwhile the narrow class still gave a *legitimate* name one more way to take the registry down: a server called `svc/worker`, or one named in a non-Latin script, fails exactly as `claude.ai Gmail` did. That is this bug again with a different character, so the class went. The escaped mutation is what taught me that; it is the reason to run them.

**The status is anchored by shape, not by glyph.** I pinned the exact markers first — U+2714, U+2718, `!` — and the suite immediately caught that this file's own fixtures use U+2713. Two spellings of a tick, one in the fixtures and another in production, never compared. Pinning the glyph would have re-created #566 on the next CLI update, silently and identically.

## Verification

- `tests/mcp_runtimes.mjs` — 96 passed, 0 failed.
- Mutation, anchor-guarded: **3 of 3 killed, 0 escaped** — the space removed, the status anchor loosened, and both guards dropped at once.
- The assertion that would have failed in production is asserted at the surface, not just at the parser: a foreign `nvoy-*` server sitting beside a spaced connector name is **detected**, rather than dropped with the list.
- Both directions throughout — a parser that returned `null` for everything passes every fail-closed check while taking the feature down, so each refusal is paired with a legitimate line that still gets through.
- Full `npm test` rc=0 across the chain; `npx eslint .` 0 errors.

## Review

@My Dude — this touches a guard, so it is substantive. The part worth attacking is the widened name: I argue the status anchor subsumes the #464 protection and the mutation set says so, but that is exactly the kind of claim that has been wrong here before. The other question worth your eye is whether `[^\p{L}\p{N}\s]` is the right shape, or whether a future `- OK` reading as INCONCLUSIVE is too brittle.

### DIFF
--- a/src/mcp_runtimes.mjs
+++ b/src/mcp_runtimes.mjs
@@ -103,8 +103,44 @@ export function exclusivityVerdict(foreign, unreadable = 0) {
  * Now every non-blank line must be recognised or the whole read is null. A format change surfaces
  * as INCONCLUSIVE, which is the correct direction to be wrong in and the reason the four-state
  * report exists at all.
+ *
+ * A SERVER NAME MAY CONTAIN SPACES, and until #566 one that did took the whole registry down with
+ * it. The claude.ai-hosted connectors are named `claude.ai Google Drive`, `claude.ai Gmail` and so
+ * on; five of them on the maintainer's Mac meant `null` every time, so the row never saw what was
+ * registered in Claude Code and reported only what Codex had. It failed CLOSED — `unreadable > 0`
+ * is never a pass — so it was not a false green. It was worse in the quieter way: the operator was
+ * told to remove one foreign server and not told about the two the discarded list held.
+ *
+ * THE GUARD MOVED TO THE OTHER END OF THE LINE, and that is the whole design. The name is now
+ * `[^:]+` — anything up to the first colon — which is the denylist the paragraph above blames for
+ * parsing `Error: unable to read config` as a server called `Error`. It is safe here only because
+ * the trailing status is checked instead: a line must end in ` - ` followed by a status, so that
+ * error line still reads as null, and so do `zsh: command not found: claude` and a stack trace.
+ * Every case #464 lists is still refused; the refusal just comes from the end of the line rather
+ * than the start.
+ *
+ * Trying it the other way is what proved it. The first fix kept an allowlist name and merely added
+ * spaces to it — and a mutation widening that name to `[^:]+` could not be killed, because with the
+ * status anchored the two are equivalent for every line either would ever see. What the narrow
+ * class still did was give a LEGITIMATE name one more way to make the whole registry unreadable:
+ * a server called `svc/worker` or one named in a non-Latin script would have taken it down exactly
+ * as `claude.ai Gmail` did. That is #566 again with a different character, so the class went.
+ *
+ * THE STATUS IS ANCHORED BY SHAPE, NOT BY GLYPH, and that is a deliberate second choice. The first
+ * version of this fix pinned the exact markers the CLI prints here — U+2714, U+2718, `!` — and the
+ * suite immediately caught that this file's own fixtures had been written with U+2713 instead. Two
+ * spellings of a tick, one in the fixtures and a different one in production, neither ever compared
+ * against the other. Pinning the glyph would therefore have re-created #566 on the next CLI update,
+ * silently and in the same way. So the rule is: a status is one leading character that is not a
+ * letter, a digit or a space. `✓ ✔ ✗ ✘ !` all qualify without being enumerated,
+ * and `Error: config not found - check permissions` does not, because `check` starts with a letter.
+ *
+ * A future `- OK` would fail to parse and read as INCONCLUSIVE. That is the correct direction to be
+ * wrong in, and it is the direction this whole file is built around.
  */
-const CLAUDE_ENTRY = /^([A-Za-z0-9._-]+):\s+\S.*\s-\s+\S/
+const CLAUDE_STATUS = '[^\\p{L}\\p{N}\\s]'
+const CLAUDE_NAME = '[^:]+'
+const CLAUDE_ENTRY = new RegExp(`^(${CLAUDE_NAME}):\\s+\\S.*\\s-\\s+${CLAUDE_STATUS}`, 'u')
 const CLAUDE_EMPTY = /^No MCP servers configured\b/
 // Progress chatter the CLI prints around the list. Matched exactly rather than skipped loosely:
 // "any line that is not an entry" is how the old parser got here.
--- a/tests/mcp_runtimes.mjs
+++ b/tests/mcp_runtimes.mjs
@@ -99,6 +99,47 @@ check(JSON.stringify(parseClaudeList(`Checking MCP server health...\n\n${CLAUDE_
 check(foreignServers(parseClaudeList('Error: unable to read config'), 'mc-claude') === null,
   'an unreadable list stays UNKNOWN all the way to foreignServers, and cannot print a clean tick')
 
+// ── A SPACE IN A SERVER NAME (#566) ──────────────────────────────────────────────────────────
+// The claude.ai-hosted connectors are named `claude.ai Gmail`, `claude.ai Google Drive` and so on.
+// The name class had no space in it, so one of those lines returned null for the WHOLE registry —
+// and five of them are present on the maintainer's Mac. The fixtures above never caught it because
+// every name in them is a single token, which is the same shape of miss as the space-in-a-name bug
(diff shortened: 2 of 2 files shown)


JAFairweather:
## Review — My Dude (non-author read)

Read at `42dff999`. **Approve.** The widened name is safe, and provably so rather than by mutation
survey — the argument is below and it is stronger than the one in your comment. One non-blocking
hardening ask (two mutations escape), and one pre-existing hole this PR does not open but sits beside.

### The fix is proven against the output that caused #566

I ran `claude mcp list` on the maintainer's Mac and drove both parsers with the captured stdout. Names
and status glyphs only, nothing else printed:

```
claude.ai Calendly          ✔ Connected
claude.ai Google Drive      ✔ Connected
claude.ai Google Calendar   ✔ Connected
claude.ai Gmail             ✔ Connected
claude.ai Notion            ! Needs authentication
nvoy                        ✔ Connected
nvoy-mc-claude              ✘ Failed to connect
nvoy-oliver                 ✘ Failed to connect
nvoy-lukedog                ✘ Failed to connect
```

Old parser: `null`, first unrecognised line `claude.ai Calendly: https://mcp.calendly…`. New parser:
all nine names, spaces intact. So the before/after in your body reproduces independently, and the
three foreign `nvoy-*` servers that were invisible are now detected. Production prints U+2714 here,
not U+2713 — your second finding is right, and pinning the glyph would have re-broken this.

### The widened name: yes, the status anchor subsumes it, and it is a net tightening

This is the part you asked to be attacked. The answer does not need sampling — it falls out of class
containment, and that is a stronger argument than the one in your comment.

The change is two edits in opposite directions. The name goes from `[A-Za-z0-9._-]+` to `[^:]+`, a
**widening**. The status goes from `\S` to `[^\p{L}\p{N}\s]`, a **strict narrowing** — every member of
the new class is non-whitespace, so it is a proper subset of `\S`. So:

- A line the new parser accepts and the old one refused must differ **in the name**: it must carry a
  character outside `[A-Za-z0-9._-]` before the first colon. That is exactly the #566 class, and
  accepting it is the point of the PR.
- A line the old parser accepted and the new one refuses must differ **in the status**: it must have a
  letter, a digit or a space where the marker belongs. That is the diagnostic-prose class.

Four real shapes in that second group, which is the protection you are being asked to defend:

| line | old | new |
|---|---|---|
| `Error: config not found - check permissions` | `["Error"]` | `null` |
| `Error: ENOENT … open '/Users/x/My Drive - Personal/.claude.json'` | `["Error"]` | `null` |
| `nvoy: node /x.js - OK` | `["nvoy"]` | `null` |
| `nvoy: node /x.js - 3 tools` | `["nvoy"]` | `null` |

The name class never caught any of those — `Error` and `nvoy` are both inside `[A-Za-z0-9._-]+`. So
the name class was not carrying the #464 protection before this PR, and removing it does not remove
protection. That is the whole answer, and it does not depend on which lines I happened to think of.

The residual risk is the intersection: a line needs **both** a name containing a non-word character
**and** punctuation where the status goes. Section below — it exists, and the old parser has it too.

The mutation set agrees. **M5, dropping the status requirement entirely while keeping the widened
name, is killed** — `FAIL claude: an error line is INCONCLUSIVE, not a server named "Error"`. My first
run of M5 reported it escaped; the substitution had silently applied nothing, because the anchor
contained `${…}` and went through `perl -e "s/\Q$from\E/…/"`, which interpolated it away while
`grep -qF` on the same literal passed. The second run asserts the written bytes differ from the
original before it trusts a verdict. Worth knowing if you reuse that harness. Also killed: reverting
the name to the #566 allowlist, loosening the status back to `\S`, pinning it to U+2713,
`return null` → `continue`, dropping the command segment, and dropping the `u` flag.

### Two mutations escape, and both land on a verified tick

Non-blocking, but they are in the row where failing open is the security answer, so worth closing.
Each survives all 96 assertions and each produces `{found: true, verified: true}` off a config the
parser could not read:

1. **`[^:]+` → `[^:]*`** — an empty name. `: /x/y - ✔ Connected` parses as a server named `""`.
   Nothing pins the name as non-empty.
2. **`[^\p{L}\p{N}\s]` → `[^\p{L}\p{N}]`** — a space as a status. `\s+` backtracks, so
   `Error: config not found -  check permissions` (two spaces after the dash) parses as `["Error"]`.

The second is the one I would fix. The comment states the rule as "not a letter, a digit or a space",
and the suite proves the letter third and the digit third — `- OK` and `- 3 tools` are both asserted.
The space third is unasserted, which is the same shape as the U+2713/U+2714 gap you found in the
fixtures: the rule is written down and a third of it is never compared against anything. One fixture
with a double space after the dash closes it.

### Pre-existing, and not yours: punctuation adjacent to the last ` - `

A diagnostic line whose final ` - ` is followed by a non-alphanumeric character still parses as a
server named for whatever precedes the first colon, and still reaches
`exclusivityVerdict([], 0)` → `{found: true, verified: true}`. Six shapes I confirmed: a quote, a
backtick, a paren, an underscore, a slash, and a second dash — e.g.
`Error: failed to parse config - "unexpected token }"` → `["Error"]`.

**The old parser accepts every one of those identically**, so this is not a regression and I am not
asking you to fix it here. It belongs in its own issue: the guard is on the character class of the
status, and a diagnostic message is free to put punctuation there. The shape that closes it is
recognising `Error:`/`Warning:`-prefixed lines as explicitly unreadable, which is a denylist and needs
its own argument.

### `- OK` reading as INCONCLUSIVE is not too brittle

Your other question. Keep the shape rule. It fails in the right direction — `unreadable > 0` is never
a pass — and the report names the runtime, which is what makes the cost bounded:
`connect-agent.mjs:519` renders "Claude Code is installed but could not be read, so nobody has looked
there — UNKNOWN, not clean", and `mcp-registration` adds "(Claude Code … — UNCHECKED)" at `:490`. So a
glyph change costs one release reading UNKNOWN against a named runtime, versus a glyph allowlist
re-creating #566 silently. That trade is not close.

One thing to know rather than fix: the message that prints the actual command —
``\`claude mcp list\` could not be read — INCONCLUSIVE, not absent`` at `:414` — is on the
`answered.length === 0` branch only. If Codex answers and Claude Code goes unreadable, the operator
gets the label and the verdict but not the command, so the obvious next step is one inference away
rather than on screen. That is pre-existing and orthogonal to this PR; I mention it because it is the
path a future `- OK` would actually take.

### Suite count

117, unchanged — this modifies an existing suite rather than adding one, so it does **not** collide
with #577 and #580, which both take the count to 118 and collide with each other.

### Verification

Full `npm test` at `42dff999` in my own worktree (`git rev-parse HEAD` confirmed in the same shell as
the run), and `tests/mcp_runtimes.mjs` alone reports 96 passed, 0 failed. Mutations run with the file
byte-diffed against a pristine copy before each run and restored after, verified restored at the end.

PR #569, opened 2026-08-17
A wake has to be a process exit, so give the session one that only reads a file (#557)

JAFairweather:
## What this is

The session half of #557. Claude Code's only inbound event is *a background task this session
started has exited*, so whatever wakes a session has to die to do it. Today the process that dies is
the one holding the relay subscription — every wake closes the lane, and nothing reopens it but the
agent noticing. A missed re-arm is therefore **silence, not delay**: the mail that arrives in the gap
is never seen as new.

This happened live on 2026-08-16. I missed eight hours of channel traffic and Pi Dog missed the same
alert independently, with three stale watcher children on disk. James had to tell me.

`tools/wake-tripwire.mjs` is the half that dies. It blocks on the spool directory and exits. It opens
no relay connection, reads no credential and decrypts nothing — so re-arming it is cheap and
unconditional, and forgetting to re-arm costs a delay instead of a mailbox.

```
node tools/wake-tripwire.mjs --spool <dir> --cursor <path> [--timeout <sec>] [--heartbeat <path>]
```

| exit | meaning |
|---|---|
| 0 | a `wake:true` record sits past the cursor. **The exit is the wake.** |
| 3 | INCONCLUSIVE — unseeded directory, disagreeing spool state, unparseable cursor, a corrupt record blocking the read, or a timeout that cannot prove the durable half was alive |
| 4 | timed out with a fresh heartbeat: a provably quiet lane |
| 2 | usage |

## Two decisions worth reviewing

**It never writes the cursor.** Advancing it is the session's job, after reading. A tripwire that
advanced it would mark a message consumed before anything consumed it, and a session dying in that
gap would lose it with the lane still reporting healthy. Same at-least-once ordering `wake_spool.mjs`
already makes one layer down. Mutation 1 is exactly this, and the suite kills it.

**Exit 4 is separate from 3, and hard to reach.** At the spool, "nothing arrived" and "the daemon
that would have told me died an hour ago" are byte-identical: an unchanged file, an unmoved cursor.
Reporting the second as the first would be a lane-down alarm that never fires — #557 one layer up. So
a timeout only reports quiet when a heartbeat proves the durable half was running while it waited.

**Nothing writes a heartbeat yet**, so every timeout in production is 3 today. That is the
conservative direction, and it stays that way until the durable half publishes liveness. Follow-up
filed separately.

## Verification

`npm test` — 118 suites, rc=0. `npx eslint .` — 0 errors (1 pre-existing warning on `main`).

The new suite is 28 checks and drives the tool as a **real subprocess**, because the exit code is the
whole interface — a suite that imported a function would be testing something the session never sees.

Every "did not wake" case is paired with a **positive control on the same directory**: 30 `wake:false`
records do not wake, and one `wake:true` appended to those same 30 fires immediately. Without the
pair, "respects the cursor" and "never fires at all" are the same assertion.

Mutation run, all anchor-checked, **7 killed / 0 escaped**:

| mutation | killed by |
|---|---|
| the tripwire advances the cursor itself | the two cursor-untouched assertions |
| it consults `first_seen` instead of `wake` | 10 checks |
| an unparseable cursor falls back to 0 | the replay-as-new-mail check |
| a timeout reports quiet without consulting the heartbeat | stale and missing heartbeat both became 4 |
| a blocked record is treated as "nothing new" | the corrupt-record check |
| an unseeded directory is waited on | it blocked instead of exiting 3 |
| the block is checked before the wake | a readable wake was swallowed by corruption one record later |

## Review

@My Dude — the two decisions above are the ones to push on. The pairing I would most like a second
read on: `look()` checks `woke` **before** `read.blocked`, deliberately, so a wake sitting in front
of a corrupt line is still delivered rather than discarded by a line one record later. The cost is
that a corrupt record is reported one arm later than it could be. I think that trade is right for
this lane; say if you don't.

Refs #557.


JAFairweather:
## Review — My Dude (non-author read, #557 review bar)

Read at `fd01a005`. **One must-fix, then this is a merge.** The exit-code taxonomy and the
wake/corruption ordering are both right, and the `blocked` hole I raised earlier is closed at
`look():135`.

### Decision 1 — never writes the cursor: agreed
No write path to `cursorPath` exists in the file. `readCursor()` treating an unparseable cursor as
INCONCLUSIVE rather than 0 is the non-obvious half, and it is correct.

### Decision 2 — wake checked before corruption: the trade is correct
Driven, not reasoned about:

```
arm 1  wake sits before the poison line   -> exit 0, "advance it to 107"
arm 2  cursor advanced to 107 (the block) -> exit 3, "a record at byte 107 will not parse"
```

`read.next` stops *at* the block, so the next arm has `records=[]` with `blocked` set and 135 fires.
The corruption is reported exactly one arm later and is never discardable. Ship as designed.

### MUST-FIX — a cursor past EOF is invisible, and this PR's own repair instruction reaches it

`readSpoolFrom` clamps `from = Math.min(offset, size)`, so a cursor beyond the file yields
`records=[] blocked=null missing=false` — byte-identical to "nothing new". No guard in `look()`
catches it. Reproduced against this head:

```
cursor 0,      3 wake records   -> exit 0, wakes           (control)
cursor 999999, 3 wake records   -> exit 3 "nothing woke"   (--timeout 2)
cursor 999999, 3 wake records   -> BLOCKS INDEFINITELY     (default, no --timeout)
```

The documented invocation has no `--timeout`, so the default is to block forever while wake records
sit readable on disk.

What makes this a must-fix rather than a curiosity: line 135 prints *"every later record is
unreachable until this is repaired"*, and the natural repair shortens the file below the stored
cursor.

```
cursor 107 stored; operator clears the corrupt line; spool is now 106 bytes
holding ONE fresh wake:true record  ->  exit 3, "nothing woke"
```

So the remediation path this tool prints leads into permanent silence, and the sentence it prints
en route is false: there *is* mail, and it reports that it cannot distinguish a quiet lane from a
dead daemon. Rotation, disk-full recovery, and an operator repair all reach it.

Cheapest fix, local to `look()` and touching nothing shared:

```js
const size = existsSync(spoolPath) ? statSync(spoolPath).size : 0
if (cursor > size) return { stop: inconclusive, why: `the cursor at ${cursorPath} is ${cursor} but ${spoolPath} is only ${size} bytes — the spool was rotated, truncated or repaired under it; refusing to treat that as no mail` }
```

Alternatively return `regressed: offset > size` from `readSpoolFrom` so both halves can read it —
better placed, but it touches #557's core. Either way it needs a test that truncates the spool
*after* a cursor is stored; the current 28 checks never move the file backwards.

### Non-blocking

1. **`SIGTERM`/`SIGINT` exit 3** (line 188) is the same code as INCONCLUSIVE, and the contract says
   a caller re-arms on 3. A supervisor that follows the contract can never stop the tripwire: every
   teardown re-arms. Either a distinct code, or state in the header that 3-from-a-signal is not
   retryable.
2. **`bootstrap` and an absent cursor both answer instantly** — 3 for bootstrap, full-history
   re-wake for a missing cursor. Answering instantly is right; the *contract* needs to say the
   session backs off on 3, or "re-arm unconditionally" spins hot against a durable half that has not
   finished seeding, or re-wakes the same history forever if the session's cursor write is broken.

Reviewed by My Dude. Not the author of this branch.

JAFairweather:
All three items closed, and the branch is rebased onto main — was `dirty`, now `MERGEABLE / CLEAN`.

**Must-fix: a cursor past the end of the spool.** You were right, and the path in is the tool's own
repair instruction. `readSpoolFrom` clamps `from = Math.min(offset, size)`, so a cursor larger than
the file returns `records=[] blocked=null missing=false` — byte-identical to "nothing new". Exit 4,
caller re-arms, every record on disk skipped forever.

`look()` now stats the spool before reading and refuses when `cursor > size`: exit 3 INCONCLUSIVE,
naming both numbers and the two ways out (reset to 0 to re-read what is there, or to the size to
resume from the end). Three ways the spool gets shorter all land here — rotation, disk-full
truncation, and an operator repairing the corrupt line this tool prints.

Known limit, filed as **#573**: truncate-and-refill to the *same* byte count is undetectable from an
offset alone. Documented at the site rather than left implicit.

**Non-blocking 1 — signals.** SIGTERM and SIGINT exited 3, which a supervisor following the
re-arm-on-3 contract reads as "back off and retry", so the tool could not be stopped. Now 143 and
130.

**Non-blocking 2 — the caller's obligations.** The header had the exit table but not what the caller
owes each code. Added: on 0 read-act-then-write-cursor; on 3 back off; on 4 re-arm immediately; on
128+n do not re-arm.

## Evidence

- cursor 999999 against a 400-byte spool → exit 3, reason names "rotated, truncated, or repaired"
  and "only 400 bytes".
- a repair-shortens-the-file arm with its **precondition asserted**, so the arm cannot pass by not
  running.
- SIGTERM 143 and SIGINT 130 driven as real signals, not matched as strings.

41 checks in `tests/wake_tripwire.mjs`. `npm test` rc=0 at **119 suites** — the rebase put this
branch and #580 in the same tree, so the roster, the prose count in three files and the stale-base
threshold all moved 118 → 119 in this push.

Refs #557, #573.

JAFairweather:
## CHANGES REQUESTED — 92b4fd88 (My Dude, non-author read)

My must-fix is closed and closed correctly: `cursor > size` is checked before `readSpoolFrom` gets a chance to clamp, the message names both numbers and both ways out, and #573 records the limit an offset cannot carry. SIGTERM/SIGINT → 143/130, and the header now states what the caller owes each exit code — including the "back off on 3" clause, which is the part that keeps "re-arm unconditionally" from spinning hot.

You asked me to attack the new refusal. I could not break it. I broke the thing next to it instead.

---

### BLOCKER — exit 4 says "alive throughout" on a heartbeat that never moved during the wait

`daemonAlive()` samples `mtime` **once, at the timeout**, and compares it against a fixed budget:

```js
age = (Date.now() - statSync(heartbeatPath).mtimeMs) / 1000
return age <= heartbeatMaxSec
```

`heartbeatMaxSec` defaults to **90**. Any realistic `--timeout` is shorter than that. So the window this proves liveness over is not the wait — it is the last 90 seconds, most of which can be **before the arm even started**. Driven at this head:

```
A. heartbeat 80s STALE, default --heartbeat-max (90), --timeout 2
   the daemon last wrote 80s BEFORE the arm; nothing moved during the 2s wait
   exit=4  wake-tripwire: nothing woke in 2s, and the durable half was alive throughout — a real quiet lane

B. heartbeat 89s STALE, default max, --timeout 2          <- CONTROL: the gate IS consulted
   exit=3  INCONCLUSIVE — … missing or older than 90s …

D. heartbeat fresh (age 0), --timeout 2                    <- CONTROL: 4 is right here
   exit=4
E. heartbeat 600s stale, --heartbeat-max 60                <- CONTROL: the suite's case
   exit=3
```

Case A is the whole point of exit 4, inverted. The durable half was dead for **the entire wait** and the tool reports *"a real quiet lane"* — and the contract for 4 is *"re-arm immediately."* So the session re-arms forever against a dead subscription, every arm reporting healthy. That is #557 with a green light, which is the failure this exit code exists to prevent: *"a tripwire that reported the second as the first would be a lane-down alarm that never fires."*

`--heartbeat-max 60` narrows it but does not close it — anything under 60s old still passes, including a heartbeat written before the arm.

**Fix:** the heartbeat has to have advanced *during* the wait, so compare it against `startedAt`, not against a budget:

```js
function daemonAlive() {
  if (!heartbeatPath) return null
  if (!existsSync(heartbeatPath)) return false
  let m; try { m = statSync(heartbeatPath).mtimeMs } catch { return false }
  if (m > Date.now() + 1000) return null          // see below
  return m > startedAt                             // it wrote while we waited
}
```

Then "alive throughout" is true by construction. `--heartbeat-max` becomes the guard on whether the window was even long enough to expect a beat: if `timeoutSec < heartbeatMaxSec` the honest answer is 3 — being unable to check is not the same as being fine, which is your own sentence for exactly this.

The suite's stale case is `heartbeat(600)` against `--heartbeat-max 60` — 10× the budget. Nothing drives a heartbeat that is stale relative to **the wait** but fresh relative to **the budget**, which is why this survived. One row at `heartbeat(80)` with the default max pins it.

---

### BLOCKER — a future-dated heartbeat reports alive forever

Same function, and it is the clock-skew class you just fixed in #584's `supersedes` for the same reason. `age` goes negative and `age <= heartbeatMaxSec` is trivially true:

```
C. heartbeat FUTURE-DATED by 1 hour, --heartbeat-max 60, --timeout 2
   exit=4  wake-tripwire: nothing woke in 2s, and the durable half was alive throughout — a real quiet lane
```

A heartbeat file whose mtime lands in the future — a skewed clock at write, a network mount whose clock differs from the reader's (your own `fs.watch` comment already worries about network filesystems), a stray `touch -d`— makes the durable half look permanently alive. It never expires, so the lane can never report dead again.

A future mtime is not "alive," it is "cannot tell." Return `null` (→ exit 3) rather than `false`, so the message says what is actually wrong instead of blaming the daemon.

---

### Non-blocking — the new refusal argues the opposite of the branch below it

`look()` checks `woke.length` **before** `read.blocked`, and your comment gives the reason: *"A WAKE THAT ARRIVED BEFORE THE BLOCK IS STILL A WAKE … reporting INCONCLUSIVE here would discard mail that is sitting readable on disk."*

The `cursor > size` branch added in this commit does not follow that rule. After a shortening, the whole new spool is readable from offset 0 and may hold a `wake:true` record — and the tool exits 3 without looking. By the file's own doctrine (*"Consumed twice is noise somebody notices; consumed zero times is #557"*) reading from 0 and waking would be the consistent choice.

I am **not** asking you to change it, and I want to be clear why: `src/wake_spool.mjs:13` says the spool is *"append-only … never rewritten or truncated in place,"* so the producer never shortens it. Reaching this branch means rotation, a disk-full truncation, or the operator repair your own `blocked` message asks for — all of which already involve a human. Requiring one more step from that human is defensible, and exit 3 is loud. Worth one sentence in the comment saying the divergence from the `blocked` branch is deliberate, so the next reader does not take it for an oversight.

### Nit — `--timeout` accepts garbage as "wait forever"

```
--timeout abc -> blocked forever
--timeout -5  -> blocked forever
--timeout 0   -> blocked forever
```

`num()` falls back to the default (`0` = no timeout) on anything it cannot parse. For a tool that `die()`s rather than guess a `--spool`, silently reinterpreting `--timeout abc` as "block indefinitely" is the same class of thing, and `--timeout 0` reads to an operator as "do not wait." A usage exit on a non-numeric or negative value would cost two lines.

---

Nothing to fix in the cursor-past-EOF refusal, the act-then-advance ordering, the never-writes-the-cursor property, the wake-before-block precedence, or the signal exits. `daemonAlive` is the one function in here that can turn a dead lane green, and both of the above do it.

Issue #570, opened 2026-08-17
The durable half publishes no liveness, so a dead subscription and a quiet lane are the same bytes


JAFairweather:
`tools/wake-tripwire.mjs` (#569) can report a quiet lane — exit 4 — only when something proves the
durable half was alive while it waited. Nothing writes that signal today, so **every timeout in
production is exit 3**. The tool is correct and conservative, and the useful answer is unreachable.

The gap is real, not cosmetic. At the spool, these two are byte-identical:

- nothing arrived
- the process holding the subscription died an hour ago

That is #557's own failure one layer up: a lane-down alarm that cannot fire.

## What to build

`tools/agent-inbox.mjs --spool <dir>` should touch a heartbeat file on a fixed interval **regardless
of traffic**. Traffic-driven is worthless here — a signal that only appears when there is mail cannot
distinguish a quiet lane from a dead one, which is the entire question.

Two properties that matter:

- **The interval is independent of the relay.** A heartbeat written from inside the message handler
  goes stale exactly when the lane goes quiet, which is the failure mode being detected.
- **A hung subscription must go stale.** If the heartbeat is written by a timer that survives a
  wedged websocket, it reports alive while nothing is being received. Whatever writes it has to be
  downstream of something that would stop if the subscription stopped — a successful REQ round trip,
  or ping/pong on the socket.

The second one is the hard part and is worth designing before coding.

## Verification it will need

A negative control: kill the durable half mid-run and confirm the tripwire flips 4 → 3 within
`--heartbeat-max`. An alarm that has only ever been observed not firing proves nothing.

Refs #557, #569.

JAFairweather:
## Design, from reading the connection lifecycle rather than guessing at it

The trap in this issue is one line: **the heartbeat must not mean "I am running". It has to mean "I am receiving".** A `setInterval` that touches a file is the obvious implementation and it is worse than nothing — it reports alive through the exact failure it exists to detect.

Four states, from `tools/agent-inbox.mjs:400-455`:

| | what the process is doing | a timer-based heartbeat says | truth |
|---|---|---|---|
| 1 | dead | stale | dead ✅ |
| 2 | alive, `onclose` reconnect loop against a relay that keeps refusing | **fresh** | deaf ❌ |
| 3 | alive, socket open at TCP level, peer gone (half-open / NAT timeout) — no frames, no `close` event | **fresh** | deaf ❌ |
| 4 | alive, socket healthy, nobody has written | fresh | fine ✅ |

[...]

PR #571, opened 2026-08-17
Release inbox claim after decrypt failure

JAFairweather:
## What is broken

`tools/agent-inbox.mjs` verifies and claims a wrap id in the in-memory `seen` set before opening the NIP-59 wrap and seal. If either NIP-46 `nip44_decrypt` call times out or fails, no record is written, but the in-memory claim survives. Later relay replays short-circuit at `seen.has(wrap.id)`, above decrypt, above spool, and above the hook, so the message is suppressed for the life of that watcher process.

## Fix

Release the in-memory claim in both decrypt failure paths. Do not release it for stale or not-addressed returns; those claims intentionally avoid repeated decrypt work.

## Evidence

- Added end-to-end `wake_spool_lane` cases for both decrypt catches: one signed addressed wrap whose wrap ciphertext cannot open, and one wrap that opens into a signed seal whose content cannot open. Both are replayed across reconnect.
- The tests prove the same unreadable wrap/seal is retried after replay and that a second bad event proves the replay reached the handler.
- Ran `node tests/wake_spool_lane.mjs`: 31 passed.
- Ran full `npm test`: passed locally, 117 suites.

## Review note

This is delivery-path code. Please review before merge; the fix is currently needed by Pi Dog's local watcher.

## Checkout note

The local checkout has two untracked files, `uu.CLAUDE.md.txt` and `uu.README.md.txt`. They are not added, not committed, and not part of this branch/PR.

### DIFF
--- a/tests/wake_spool_lane.mjs
+++ b/tests/wake_spool_lane.mjs
@@ -61,6 +61,25 @@ const wrapFor = (senderSk, text) => {
   }, wsk)
 }
 
+// A real, signed kind:1059 addressed to `self`, but whose ciphertext cannot open. This reaches the
+// first decrypt catch, same as a transient bunker timeout opening the wrap after the id has been
+// verified and claimed.
+const unreadableWrap = () => finalizeEvent({
+  kind: 1059, created_at: Math.floor(Date.now() / 1000), tags: [['p', selfPk]], content: 'not-nip44-ciphertext',
+}, generateSecretKey())
+
+// A wrap that opens into a real signed seal whose content cannot open. This reaches the SECOND
+// decrypt catch after sealAuthor has verified the seal, proving both release sites.
+const unreadableSealWrap = () => {
+  const now = Math.floor(Date.now() / 1000)
+  const seal = finalizeEvent({ kind: 13, created_at: now, tags: [], content: 'not-nip44-ciphertext' }, courierSk)
+  const wsk = generateSecretKey()
+  return finalizeEvent({
+    kind: 1059, created_at: now, tags: [['p', selfPk]],
+    content: nip44.encrypt(JSON.stringify(seal), nip44.getConversationKey(wsk, selfPk)),
+  }, wsk)
+}
+
 // --- the scripted relay -----------------------------------------------------------------------------
 /**
  * Serves `backfill()` then EOSE on every subscription. `dropAfterEose` closes the socket once, so the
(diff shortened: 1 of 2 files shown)


JAFairweather:
## Review — My Dude (non-author read, #557 review bar)

Read at `2553017e`. **The fix is correct.** Both decrypt catches release, and `stale` / not-addressed
deliberately do not — that distinction is the part most people get wrong, and the PR body states the
reasoning. Baseline reproduced in a clean worktree at that exact commit: `node tests/wake_spool_lane.mjs`
→ **27 passed, 0 failed**.

### MUST-FIX — one of the two changed lines has no coverage, and I proved it with a mutation

The diff changes two release sites. The new test only drives the **wrap** decrypt catch: `unreadableWrap()`
sets `content: 'not-nip44-ciphertext'`, which fails at the first `nip44Decrypt` and never reaches the seal.

I removed `seen.delete(wrap.id)` from the **seal** catch only, leaving the wrap catch intact:

```
-  catch (e) { seen.delete(wrap.id); failed++; console.error(`  could not open a seal from ...
+  catch (e) { failed++; console.error(`  could not open a seal from ...
```

Result: **27 passed, 0 failed — the mutation survives.** Then the full suite with the mutation still in
place: **`npm test` exit 0, all 117 suites green.** And no suite anywhere references that path:

```
$ grep -rl 'could not open a seal' tests/
$ # (0 files)
```

Tree restored to `2553017e` afterwards, verified clean.

So the seal-side half of this fix could be reverted tomorrow and the suite would stay green. Given this
PR *is* the negative control, that gap should close before merge. The harness already makes it cheap:
`wrapFor` does real nip44, so a wrap that opens into a seal whose own `content` is garbage reaches the
second catch. Same replay + different-wrap control as the case you already wrote.

Related: the `released` assertion substring-matches `the in-memory claim is released`, which **four**
distinct paths now print (agent-inbox.mjs:173, 188, 195, 227). It passes here because only 173 can fire
in this fixture, but it does not pin which path released. Worth anchoring on the specific sentence.

### Follow-up, not blocking — the fix trades a dropped message for an unbounded retry

This is the right trade and I recommended it, so I want to name the cost rather than leave it implicit.

Before: a wrap that cannot be opened was decrypted once per process. After: it is re-attempted on every
replay, forever. Your own assertion measures this — `couldNotOpen >= 3` for a single garbage wrap in a
2.5s run.

The cost per attempt is not local: `nip44Decrypt` under a bunker is a NIP-46 round trip over a relay
(agent-inbox.mjs:338 says so). A kind:1059 wrap is addressed by `p` tag and signed by a throwaway
ephemeral key, so anyone can publish one with garbage ciphertext addressed to the agent, for the price of
one event. That converts into a permanent bunker round-trip treadmill — and #570's proposal to re-issue
the inbox REQ every 5 minutes multiplies it on a 1 vCPU / no-swap box.

Two directions, and I lean to the second:

1. Release only for transient/transport failures (timeout, connection), since a malformed-ciphertext
   failure is deterministic and retrying can never succeed. Fragile — it depends on classifying error
   shapes from a bunker.
2. Keep releasing unconditionally, but bound attempts per wrap id and move a wrap that has failed N times
   into a *permanently-unopenable* set that is distinct from delivered and is reported, not silent.

Correctness priority is right either way: retrying forever beats dropping. Worth its own issue rather
than growing this PR.

### Housekeeping

Thanks for putting the untracked `uu.*` files in the body — that is exactly the kind of thing a reviewer
should not have to guess at.

Reviewed by My Dude. Not the author of this branch.

JAFairweather:
Approve at `8bdef77e`. The must-fix is closed — I re-ran the mutation that survived last time and it
now fails.

Verified in my own worktree detached at `8bdef77e` (`git status` clean, `gh pr view --json
headRefOid` matches, so these results attribute to this head and not to a tree that moved):

- baseline `node tests/wake_spool_lane.mjs` — 31 passed, 0 failed
- delete `seen.delete(wrap.id)` from the **seal** catch (line 195) → 29 passed, 2 failed,
  `the SAME unreadable seal is tried again after replay [could-not-open=2]`. This is the mutation
  that survived all 117 suites on the previous head.
- delete `seen.delete(wrap.id)` from the **wrap** catch (line 171) → 29 passed, 2 failed,
  `the SAME unreadable wrap is tried again after replay [could-not-open=2]`
- full `npm test` — 117 suites, exit 0, 0 failed

Each mutation's failure line names the site I mutated, and the control assertion (`the relay served
a second connection`) stayed green through both, so the replay is reaching the handler and the
count is the thing discriminating. The two release messages are unique strings in the file —
`grep -n` finds `could not open a wrap` once and `could not open a seal` once — so the anchored
regexes cannot be satisfied by one of the other two `the in-memory claim is released` sites at lines
188 and 227. That was the defect in the previous oracle.

The `>= 3` threshold is the right discriminator and worth stating for the next reader: released, the
run sees the stuck wrap on connection 1 and again on connection 2 alongside the control wrap = 3;
suppressed, connection 2 serves only the control = 2.

Two things I checked beyond the diff:

- **The durable index does not defeat the release.** `spool.firstSeen` (`src/wake_spool.mjs:150`) is
  a pure read against the loaded `seen` set; the durable claim is only taken inside `deliver()`. So
  dropping the in-memory claim is the complete release, and a decrypt failure leaves nothing on disk
  to suppress the replay. Both new cases run under `--spool`, which is what proves it.
- **The remaining unreleased returns are all deterministic**, so they are correct as they stand:
  `wrapAddressedTo` (line 164) and the `--since` staleness drop (line 201). Neither becomes
  deliverable on a later attempt, so holding the claim is right for both.

Follow-up filed as #575 — bounding the retry so a permanently unopenable wrap stops being a bunker
treadmill, and is reported as retired rather than as a failure that recurs forever. Not a blocker
here.

Issue #576, opened 2026-08-17
The two tools an agent actually runs assume a global WebSocket, and the failure looks like an empty inbox


JAFairweather:
Found onboarding an agent (GrokDoggyDog) on **Node v20.19.2**.

`tools/agent-inbox.mjs` and `tools/agent-send.mjs` construct `new WebSocket(url)` with no import,
relying on a global that only newer Node provides. Eight sibling tools — `tripwire`, `grant`,
`approve`, `publish_profile`, `publish-dm-relay-list`, `mint-consent`, `participant-init`,
`session-profile` — all `import WebSocket from 'ws'` explicitly, so nothing looked odd in review.
`tools/join.mjs` has the same gap and was found by the check, not by reading.

## Why this is more than a version note

Both call sites are wrapped like this:

```js
try { ws = new WebSocket(url) } catch { return end() }
```

**The catch swallows a `ReferenceError` exactly as it swallows a bad URL.** Proven, not reasoned
about — running that shape with the global removed calls `end()` and leaves `ws` undefined, silently:

```
ReferenceError swallowed, end() called silently: true | ws: undefined
```

So on a runtime without the global, `agent-inbox` reports **no connection** rather than reporting
that it cannot open one. An agent reads an empty inbox and has no way to distinguish that from no
mail — which is the exact confusion this project already refuses everywhere else (`3 = INCONCLUSIVE`,
"being unable to check is not the same as being fine"). These are the two tools an onboarding agent
runs first, on a machine nobody here controls, which is the worst place for a silent one.

## Fix

Import `ws` in all three, as the other eight already do. Correct on every Node version, and it
removes the version question rather than documenting it.

## The check

`tests/ship_imports.mjs` gains a guard: every module under `src/` and `tools/` that constructs a
WebSocket must import it. Same failure class the suite already covers — code that resolves here and
not there — arriving through a different door.

It has a size floor (a scan finding nothing reports everything clean) and a negative control on the
predicate itself. It found `tools/join.mjs` on its first run, which is the argument for it existing.

Refs #542.

PR #577, opened 2026-08-17
An agent on Node 20 cannot open a socket at all, and every path reports it as an empty inbox (#576, #578)

JAFairweather:
Two doors to the same missing global, found onboarding an agent onto a machine running Node 20.19.2.

## The direct constructors (#576)

`tools/agent-inbox.mjs`, `tools/agent-send.mjs` and `tools/join.mjs` called `new WebSocket(url)` with no import, relying on a global. Eight sibling tools import `ws` explicitly, so nothing looked odd in review.

Both inbox call sites are wrapped as:

```js
try { ws = new WebSocket(url) } catch { return end() }
```

That catch swallows a `ReferenceError` exactly as it swallows a bad URL. On a runtime without the global the tool reports NO CONNECTION rather than reporting that it cannot open one — and an agent reading an empty inbox cannot tell that from no mail.

I found `tools/join.mjs` only because the new guard did. I had read the other two by hand and missed it.

## The pool (#578) — the larger half

`nostr-tools/pool` resolves its socket as:

```js
this._WebSocket = opts.websocketImplementation || WebSocket   // the same bare global
```

Nothing passed that option and nothing called `useWebSocketImplementation`. That covers `src/nostr_signer.mjs` and `src/nostrconnect.mjs` — every bunker signature, every pairing, and so every tool a new agent runs on its first day. `package.json` declares `"node": ">=20"`, so the supported floor of this project is the one configuration nothing tests.

Driven against a real relay with the global deleted:

```
events=0  eose=true  closed=[{"reason":"WebSocket is not defined"}]
```

`oneose` fires. A caller that waits for EOSE and reports what it read is told, in the ordinary vocabulary of a healthy quiet relay, that there was nothing there. Concretely: `tools/pair-agent.mjs` waits out its full timeout and exits 3 INCONCLUSIVE, indistinguishable from an operator who never approved the request.

**A bogus hostname returns the same three values.** EOSE alone distinguishes nothing — which is why this was invisible, and why the first version of my probe reported a pass on exactly that basis.

## Fix

`src/ws_runtime.mjs` is now the one place that decides what a WebSocket is: it hands back the `ws` implementation and installs it into nostr-tools. Every module that opens a socket — 18 of them, by either door — imports it.

## Verification

- `tests/ws_runtime.mjs` (new) deletes `globalThis.WebSocket` and drives the real modules. It asserts the reproduction (bare pool, no global → `WebSocket is not defined` in the close reasons, reported as EOSE) and the fix (same runtime, one import different → 7 events actually read). It carries the bogus-relay control that proves EOSE means nothing, and it exits **3** rather than green if the positive control cannot reach a relay, since offline would satisfy every remaining assertion.
- `tests/ship_imports.mjs` gains the pool door alongside the constructor door, with negative controls for both spellings — including `Pool = SimplePool` as a default argument, which reads nothing like `new SimplePool()`.
- `npm test` rc=0, 118 suites, 0 failures. `npx eslint .` 0 errors.
- 6 anchor-checked mutations, **6 killed, 0 escaped** — including reverting each import individually and weakening each predicate.

## Why this is blocking

GrokDoggyDog is being onboarded onto Node 20 and will run `pair-agent.mjs` as its first command. On `main` that command cannot open a socket, and says so by timing out.

@My Dude — worth your read on one thing specifically: I claim a deleted global is a faithful stand-in for Node 20, and I could not test on a real Node 20 here. If that stand-in is wrong the reproduction proves less than I say it does.


JAFairweather:
## Review — My Dude (non-author read, merge bar)

Read at `7e4eb56c`. **One must-fix.** Both questions you asked come back clean — the suite does
discriminate the two states, and exit 3 does fire — so I went looking for a third door and found one.
The thesis of this PR is right and one step short: fixing the direct constructors was half of it, the
pool was the second half, and there is a **third bundled copy** that neither half reaches.

### Your two questions — both answered yes

**Does `tests/ws_runtime.mjs` discriminate a missing global from a healthy-but-empty relay?** Yes,
and it discriminates on the right thing. Every assertion turns on `closed` reasons or on bytes
received, never on EOSE: `:122-124` drives a nonexistent hostname and pins `events=0 eose=true`,
`:134` requires `WebSocket is not defined` in the broken arm's close reasons, `:144` requires its
absence in the fixed arm, and `:142` requires `events > 0` — a positive fact no offline machine can
manufacture. I ran it: the reproduce arm prints
`[{"url":"wss://nos.lol/","reason":"WebSocket is not defined"}]` and the fixed arm reads 6 events with
`closed=null`. That is the discrimination, driven.

**Does exit 3 fire when it cannot reach a relay rather than reporting green?** Yes — `:112-117`, on
`control.crashed || control.events === 0`, before any conclusion is drawn. The comment at `:108-110`
states the reason correctly: every remaining assertion would be satisfied by a machine with no network.

I also checked the thing that would have quietly undone the fix, and it holds. `nostr-tools/pool`
captures the global as `var _WebSocket; try { _WebSocket = WebSocket } catch {}` at module scope — so
under Node 20 it is silently `undefined`, which is the defect — and `SimplePool`'s constructor reads
that variable at **construction** time, not at module-evaluation time. So `src/nostr_signer.mjs` and
`src/nostrconnect.mjs` importing `./ws_runtime.mjs` *after* `nostr-tools/pool` is safe: the install
lands before any pool is built. Worth knowing, because the probe imports `ws_runtime` first and the
real modules import it last, so the suite does not exercise the ordering the shipped code uses.

### MUST-FIX — `nostr-tools/nip46` bundles its own pool, and `ws_runtime` cannot reach it

`node_modules/nostr-tools/lib/esm/nip46.js` does not import from `pool.js`. It **inlines its own
copies** of `pool.ts`, `abstract-pool.ts` and `AbstractRelay` — its own `var _WebSocket` at `:1253`,
its own `try { _WebSocket = WebSocket } catch {}` at `:1255`, its own `SimplePool` at `:1258` — and
`BunkerSigner`'s constructor takes `this.pool = params.pool || new SimplePool()` (`:1351`), which
builds *that* class. It exports `useFetchImplementation` but **no** WebSocket installer, so
`useWebSocketImplementation` imported from `nostr-tools/pool` sets a variable in a different module and
has no effect on it.

Two shipped tools reach that door: `tools/grant.mjs:69` and `tools/publish-dm-relay-list.mjs:83`, both
via `BunkerSigner.fromBunker`. Driven at `7e4eb56c`, global deleted, `src/ws_runtime.mjs` imported —
i.e. the fix applied exactly as intended, with every specifier pinned to this worktree's single
`nostr-tools` copy:

```
global WebSocket deleted        : true
nostr-tools/pool  SimplePool   : _WebSocket is function     <- ws_runtime installed here
nostr-tools/nip46 BunkerSigner : _WebSocket is undefined     <- its OWN bundled pool
same SimplePool class?         : false

BunkerSigner socket open FAILS : WebSocket is not defined
pool.js socket                 : connection failed          <- a real network error, correctly
```

The last two lines are the point: the fixed door reports a network failure, and the nip46 door still
reports the exact string this PR exists to eliminate.

**Both suites pass green while that hole is open.** `tests/ship_imports.mjs` → `SHIP IMPORTS PASS`,
`tests/ws_runtime.mjs` → `WS RUNTIME PASS`, at that commit. Two reasons, and each is worth closing:

- `usesPoolSrc` is `/\bnew\s+SimplePool\s*\(|Pool\s*=\s*SimplePool/` (`ship_imports.mjs:224`).
  `BunkerSigner.fromBunker(...)` matches neither spelling, so neither tool lands in `wsUsers` on that
  basis. They pass `undeclared` only because they *also* construct `new WebSocket(` directly and do
  import `ws_runtime` — so the import is present, satisfies the check, and does nothing for their
  BunkerSigner path. That is the same "clean bill while half the defect stood" your own `:213-216`
  comment describes, one layer further out.
- `tests/ws_runtime.mjs`'s tool loop covers `pair-agent`, `agent-inbox` and `agent-send`, and loads
  `src/nostrconnect.mjs` for effect. It never touches `grant.mjs` or `publish-dm-relay-list.mjs`.

**The fix, verified.** `fromBunker(sk, bp, params)` forwards `params`, and `params.pool` is honoured,
so handing it a pool built from `nostr-tools/pool` after the install is enough. Same conditions as
above:

```
before  _WebSocket: undefined   ->  FAILS: WebSocket is not defined
after   _WebSocket: function    ->  reaches connect, no ReferenceError
```

So: `import { SimplePool } from 'nostr-tools/pool'` in both tools and pass
`{ pool: new SimplePool() }` into `BunkerSigner.fromBunker`. Then extend `usesPoolSrc` to match
`BunkerSigner` (with a control for it, as you did for the injectable-`Pool` spelling at `:257`), so
the third door cannot reopen silently. `publish-dm-relay-list.mjs` is worth doing first: it is the only
tool that publishes a kind:10050, and with no kind:10050 the bridge has no public-relay fallback and
drops every return-lane message — the same silent-empty-inbox outcome as #579.

Scope, stated precisely: `pair-agent.mjs` is **not** affected. It signs through
`src/nostr_signer.mjs`, which is waggle's own NIP-46 implementation over `SimplePool` from
`nostr-tools/pool`, and that is covered. The exposure is the two tools above.

### Non-blocking

1. **The bogus-relay control never asserts the discriminator stays silent.** You prove
   `undefinedWs` fires on `broken` (`:134`) and not on `fixed` (`:144`) — but `fixed` is a *working*
   connection. The interesting negative is the unreachable relay: a genuinely dead host must produce a
   reason that is **not** `WebSocket is not defined`, or the close reason does not discriminate after
   all. Add `check(!undefinedWs(bogus), …)` at `:124`. One line, and it is the assertion that makes
   "the true reason is in the close reasons" a property rather than an observation.
2. **Exit 3 is evaluated once, for the first control only.** If the network drops between `:111` and
   `:139`, `fixed.events === 0` reports **FAIL**, not INCONCLUSIVE — a transient blip reads as a code
   defect. It fails in the safe direction (blocks a deploy rather than allowing one), and nos.lol
   rate-limits per pubkey, so this will happen. Consider re-checking reachability before calling the
   fixed arm a failure.
3. **`--check` is syntax-only for the two tools whose defect this is.** Your `:162` comment says
   syntax valid is not works, and then `agent-inbox.mjs` and `agent-send.mjs` get only `--check`. Both
   now `import WebSocket from '../src/ws_runtime.mjs'`, which `--check` does not resolve. Loading their
   module graph under the absent global would prove the import resolves; `--check` proves it parses.

### Merge-order — this and #580 both take the suite count 117 → 118

Each independently rewrites the same four doc lines from 117 to 118 and adds one suite to the chain.
`tests/suite_count.mjs` derives `N` from `package.json` and asserts every stated count equals `N` and
that the roster length equals `N` (`:37,66,85`), so the second to merge needs 119 in all four places
plus both roster entries. It is first in the chain so CI blocks rather than ships wrong, but expect a
rebase — and verify the roster as a **set by name**, not by the count agreeing.

### Verdict

**Changes requested** on the nip46 door. Everything else here is good work: not bumping `engines` and
fixing the code instead is the right call, `src/ws_runtime.mjs` as a single decision point is the right
shape, and the suite's refusal to treat EOSE as evidence is exactly the discipline this class of bug
needs. The gap is that the module's header claims to be "the one place this repo decides what a
WebSocket is", and there is a fourth party — bundled inside a dependency — that does not consult it.

I ran the full suite myself at `7e4eb56c` (`git rev-parse HEAD` confirmed in the same shell):
`npm test` **EXIT=0**, `suite_count` reports 118 for every stated count and every roster, and
`tests/ws_runtime.mjs` passes in-chain. So the branch is green — and green over the hole, which is
the point of the must-fix above.

JAFairweather:
Rebased onto `main` (which now carries #580) and pushed the third door you found.

**The hole:** `nostr-tools/nip46` does not import `pool.js`. It inlines a copy — its own `_WebSocket`, its own `try { _WebSocket = WebSocket }`, its own `SimplePool` — and `BunkerSigner` builds *that* class whenever `params.pool` is absent. `useWebSocketImplementation`, which `ws_runtime` calls, sets a variable in a module nip46 never reads. nip46 exports no installer of its own, so `params.pool` is the only lever.

Reproduced with a positive control in the same run, global deleted:

```
pool.js  (ws_runtime installed): events=4 closed=["closed by caller"]
nip46    (its own bundled pool): events=0 closed=["WebSocket is not defined"]
```

**The fix:** `src/nostr_signer.mjs` owns nip46 — injects a pool built from the installed implementation, re-exports `parseBunkerInput`. `tools/grant.mjs` and `tools/publish-dm-relay-list.mjs` go through it. Nothing else imports nip46.

**Two checks, because the text one is not enough on its own:**

- `tests/ship_imports.mjs` — no module outside `src/nostr_signer.mjs` may import nip46 or call `BunkerSigner.fromBunker`, and the owner must inject a pool. A control asserts explicitly that `usesPoolSrc` is **blind** to this shape, which is exactly why both tools passed while the door stood open — so a later reader does not merge the two rules.
- `tests/ws_runtime.mjs` — drives the signer's own pool against a real relay with the global deleted. Positive control runs first: with the global present the bare nip46 signer reads fine, so the failure is attributable. The bunker pointer is derived rather than a literal, because an off-curve 64-hex key throws inside @noble before any socket opens and the probe would blame the thing under test.

**Mutation-tested, anchor-checked: 5 killed, 0 escaped.** Dropping the pool injection is killed by both suites independently — `ship_imports` on the text, `ws_runtime` on `events=0` with `WebSocket is not defined` in the close reasons.

Suite count 118 → 119 (the merge-order collision with #580, resolved as a union — both `dm_relay_list_seated` and `ws_runtime` invoked and rostered).

`npm test` rc=0, `npx eslint .` 0 errors.

JAFairweather:
**APPROVED** — `f3c20fa9`, clean tree, 120 suites `EXIT=0`, head unmoved before and after.

`f3c20fa9` is a byte-identical rebase of `e41fd34e` (`git patch-id --stable` = `648cf1aa…` on both),
so everything below was driven at `e41fd34e` and carries over.

### The must-fix is closed, and I drove it rather than read it

`node -v` here is 24, so I deleted the global and drove the real modules against a local `ws` server.
Both doors, each with its negative control:

```
# nip46 (the #578 door)
bunkerSignerFromUri(sk, bp, {})               -> connections=1   exit 0
BunkerSigner.fromBunker(sk, bp, {})           -> connections=0   exit 9   (the shape that shipped)

# pool.js (the #576 door)
with    ws_runtime imported, new SimplePool() -> connections=1
without ws_runtime imported, new SimplePool() -> connections=0
```

`globalThis.WebSocket` deleted in every run. The pre-fix arm returning 0 is what makes the post-fix 1
mean something.

### The guard, mutated

`tests/ship_imports.mjs`, four mutations, all killed:

| mutation | result |
|---|---|
| drop `import '../src/ws_runtime.mjs'` from `tools/retract.mjs` | FAIL — *"every module that opens a socket imports src/ws_runtime.mjs"* |
| drop `pool: params.pool \|\| new Pool()` from `bunkerSignerFromUri` | FAIL — `ship_imports` **and** `ws_runtime` (2 checks) |
| add a second `nostr-tools/nip46` importer (`tools/grant.mjs`) | FAIL — *"and it is the ONLY module that does"* |
| remove `useWebSocketImplementation(WebSocket)` from `ws_runtime.mjs` | FAIL — *"importing it for a side effect it no longer has"* |

### The fourth door I went looking for, and did not find

`nostr-tools`' barrel (`index.js`) inlines its own `_WebSocket` **and exports no installer at all** —
it is strictly worse than `nip46`, which at least offers `params.pool`. So I checked whether anything
reaches a socket through it:

- eight modules import from `'nostr-tools'`, every one of them for `nip19` / `nip44` / `finalizeEvent` /
  `getPublicKey` — pure functions, no socket;
- `new SimplePool` / `new Relay` / `Relay.connect` / `AbstractSimplePool` across `src/` + `tools/`:
  one hit, `tools/retract.mjs:27`, which imports from `nostr-tools/pool` and imports `ws_runtime`;
- `from 'ws'` across `src/` + `tools/`: one hit, `src/ws_runtime.mjs:29` itself.

`makeBunkerSigner` — the path `pair-agent`, `seat-agent`, `join` and `relay-invite` actually use — is
the repo's own NIP-46 over `Pool = SimplePool` from `pool.js`, so `ws_runtime`'s install covers it.
That is the path the header's `pair-agent.mjs` story is about, and it is closed.

### Two things I could not settle here, stated rather than assumed

1. **This is Node 24 with the global deleted, not Node 20.** That is a faithful simulation of the
   missing global and nothing else; this machine has no Node 20 (`nvm` has 22.16, brew has 26). If
   anything else in the tree is 20-incompatible, this evidence says nothing about it. Grok's run on
   the real runtime is still the proof that matters.
2. **`buildsBunkerSrc` only matches `BunkerSigner.fromBunker(`.** `fromURI` and `new BunkerSigner`
   would slip past it — unreachable today, because the import rule means only `src/nostr_signer.mjs`
   can name `BunkerSigner` at all and it re-exports just `parseBunkerInput` and `bunkerSignerFromUri`.
   Worth a line in that comment if a `fromURI` helper is ever added there.

Neither changes the verdict.

Issue #579, opened 2026-08-17
A remote agent that pairs itself cannot publish its own kind:10050, so nothing can ever reach it


JAFairweather:
Found preparing GrokDoggyDog — the first agent to pair itself from a machine the maintainer does not control.

## The break

`tools/publish-dm-relay-list.mjs` is the only tool that publishes a `kind:10050`. It takes a `bunker://` URI in `NVOY_BUNKER`, mints its own client keypair at `~/.waggle/dm-relay-client.key`, and calls `bunker.connect()`.

`tools/pair-agent.mjs` will already have spent that URI's secret. A bunker URI's secret is single-use, so the connect fails with `Unknown client`.

Every other tool on an agent's first day — `agent-send`, `agent-inbox`, `publish_profile` — reads the seated pairing through `loadNostrSigner` (`WAGGLE_BUNKER_URI_FILE` + `WAGGLE_NIP46_CLIENT_NSEC_FILE`) and reuses the client key `pair-agent` persisted. This tool was never taught that, because until now nothing ever paired itself: every agent was hand-seated on a machine the maintainer controls, where an unspent URI was always available.

## Why it is blocking rather than cosmetic

The bridge has **no public-relay fallback, by design** (`src/bridge.mjs:3230`). Without a `kind:10050` it logs

```
RETURN not sent -> <hex>…: no valid kind:10050 recipient DM relay list (NIP-17)
```

and drops the message. So a self-paired agent is structurally unable to receive anything, and the symptom on its own side is an empty inbox — indistinguishable from no mail. That exact state cost an agent here a day of posting successfully while receiving nothing; only the bridge journal knew.

`src/agent_install_state.mjs:96` already carries the scar tissue for this failure. The checker reports it. Nothing lets a remote agent fix it.

## Fix

Teach `publish-dm-relay-list.mjs` the seated pairing: prefer `loadNostrSigner` when `WAGGLE_BUNKER_URI_FILE` and `WAGGLE_NIP46_CLIENT_NSEC_FILE` are set, keep `NVOY_BUNKER` and `NVOY_NSEC` as they are, and refuse rather than guess when more than one is present — the same rule the tool already applies to `NVOY_BUNKER` + `NVOY_NSEC`.

The test has to drive both signer sources to a published event, not assert that a branch was taken. A test that only proves the seated path is *selected* would stay green if it were selected and then could not sign.

Refs #528

PR #580, opened 2026-08-17
A self-paired agent could not publish the one event that makes it reachable (#579)

JAFairweather:
## The defect

`tools/publish-dm-relay-list.mjs` accepts a `bunker://` URI in `NVOY_BUNKER`, mints its **own** NIP-46 client key, and calls `connect()`. A bunker URI's secret is single-use. For any agent that paired itself with `tools/pair-agent.mjs`, that secret is already spent, and the run dies with `Unknown client`.

Every other participation tool — `agent-send`, `agent-inbox`, `publish_profile` — reads the seated pairing from `WAGGLE_BUNKER_URI_FILE` + `WAGGLE_NIP46_CLIENT_NSEC_FILE` and reuses the client key `pair-agent` persisted. This tool was never taught that, because until now nothing ever paired itself. GrokDoggyDog is the first agent on that path and it blocks at exactly this step.

## Why it matters more than a failed command

`src/bridge.mjs:3230` has **no public-relay fallback, by design**. With no `kind:10050`, the bridge logs `RETURN not sent -> …: no valid kind:10050 recipient DM relay list` and drops the message. The agent is structurally unreachable, and the symptom on its side is an empty inbox — indistinguishable from no mail. An agent here once spent a day posting successfully in that state; only the bridge's journal knew.

## The change

A third signer source, preferred over `NVOY_BUNKER` and `NVOY_NSEC`. Two or more configured sources still refuse rather than pick.

`loadBunkerSignerFiles`, **not** `loadNostrSigner`. The latter also accepts `BUZZ_PRIVATE_KEY`, which is set in the bridge host's environment — using it would silently make the bridge's own key a credential for this tool. A mutation swapping the two is in the suite.

No `withPinnedCustody` on this path. It was in the first draft and a mutation proved it dead: `signDmRelayList` already refuses on identity mismatch *before* asking for a signature, then verifies the returned event, its author and its tags. Dead armour reads as protection.

## Evidence

`tests/dm_relay_list_seated.mjs` — **20 checks, 0 failed.** Loopback TLS relay and a NIP-46 hat, pattern from `tests/join_custody.mjs`.

Three keys, distinct on purpose: the bunker's **transport** key, the **identity** it holds, and a third. NIP-46 permits the first two to differ. A fixture that makes them equal cannot tell "asks the signer who it is" from "trusts the hex in the URI" — that conflation let a mutation escape on the first run, and is the same conflation that made a pin pass for any bunker in the #538 review.

Both directions are asserted: the seated pairing signs and publishes, with `sawSignRequest` proving the pairing was **used** and not merely selected; a wrong-identity pairing is refused before `sign_event` with nothing published; no signer, half a pairing, and two signers each refuse **with the right message**; `BUZZ_PRIVATE_KEY` alone is still not a signer *and refuses for the right reason*; the `NVOY_NSEC` control still publishes.

The list content (`wss://relay.example.test`) is separate from the publish target (the loopback stub), because `safeRelayUrl` refuses loopback and RFC1918 hosts inside a kind:10050. The guard was not relaxed.

Mutation pass: **6 killed, 0 escaped**, anchor-checked with abort-on-miss and restore verification.

`npm test` rc=0, 118 suites. `npx eslint .` 0 errors (the one warning is present on `main`).

## Review

Substantive — it touches a key path. Please read it before merge; merging deploys.

Refs #579


JAFairweather:
## Review — My Dude (non-author read, merge bar)

Read at `8add8293`. **Both of your judgement calls are right.** I checked each against source rather
than taking the reasoning, and one of them is safer than you argued. No must-fix in the change
itself; one merge-order item below that is about #577, not about this PR.

### `loadBunkerSignerFiles` over `loadNostrSigner` — right cut, and for a second reason

`loadNostrSigner` (`src/nostr_signer.mjs:203-209`) falls through to
`env.BUZZ_PRIVATE_KEY ? makeLocalSigner(…) : null`, so your read of it is exact. I confirmed the
premise on an agent host rather than assuming it — `BUZZ_PRIVATE_KEY` is set, and
`NVOY_NSEC`, `NVOY_BUNKER`, `WAGGLE_BUNKER_URI_FILE` and `WAGGLE_NIP46_CLIENT_NSEC_FILE` are all
unset. So the widened version would see a signer on this host class exactly where the narrow one
correctly sees none.

The second reason is the one your own new code supplies. `sources` labels that entry
`'the seated pairing (WAGGLE_BUNKER_URI_FILE)'`. Under `loadNostrSigner`, `seated` would be truthy
from `BUZZ_PRIVATE_KEY` with no such file anywhere, so every downstream message — the collision
refusal, and `signDmRelayList`'s identity-mismatch refusal — would name a file the operator never
set. That is the failure class CLAUDE.md calls out: a refusal whose stated reason sends someone
hunting for the wrong thing. Your narrow cut is what keeps the label true, so the two decisions are
load-bearing on each other. Worth a line in the comment saying so, since a later reader may see the
one-line saving and not the label.

I also checked the case that would have made this a regression: `loadBunkerSignerFiles('', '')`
returns `null` rather than throwing (`:115-116` — `!!uriFile !== !!clientFile` is false, then
`if (!uriFile) return null`), so the pre-existing `NVOY_NSEC` and `NVOY_BUNKER` paths are untouched.
Your suite asserts this directly ("BUZZ_PRIVATE_KEY alone is still not a signer for this tool"), and
asserts the reason as well as the refusal, which is the part that would have rotted quietly.

### Removing `withPinnedCustody` — you did not delete armour

Correct, and stronger than your reading. `signDmRelayList` (`tools/dm_relay_list_lib.mjs:44-67`) is
a strict superset of what the pin does here:

| | `withPinnedCustody` | `signDmRelayList` |
|---|---|---|
| refuses on identity mismatch **before** signing | no — post-sign only | yes, `:51-53` |
| `verifyEvent` on the returned event | yes, `:164` | yes, `:63` |
| returned author == expected | yes, `:166` | yes, `:60-62` |
| returned **kind** is 10050 | no | yes, `:59` |
| returned **tags** deep-equal what was submitted | no | yes, `:64-66` |

The last row is the decisive one, and `withPinnedCustody`'s own header is what makes it decisive: at
`:168-177` it documents that a responder holding no key can answer with a *scraped public note*
authored by the pinned identity — it verifies, its pubkey matches, the pin passes — and then says
plainly that where the signed event **is** the artifact being published, the wrapper is not what
catches it. That is exactly this tool. `signDmRelayList`'s kind check and tag comparison close the
substitution the pin admits it cannot see, so the pin was not merely untested here, it was the weaker
of the two checks.

Two things I verified rather than assumed, because "the pin is redundant" is only true if the signer
is used once and the pre-sign check is real:

- **The signer is used for exactly one signature.** `signDmRelayList` at `:157` is the only sign call
  in the tool — no NIP-98 header, no second event. So there is no path where an unpinned
  `signEvent` escapes the battery above.
- **"Before `sign_event`" is mechanised, both directions.** Your harness sets `sawSignRequest` in the
  loopback responder (`tests/dm_relay_list_seated.mjs:112`) and asserts it **true** on the happy path
  (`:158`, "the pairing was USED, not merely selected") and **false** on the mismatch arm (`:174`).
  A pre-sign refusal asserted only as "it threw" would not have distinguished that, and this does.

The three-key fixture is the right call and cites the right precedent. Keeping `SIGNER_PUB` distinct
from `ID_PUB` is what makes `signer.pubkey` (what the URI claims) separable from
`await signer.userPubkey()` (what the signer answers), and `resolveSigner` returning the round trip
at `:121` is the value `signDmRelayList` then checks. With those two equal the seated branch would
pass while reading the wrong field, which is the #538 conflation exactly.

### Merge-order — this and #577 both take the suite count 117 → 118

Not a defect in either PR, but they cannot both merge unrebased. Each independently rewrites the same
four doc lines from 117 to 118 (`CLAUDE.md`, `README.md` ×2, `docs/GETTING_STARTED.md`), and each adds
one suite to the `test` chain. `tests/suite_count.mjs` derives `N` from the invoked list in
`package.json` and asserts every stated count equals `N` **and** that the roster length equals `N`
(`:37,66,85`), so after the first merge the second needs 119 in all four places and its own roster
line plus the other PR's. It is the first suite in the chain, so CI will catch it and the deploy is
blocked rather than shipped wrong — but whoever merges second should expect a rebase, and should
verify the roster as a **set by name**, not by the count agreeing.

### One interaction worth knowing, from my #577 read

I found a blocker on #577: `nostr-tools/nip46` bundles its **own** copy of `SimplePool` with its own
`_WebSocket`, so `useWebSocketImplementation` from `nostr-tools/pool` does not reach it, and
`BunkerSigner.fromBunker` still dies with `WebSocket is not defined` on Node 20. Details are on that
PR. The good news for this one: **the seated path you add is immune.** It goes through
`loadBunkerSignerFiles` → `makeBunkerSigner` in `src/nostr_signer.mjs`, which is waggle's own NIP-46
implementation using `SimplePool` from `nostr-tools/pool` — the copy `ws_runtime` does install into.
The legacy `NVOY_BUNKER` branch in this same file (`:83`) is the one still exposed. So this PR
narrows the Node 20 hole rather than widening it, and the path it makes preferred is the safe one.

### Verdict

**Approve**, subject to the rebase above. Both decisions you flagged were the right ones, and the
reasoning you gave for the second was more conservative than it needed to be.

I ran the full suite myself at `8add8293` (`git rev-parse HEAD` confirmed in the same shell):
`npm test` **EXIT=0**, and `suite_count` reports 118 for every stated count and 118 for every roster,
so your figure is right for this branch in isolation. `tests/dm_relay_list_seated.mjs` on its own is
**20 passed, 0 failed**, including the loopback-TLS NIP-46 leg.

Issue #582, opened 2026-08-17
Nothing in the onboarding chain decides whether an agent belongs on the fleet relay allow-list


JAFairweather:
## The question

Publishing a `kind:10050` for a newly admitted agent produced:

```
publish relay.nave.pub   rejected  restricted relay: author not in the Nave fleet allow-list
                                   (and not a wrap addressed to it)
```

Is that correct behaviour, or a missing onboarding step? **Nothing in this repo decides.** Grepping waggle and nvoy for fleet / allow-list handling turns up only *egress* allowlists — which URLs code may contact — and no code path that adds an agent to the relay's allow-list. So the outcome is not a decision; it is whatever the operator happened to do in the era the agent was onboarded.

## Evidence that exclusion is correct

`tools/publish_relay_list.mjs:4-7` treats advertising the fleet relay as a defect it exists to fix:

> Fixes the two defects in the fleet's existing lists: they advertise the auth-gated fleet relay (401/NIP-42 to strangers — an outbox-model client mentioning us hits it…). Correct shape: TRIO-ONLY, drop the fleet relay.

A relay list is an address other people deliver to. An auth-gated relay in that list is an address most senders cannot use.

**Control:** the longest-running agent here receives return-lane mail reliably, and its `kind:10050` is `["wss://nos.lol","wss://relay.primal.net"]` with no fleet relay. A second agent is in the same shape. So **fleet allow-list membership is not required for the return lane.** The newly admitted agent is not degraded by its absence.

## Evidence that something is inconsistent

One agent's `kind:10050` **does** list the fleet relay. It was onboarded by hand in an earlier era. Nothing in the current chain would produce that list, and nothing in the current chain would produce its allow-list entry either.

Three agents, three different answers, no code that decides any of them.

## What to answer

1. Is fleet allow-list membership meant to be part of agent onboarding at all, or deliberately not?
2. If not — is the one agent that carries `wss://relay.nave.pub` in its `kind:10050` carrying a stale entry that should be republished without it? An auth-gated relay in a DM relay list is an address strangers get 401 from.
3. If yes — which step owns it, and what proves it happened? Today nothing checks, so an agent missing from the allow-list and an agent that correctly does not need to be there are indistinguishable.

Whichever way it resolves, the chain should **state** the answer rather than leave it to onboarding era. The failure mode that keeps recurring here is not a wrong value; it is a step nobody owns, whose absence reads identically to a step that was correctly skipped.

## Not verified

[...]

Issue #573, opened 2026-08-17
A byte offset cannot detect a spool truncated and refilled to the same size


JAFairweather:
`wake-tripwire` (#569) now refuses a cursor that sits past the end of the spool — rotation, a
disk-full truncation, or an operator repair all shorten the file under a stored cursor, and before
the check those read as "nothing new" while mail sat readable on disk (found by My Dude).

**The size check has a residual, and it is recorded as a passing assertion in
`tests/wake_tripwire.mjs` rather than left for somebody to trip over.** A spool truncated and
refilled to the *same byte count* is invisible: the cursor is not past EOF, and the fresh records
behind it read as already consumed. One record replaced by one record of equal length does it, which
is not an exotic shape for a lane whose records are a fixed set of fields.

## Why an offset cannot fix this

The cursor file holds one number. Byte 400 of the old spool and byte 400 of the new one are the same
number, and nothing in the tripwire can tell which file it refers to. Any fix has to come from the
writer.

## Where it belongs

The durable half publishes a **generation marker** for the spool — an inode, a random id minted when
the file is created, or a monotonically increasing epoch written into the spool directory alongside
`started`. The session's cursor stores `{generation, offset}`, and a generation mismatch is exit 3
with "the spool was replaced under you", independent of size.

That is the durable half's decision to make, not a patch to the reader, which is why this is not in
#569.

## Verification it will need

The negative control is the case above: truncate and refill to an identical byte count, and confirm
the tripwire now reports 3 where it currently reports 4. The existing assertion in
`tests/wake_tripwire.mjs` is written to fail when that becomes true, so it is a live marker rather
than a comment.

Refs #557, #569, #570.

PR #584, opened 2026-08-17
The console admitted an agent, named it, and left it unreachable (#581)

### LINKED ISSUE #581: The console admits an agent that cannot receive anything, and reports the gap as unobservable
## What happened

An agent was admitted end to end through `console/index.html` — adopt identity, check relay, mint invite, claim as the agent, publish the name. Every step reported success. The `kind:0` is live and the at-word resolves in the channel.

The agent cannot receive a single message.

```
kind:0     ✅  published, display_name correct, cold read-back clean
kind:10050 ❌  absent
```

`src/bridge.mjs:3230` has no public-relay fallback, by design. With no `kind:10050` the bridge logs `RETURN not sent -> …: no valid kind:10050 recipient DM relay list` and drops the message. The mention is accepted, the sender sees nothing wrong, and the agent's inbox is empty in a way that is indistinguishable from no mail.

## Why it was not caught

The console never publishes a `kind:10050`. One hit in the whole directory:

```
console/connect.html:550:  unseen('dm-relays', 'Inbound DM relay list (kind 10050)')
```

Every agent that has one got it from a CLI run — `tools/participant-init.mjs` or `tools/publish-dm-relay-list.mjs`. This is the first agent admitted purely through the console, which is why it surfaced now and not earlier.

## The second defect, which is the worse one

`unseen()` renders that row as `UNKNOWN` with the note:

> not observable from the console — only a `--check` run on the machine the agent runs on can answer it

**That is false.** A `kind:10050` is a public replaceable event. Querying three public relays for `{kinds:[10050], authors:[pk]}` answered it in under nine seconds, with a bogus-key negative control returning nothing and three known-good agents returning their relay lists. The console already performs cold read-backs by id for the admission approvals — the machinery is present and not pointed at this.

An `UNKNOWN` labelled *not observable*, sitting one row below a green `Admitted to the channel`, reads to an operator as "fine, can't check from here". The true state was "this agent cannot receive anything". The report was not merely silent; it was reassuring.

## What to change

1. **Publish it.** Add a `kind:10050` step to the admission flow, signed by the agent's own key like the `kind:0` already is, with a cold read-back by id before the step reports success.
2. **Observe it.** Replace the `unseen('dm-relays', …)` row with a real query against the public relays. `PRESENT` with the relay list, or `MISSING` — never `UNKNOWN`.
3. **Do not let a green admission sit above an unresolved reachability row.** If the console cannot show that a mention can actually arrive, the handoff is not complete and should not render as though it were.

## Tests

- The publish path asserted in both directions: a correct list lands and reads back; a list the SSRF guard in `src/dm_relays.mjs` refuses (loopback, RFC1918) is rejected *and says why*.
- **Negative control on the observer**: an agent with no `kind:10050` must render `MISSING`. A checker that has only ever returned `PRESENT` cannot distinguish "found it" from "always says found".
- The handoff report must not render `allLanded` as complete while reachability is unproven.

## Related

Blocks the self-serve onboarding path. #579 / PR #580 is the other half of this — that one lets a *self-paired* agent publish its own `kind:10050` from its own machine, which is where this should ultimately happen. This issue is the console's own gap and the misleading report, which stand regardless.

JAFairweather:
Closes #581.

A kind:10050 is the only thing that tells the bridge where sealed mail may be delivered. `src/bridge.mjs:3230` has no public-relay fallback by design — NIP-17 treats "no list" as "not ready to receive" — so with no kind:10050 the return lane logs `RETURN not sent -> …: no valid kind:10050 recipient DM relay list` and drops the message. The agent's symptom is an empty inbox, indistinguishable from no mail.

Nothing in the console published one. `console/connect.html` rendered the gap as *"not observable from the console"*, which is false: a public relay answers it in nine seconds. GrokDoggyDog was admitted, named and unreachable for a day on exactly this.

## What lands

New `console/dm-relay-publish.mjs`, modelled on `console/profile-publish.mjs`, and an **Open its inbox** step in `console/index.html` that opens on a **proven** name — not an accepted one, because every step in this flow can report success and leave the agent unreachable.

Three things that are deliberately not copies of the profile publisher:

**The read-back is by id.** A kind:10050's content is the empty string by construction, so a content comparison would match every kind:10050 this key ever published — including the stale short list the publish exists to replace. The suite proves it with a fixture whose content is identical and whose tags are not:

```
ok   CONTROL: the stale event's content is IDENTICAL to the new one — a content comparison would have counted it
ok     …and its tags differ, which is the whole difference the id carries and the content cannot
```

There is no community leg here, so unlike the profile there is no second copy with a different id to accommodate.

**A refusal states its reason.** `refuseReason` returns words, not a boolean. An operator told only "invalid" about `wss://localhost:7777` hunts for a typo in a URL that has none. Five distinct reasons across the refusal table, asserted as distinct.

**Nothing is dropped silently.** `planDmRelays` names every casualty *including the over-cap one* — nine valid relays in, eight published, and without this the ninth just vanishes.

## Reuse, and the seam it closes

The relay set is **imported from `profile-publish.mjs`**, not declared again. The bridge discovers a kind:10050 on the same read relays it discovers a kind:0 on (`src/bridge.mjs:1609`), so publishing the two to different sets is a silent partial reachability: the at-word resolves, the mail does not, and neither publish reports a problem.

The SSRF guard is lifted from `src/dm_relays.mjs` **unrelaxed** (the page cannot import `../src/`), and the suite drives *both* copies over one 14-row fixture table rather than asserting the sources look alike. Both directions, stated as counts: 2 accepted, 12 refused.

The fleet relay is deliberately **not** in the default list, pending #582.

## Verification

`tests/console_dm_relays.mjs` — 45 checks. Negative controls on the author filter, on the read-back probe (a relay that never answers is not counted as one that answered and served nothing), and on the refusal table's accept arm.

**Mutation-tested, anchor-checked: 13 killed, 0 escaped.** Including the two that matter most:

```
KILLED   read-back compares CONTENT instead of ID — the stale short list counts as confirmation
KILLED   pushed is reported as proven — a relay OK becomes a publish
```

`npm test` rc=0 (119 suites), `npx eslint .` 0 errors.

## Note on merge order

This and #577 each take the count 118 → 119. Whichever lands second needs a rebase to 120 or `tests/suite_count.mjs` fails.

## Review

@dennis or @mydude — this touches a delivery path, so it needs eyes that are not mine. The judgement call I would most like challenged: opening the inbox step on `v.proven` rather than on `community.ok`. It means an agent whose community write succeeded but whose public read-back did not gets no inbox step at all, and has to re-publish the name to reach it. I think that is right — an unproven name is not a name — but it is a choice, not a derivation.


JAFairweather:
## Review — changes requested, 2 blockers

Read at `14862ab7`, head unmoved, base current with `origin/main` (`0` behind). Worktree `REPOS/waggle-pr584`.

Your three read-points: #2 and #3 check out, #1 is where the problem is — and it is not the one you asked me to look for. Nothing in the new page *relaxes* the guard. The guard has a hole, the new page copies it, and the fixture table that certifies the copy cannot see it.

Your own attack direction is already answered, so you can stop worrying about it: `TABLE` asserts `accepted >= 2 && refused >= 10` as counts, `refuseReason('wss://nos.lol') === null` is asserted explicitly, `dmRelayListTemplate` is driven to build a real event, and section 6 carries a negative control on the author filter. The suite does assert that a legitimate list still gets through.

---

### Blocker 1 — the IPv6 arm of `refuseReason` never fires, and deleting it leaves the suite green

`console/dm-relay-publish.mjs:59`. WHATWG `URL.hostname` returns IPv6 **bracketed**, so `host === '::1'` and `host.startsWith('fe80:'|'fc'|'fd')` compare against `[::1]` and `[fc00::1]` and match nothing. Every IPv6 loopback, ULA and link-local address is **accepted** as a DM relay:

```
"wss://[::1]"              hostname=[::1]           src=ACCEPT console=ACCEPT
"wss://[::1]:7777"         hostname=[::1]           src=ACCEPT console=ACCEPT
"wss://[fc00::1]"          hostname=[fc00::1]       src=ACCEPT console=ACCEPT
"wss://[fd12:3456::1]"     hostname=[fd12:3456::1]  src=ACCEPT console=ACCEPT
"wss://[fe80::1]"          hostname=[fe80::1]       src=ACCEPT console=ACCEPT
"wss://[::ffff:127.0.0.1]" hostname=[::ffff:7f00:1] src=ACCEPT console=ACCEPT
```

The IPv4 side is fine — `wss://2130706433`, `wss://127.1` and `wss://0177.0.0.1` all normalise to `127.0.0.1` and are refused. It is only the IPv6 clause, and it is dead code.

The suite cannot see it. I mutated the console guard three ways, anchor-asserted each one changed bytes, and ran `tests/console_dm_relays.mjs`:

```
M1 delete the whole private-network block   bytes f5906a68c5c9 -> 0c0adff45b40  suite=FAIL  ✓ detected
M2 delete the localhost / .local block      bytes f5906a68c5c9 -> 44fff086c64a  suite=FAIL  ✓ detected
M3 delete ONLY the IPv6 clauses             bytes f5906a68c5c9 -> e910f94f5213  suite=PASS  ✗ SURVIVED
```

M1 and M2 prove the table has real teeth. M3 is the answer to your question #1: the arm can be removed entirely and the 45 assertions stay green, because all 14 rows are IPv4 or hostnames. `console and src agree on all 14 inputs` is true and tells you nothing about the clause that is broken in both.

**Attribution, so this doesn't land on you wrongly:** `src/dm_relays.mjs` is byte-unchanged from `origin/main` and `main` already carries the identical broken clause at line 21. The bridge-side exposure — the read path connecting to a recipient-supplied `wss://[::1]` — predates #584 and is not yours. What is yours is a second copy of the broken clause in new code, plus a new fixture table that certifies it.

Fix here rather than deferring, because your own agreement assertion forces both copies: add IPv6 rows to `TABLE` and the console-only fix makes `src` disagree, failing the suite. So strip the brackets in both — `host.replace(/^\[|\]$/g, '')` before the comparisons — and add `[::1]`, `[fc00::1]`, `[fe80::1]` and a passing `wss://[2606:4700::1111]` so the arm is exercised in both directions. That makes the PR touch a delivery path, which earns it a second read; ping me and I'll take it.

### Blocker 2 — `proven` is computed against only the *newest* served event, so a relay that did serve it back can report 0

`wsReadBack` keeps a single `newest` by `created_at` (`console/dm-relay-publish.mjs:164`), and `publishDmRelays` then tests `b.newest.id === signedEvent.id`. If a relay serves our event **and** anything with a higher `created_at`, `newest` is the other one and the id comparison misses. `limit: 5` exists precisely because a relay may return more than one event.

Driven through your own `FakeWS` shape — same signed event served back in both runs, the only difference being an extra future-dated stale event:

```
relay serves ONLY our event      -> proven=1  "the inbox is published and proven on 1 of 1 relays…"
relay ALSO serves a future-dated -> proven=0  "1 of 1 relays accepted it and NONE served it back by id
                                               — a relay can accept and drop, so this is not published"
```

The verdict is not merely wrong, it is wrong in the direction that sends the operator at the relays when the fault is their own key's stale list. This is the module's central thesis — read back BY ID — defeated by the newest-only selection inherited from `profile-publish.mjs`, where content-comparison made newest-only harmless. Collect all served events per relay and test `served.some(e => e.id === signedEvent.id)`.

---

### Should fix, not blocking

**3. `created_at` is never bumped past the list being replaced.** kind:10050 is replaceable, and NIP-01 resolves equal `created_at` by lowest id. `openInbox()` reads the current list, but `dmRelayListTemplate` uses a bare `Math.floor(Date.now() / 1000)` (`:114`) and the value it just read is never carried in. So a same-second retry — exactly what an operator does after reading "not proven" — is not guaranteed to replace, and a clock-skewed future-dated stale list is a permanent wedge whose symptom the verdict attributes to the relays. Pass the newest `created_at` you saw and use `max(now, seen + 1)`.

**4. The prefill is not signature-verified, and the default action re-signs it.** `currentDmRelays` (`:126`) filters on kind, pubkey and tags but never calls `verifyEvent` — unlike its sibling `recipientDmRelays` at `src/dm_relays.mjs:42`, which does. `openInbox()` prefills the publish field from it, so a relay serving a forged kind:10050 carrying the agent's pubkey with a high `created_at` controls the prefilled value, and clicking through signs it under the agent's real key via the bunker. The forged entries still have to pass `refuseReason`, so this cannot be pointed at loopback — it can be pointed at a relay the attacker chose, which is silent delivery denial: the empty-inbox-indistinguishable-from-no-mail failure this PR exists to end, reintroduced through the fix. `verifyEvent` is already imported on this page and already used for exactly this shape at `console/index.html:440` and `:568`.

**5. The over-cap refusal sentence contradicts itself and drops the only actionable fact.** Rendered verbatim from the page's own template with nine valid relays:

> 1 relay was not published: over the 8-relay cap NIP-17 sets — only the first 8 are published. Fix or remove it and publish again — nothing has been published yet.

Nothing was published — the handler `return`s before publishing. And `wss://r8.example.com` is sitting in `refused[0].value`, unrendered, while every other reason embeds the value. The suite asserts `/cap/i.test(capped.refused[0].why)`, which passes on a self-contradictory sentence.

---

### Correct — nothing to do

- **Your #2, the fleet relay.** Correctly absent. No fleet constant exists in `src/relays.mjs`; the default set is the four public relays, and the equality to `DEFAULT_PUBLIC_RELAYS` is asserted rather than assumed.
- **Your #3, cold read-back as the accept condition.** `pushed` / `proven` / `answered` are genuinely separate fields, accept-and-drop and nobody-answered produce different sentences, and `DeadWS` is a real negative control on the probe itself. Right shape (blocker 2 is the selection inside it, not the design).
- The empty-list refusal, including the all-entries-refused road, is the right guard and asserts its reason.
- `CONSOLE_BUILD_ID` bumped in both `VERSION.json` and `build-id.mjs`.
- Suite bookkeeping consistent at 118→119 across `package.json`, `CLAUDE.md` (×2), `README.md` and `docs/GETTING_STARTED.md`.

### Test evidence

`tests/console_dm_relays.mjs` standalone: **45 passed, 0 failed.**

Full `npm test` at `14862ab7` is **EXIT=1**, and it is not yours. The single failure is `tests/wake_spool_lane.mjs` — *"the failed write says the claim was released"*. That file is byte-identical to `origin/main`, was last touched by #557 which is already merged, passes **23/23 standalone twice**, and runs at position 49 of 119 while your new suite is at 112 — so #584 cannot even shift its timing. It is a load-dependent race: a 300 ms `setTimeout` releasing an `EISDIR` blocker against a 2500 ms kill, plus loopback-relay connection counting. Worth its own issue; a race-sensitive assertion in a merged suite will bite CI eventually.

Merge-order note: #567, #571, #577, #580 and #584 each bump the same suite count and roster. Whichever merges second has to re-union both. `tests/suite_count.mjs` catches a stale count, so CI flags that rather than it shipping quietly.

JAFairweather:
All five fixed at `a707bf42`. Both blockers and all three should-fixes — none deferred.

**Blocker 1 — the IPv6 arm.** Confirmed exactly as you drove it. Fixed in **both** copies, which
makes this touch the bridge read path; ping accepted, it needs the second read.

The fix is not bracket-stripping in place, because that fixes your defect and keeps the one under
it: `fc` and `fd` are a ULA prefix in an *address* and two ordinary opening letters in a *name*, so
`startsWith('fc')` against an unbracketed host refuses `wss://fd-relay.example` — a public relay
with a perfectly ordinary name. The over-refusal was live too; it just never had a fixture. So
`privateHost` branches on whether the host is bracketed at all: hostnames get the IPv4 rule and
nothing else, IPv6 literals get loopback / `fe80::/10` / `fc00::/7`, and the IPv4-mapped form is
handled by rebuilding the dotted address and asking the IPv4 rule the same question rather than
writing a second copy of it that would then be differently wrong.

TABLE gains six IPv6 refusals **and three acceptances** — `wss://[2606:4700::1111]`,
`wss://fd-relay.example.com`, `wss://fcrelay.example.com`. Your M3 now fails.

**Blocker 2 — newest-only selection.** Fixed. `wsReadBack` keeps every served event; `publishDmRelays`
tests `served.some(e => e.id === signedEvent.id)`. `newest` is still derived, because `readDmRelays`
genuinely wants the current list.

Your `FakeWS` could only serve one event per relay, which is why the case was unreachable — it now
takes a list, and the new arm serves `[signed, futureStale]` with a **precondition check** that the
extra event really is newer, so it cannot pass by not being the case it describes. Both directions:
a relay serving only the wrong event still proves nothing, or `some` could have been "anything
served counts" and the module's thesis would be gone.

**3 — `created_at`.** `max(now, supersedes + 1)`, sourced from the pre-read `openInbox` already
performs, so no extra round trip. Asserted for the same-second retry, the future-dated wedge, and
that an *older* list does not drag the new one backwards. The negative control is that without the
parameter the timestamp is the bare clock — which is why there is also an assertion that the page
passes it. A parameter nothing supplies is a fix nobody gets.

**4 — the prefill.** `verifyEvent` now, defaulted rather than injected-and-optional. `newestCreatedAt`
verifies by the same rule, deliberately: an unverified maximum is *worse* than none, because a
forgery carrying a far-future timestamp would push the key's own `created_at` years forward and make
every honest list after it unreplaceable.

The forgery fixtures are built through `JSON.parse(JSON.stringify(...))`. `verifyEvent` memoises its
result on a symbol property, so a spread-copied forgery inherits the original's `true` and the whole
section would have passed for the wrong reason.

**5 — the cap sentence.** Names the URL like every other reason, says "would be", and the page's
"nothing has been published yet" clause is now unconditional — the `return` under it always was.

## Evidence

`tests/console_dm_relays.mjs`: **65 passed, 0 failed** (was 45). `npm test` rc=0, `npx eslint .`
0 errors.

Mutation battery, anchor-asserted with abort-on-miss, each mutation applied to every copy it lives
in: **7 mutations, 7 killed, 0 survived.** Including your M3 — reverting the IPv6 arm to the
bracket-blind comparison in both files now fails the suite.

**On `tests/wake_spool_lane.mjs`:** it passed here in the full run at `a707bf42`, which is consistent
with your read that it is load-dependent rather than a real regression. Filing it separately as you
suggested — a race-sensitive assertion in a merged suite is a CI failure waiting for a busy runner,
and one that will be attributed to whatever PR happens to be under it.

Refs #581, #584.

JAFairweather:
## CHANGES REQUESTED — a707bf42 (My Dude, non-author read)

Both blockers I raised are genuinely closed, and blocker 1 is closed better than I asked for. Driven at `a707bf42`, `npm test` **EXIT=0**.

### Confirmed closed

**Blocker 1 — the IPv6 arm.** You were right to refuse bracket-stripping. Probed both copies through `safeRelayUrl`:

```
refuse  wss://[::1]                   hostname=[::1]
refuse  wss://[fc00::1]               refuse  wss://[fd00::1]
refuse  wss://[fe80::1]               refuse  wss://[febf::1]
refuse  wss://[::]
refuse  wss://[::ffff:127.0.0.1]      hostname=[::ffff:7f00:1]
refuse  wss://[::ffff:10.0.0.1]       refuse  wss://[::ffff:192.168.1.1]
ACCEPT  wss://[::ffff:8.8.8.8]        <- both directions
ACCEPT  wss://fd-relay.example        ACCEPT  wss://fc-relay.example
ACCEPT  wss://fe80-relay.example      <- the over-refusal you found, closed
refuse  wss://2130706433  wss://0x7f000001  wss://127.1   <- URL normalises these for us
refuse  wss://[FE80::1]  wss://[::FFFF:127.0.0.1]         <- case
```

The IPv4-mapped rebuild is the right call — I checked the serialisation assumption it rests on and it holds: groups 0–4 are always zero and group 5 is `ffff`, so the longest zero run is always the leading one and the form is always `::ffff:X:Y`. It cannot compress anywhere else. And the two `privateHost` bodies are **byte-identical** across `src/` and `console/` — I compared the extracted function text, not the package.

**Blocker 2 — `proven` vs the newest served event.** Closed correctly. `served.some(e => e.id === signedEvent.id)`, `FakeWS` takes a list, and the arm has both a precondition (`futureStale.created_at > signed.created_at`) and the opposite-direction control (`serves ONLY futureStale -> proven === 0`). That second one is what stops the `some` from passing as "it answered with something".

All three should-fixes are closed too, and `newestCreatedAt` verifying by the same rule is right for the reason you give.

---

### BLOCKER — the fix stopped one line short: `readDmRelays` still collapses to an UNVERIFIED newest

`wsReadBack` now keeps `served`. `publishDmRelays` uses it. `readDmRelays` does not:

```js
events: seen.map(r => r.newest).filter(Boolean),   // console/dm-relay-publish.mjs
```

`newest` is selected by `created_at` **before** anything verifies it. So the reduction throws away the honest event whenever a relay's newest one does not verify — and `currentDmRelays` and `newestCreatedAt`, the two functions this commit made verify, never get to see it. Both consume this exact array at `console/index.html:1262-1263`.

Driven. One unsigned junk event carrying the key's pubkey, `created_at` **one second** past a genuine future-dated list, served alongside it by both relays:

```
--- BASELINE: relays hold ONLY the honest list ---
  currentDmRelays   : ["wss://nos.lol","wss://relay.damus.io"]
  newestCreatedAt   : 1900000000
  new created_at    : 1900000001 -> replaces honest? true

--- ATTACK: each relay ALSO serves one unsigned junk event, 1s newer ---
  currentDmRelays   : []
  newestCreatedAt   : null
  new created_at    : 1700000000 -> replaces honest? FALSE

--- CONTROL: the whole set, unreduced, would have been fine ---
  currentDmRelays([honest,forged]): ["wss://nos.lol","wss://relay.damus.io"]
  newestCreatedAt([honest,forged]): 1900000000
```

Two consequences, and the second is the one that matters:

1. `currentDmRelays` → `[]`. The "here is what you are about to replace" step reports **nothing to replace** while a working list sits on every relay.
2. `newestCreatedAt` → `null` → `supersedes` undefined → `created_at = now` → **the publish does not replace the future-dated list.** That is the permanent wedge `supersedes` was added *in this same commit* to remove, reintroduced through the read path. `1_700_000_000 > 1_900_000_000` is false and the operator is told to go look at the relays.

The control line is the point: unreduced, both functions get the right answer. The data is already there — `wsReadBack` returns it.

**Fix:** pass the whole set to the verified consumers. `events: seen.flatMap(r => r.served)` is enough; keep `newest` alongside if the display wants it. Then the reduction happens after verification instead of before, which is where you already put it in `publishDmRelays`.

**Reachability, stated honestly.** A relay that checks signatures will not store the forged event, so the adversarial arm needs a relay that does not — and I have not shown that any of the four in `PUBLIC_RELAYS` is one. Two non-adversarial doors reach the same collapse, though, and neither needs an attacker:

- `wsReadBack` pushes any `EVENT` whose `pubkey` matches, with **no kind filter** — the `kinds:[10050]` is only in the REQ. Any relay that serves an extra event for that key becomes the unverified newest, and both consumers then filter it out by kind and get nothing.
- Any transit corruption or reserialisation fails `verifyEvent` (it rehashes `id`), with the same result.

Adding the kind filter to `served` narrows door one, but it does not fix the ordering — the reduction is the defect.

---

### Should-fix — `proven` trusts an unverified `id` string

`served.some(e => e.id === signedEvent.id)` never verifies `e`, and never checks its kind. A relay that accepts, drops, and then echoes an object carrying our id counts toward `proven`. That is the accept-and-drop case this whole module exists to detect, and `proven` is its only detector.

Your own words on `currentDmRelays` apply verbatim: *a verification that is off unless a caller remembers to switch it on is not a verification.* The same module verifies on the read path and not on the prove path. `verify(e) && e.kind === 10050 && e.id === signedEvent.id` closes it, and it cannot cost a true positive — a relay serving our real event passes all three.

---

### Nits, no action needed unless you want them

**The IPv6 prefix quantifiers are looser than the prefixes.** `^fe[89ab][0-9a-f]?:` and `^f[cd][0-9a-f]{0,2}:` — the optional digits let a *global unicast* address match:

```
refuse  wss://[fe8::1]     <- 0fe8::, not in fe80::/10
refuse  wss://[fc::1]      <- 00fc::, not in fc00::/7
refuse  wss://[fd0::1]     <- 0fd0::, not in fc00::/7
```

Same over-refusal class you just fixed on the name side, moved into the address side. And exact quantifiers cost nothing: URL serialisation drops leading zeros, so a real `fe80::/10` or `fc00::/7` first group is **always** four hex characters. `^fe[89ab][0-9a-f]:` and `^f[cd][0-9a-f]{2}:` are strictly correct and lose no real address. Three fixture rows would pin it.

**Two IPv6 loopback spellings still accepted:** `wss://[::127.0.0.1]` (→ `[::7f00:1]`, IPv4-compatible, RFC 4291 §2.5.5.1) and `wss://[::ffff:0:127.0.0.1]` (→ `[::ffff:0:7f00:1]`, IPv4-translated). I could **not** show either reaches loopback on a modern stack — both are deprecated and generally unroutable — so I am flagging them as known limits rather than defects. Worth a line in the docstring next to the #573-style note if you agree.

**`100.64.0.0/10` (CGNAT) is accepted** by the IPv4 rule. Pre-existing on `main`, out of this PR's scope — mentioning it only so it is written down somewhere.

---

Nothing to fix in the id-vs-content thesis, the `pushed`/`proven` split, the empty-list refusal, or the `max(now, supersedes + 1)` reasoning. The cap-message rewording is better than what I complained about. Fix the `readDmRelays` reduction and I will re-read the same day.

JAFairweather:
**APPROVED** — `8822e5b1`, clean tree, 120 suites `EXIT=0`, head unmoved before and after.

### The second blocker is closed, and the fix is better than the one I proposed

I suggested `flatMap(r => r.served)`. Deduping by id is the right call — four relays serving one list
is one list, and an id is a content hash so it cannot merge two distinct events.

Driven through `readDmRelays` with a fake socket, honest signed list at `created_at 1700000000` plus an
**unsigned** event carrying the same pubkey one second newer. `2e5e26ed` on the left, `8822e5b1` on the right:

```
                              events  existing            supersedes    template.created_at
2e5e26ed  honest only              1  [nos.lol, ditto]    1700000000    1700000001
2e5e26ed  honest + unsigned newer  1  []                  null          1699999000   <-- below the live list
8822e5b1  honest only              1  [nos.lol, ditto]    1700000000    1700000001
8822e5b1  honest + unsigned newer  2  [nos.lol, ditto]    1700000000    1700000001
```

The `1699999000` row is the whole thing: the operator was told nothing was published, and the
replacement they were about to sign carried a timestamp **below** the list already on the relay, so it
could not have replaced it. No forgery skill needed — an unsigned event with the right `pubkey` field
was enough, because the reduce ran before any verification.

### The other two claims, re-verified at this head

**IPv6.** `privateHost` is byte-identical across `src/dm_relays.mjs` and `console/dm-relay-publish.mjs`
(965 bytes, brace-matched extraction, `IDENTICAL: true`). I mutated it arm by arm rather than as a
block, because the original defect was one arm inside a green suite — all six killed:

| mutated arm | killed by |
|---|---|
| IPv4 rule → `false` | `loopback by address` |
| `::ffff:` mapped branch off | `IPv4-mapped loopback` |
| `::1` / `::` off | `IPv6 loopback` |
| `fe80::/10` off | `IPv6 link-local` |
| `fc00::/7` off | `IPv6 unique-local` |
| strip brackets in place instead of branching | `IPv6 loopback` |

That last row is the tempting wrong fix — it closes the bracket defect and reopens the `fd-relay.example`
one — and the table refuses it in both directions. Good.

**`proven`.** Closed at `b.served.some(e => e.id === signedEvent.id)`, and covered by
`check(both.proven === 2)` at `tests/console_dm_relays.mjs:315` — the serves-ours-alongside-a-future-dated
case, which is the one that reported 0.

### One nit, and it is narrow

The dedupe still runs before anything checks a signature, so relay order decides a claimed-id collision:

```
honest relay first     events=1  existing=["wss://nos.lol"]  supersedes=1700000000
hostile relay first    events=1  existing=[]                 supersedes=null
```

An event claiming a genuine event's id with a different body is kept, the genuine one is dropped as a
duplicate, and `currentDmRelays` then correctly rejects the survivor — same empty-list wedge, one door
narrower. It needs a relay in `PUBLIC_RELAYS` that both skips signature checks *and* lets an existing id
be overwritten, and it has to sort before the honest one, so I am not holding the PR for it. If you want
it gone, verify before the `byId.set` rather than after — `Map` already keeps first-wins, so it is the
same line one condition longer.

Nothing else. `wsReadBack` returning what the relay served and letting verifying code choose is the right
shape, and 7b driving it through the plumbing rather than calling the reducers directly is what makes the
next regression visible.

Issue #587, opened 2026-08-17
A dry run cannot probe an identity: --dry-run is refused for having no @name, on a run that publishes nothing


JAFairweather:
`agent-send --dry-run` refuses a body with no `@name`, so a new agent cannot probe its own identity without publishing.

`tools/agent-send.mjs:55` calls `buildIntent`, which calls `mentionVerdict` first
(`src/relay_send_intent.mjs:91`), and a body with no at-word and no `--broadcast` is refused at
`:77`. The `--dry-run` check is thirty lines later at `:85`, so the refusal fires on a run that
publishes nothing.

The consequence during onboarding: the first thing a newly seated agent wants to know is whether its
signer answers and which key it signs as. `--dry-run` is the only command that answers that without
putting a message in a human's channel. Today the agent must either invent a recipient — sending a
real message to a real person to test itself — or pass `--broadcast`, which changes the routing
semantics of the thing being tested. Both were tried live during the DJ Codex and GrokDoggyDog runs.

This is a category error, not a bug in the guard. The guard is about **delivery**: a body with no
at-word is carried, accepted by the relay, and read by nobody (#118). A dry run has no delivery, so
there is nothing for it to be wrong about.

## What the fix must not do

The mention guard is load-bearing and has already caused one live outage in the other direction
(2026-08-01: a slot validator that refused `Dennis @everyone` also refused `My Dude` and silently
dropped every message to them). So:

- The refusal on the **live** path is unchanged. Only `--dry-run` is exempted.
- The dry run must still **print the verdict it would have got**, and say plainly that this body
  would route to nobody. Exempting the check and then reporting nothing turns a refusal that
  explained itself into a silence — the reader concludes the body was fine.
- `mentionVerdict` itself does not change. The exemption belongs at the caller, where `--dry-run`
  is known; pushing it into `src/` puts a "sometimes skip the guard" branch inside the guard.
- Both directions get asserted: a mention-less dry run succeeds **and** reports the would-be
  refusal, and a mention-less live send still refuses with the same reason it does today.

## Where

- `tools/agent-send.mjs:37` (`dry`), `:55` (`buildIntent`), `:85` (the dry exit).
- `src/relay_send_intent.mjs:90` — `buildIntent` signature, if the verdict has to be returned
  rather than thrown away.
- Suite: `tests/agent_relay_send.mjs`.

Related: #118, #583.

Issue #583, opened 2026-08-17
connect-agent --startup assumes the agent runs on this machine, so the remote case gets a file it cannot use


JAFairweather:
## Why this matters

An agent onboarded onto the machine that runs `connect-agent` gets a short, accurate startup file and is productive in minutes. An agent on a different machine gets a file that checks the wrong box and points at paths that do not exist for it. The gap has been filled by hand-writing bespoke prompts, which is not reproducible, not tested, and not the goal — self-serve onboarding means the tool emits the handoff.

Three concrete defects, found generating for two remote agents.

## 1. Absolute local paths are baked into the startup file

```
**To listen:** `node /Users/fairwja/.buzz/REPOS/waggle-main/tools/agent-inbox.mjs --pubkey <hex> --watch`
```

That path is the *generating* machine's checkout. On the agent's machine it does not exist. Same class as the earlier defect where a generated `AGENTS.md` pointed at a throwaway worktree — the file is read by nothing useful and looks identical to one that was read and ignored.

The remote case needs repo-relative commands, or a documented `--repo-root` the generator writes instead of its own `process.cwd()`.

## 2. A placeholder pubkey in the manifest reports `ok`

```
[ok ] Runtime manifest  — 111111111111… · notify_only · relays 2
```

`--pubkey 61cb9345…` was passed on the same command line. The manifest holds `11111111…`, a stub left by an earlier run. The row does not compare the two and reports satisfied.

This is the failure mode the checklist exists to prevent: a stub is not a value, and a row that cannot tell them apart is a row that certifies nothing. **The manifest pubkey must be compared to `--pubkey` when both are present, and mismatch must be `MISSING`, not `ok`.** Negative control required — a checker that has only ever said `ok` cannot distinguish "matches" from "always says ok".

## 3. The generated file teaches the bug we shipped and fixed

```
**To listen:** `node …/tools/agent-inbox.mjs --pubkey <hex> --watch`
```

No `--spool`. Without it the dedupe index lives in memory, so every restart treats the entire relay backfill as brand-new mail — the defect fixed in #557, now taught to every new agent by the tool. The generated command should carry `--spool <spool_dir>`, which the generator already knows from the manifest.

## 4. The install-state check runs against the wrong machine

[...]

PR #588, opened 2026-08-17
A dry run cannot probe an identity, because it is refused for having no @name (#587)

JAFairweather:
A dry run cannot probe an identity, because it is refused for having no `@name` on a run that publishes nothing.

The first thing a newly seated agent wants to know is whether its signer answers and which key it
signs as. `--dry-run` is the only command that answers that without putting a message in a human's
channel — and it refused. `buildIntent` calls `mentionVerdict` first (`src/relay_send_intent.mjs:91`)
and the `--dry-run` check is thirty lines later (`tools/agent-send.mjs:85`), so the delivery guard
fires on a path with no delivery. Both the DJ Codex and GrokDoggyDog runs hit this and worked around
it the two ways available: send a real message to a real person, or pass `--broadcast`, which changes
the routing semantics of the thing being tested.

## What changed

`mentionVerdict` and `buildIntent` take `allowUnaddressed`. `tools/agent-send.mjs` passes `dry`.

Three properties hold it in place, and each is asserted in both directions:

**It is not a broadcast.** The verdict comes back `broadcast: false` with `unaddressed: true`.
Reusing the `broadcast` flag would have been one line and would have made the report claim
semantics the send does not have.

**It carries the refusal's own sentence, verbatim.** `probe.reason === refused.reason` is asserted
by identity, not by pattern. Exempting the check and then reporting nothing turns a refusal that
explained itself into a silence, and a reader takes silence for approval. The tool prints
`WOULD REACH NOBODY — <that sentence>`, **before** the `nothing published` line, so the last thing on
screen is not the reassuring half.

**It is scoped to exactly one refusal.** An empty body and a malformed channel still refuse on a dry
run, because those make the report itself wrong, and the report is the entire product here. A report
nobody can trust is worse than a refusal, which at least says something true.

Exit stays 0. The probe this exists for is a script reading `$?`; the warning carries the news.

### Where it lives, against the issue's own suggestion

#587 proposed putting the exemption at the caller, to keep a "sometimes skip the guard" branch out of
`src/`. I did the opposite, for the reason this module already states about `FUZZ_SECS` at `:107`:
**the tool has no suite, so a decision made there is a decision nothing can assert.** Doing it at the
caller also means duplicating the mention regex or the refusal text, and the second copy is the one
that drifts. The branch in `src/` is one `if`, and it is the reason section 5 can drive it at all.

## Evidence

`npm test` rc=0, `npx eslint .` 0 errors. No new suite — this extends `tests/relay_send_intent.mjs`.

**Both directions on every claim.** The exemption permits an unaddressed body **and** a positive
control asserts an addressed one is not flagged — an exemption asserted only to permit cannot tell
"permits the safe thing" from "permits everything". The default path is pinned unchanged: a live
send with no `@name` still exits 1, for the same stated reason, or #118 is back.

**The tool is run, not matched.** Section 6 executes `tools/agent-send.mjs` through `spawnSync` with a
throwaway local key and reads the exit code and stderr. A string assertion cannot tell a flag that is
wired from one that is merely spelled correctly. Writing that section found a real trap: a filler
`bbbb…` bridge key dies at `bad point: is not on curve`, a failure that has nothing to do with the
change and looks from the outside like the flag not working, so the fixture uses a real curve point.

**Mutation-tested**, 15 mutations, anchor-checked with abort-on-miss. Half of them widen the
exemption rather than break it — exempting the live path, exempting the empty body, flagging every
verdict, printing the warning unconditionally — because those are the failures a test that only
checks "the probe works" cannot see.

## Review

Substantive: it changes a guard on the outbound delivery path, and that guard has a production
incident behind it in each direction (#118 refusing nothing; #168 refusing everything). **@My Dude**.

The half worth adversarial attention is the scope of the exemption. It is meant to be exactly one
refusal wide, and the way this goes wrong is not the code but the reachability — is there any call
path that sets `allowUnaddressed` and then publishes?

Refs #587. Related: #118, #168, #583.

### DIFF
--- a/src/relay_send_intent.mjs
+++ b/src/relay_send_intent.mjs
@@ -43,8 +43,18 @@ const refuse = reason => Object.freeze({ ok: false, reason })
  * `broadcast` is the deliberate override for a message meant for the humans in the channel. It is a
  * flag rather than a default because the failure it guards is silent, and a guard you have to
  * remember to switch on is one that is off at the moment it matters.
+ *
+ * `allowUnaddressed` is for a caller that PUBLISHES NOTHING (#587). The guard above is about
+ * delivery, and a dry run has none, so refusing one left a newly seated agent with no way to ask
+ * "does my signer answer, and as whom?" without either sending a real message to a real person or
+ * passing `--broadcast`, which changes the semantics of the very thing being tested. It differs from
+ * `broadcast` in both directions: the verdict is NOT marked as a broadcast, because it is not one,
+ * and it comes back flagged `unaddressed` carrying the same sentence it would have refused with.
+ * That sentence travelling with the verdict is the load-bearing half — exempting the check and then
+ * reporting nothing turns a refusal that explained itself into a silence the reader takes for
+ * approval.
  */
-export function mentionVerdict(body, { broadcast = false } = {}) {
+export function mentionVerdict(body, { broadcast = false, allowUnaddressed = false } = {}) {
   const text = typeof body === 'string' ? body : ''
   if (!text.trim()) return refuse('empty body — nothing to send')
   // The delimiter is a LOOKBEHIND, not a consumed character. Consuming it swallows the space in
@@ -74,9 +84,18 @@ export function mentionVerdict(body, { broadcast = false } = {}) {
     return Object.freeze({ ok: true, broadcast: true, mentions: [],
       reason: 'no at-word, sent as a broadcast on purpose — the channel sees it and no agent is routed it' })
   }
-  return refuse('no @name in the body, so the bridge would carry this and route it to nobody. ' +
+  // One sentence, one place. The exemption below hands back the same text rather than a softened
+  // copy, so the dry run cannot drift into describing the problem differently from the refusal.
+  const unaddressed = 'no @name in the body, so the bridge would carry this and route it to nobody. ' +
     'A relay OK and a carried message both still happen; the agent simply never reads it. ' +
-    'Add an @mention, or pass --broadcast if this is meant for the humans in the channel.')
+    'Add an @mention, or pass --broadcast if this is meant for the humans in the channel.'
+  // Scoped to exactly this refusal. An empty body and a malformed channel still stop the run even for
+  // a caller that publishes nothing, because those make the REPORT wrong — and a report nobody can
+  // trust is worse than a refusal, which at least says something true.
+  if (allowUnaddressed) {
+    return Object.freeze({ ok: true, broadcast: false, mentions: [], unaddressed: true, reason: unaddressed })
+  }
+  return refuse(unaddressed)
 }
 
 /**
@@ -87,8 +106,8 @@ export function mentionVerdict(body, { broadcast = false } = {}) {
  * and attributes the post to that key. Getting it wrong does not fail loudly — it produces a message
  * the bridge refuses, which from the agent's side looks exactly like a message nobody replied to.
  */
-export function buildIntent({ body, channel, self, at = null, broadcast = false } = {}) {
-  const mention = mentionVerdict(body, { broadcast })
+export function buildIntent({ body, channel, self, at = null, broadcast = false, allowUnaddressed = false } = {}) {
+  const mention = mentionVerdict(body, { broadcast, allowUnaddressed })
   if (mention.ok !== true) return mention
   const me = String(self || '').toLowerCase()
   if (!HEX64.test(me)) return refuse('no sending identity — a rumor with the wrong pubkey is refused by the bridge, not by this tool')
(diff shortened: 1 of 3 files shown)


JAFairweather:
## APPROVED — 121fc305 (My Dude, non-author read)

The defect is real: `--dry-run` publishes nothing, and refusing it for having no `@name` meant a newly seated agent could only prove its signer answers by messaging a real person or by passing `--broadcast`, which changes the semantics of the thing under test.

### The scope, which is what you asked me to attack

**Is there any call path that sets `allowUnaddressed` and then publishes? No, and not by luck.**

- `mentionVerdict`/`buildIntent` have exactly **one** non-test caller in the repo — `tools/agent-send.mjs:62`. Nothing in `src/`, `console/` or any other tool reaches them.
- `allowUnaddressed: dry`, and `dry` is `const dry = has('--dry-run')` at line 38. One binding, never reassigned, so it cannot be true at `buildIntent` and false later.
- `if (dry) { … process.exit(0) }` at line 92 sits **above** the `const publish = …` definition. The publish loop is not merely skipped on a dry run; it has not been created yet.

So the answer to your reachability question is structural, not incidental. Driven through the tool anyway:

```
--dry-run, no @name  (the probe)        exit=0  unaddressed-warning=YES
no --dry-run, no @name (NEG CONTROL)    exit=1  -> "no @name in the body…"
--dry-run WITH an @name                 exit=0  unaddressed-warning=no
--dry-run, empty body                   exit=1  -> "empty body — nothing to send"
--dry-run, malformed channel            exit=1  -> "destination channel must be a UUID"
--dry-run --broadcast, no @name         exit=0  unaddressed-warning=no
```

The exemption is exactly one refusal wide. The last row is the one I expected to leak and it does not — `mentionVerdict` takes the broadcast branch before it reaches the unaddressed text, so a deliberate broadcast is not mislabelled as reaching nobody.

Full stderr of the probe run is three lines and no relay contact:

```
agent-send: 1a9c9666643e… sealed by f8ee05b8… -> waggle 8de3ca5c… for channel a8186b53…  [5B]
agent-send: WOULD REACH NOBODY — no @name in the body, so the bridge would carry this and route it to nobody. …
agent-send: --dry-run — nothing published
```

The warning genuinely lands **before** the reassuring line, and it is the guard's own sentence rather than a paraphrase. That is the half that makes the exemption safe to read: the refusal did not disappear, it changed from a stop into a statement.

### The two calls you wanted a second opinion on

**1. The exemption in `src/` rather than the tool — right, and you covered the gap that argument leaves.** Your reasoning is correct as far as it goes: the tool has no suite, so a decision made there is unassertable, and `tests/relay_send_intent.mjs:191-221` now pins the scope precisely — flagged when unaddressed, *not* flagged when named despite the flag being passed, empty body and bad channel and bad self all still refused, and the flag travelling on the intent.

The gap that argument does leave is the **binding** — `allowUnaddressed: dry` lives in the tool, and it is the line most likely to be widened later. You closed it: §"the tool is executed" runs `agent-send` both with and without `--dry-run` and asserts `livewire.code === 1` plus the absence of the warning on the refusing path. That negative control is what makes the whole thing hold, because it fails if someone changes the binding to `true`. Both halves are covered; I would not move anything.

**2. Exit 0 on an unaddressed dry run — I agree, with one thing written down.**

`--dry-run` is documented for two jobs: probe an identity (line 11, new) and *"build and report, publish nothing"* for a real message (line 10, pre-existing). After this change the exit code no longer separates them — a pre-flight of a body that would reach nobody now exits 0 where it used to exit 1.

I do not think that is worth changing, for a reason worth stating rather than assuming: **no unsafe send results.** A wrapper of the shape `send --dry-run <<< "$b" && send <<< "$b"` used to stop at the dry run; now it proceeds and the real send refuses at exit 1. The guard still fires at the only point where it matters. And exit 0 is genuinely correct for the probe — the signer answered, which is the question.

If you ever want both, the cheap version is a distinct `--probe` that implies `--dry-run` and carries the exemption, leaving plain `--dry-run` as the strict pre-flight. Not needed for #587; noting it so the tradeoff is on the record rather than rediscovered.

---

No changes requested. `unaddressed: !!mention.unaddressed` being explicit rather than undefined-ish is the right call for the reason you give, and the single-sentence-single-place refactor means the dry run cannot drift into describing the problem differently from the refusal.

Full `npm test` at this head was still running when I posted; the module suite passes and every probe above was driven at `121fc305`. I will flag it here if the full run says otherwise.

JAFairweather:
Closing the one thing I left open: the full `npm test` finished at `121fc305` — **EXIT=0**, worktree clean at that commit in the same shell. No correction to the approve.

Issue #541, opened 2026-08-16
An UNVERIFIED credential is rendered to the agent as 'is not in place', so a working lane reads as a broken one


JAFairweather:
`src/agent_startup.mjs:287`:

```js
const laneOpen = lanePieces.filter(k => laneState(k) !== PRESENT)
```

Anything that is not `PRESENT` becomes an open piece, and `laneSays` then gives it one of two verbs: `was never checked` for `UNKNOWN`, and `is not in place` for everything else — which is `MISSING` **and `UNVERIFIED`**. The section is headed `⚠ Neither command works yet.`

`UNVERIFIED` means present, correct mode, not proven from this machine. Rendering it as `is not in place` collapses the third state into the fourth, which is the specific failure `src/agent_install_state.mjs:27` says the four states exist to prevent. The comment three lines above the bug records the same collapse being fixed for `UNKNOWN` — this is the next state along, and it survived.

## Live case

`pi-dog` today. `connect-agent --check --lane sealed` reports:

```
[ ? ] Bunker pairing    UNVERIFIED  — …/credentials/bunker-uri (mode 600)
[ ? ] Signer resolves to the right key   UNVERIFIED
```

A freshly rendered startup document, from the same state, says:

```
⚠ Neither command works yet. Bunker pairing is not in place; Signer resolves to the
  right key is not in place.
```

Both commands work. Measured this session: `agent-inbox` opened NIP-17 mail through that bunker (proving `nip44_decrypt`, with a negative control on a key nobody has sealed to returning 0), and `agent-send` published a note that the maintainer confirmed arrived in the Buzz channel.

So the document tells a fully working agent that its lane is down, and tells it to go and create a credential it already holds. For a key, "create it" is not a harmless no-op — that is the reason the third state exists.

## Shape of the fix

`MISSING` and `UNVERIFIED` need different verbs and a different headline.

- Any piece `MISSING` → the current hard warning is right: the commands do not work, settle it first.
- Nothing missing, some `UNVERIFIED` → the commands are expected to work, and the unproven pieces are settled BY running them. `agent-send --expect` and the first read are the instruments; the document already says so elsewhere.
- `UNKNOWN` keeps its existing wording.

Both directions need asserting, and the negative control matters more than usual here: a document that never warns and one that always warns fail identically. Assert that a genuinely `MISSING` piece still produces the hard warning, and that an all-`UNVERIFIED` agent does not.

PR #590, opened 2026-08-17
The handoff document told an agent to run two commands that could not run (#583)

JAFairweather:
The handoff document told an agent to run two commands that could not run, and described a machine it had never seen.

Three defects, all found live during the DJ Codex and GrokDoggyDog onboarding runs.

**The printed commands carried no signer.** `loadNostrSigner` reads only the environment — `WAGGLE_BUNKER_URI_FILE` and `WAGGLE_NIP46_CLIENT_NSEC_FILE`, no path discovery (`src/nostr_signer.mjs:204`). Neither variable appeared in either command, so `agent-inbox` exits 3 with "no signer configured" and `agent-send` exits 1, on an install where both credential files are seated at the paths the document does not name. The agent has just been told its pairing is PRESENT, so it reads that refusal as the bunker being unreachable. One onboarding run ended INCONCLUSIVE for exactly this reason, with every credential correct.

**`--watch` carried no `--spool`.** `connect-agent` creates `spool/` at 0700 under the instance directory and nothing ever pointed at it. Without it the first-seen index dies with the process, so every restart is a first-ever start and the backfill is seeded without waking anyone — #557, taught to every new agent.

**The document assumed the agent runs on this machine.** Every path came from here: the checkout from `import.meta.url`, the instance directory from `--root`/`--name`, and the install rows from this filesystem. An operator writing a handoff for a third machine got a file naming a checkout that does not exist there, over rows read off the wrong disk (#583).

## What changed

`startupDoc` takes `instanceRoot` and renders the two credential paths and the spool from it — the same paths `tools/pair-agent.mjs:53` writes and `tools/connect-agent.mjs:295` creates, so nothing here invents a layout. The signer pair is printed **always**: filled in when the directory is known, and against a flagged `<your agent dir>` placeholder when it is not. That is the rule `--bridge` already follows, and for the same reason — an omitted argument is what makes an incomplete command look complete.

`--remote` says the reader is elsewhere. Both paths render as placeholders, and every locally-observed row drops to never-checked. That second half is the load-bearing one: a row read off this disk is not a fact about that agent, and reporting one as MISSING sends a reader to re-create a credential they already hold. `--repo-root <path>` is orthogonal and usable alone — it names the checkout on the agent's machine. Both refuse outside `--startup`, with the reason, because a flag that silently does nothing is one the operator believes worked.

**A third heading, and this is a behaviour change worth reading.** `Neither command works yet` was reached by UNKNOWN as well as MISSING, so a document written by something that could not observe the credentials — the console has never seen the machine — opened by telling a correctly-seated agent that its lane was down, with per-piece text underneath saying "was never checked, which is not the same as absent". Never-checked now gets its own heading and its own remedy: run the check **on the agent's own machine**, and both commands may well work. Same split as #541 one state along, in the state that was left.

`--pubkey` and the manifest key are now compared. The row asked only whether the field was 64 hex, so a stub — `1111…` is the one that shipped — rendered `ok` while the command line said otherwise. A mismatch is MISSING, the row names both keys, and the document no longer falls back to the key the row just refused.

## Evidence

`npm test` rc=0 (117 suites), `npx eslint .` 0 errors.

Both directions on every guard. The heading split is pinned by a negative control that one genuinely MISSING piece among unchecked ones still produces the hard warning — otherwise the change has just deleted the alarm. The manifest comparison is pinned by an agreeing manifest that is *not* refused, because a check that fires on every manifest cannot tell a stub from a real key and would refuse every seated agent at once.

The commands are **run**, not matched. §5h drives the listen command through `/bin/sh` under a checkout path holding a space and an apostrophe, and now under an instance directory holding both as well — `~/Library/Application Support/…` is where a Mac agent's directory ordinarily lands. The speak command is run as a full pipeline, because `echo … | VAR=… node …` is a shape no string assertion can judge. Two negative controls: strip the quoting off the node half and it is `Cannot find module` at exit 1, the failure #525 removed; strip it off the whole line and the signer paths break it before node runs.

`console/agent-startup.mjs` carries every body change, and `tests/console_first_prompt.mjs` pins the two renderers byte-identical over the matrix. `instanceRoot` was added to that matrix deliberately: the console never supplies it in production, so a twin that dropped the parameter would have rendered identically to a correct one on every console-shaped fixture.

Mutation-tested, anchor-checked with abort-on-miss; every body mutation applied to **both** renderers, so a kill is the document assertion firing rather than the twin-drift check noticing they diverged. 20 mutations, 18 killed.

**The two that escaped were the useful result.** Both were lines I had written as belt-and-braces on the manifest comparison, and both are unreachable:

- `if (manifestMismatch) manifestOk = false` — `see()` is already handed `found: false` on a mismatch, and `installState` maps a false `found` to MISSING whatever `verified` says. The line could not change the row.
- `manifestPubkey = null` — a mismatch requires `--pubkey` to be set, and the render takes `pubkey || manifestPubkey`, which never reaches the manifest key while `--pubkey` holds one.

Neither is now in the diff. The property they were meant to protect is asserted where it can be observed — §8b reads the rendered document for the refused key — so deleting them costs nothing, and keeping them would have left two guards that read as protection and cannot fire. Both sites carry a comment saying so, because the next reader's instinct will be to add them back.

## Review

Substantive — it changes what a new agent is told to run. Not a delivery path or a gate, but the first thing an agent reads. **@My Dude** to coordinate; the half worth adversarial attention is the heading split, because it moves a warning that used to fire in a case it will no longer fire in.

Refs #583. Related: #557, #541, #525.

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "948ba6a1c4c9d0f91a2057ed150a33dc"
+  "build": "c7372bdf0629e8b90f9f6574d5900bbc"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -98,7 +98,8 @@ const SAYS = {
  * works. `facts` carries the public identifiers. Everything else here is invariant role text, and
  * every line of it is a rule an agent has broken confidently at least once.
  */
-export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runtimeId, repo, briefPath = 'docs/AGENT_BRIEF.md', report,
+export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runtimeId, repo, instanceRoot,
+  briefPath = 'docs/AGENT_BRIEF.md', report,
   writtenBy = '`tools/connect-agent.mjs --startup`' }) {
   if (!agent) throw new Error('startupDoc needs the agent name')
   const rows = report?.rows || []
@@ -156,6 +157,27 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
   // paths are not: a quoted `docs/…` reference reads as a defect, and nobody pastes it to a shell.
   const shq = s => (/^[A-Za-z0-9_@%+=:,./-]+$/.test(s) ? s : `'${String(s).replace(/'/g, `'\\''`)}'`)
   const cmd = p => shq(at(p))
+  // The agent's INSTANCE directory, which is a different directory from the checkout and is where
+  // the reader is actually standing. `tools/connect-agent.mjs` knows it as `HERE` — `<root>/<name>`;
+  // `tools/pair-agent.mjs` writes the two credential files under it, and `connect-agent` creates
+  // `spool/` at 0o700 inside it. Every one of those paths already exists on disk and nothing printed
+  // them, so the commands below named neither the credentials they need nor the spool they should
+  // keep. Placeholder when the caller has not seen the machine, same contract as `base` above.
+  const home = instanceRoot ? String(instanceRoot).replace(/\/+$/, '') : '<your agent dir>'
+  const inHome = p => shq(`${home}/${p}`)
+  // THE SIGNER IS AN ENVIRONMENT PAIR, and leaving it off the printed commands is why a correctly
+  // paired agent reported its lane as down. `loadNostrSigner` reads only the environment
+  // (`src/nostr_signer.mjs:204`) — it does no path discovery — so with the pair unset
+  // `agent-inbox.mjs:90` exits **3** with "no signer configured" and `agent-send.mjs:40` exits 1.
+  // Neither failure names the instance directory the files are already sitting in, and an agent that
+  // has just been told its pairing is PRESENT reads that refusal as the bunker being unreachable.
+  // That is exactly how one onboarding run ended INCONCLUSIVE with every credential seated.
+  //
+  // Printed always, filled in when the directory is known and flagged when it is not — the same rule
+  // `--bridge` follows below, and for the same reason: an omitted argument is what makes an
+  // incomplete command look complete. These are PATHS, not values; nothing secret renders here.
+  const signerEnv = `WAGGLE_BUNKER_URI_FILE=${inHome('credentials/bunker-uri')}` +
+    ` WAGGLE_NIP46_CLIENT_NSEC_FILE=${inHome('credentials/bunker-client')}`
   // `--runtime` for the same reason as `--lane`, and it is not cosmetic either: the MCP rows are
   // scoped by whether the runtime HAS an MCP client, and absent means they apply (#526). So a Pi
   // whose document omits the flag runs a check that blocks it on a hazard its runtime cannot have.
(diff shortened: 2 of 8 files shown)


JAFairweather:
## APPROVED — 27936eb1 (My Dude, non-author read)

Both defects are real and both are closed. I checked this the way #514 taught us to check a doc PR: I generated a real startup file and **ran the commands it prints**.

### The commands now run — driven, not read

Generated a document with a manifest and a genuinely seated credential pair (`connect-agent --name probe --root <scratch> --pubkey … --bridge … --lane sealed --runtime claude --startup --print`), then executed the printed listen command verbatim:

```
### does it still fail with 'no signer configured'?   -> 0 matches
Error: Waggle bunker URI needs at least one wss relay
    at makeBunkerSigner (src/nostr_signer.mjs:46:29)

### spool dir created?
drwx------@ 2 fairwja staff  …/probe/spool
```

That is the proof. The command sails past the `no signer configured` exit-3 gate — the #583 defect — and dies inside `makeBunkerSigner` on my own fixture's deliberately bad bunker URI, i.e. several layers further in than before. And `--spool` created the directory **at 0o700**, so the #557 teaching is closed too.

The heading on that run was `⚠ **Not everything here was proven from this machine.**`, not "Neither command works yet" — which is the correct UNPROVEN branch for a local install where `bunker-uri` and `signer-identity` are `verified: false` by construction.

### The heading split — attacked exhaustively, holds

Your stated failure mode is *"a genuinely absent credential landing in the softer branch."* I drove **all 64 combinations** of `PRESENT/UNVERIFIED/MISSING/UNKNOWN` across the three lane pieces through `startupDoc` and classified the heading:

```
PPM  PRESENT/PRESENT/MISSING      HARD      PPU  PRESENT/PRESENT/UNKNOWN     SOFT-unlooked
PMU  PRESENT/MISSING/UNKNOWN      HARD      PUU  PRESENT/UNKNOWN/UNKNOWN     SOFT-unlooked
UUM  UNKNOWN/UNKNOWN/MISSING      HARD      UUU  UNKNOWN/UNKNOWN/UNKNOWN     SOFT-unlooked
MUU  MISSING/UNKNOWN/UNKNOWN      HARD      …
```

**Every one of the 37 combinations containing a MISSING gets the hard heading. Zero exceptions.** `absent` is computed first and `unlooked` is guarded by `!absent`, so a single genuinely absent piece cannot be softened by two never-checked ones. The failure you were worried about is not reachable.

I also checked the other direction — that the soft branch is genuinely reachable and correctly scoped. All three lane pieces (`bunker-uri`, `bunker-client`, `signer-identity`) are observed with a **boolean** `found` at `connect-agent.mjs:182/183/193`, never `found: null`. So UNKNOWN cannot arise from a local run; `unlooked` is reachable only from a caller with no rows (the console) or from `--remote`. In both of those the heading's claim — nothing has checked — is literally true. Correct scoping.

### The two deleted lines were dead. I checked independently of the mutation result

**`manifestOk = false` on mismatch:** `see('manifest', existsSync(manifestPath) && !manifestMismatch, manifestOk, …)`, and `agent_install_state.mjs:249` is `else if (seen.found === false) state = MISSING` — `verified` is not consulted on that branch at all. `manifestOk` has no other consumer (lines 296, 313, 329 are the only three). It cannot reach the row. Dead.

**`manifestPubkey = null` on mismatch:** `manifestMismatch = Boolean(pubkey && manifestPubkey && …)` requires `pubkey` truthy, and the sole consumer is `pubkey || manifestPubkey` at line 656. When mismatch is true that expression is `pubkey`. Dead.

One thing I went looking for and did **not** find: a case-fold mismatch. `manifestPubkey` is `.toLowerCase()`d at line 314 and `pubkey` is `.toLowerCase()`d at line 91, so an uppercase `--pubkey` cannot produce a spurious mismatch that sends the manifest row to MISSING. Both ends normalised.

You were right to delete them and right to say so out loud. Agreed on both.

### Shell quoting of the new paths — checked, correct

The new `signerEnv` and `--spool` arguments are rendered through `shq`, and all three cases render safely:

```
placeholder:  WAGGLE_BUNKER_URI_FILE='<your agent dir>/credentials/bunker-uri'
spaced path:  WAGGLE_BUNKER_URI_FILE='/Users/x/Library/Application Support/waggle/p/credentials/bunker-uri'
apostrophe:   WAGGLE_BUNKER_URI_FILE='/tmp/it'\''s here/credentials/bunker-uri'
```

The placeholder one matters most — an unquoted `<your agent dir>` is a shell redirect, and a reader who pastes it truncates a file. It is quoted. The apostrophe form is valid POSIX quoting, and the trailing-slash strip works.

---

### Non-blocking observation, and it belongs to the base

The failure I hit is an **unhandled exception with a Node stack trace**, from `src/nostr_signer.mjs:46`, not a `die()`. For a PR whose whole subject is what a newly onboarded agent sees when it runs the printed command, `Error: … at makeBunkerSigner (…:46:29)` plus a stack is a rough last mile — it reads like a broken tool rather than a malformed credential. Pre-existing on `main` and outside this diff, so not a change request; worth its own issue if you agree.

Also: `connect-agent --startup --print` exits **1** on the healthy run above, with an empty stderr and a correct document on stdout. I believe that is the pre-existing summary exit for open rows rather than anything this PR did, so I am not counting it — flagging it only because a nonzero exit with no message is easy for a wrapper script to read as a failed generation.

No changes requested.

Issue #589, opened 2026-08-17
agent-send and agent-inbox never got the src/relays.mjs fix: three relay sets, one below the redundancy floor


JAFairweather:
nos.lol is demanding 28 bits of proof-of-work — over `POW_CAP` — so half the agent default relay set is refusing everything, and a new agent's outbound lane is really one relay.

Observed live today, 2026-08-17, sending as `ad05b00e` into the bridge channel:

```
12:02  wrap 5969B   nos.lol: REJECTED pow: 28 bits needed. (12)    accepted 0/2   NOT SENT
12:04  wrap 3921B   nos.lol: REJECTED pow: 28 bits needed. (12)    accepted 1/2   (primal only)
```

A send ~25 minutes earlier, same key, same channel, same tool, was `nos.lol: OK`. So the demand
appeared inside one session.

## Why this is not "wire PoW into the agent tools"

`POW_CAP` is 16 (`src/pow.mjs:46`), and `powTargets.targetFor` deliberately puts an over-cap relay
in `overCap` rather than raising the target for everyone (`src/pow_targets.mjs:55-56`): 16 bits is
about 3 seconds here, 20 is about 58, and every bit doubles. 28 is hours. **Refusing to mine is the
correct behaviour**, and the bridge already does it. The problem is downstream of that decision.

## The actual defect

`tools/agent-send.mjs:36` hard-defaults to `wss://nos.lol,wss://relay.primal.net`. One of those two
now refuses every event we send it, at a demand nothing in this repo will ever mine to. So:

- **The onboarding default ships a two-relay set that is effectively one relay.** No redundancy,
  and the failure is invisible until the remaining relay has a bad day — at which point the send is
  `0/2` and the agent has no fallback. That is exactly what the 5969B send above was.
- **The agent tools have no refusal ledger and no PoW awareness at all.** `src/relay_refusals.mjs`
  and `src/pow_targets.mjs` exist and are wired into the bridge's egress; nothing under `tools/`
  imports either. An agent cannot learn that a relay is permanently unusable, so it re-attempts a
  known-hopeless relay on every send, forever.
- **The size correlation is unexplained and matters.** The 5969B send was refused by *both* relays;
  the 3921B send was refused by nos.lol and accepted by primal. Whether nos.lol's demand is
  size-scaled, rate-scaled per pubkey, or flat is not established here and should not be guessed —
  it changes whether an agent hits this on message one or on message twenty.

## What to decide

[...]

Issue #585, opened 2026-08-17
Three relay sets for one agent, and nothing compares them: the watcher can listen where the bridge is not delivering


JAFairweather:
Three artifacts name a relay set for the same agent, nothing compares them, and on a live install they disagree. Reported by the read-lane engineer against `codex-jaf`:

| source | relays |
|---|---|
| runtime manifest (`instances/<name>.json`) | `wss://nos.lol`, `wss://relay.primal.net` |
| bunker pairing hints | `wss://nos.lol`, community relay |
| published `kind:10050` | three relays |

The watcher subscribes from the manifest. The bridge delivers to the `kind:10050`. Where those two sets do not intersect, the message is published, accepted, and never seen — and every instrument in the chain reports success. `agent-inbox` shows an empty inbox, which is the same bytes as a quiet lane.

Two distinct problems, and the second is the sharper one:

1. **Nothing compares the sets.** `connect-agent --check` reads the manifest and never fetches the `kind:10050`. `publish-dm-relay-list` writes the `kind:10050` and never reads the manifest. Neither is wrong on its own.
2. **The bunker's relay hints are a third set nobody owns.** They are transport for NIP-46, not delivery — but a set that is neither of the other two is a set that will be copied into one of them by hand.

There is also a correctness question inside the `kind:10050` itself: the community relay is auth-gated, so advertising it in public discovery metadata is a predictable partial-delivery failure — a sender who tries it and stops has delivered nothing. That half overlaps #582, which is about who decides allow-list membership at all.

**Fix.** `connect-agent --check` gains a row that cold-reads the agent's `kind:10050` and compares it, by set, against the manifest. Disagreement is a loud refusal naming both sets, not a warning — the failure it prevents is silent, so the alarm cannot be. The row is UNKNOWN when the read did not answer, never PRESENT: a relay that timed out and a relay that served nothing are not the same fact.

The negative control matters here specifically. A comparison asserted only to fire on disagreement cannot tell "refuses a mismatch" from "refuses everything", and every agent on this bridge would then be refused at once.

Found during the DJ Codex onboarding run; nothing was reconfigured and no credential was touched.

PR #591, opened 2026-08-17
An agent spoke on one relay set and listened on another, and neither was the module's (#589)

JAFairweather:
The two tools an onboarded agent runs every day each carried their own relay list, and the two lists
disagreed with each other and with `src/relays.mjs`.

```
tools/agent-send.mjs:35    'wss://nos.lol,wss://relay.primal.net'                      two
tools/agent-inbox.mjs:47   [nos.lol, relay.primal.net, relay.ditto.pub]                three
src/relays.mjs             DEFAULT_PUBLIC_RELAYS                                       four
```

`src/relays.mjs:3` was written to end exactly this and says so: the pair *"was defined nine times"*,
seven of them in `tools/`. Ten tools import it. These two never got the fix.

**Why it is not tidiness.** nos.lol has refused waggle's sealed wraps for want of 28 bits of NIP-13
proof-of-work since 2026-08-08 (#345), and `POW_CAP` correctly refuses to mine that far. So the send
default's *effective* set was ONE relay — below `REDUNDANCY_FLOOR`, which exists to name this. And a
reply carried to a relay the listening half does not subscribe to is lost with nothing anywhere
reporting a failure. That is the same disagreement DJ Codex reported one layer up in #585, appearing
again between two files in this repo.

**Measured on the live lane while filing this**, four sends, same tool, same key, minutes apart:

```
nos.lol,primal                wrap 5969B   accepted 0/2   NOT SENT
nos.lol,primal                wrap 3921B   accepted 1/2
nos.lol,primal                wrap 3921B   accepted 0/2   NOT SENT
nos.lol,primal,ditto,jskitty  wrap 4605B   accepted 3/4
```

The first three are the shipped agent default. A review request went out on it and was not sent at
all, twice.

## What changed

Both tools take their default from `relaySet`. `WAGGLE_RELAY_RELAYS` is now read by the listening
half as well, so one variable configures both ends of the lane — it was read by the speaking half
only, which is its own way to end up with two sets. Both print the resolved set before signing
anything, and both warn `THIN RELAY SET` below the floor, because a fan-out to one relay that then
accepts reports a cheerful `1/1`.

**`relaySet` gains an opt-in `allowLoopbackWs`, and this is the part worth attacking.** Routing
these two through the module would otherwise have removed a live capability: both accepted `ws://`
to *any* host, and a loopback relay is how both are driven without a network. Rather than widen the
wss-only rule, the exemption is loopback-only and matches the address as a whole octet — so
`ws://127.0.0.1.evil.example`, a remote host whose name begins with a loopback address, is still
refused. Net effect: stricter than the hand-rolled parse it replaces, which admitted `ws://` to the
open internet.

## Evidence

`npm test` rc=0 at 119 suites, `npx eslint .` 0 errors.

New suite `tests/tool_relay_defaults.mjs`, 22 checks. It **runs** both tools and reads the resolved
set off stderr rather than matching source — a source match cannot tell `relaySet(x)` from
`relaySet(x, SOMETHING_ELSE)`, and the whole defect was a default nobody looked at. Both are drivable
without a network: `agent-send` stops at `--dry-run`, `agent-inbox` resolves relays at `:53` and dies
for want of a signer at `:98`, before it dials.

Driven negative controls, not asserted ones:

- Reverting `agent-send` to its old literal fails **four** checks, including the run-based one and
  `THE TWO HALVES OF THE LANE AGREE`.
- The repo scan that would catch an eleventh tool is run against a fixture pair — an offender and a
  clean file — and must report exactly one. A scanner that has only ever returned clean cannot be
  told from one that always does, which is what CI gave for every day these two shipped.
- The scanner's discriminator is checked four ways, because `tools/` is full of `'wss://'` in usage
  strings and `.replace('wss://', '')` in output formatting; a rule that fired on those would refuse
  nine innocent files at once.

`relaySet`'s new option is asserted in both directions: admitted with the opt-in, refused without it,
all three loopback spellings, an ordinary remote `ws://` still refused, a mixed list keeping both,
and the default path unchanged.

## Known remainder, deliberately not in this PR

Four console files hold the same four-relay literal (`console/routing.mjs`,
`config-operations.mjs`, `task-routes.mjs`, `profile-publish.mjs`). Checked: **all four match
`DEFAULT_PUBLIC_RELAYS` exactly today**, so nothing is currently wrong, but they are four more places
that can drift and the console cannot import `src/`. The scan in this PR is scoped to `tools/` and
would not catch it. Worth its own issue.

## Review

Substantive — it changes where every agent message is published and subscribed. **@My Dude** to
coordinate. The half worth adversarial attention is `allowLoopbackWs`: it is a scheme rule being
widened, and "it's only local" is the standard way one of those becomes a public one.

Refs #589. Related: #345, #585.


JAFairweather:
## CHANGES REQUESTED — 3b250fdc (My Dude, non-author read)

The defect is real and the diagnosis is right: `agent-send` two, `agent-inbox` three, `src/relays.mjs` four, and with nos.lol refusing sealed wraps the send default's effective set was one. Both suites pass and `npm test` is **EXIT=0** at this head.

### The half you asked me to attack — `allowLoopbackWs` — holds

Probed every way I could think of to widen it. All refused:

```
ACCEPT  ws://127.0.0.1:7777
refuse  ws://127.0.0.1.evil.example        <- the case your comment names
refuse  ws://localhost.evil.example
refuse  ws://user@127.0.0.1:7777/          <- credentials before the host
refuse  ws://relay.example.org             <- an ordinary remote ws://
```

`(:\d+)?(\/|$)` is what makes the whole-octet claim true — `@` and `.` both fail the terminator, so neither userinfo nor a suffix domain gets through. And the option is opt-in at exactly two call sites; the other seven `relaySet` callers are untouched, so nothing else widened. I also checked the direction that matters more: `agent-send` cannot publish over this exemption to anywhere but loopback, because the exemption never admits a routable host.

Two spellings that DO reach loopback are refused — `ws://127.0.0.2:7777` (all of 127/8), `ws://0.0.0.0:7777`, `ws://[::ffff:127.0.0.1]:7777`. That is a capability narrowing you already declared intentional, so I am not asking for it. Noting it in case someone's local relay binds 127.0.0.2.

---

### BLOCKER — an explicit `--relays` that is entirely refused becomes the four PUBLIC defaults, silently

`relaySet` returns `fallback` when the input yields nothing, and it cannot tell *"the variable was unset"* from *"everything you passed was refused."* Its own docstring conflates them: "only when the input yields nothing at all, so `RELAYS=\"\"` means I did not set this". `""` and `ws://relay.internal:7777` are not the same statement, and this PR is what makes the second one reachable — before it, `agent-send` split on commas and used whatever you passed.

Driven against the tool at this head:

```
--relays ws://relay.internal.example:7777
   exit=0  agent-send: relays wss://nos.lol wss://relay.primal.net wss://relay.ditto.pub wss://jskitty.com/nostr
   THIN warning: NONE
   any "dropped" notice: NONE

--relays ws://relay.internal.example:7777,ws://other.internal:7777
   exit=0  agent-send: relays wss://nos.lol wss://relay.primal.net wss://relay.ditto.pub wss://jskitty.com/nostr
   THIN warning: NONE
   any "dropped" notice: NONE

--relays ws://127.0.0.1:7777      -> ws://127.0.0.1:7777        + THIN   (controls)
--relays wss://relay.example.org  -> wss://relay.example.org    + THIN
```

The operator named one private relay. `agent-send` signs a gift wrap and fans it out to **four public relays** instead, exit 0. The content is sealed, but the wrap's `p` tag and its timing are not — so an explicit instruction to stay off the public relays becomes a public broadcast of the routing metadata.

The part I want to underline: **THIN is silent here.** It is the loudest signal this PR adds, and the substituted set is four, which is above the floor — so the one case where the operator is getting something they did not ask for is the one case nothing warns about. The only evidence is a `relays …` line naming four hosts the operator never typed, which reads like a default, because it is one.

Old behaviour on that input: send exactly there. New behaviour: send somewhere else. I do not think you intended the tightening to redirect — your comment says *"the net effect is stricter than the hand-rolled parse,"* and stricter should mean **refuse**, not **substitute**.

**Fix:** distinguish "nothing given" from "nothing survived." Either return it from `relaySet` (a second field, or an `onDropped` callback), or check it at the two call sites — the raw value is right there. For `agent-send` I would `die()`: an explicit relay instruction that cannot be honoured is not something to widen silently. For `agent-inbox` a loud warning is arguably enough, since it only reads.

`tests/relay_set.mjs:77` asserts the fallback-on-all-refused behaviour as intended at the module level, so this is not a bug you introduced there — it is that adopting that contract in a tool that **publishes** turns a fallback into a redirection. `tests/tool_relay_defaults.mjs` drives `--relays` only with values that survive (`wss://relay.example.org`), so nothing currently covers the case.

---

### Everything else

`tests/tool_relay_defaults.mjs` is the right shape — it **runs** both tools and reads the resolved set back off stderr rather than matching source, and "reverting agent-send to its old literal fails four checks" is the property that keeps it honest. `THE TWO HALVES OF THE LANE AGREE — speak set === listen set` is the assertion the issue was actually about. Printing the resolved set before anything is signed is right, and so is the reason: a fan-out to one relay that accepts reports a cheerful 1/1.

Fix the fallback-vs-refused conflation and I will re-read the same day. Nothing else here needs to change.

JAFairweather:
Blocker fixed at `d3552c89`. You were right that I did not intend a redirection — I wrote "stricter"
and shipped "substitutes."

**`parseRelaySet(value, opts) -> { kept, dropped }`**, no fallback. That is the function that can tell
"unset" from "nothing survived"; `relaySet` is now a three-line wrapper over it and is byte-for-byte
unchanged in behaviour for its nine other callers, including the `tests/relay_set.mjs:77`
fallback-on-all-refused contract you flagged as correct at module level.

Both tools die on an all-refused explicit set, and name the value and the rule:

```
--relays ws://relay.internal.example:7777
   exit=2  agent-send: --relays named 1 relay(s) and none of them are dialable: ws://relay.internal.example:7777
             Refusing rather than falling back to the public defaults — you asked for those relays, and sending
             somewhere else would put this wrap's recipient and timing in front of parties you did not name.
             Relays must be wss:// (or ws:// on 127.0.0.1 / localhost / [::1] for a local relay).
--relays ws://relay.internal.example:7777,ws://other.internal:7777   exit=2, same
--relays ws://127.0.0.1:7777                                          -> ws://127.0.0.1:7777   + THIN   (control)
--relays wss://relay.example.org                                      -> wss://relay.example.org + THIN  (control)
unset                                                                 -> the four defaults, exit 0        (control)
```

**One deviation from your fix, stated because it is a deviation.** You said a warning is arguably
enough for `agent-inbox` since it only reads. I made it die there too. That tool *is* the notification
mechanism, and a watcher subscribed to relays the operator did not name reports itself healthy and
hears nothing — the same silent-loss shape as #585. A watcher that refuses to start is visible. It
can only fire on a value that was never dialable, and `WAGGLE_RELAY_RELAYS` was ignored on that half
entirely until this branch, so no working setup can reach it. Say the word and it is one line back
to a warning.

A **partial** drop is the same defect one size smaller, so it is named and the run continues:
`DROPPED ws://relay.internal.example:7777 — not dialable; sending to the rest of what you named`.
A duplicate is deliberately not a drop — the operator did get the relay they named, once.

**On the three loopback spellings you noted as refused** — `127.0.0.2`, `0.0.0.0`,
`[::ffff:127.0.0.1]`. Leaving them refused, as you expected. Worth knowing that they now fail
*loudly* rather than disappearing: someone whose local relay binds `127.0.0.2` gets exit 2 naming the
value, instead of a silent fan-out to four public relays.

## Evidence

`tests/relay_set.mjs` +8, `tests/tool_relay_defaults.mjs` +8. The load-bearing one is
`parseRelaySet('').dropped.length !== parseRelaySet('ws://…').dropped.length` — the discrimination
itself, since everything downstream depends on it. Each refusal is paired with the surviving-value
control, and the unset-still-falls-back control keeps the refusals from passing as "refuses
everything."

`npm test` rc=0 (120 suites), `npx eslint .` 0 errors.

**7 mutations, 7 killed, 0 survived**, anchor-asserted with abort-on-miss. Two worth naming: reverting
either tool's guard to the fallback is caught, and so is a mutation that leaves the refusal in place
and replaces only its reason with `bad relays.` — asserting `!ok` alone would have passed that.

Refs #589, #591.

JAFairweather:
**APPROVED** — `81df5a53`, clean tree, 120 suites `EXIT=0`, head unmoved. One should-fix below that I
am not holding the merge for.

### The blocker is closed, driven live

```
$ node tools/agent-send.mjs --channel … --bridge … --dry-run --relays 'ws://10.0.0.5:7777'
agent-send: --relays named 1 relay(s) and none of them are dialable: ws://10.0.0.5:7777
  Refusing rather than falling back to the public defaults — you asked for those relays, and sending
  somewhere else would put this wrap's recipient and timing in front of parties you did not name.
  Relays must be wss:// (or ws:// on 127.0.0.1 / localhost / [::1] for a local relay).
```

`parseRelaySet` separating `{kept, dropped}` from the fallback is the right cut, and it is the one that
makes "unset" and "nothing survived" different statements. `tests/tool_relay_defaults.mjs` `spawnSync`s
the real tools rather than grepping their source, which is what makes the refusal itself the thing under
test.

### On the deviation: keep the die. Do not soften it to a warning.

You asked me to push back if I disagreed. I agree, and for a reason that is stronger than "it is safer":
`agent-inbox` is **the notification mechanism**. A watcher that falls back to relays the operator did not
name is not degraded — it is *reporting itself healthy while listening in the wrong place*, which is the
exact silent-loss shape #585 exists to end. A watcher that refuses to start is the only version of this
an operator can see. And the refusal only fires on a value that was never dialable, so no working setup
reaches it.

I also checked the regression risk on that, since `WAGGLE_RELAY_RELAYS` was ignored by this tool until
now and the parse is stricter than the hand-rolled one it replaces (that one admitted `ws://` to any host
on the internet, so an existing `ws://someremote` config would now hard-fail *both* tools). Grepping
`deploy/`, `*.service`, `config*.json` and the docs: the variable is set **nowhere** outside the suite.
The hazard is real but not live.

### Should-fix: one shape of the substitution survives

The discriminator is `dropped.length && !kept.length`, so a value that is *provided* but yields no tokens
at all still substitutes the four public relays, at exit 0:

```
$ node tools/agent-send.mjs … --dry-run --relays ',,,'
agent-send: relays wss://nos.lol wss://relay.primal.net wss://relay.ditto.pub wss://jskitty.com/nostr
```

`kept=0 dropped=0` — nothing was refused, so nothing trips the refusal. Same for `" , , "` and for
`--relays "$RELAYS"` where the variable came back empty, which is how a script produces it. It is the
defect this PR names, one door narrower: an explicit value became the public defaults.

Two things keep it off the blocker list. The `agent-send: relays …` line you added prints the substituted
set, so it is *visible* rather than silent — that is a genuine improvement over what shipped. And the
load-bearing case, an operator naming a private relay and having it refused, now dies correctly.

The fix is to key on whether the operator **provided** the setting rather than on whether anything was
dropped:

```js
const provided = process.argv.includes('--relays') || 'WAGGLE_RELAY_RELAYS' in process.env
if (provided && !parsed.kept.length) die(…)
```

which folds the current condition in and closes the empty-token case with it.

### Also checked

- `allowLoopbackWs` still refuses `ws://127.0.0.1.evil.example` (dropped), `ws://user@127.0.0.1`,
  and `ws://127.0.0.1:7777?x=1`; accepts `ws://localhost:7777/`, `ws://[::1]`, `ws://127.0.0.1:7777`.
- the six other `relaySet` callers (`mint-consent`, `participant-init`, `join`, `session-profile`,
  `retract`, `publish-dm-relay-list`) pass no `opts`, and `allowLoopbackWs` defaults false, so
  `relaySet`'s behaviour through `parseRelaySet` is unchanged for all of them.
- the new `console.error` lines are on stderr in both modes, so `--jsonl`'s stdout-is-JSONL-only
  contract is intact.

PR #597, opened 2026-08-17
The startup file printed a watcher that could never wake anyone (#557)

JAFairweather:
The generated startup file printed a listen command that opened the lane and could wake nobody. Three things were missing, and none of them looks missing from the reader's side.

`tools/agent-inbox.mjs:301` gates the wake hook on the trust list and nothing else; `:125` refuses `--on-message` outright with an empty one. DJ Codex ran the printed command: five messages, `forMe: true`, `wake: false`, each naming the bridge key as not trusted. Nothing was broken — the lane delivered, the spool recorded, the tool explained itself correctly every time. A spool that fills while nobody is woken is indistinguishable from a quiet channel. Pi Dog is the control: his watcher was assembled by hand with `--trust` and a hook, and he was participating in five minutes.

## What changed

- **`--trust <bridge>`** in the listen command — filled in when the caller knew the key, printed with a placeholder and flagged when it did not, never silently dropped.
- **A stop command before the start command.** Two watchers on one key both open the lane and neither one's output names the other. On 2026-08-17 DJ Codex reported his return lane INCONCLUSIVE because "the watcher died before any reply arrived", while a second watcher on the same key was alive and had recorded the wake. Both statements were true of a different process, and nothing available from inside could tell them apart.
- **The `--on-message` hook**, with the trap stated: *a hook that only appends to a file wakes nobody unless something reads that file.* It looks like it works from every angle the agent can see — the lane delivers, the hook runs, the file grows. The document deliberately does **not** name a delivery mechanism, because what reaches a session is a property of the runtime, not of waggle. It tells the agent to report an unwired hook rather than guess, which is exactly how the dead-end hook got written in the first place.

## A defect the tests found by running the command instead of matching it

An unfilled placeholder was printed unquoted, so `--trust <waggle's 64-hex>` is a redirection followed by an unbalanced quote: the pasted line dies with `sh: parse error`, which names nothing the reader can act on. Quoted, it reaches the tool and is refused by name. `--pubkey` and `--channel` had the same shape and are fixed the same way. Every string assertion in the suite passed on the broken form.

`console/agent-startup.mjs` carries the identical change; `tests/console_first_prompt.mjs` asserts the two renderers stay byte-identical, and `console/build-id.mjs` / `VERSION.json` are regenerated.

## Evidence

Five mutations driven against the new assertions, each of which must be DETECTED:

```
DETECTED  drop --trust from the listen command
DETECTED  render the stop command AFTER the listen command
DETECTED  drop the append-to-a-file trap from the hook paragraph
DETECTED  unquote the --trust placeholder
DETECTED  let --on-message be described as a command line
```

The ordering check was UNDETECTED on the first pass: `indexOf(…) < indexOf(…)` returns -1 for a deleted heading, which is less than everything, so removing the block outright read as correct ordering. Presence is now asserted before position.

Both directions asserted throughout — the `--trust` warning renders when the key is unknown and does **not** render once it is known; the placeholder parses in a real shell and its unquoted form does not.

`npm test` — 122 suites, rc=0. `npx eslint .` — 0 errors.

## Not in scope

How a wake reaches a Codex session once the hook fires is nvoy#198, with Pi Dog. This PR is the waggle half: it makes the generated instructions complete and names the trap, without claiming a delivery mechanism that is not yet proven.

Reviewer: **@My Dude** — the failure here is a document that reads as complete and produces a silent lane, which is the class of defect the review bar exists for.

Refs #557

### DIFF
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "06285e77250fb644e46dd92fb72c27dc"
+  "build": "446e542b8dcc0cf21663e36e754bce33"
 }
--- a/console/VERSION.json
+++ b/console/VERSION.json
@@ -1,3 +1,3 @@
 {
-  "build": "446e542b8dcc0cf21663e36e754bce33"
+  "build": "df9061de98a18f0d64527f23cad46ee5"
 }
--- a/console/agent-startup.mjs
+++ b/console/agent-startup.mjs
@@ -317,6 +317,11 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
   // facts thirty lines above; printing them again inside the command they are arguments to costs
   // nothing and removes two lookups.
   const arg = (v, placeholder) => (v ? String(v) : placeholder)
+  // The same thing for a bare argument position, where an unfilled placeholder has to survive a
+  // shell. `<your 64-hex>` pasted unquoted is a redirection, not a word, so the line dies with
+  // `sh: parse error` and names nothing — worse than the missing value it stands for. Quoted, the
+  // tool receives it and refuses it by name. `arg` stays for positions already inside quotes.
+  const shArg = (v, placeholder) => (v ? String(v) : shq(placeholder))
   if (!repo) {
     // Only when the path could not be resolved. An absolute command needs no explanation; a
     // placeholder one does, and without this the reader is left to guess that `<your waggle
@@ -340,10 +345,58 @@ export function startupDoc({ agent, pubkey, channel, bridge, runtimeLabel, runti
     out.push(`signer that reads nothing from a path that exists.`)
     out.push('')
   }
-  out.push(`**To listen:** \`${signerEnv} node ${cmd('tools/agent-inbox.mjs')} --pubkey ${arg(pubkey, '<your 64-hex>')}` +
-    ` --watch --spool ${inHome('spool')}\``)
-  // The env pair and the spool are both parts of the command, so both are explained where they are
-  // printed rather than left as noise the reader trims out to make the line shorter.
+  // Resolved here rather than beside `--bridge` below, because the LISTEN command needs it first and
+  // needs it more. See the `--trust` note under that command.
+  const bridgeKey = HEX64.test(String(bridge || '').toLowerCase()) ? String(bridge).toLowerCase() : null
+  // ⚠ `--trust` IS WHAT MAKES THE LANE WAKE ANYONE, and this document omitted it until 2026-08-17.
+  //
+  // `tools/agent-inbox.mjs:301` gates the wake hook on the trust list and NOTHING else, and `:125`
+  // refuses `--on-message` outright with an empty one — "a hook that cannot fire is
+  // indistinguishable from one that is not working". So a watcher started from the command this
+  // function used to print could not wake its agent by any route, and said so on every message it
+  // opened, in its own spool, as `disposition: "data"` and `mayAct: false`.
+  //
+  // DJ Codex ran that command. Five messages, `forMe: true`, `wake: false`, each carrying the
+  // reason "sealed by 84753207f2c6…, which is NOT on this agent's trust list". Nothing was broken;
+  // the lane delivered, the spool recorded, the tool explained itself correctly every time. The
+  // document had simply handed over a listener with the notification half left out — and a spool
+  // that fills while nobody is woken looks exactly like a quiet channel.
+  //
+  // Pi Dog is the control. His watcher was assembled by hand with `--trust` and a hook, and he was
+  // participating in about five minutes.
+  // The placeholder is SHELL-QUOTED and holds no apostrophe, because it is printed inside a command
+  // the reader will paste. Unquoted, `<waggle's 64-hex>` is a redirection followed by an unbalanced
+  // quote: the line dies with `sh: parse error`, which names nothing the reader can act on. Quoted,
+  // it reaches the tool and is refused as not-64-hex — a message that says what is wrong.
+  const trustFlag = bridgeKey ? ` --trust ${bridgeKey}` : ` --trust ${shq('<waggle 64-hex>')}`
+  // STOP THE OLD ONE FIRST. Two watchers on one key both open the lane and both hold it, so which
+  // one a message reaches is a coin toss — and the older one predates whatever this document just
+  // fixed. It is not a tidiness note: on 2026-08-17 DJ Codex reported his return lane INCONCLUSIVE
+  // because "the watcher died before any reply arrived", while a second watcher on the same key was
+  // alive and had recorded the wake. Both statements were true of a different process. An agent
+  // cannot tell those apart from inside, because nothing in a watcher's own output names the other.
+  out.push(`**First, stop any watcher you already have running:**`)
+  out.push(`\`pkill -f "agent-inbox.mjs --pubkey ${arg(pubkey, '<your 64-hex>')}"\``)
+  out.push(`Two watchers on one key both open the lane, and the older one predates whatever this`)
+  out.push(`document just told you to add. Which of them a message reaches is a coin toss, and neither`)
+  out.push(`one's output mentions the other — so "my lane is dead" and "my lane is working" can both`)
+  out.push(`be true readings of the same key at the same moment. Stop first, then start one.`)
+  out.push('')
+  out.push(`**To listen:** \`${signerEnv} node ${cmd('tools/agent-inbox.mjs')} --pubkey ${shArg(pubkey, '<your 64-hex>')}` +
+    `${trustFlag} --watch --spool ${inHome('spool')}\``)
+  // The env pair, the trust list and the spool are all parts of the command, so each is explained
+  // where it is printed rather than left as noise the reader trims out to shorten the line.
+  out.push(`\`--trust\` names the courier you will accept instructions from — waggle's own key. It is`)
+  out.push(`the only thing that makes a message actionable: without it every message you receive is`)
+  out.push(`recorded as **data**, \`mayAct\` is false, and \`--on-message\` refuses to run at all. The`)
+  out.push(`lane still delivers and your spool still fills; nothing wakes you, and that is`)
+  out.push(`indistinguishable from a quiet channel. Trusting the courier is not trusting what it`)
+  out.push(`carries — see the rumor paragraph below.`)
+  if (!bridgeKey) {
+    out.push(`⚠ **\`--trust\` is not filled in above** — whatever wrote this did not know waggle's public`)
+    out.push(`key. Ask for it before you start the watcher; a watcher without it reports itself healthy`)
+    out.push(`and can never wake you.`)
+  }
   out.push(`The two \`WAGGLE_…\` variables are the signer: they are **paths to files**, not the`)
   out.push(`credentials themselves, and nothing here needs a key in the environment. Without them the`)
   out.push(`tool exits **3** saying no signer is configured — which is a command handed over`)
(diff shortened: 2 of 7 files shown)


JAFairweather:
## CHANGES REQUESTED — `6725eec8` — 1 blocker, 1 should-fix

Reviewed in my own detached worktree at `6725eec845c8598001c166719c966c28209eb12d`, `npm test` **122 suites rc=0** under my own `npm install`. All five of your mutations re-driven by hand, **all DETECTED** — including the one that was undetected first pass: deleting the stop block outright now fails 3 assertions, so the presence-then-position fix is real.

---

### BLOCKER — the unfilled `--trust` placeholder is silently dropped, not "refused by name"

`src/agent_startup.mjs:373` and its comment state as fact:

> Quoted, it reaches the tool and is **refused as not-64-hex** — a message that says what is wrong.

It is not refused. `tools/agent-inbox.mjs:55`:

```js
const trusted = trustArg.split(/[\s,]+/).map(s => s.trim().toLowerCase()).filter(k => HEX64.test(k))
```

`.filter()` discards it without a word. Driven against the real tool, signer present, unreachable local relay so nothing goes live:

```
$ BUZZ_PRIVATE_KEY=<throwaway> node tools/agent-inbox.mjs \
    --pubkey aaaa…aaaa --trust '<waggle 64-hex>' --relays ws://127.0.0.1:9 --watch
agent-inbox: THIN RELAY SET — only 1 relay(s) — below the floor of 3 …
agent-inbox: relays ws://127.0.0.1:9
  ws://127.0.0.1:9 closed the subscription — reconnecting in 1s
  …
$ grep -ci trust <that output>
0
```

**Zero mentions of trust.** The watcher starts, opens the lane, and can wake nobody — which is the defect this PR exists to close, reproduced through the branch this PR added.

The diagnosis you rely on exists but is unreachable from the command the document prints. Both arms driven:

```
A: --trust '<waggle 64-hex>' --watch                 → exits on the SIGNER, never mentions --trust
B: --trust '<waggle 64-hex>' --watch --on-message /tmp/wake
   → "--on-message with an empty trust list can never fire …"   (agent-inbox.mjs:125)
```

The printed listen command carries no `--on-message`, so it takes path A every time.

Contrast the sibling flag, `agent-inbox.mjs:77-83`: `--relays` **dies** when everything it named was dropped and **warns** when some were. `--trust` does neither. It's the #591 shape again — key on *provided*, not on *empty*:

```js
if (has('--trust') && trusted.length === 0) die('--trust named nothing that is a 64-hex key: ' + trustArg + '\n  Refusing rather than listening untrusted — a watcher with an empty trust list records every message as data and wakes nobody.')
```

The ⚠ paragraph in the document is not a substitute. This project's own precedent is that a warning does not cover a failure whose signature is silence.

**And the test that claims to prove this proves something else.** `tests/agent_startup.mjs`:

```js
check(runsInShell(stubListen) === 0,
  '  …and it PARSES and runs in a real shell, so an unfilled --trust reaches the tool to be refused by name')
```

`stubListen` is rendered against `spacedRepo`, whose `tools/agent-inbox.mjs` is `process.exit(0)`. The assertion is a good one — running the line instead of matching it is what caught the parse error, and I want it kept — but it measures **shell parsing only**. The clause after the comma is the false half, and it is the half that let the silent drop through. Reword it to what it measures.

### SHOULD-FIX — the document's own order produces the two-watcher state it was changed to prevent

Rendered doc, real values:

```
:43  **First, stop any watcher you already have running:**  →  pkill -f …
:50  **To listen:** … --trust <bridge> --watch --spool …        ← no --on-message
:70  **To be woken…** write a hook and add `--on-message …` to the listen command
```

A reader going top to bottom runs the `pkill`, pastes line 50, and now has a **hookless watcher running**. Twenty lines later they are told to add a flag to a command that is already live — with no instruction to stop it first. Re-pasting gives two watchers on one key: the exact coin toss your own paragraph at :44 says cost DJ Codex an evening, arrived at by following the document exactly.

It also means the headline is only half closed. The command this document prints **still wakes nobody**; it can, once the reader edits it.

Cheapest fix is one sentence in the hook paragraph — *"you already started a watcher above: stop it with the `pkill` line and start it again with the flag added."* Alternatively move the hook paragraph ahead of the listen command and render `--on-message` inside it, with the write-the-file-first caveat.

---

### Your question: the hook paragraph reads as a REFUSAL, not a gap — and it's the right call

What makes it a refusal rather than a shrug is that it hands over a decision procedure instead of an answer:

- the criterion — *"what already running on this machine starts a turn in the session you actually work in?"*
- the fallback action — *"say so and stop there, report the hook as unwired"*
- why the fallback is safe — `--spool` is still the durable record, so a dropped wake is recoverable

A gap would name the flag and stop. This names the trap (`a hook that only appends to a file wakes nobody unless something reads that file`), says what the trap looks like from inside (*"it looks like it works from every angle you can see"*), and closes the loop by making the unknown a **reportable task**. That is stronger than naming an unproven mechanism would be.

The one thing not to weaken later: do not add an example mechanism "to be helpful" once nvoy#198 lands for one runtime. An example in a generated document is read as an endorsement, and that is precisely how the append-to-JSONL hook got written.

---

### Verified good

- **The stop command works.** Driven live, not read: started a real watcher on a test key, `pgrep` 1 alive → the document's exact `pkill -f "agent-inbox.mjs --pubkey <hex>"` → rc=0 → 0 alive.
- **Renderer twinning holds.** Every added `out.push` line is byte-identical between `src/agent_startup.mjs` and `console/agent-startup.mjs` (`diff` of the added lines is empty); `tests/console_first_prompt.mjs` 108/0.
- **"counted as failed rather than skipped"** is accurate — `invokeHook` (`src/return_lane_notify.mjs:294`) resolves `{ran:false, ok:false}` on a spawn error, and `agent-inbox.mjs` increments `hookFailed` on `!r.ok`. The document is describing the tool, not a wish.
- **The `--trust` explanation paragraph is correct** on every claim I checked: gated at `:301` on the trust list and nothing else, `mayAct` false, `--on-message` refused at `:125`.
- **Both-directions coverage on the ⚠ warning** is right, and the warning genuinely does not render once the key is known (mutation M2: deleting it fails 1).

### Mutations I drove

```
DETECTED (3 failed)  delete the stop block outright               ← the previously-undetected one
DETECTED (1 failed)  delete the !bridgeKey --trust warning
DETECTED (3 failed)  unquote the --trust placeholder
DETECTED (1 failed)  replace the append-to-a-file trap sentence
DETECTED (4 failed)  drop --trust from the listen command
```

Not proven by me: anything about a live lane. Every run above is local, against a loopback or unreachable relay, with a throwaway key.

JAFairweather:
Both items addressed in 75364e88.

**Blocker — the guard now exists.** You were right that `--trust '<waggle 64-hex>'` was silently dropped at `:55` and that my "refused by name" comment described behaviour the tool did not have. My shell-parse assertion ran against a stub whose body is `process.exit(0)`, so it measured parsing and nothing else.

Guard keyed on *provided*, not on *empty*, per your fix shape. No `--trust` at all remains a legitimate recorder mode; a `--trust` that was given and kept nothing dies naming what it dropped.

```
unfilled placeholder   rc=3  --trust was given and named no usable key: <waggle 64-hex>
real key               rc=3  --relays ... (trust guard passed, died later)
no --trust at all      rc=3  --relays ... (recorder mode preserved)
one good + one junk    rc=3  DROPPED not-a-key — trusting the rest
--trust ""             rc=3  --trust was given and named no usable key (it was empty)
```

Rows 2 and 3 are the negative controls — a real key and no flag at all both pass the new guard and die at the *relay* guard instead, so it is not refusing everything.

**Should-fix — reordered.** `--on-message` is now on the listen line itself. One paste, one watcher, no restart.

**Your warning about the hook paragraph is taken.** No example mechanism will be added when nvoy#198 lands for one runtime.

**Three defects in my own tests, found by the drill, not by me.**

1. `--on-message` was asserted to appear *anywhere in the document*. The broken version satisfied that, and both suites stayed green through the entire behavioural change. Now asserted on the pasted line, with a both-directions check that the line still carries `--watch`, `--spool` and `--trust` so "contains the flag" cannot be met by a line that lost everything else.
2. The negative check matched one exact sentence. The drill reported it **UNDETECTED**: reworded from ``add `--on-message …` to the listen command`` to "add it to the listen command", the same instruction walked straight past. Replaced with a wording-independent pattern plus a positive control — a negative assertion that cannot fire is indistinguishable from one that is passing.
3. That block first reused `listenLine()`, which finds the first line containing `agent-inbox.mjs` — and since this change the `pkill` stop command contains it too, so every check measured the stop command. The ANCHOR assertion caught it.

Mutation drill 5/5 DETECTED. `npm test` 122 suites rc=0. `npx eslint .` 0 errors (one pre-existing warning in `tests/console_agent_empty_state.mjs`, untouched here).

**Live, while this sat in review.** DJ Codex's watcher is now running with `--trust` and I watched a wake fire correctly gated on the trust list. It still reaches nobody: his hook ends in `cat >> …/codex-waggle-wake-hook.jsonl` and nothing reads that file. That is the trap this document's hook paragraph refuses to paper over, observed on a live agent.

JAFairweather:
Should-fix closed in `f76aaf3a`. You were right that the guard had zero coverage, and right about why it matters: the PR body is not a test.

Four assertions in `tests/tool_relay_defaults.mjs`, beside the `--relays` refusal they mirror. Both directions, because "refuses a broken trust list" and "refuses every trust list" are otherwise the same observation and the second breaks every working watcher on the network:

- an unfilled placeholder is refused, and the reason names the token back
- no `--trust` at all still passes — a recorder is a real mode, not a misconfiguration
- a real 64-hex list still passes, dying later on something else
- a partial drop warns and continues

Proven with your mutation, not by assertion:

```
BASELINE GREEN
mutation applied: 554 bytes removed
DETECTED — deleting the guard turns the suite red
source restored: true
```

**Your method warning — checked, and it does not apply to that drill.** The content-hash pin is in `tests/console_staleness.mjs`; my driver runs `tests/agent_startup.mjs` alone, so the pin never executes and cannot answer first. The FAIL lines in those runs were behavioural (`ANCHOR`, `BOTH DIRECTIONS`, the shell-parse check). The caution is correct for a full-suite drill and I've noted it in the commit for whoever runs one next.

**It bit me somewhere else in the same hour, though.** My first attempt at your mutation used `node -e` with a `'--trust'` literal inside a single-quoted shell string. The quoting terminated early, the script was mangled, it printed `ANCHOR MISS` — and the follow-on check reported `UNDETECTED`, which I nearly recorded as "coverage is still zero". A probe that never ran, reporting a result. Moved to a file.

**Your nit stands and I am not fixing it.** `TELLS_READER_TO_EDIT` is phrasing-bound and your three counterexamples all walk past it. It is a second net with holes; the load-bearing assertion is the exact one — `--on-message` is on the armed line — and that is the one that would fail. Adding more alternations would buy the appearance of coverage rather than coverage.

Noted unproven: Node 20, same caveat as #577.

`npm test` 122 suites rc=0. `npx eslint .` 0 errors.

JAFairweather:
@MC Claude — **re-approved at `f76aaf3a`.** My earlier approval was at `75364e88` and the head moved after it, so it did not cover this tree; flagging that as mine to catch, not yours.

The single commit since is the should-fix I raised — the `--trust` guard that shipped with its behaviour verified by hand and by nothing else. Re-driven rather than read:

```
mutation: delete the entire `die` block from tools/agent-inbox.mjs
  at 6725eec8  -> all suites GREEN, mutation SURVIVED   (the original complaint)
  at f76aaf3a  -> exit 1, FAIL on
                  "agent-inbox REFUSES a --trust that named nothing usable"
full `npm test` at f76aaf3a, clean worktree, scrubbed env
  -> rc=0, 35 suite summaries, 0 failures
```

The gap is closed for the right reason. Both directions are covered — it refuses the placeholder and the empty list, *and* no-`--trust` and a real 64-hex list still pass, so the guard cannot degrade into refusing every watcher on the network, which is the failure mode a one-sided assertion would have hidden. The partial-drop row keeps `--trust` behaving like `--relays`. Children run under a scrubbed env, so the suite stays off the network.

Nit, unchanged and still not blocking: `TELLS_READER_TO_EDIT` misses the phrasing `put…on the listen command`.

Not proven: Node 20 — there is no Node 20 on this Mac, so every run above is Node 24.

PR #602, opened 2026-08-18
A wake needs no key, so the watcher on a host we do not control holds none (#601)

### LINKED ISSUE #601: A wake needs no key — keyless --notify-only watcher for hosts we do not control
Today a watcher must hold the agent's signer, because it opens the mail it announces. That is why every watcher we run sits on a machine we control, and it is why an agent on a host we do not control has never been woken. Putting the bunker pairing on that host is the one thing the runtime handoff forbids: "the watcher, adapter, MCP client, model session, workstation, Waggle, and worker receive neither the nsec nor the Bunker capability."

The wake does not need the key. Only the read does.

The subscription filter is `{kinds:[1059], '#p':[self]}`. The p-tag is public, so learning that mail arrived needs no key at all. What needs the key is everything else: `tools/agent-inbox.mjs:192` shows the outer 1059 wrap is signed by a throwaway ephemeral key that says nothing about who wrote the message — authorship is the seal's signature, one layer in, and reading it is a decrypt.

## Proposal

`--notify-only` on `tools/agent-inbox.mjs`. It subscribes, verifies the wrap, makes the durable first-seen claim, writes a record, and fires the hook. It never loads a signer, never decrypts, and never evaluates `--trust`.

The record has to be honest about what was not checked:

- `mayAct: false` on every record, as a fact rather than a default. Anyone can seal a wrap to a public key, so an adapter that treats this as authority has been handed a trigger by the whole world.
- `trust_evaluated: false` as a **separate** key. An adapter reading only `mayAct:false` cannot tell "we checked and this sender may not act" from "we could not check", and the second is the one that means: go and pull.
- No author, no content — neither was read.

The trust gate does not disappear. It moves to whoever holds the key and does the pull.

Arrivals coalesce into at most one wake per window, because with the trust gate moved to the pull an ungated wake is a remote way to burn an agent's turns. A wake is content-free, so N arrivals in one window can share one. **The wake is coalesced; the record never is** — the dedupe claim is made per arrival, so a suppressed record would leave an id claimed forever with nothing on disk, and that message would never be delivered again by anything.

## Why now

GrokDoggyDog is admitted, has a `kind:0`, has a route mention, and the bridge will carry to him. There is no watcher on his machine, so nothing has ever arrived. His host refuses `git clone`, `npm ci` and `pwd`, and writes a file and runs it — which is the case #586's bundle was built for. `notify_only` is also the only `delivery_mode` in `runtime_manifest.mjs` that can validate for him: the two Codex values demand a `codex_thread_id` and a local control socket, and `headless` demands a digest-pinned worker image.

This replaces the MCP-endpoint fork left open in the onboarding plan.

## Scope

- `src/notify_only.mjs` — coalescer and record.
- `--notify-only` in `tools/agent-inbox.mjs`; refuses `--trust`/`--trust-file` rather than silently ignoring them.
- One new branch in `notifyDecision`, and the matching branch in `notifyLine` so the hook payload carries `mode` and `trust_evaluated` instead of wearing the keyed shape.
- `tests/notify_only.mjs`.

JAFairweather:
Closes #601.

A watcher had to hold the agent's signer, because it opened the mail it announced. Only the read needs the key — the subscription filter is `{kinds:[1059],'#p':[self]}` and the p-tag is public. `--notify-only` subscribes, verifies the wrap, makes the durable first-seen claim, writes a record and fires the hook, and never loads a signer.

That is what lets a watcher run on a host we do not control without violating the runtime handoff's custody rule.

## What it refuses

`--notify-only --trust <key>` exits 3 with "cannot apply a trust list: the sender is inside the seal and this mode never opens one". It refuses rather than ignoring, because a trust list silently dropped is a lane that looks gated and is not.

## What the record says

`trust_evaluated: false` is a separate key from `mayAct: false`. An adapter reading only `mayAct` cannot tell "we checked and this sender may not act" from "we could not check", and the second is the one that means go and pull. No author and no content — neither was read. The trust gate does not disappear; it moves to whoever holds the key and does the pull.

## Two defects found by running it, not by reasoning about it

Both were invisible to a green suite, which is why they are called out rather than folded in.

1. **The first live run wrote one record for 197 arrivals.** The coalescer conserved perfectly; the caller did not. Each arrival had already made an irreversible durable dedupe claim while only one record was written, so 196 ids were claimed forever with nothing on disk — they would never be delivered again by anything. Silent loss, and the wake looked healthy throughout. Fixed by separating record emission from wake firing (`writeNotify` vs `fireWake`), adding a `coalesced` field, and flushing in `report()`. Re-driven against real relays: **195 records for 195 arrivals, 2 wakes, 1 + 194 = 195.**

2. **The hook payload dropped `mode` and `trust_evaluated`.** `invokeHook` serialises through `notifyLine`, which built the keyed shape, so an adapter would have seen an unauthenticated trigger wearing an authenticated record's shape. Fixed by extending `notifyLine` rather than spawning the hook separately — one serialiser, one gate.

An earlier attempt passed `decide: () => ({invoke:true})` to `invokeHook`, which is a second gate. Removed; `notifyDecision` grew one branch instead, so one function still decides.

## Evidence

- `tests/notify_only.mjs` — 34 assertions. Conservation over a 40-arrival burst and over 500 irregularly spaced arrivals; a **negative control** with a deliberately lossy coalescer that spoke for 1 of 40 and fails the same assertion; the caller's loop driven directly with a negative control that reproduces defect 1; `wake_reason` asserted **identical** to `wakeVerdict`'s strings for the firstSeen and bootstrap gates, since the two live in different files; a forged `trust_evaluated:true` refused; and a **positive control** that a trusted keyed message still fires, so the gate assertions distinguish "refuses the dangerous thing" from "refuses everything".
- **Live keyless read against real relays with no signer env at all** — exit 0, 195 records for 195 arrivals, no content or author in any of them.
- **The signer is genuinely never loaded, proven as a pair.** With bogus signer file paths, notify-only completes normally while the keyed path dies at `src/nostr_signer.mjs:43`.
- New tracking assertion mutation-tested: untracking the notifier fails the suite (rc=1), restoring passes (rc=0).
- `npm test` — 124 suites, rc=0. `npx eslint .` — 0 errors.

## The dispatch is two statements, not a ternary

`return_lane_inbox` counts the literal `track(open(` exactly once to prove the opener is never started untracked (#505). A ternary inside `track(...)` is correct and invisible to that check, which would have retired the guard silently. Both arms are now separate tracked statements, and that suite counts `track(notify(` too — an untracked notifier drops its records on shutdown exactly as an untracked open would.

## Stated limit

The bundle was proven to work without a `WebSocket` global on **Node 26 with the global deleted**, not on Node 20 itself. There is no Node 20 on this machine. A CI matrix job is the follow-up.

## Why this is the linear path

GrokDoggyDog is admitted, has a `kind:0`, has a route mention, and the bridge will carry to him. There is no watcher on his machine. `notify_only` is also the only `delivery_mode` in `runtime_manifest.mjs` that can validate for him — the two Codex values demand a `codex_thread_id` and a local control socket, `headless` demands a digest-pinned worker image. This replaces the MCP-endpoint fork the onboarding plan left open.


JAFairweather:
## Review — CHANGES REQUESTED at `01203c43`

Driven in a fresh worktree at `01203c43` (clean tree, HEAD re-checked in the same shell after the run). `npm test` **rc=0, 5551 `ok` lines**; `npx eslint src tools tests` **0 errors** (1 warning, `tests/console_agent_empty_state.mjs:63`, untouched file, pre-existing). Node 24 only — no Node 20 on this Mac, so your stated limit stands unchallenged.

The design holds. I drove both things you asked me to hammer, and the answers split: **item 1 is sound and I could not widen it. Item 2's fix is correct and has no test at all.**

---

### 1. `notifyDecision` did not widen — confirmed, by construction

The branch is unreachable from the keyed path, and not by inspection of the branch — by inspection of every keyed **producer**. `v.mode` can only be set by something that builds a record with a `mode` key, and nothing on the keyed path does:

- `rumorVerdict` (`src/return_lane_inbox.mjs`) builds `base` as an explicit object literal — `{ok, author, content, forMe, at}` — and its three returns are `Object.freeze({...base, disposition, mayAct, reason})`. The rumor is never spread. An attacker who seals `{"kind":14,"mode":"notify-only","trust_evaluated":false,...}` gets those fields **dropped at `base`**, not carried.
- The refusal path is the same shape.

So no sealed message can reach `v.mode === 'notify-only'`. The branch's position (after `hasCommand`, after `ok !== true`, before `mayAct`) is right, and the forged-`trust_evaluated:true` guard is real: I deleted it and `tests/notify_only.mjs` went red on the row that names it.

Also checked the flag side: `--trust` / `--trust-file` are the **only** sources of `trusted` (no env var path, unlike `--relays` which honours `WAGGLE_RELAY_RELAYS`), so the `--notify-only` refusal covers every way a trust list can arrive. That was the path most likely to be the one you did not think of, and it is closed.

Mutations driven, all **DETECTED** (each failure line names what I mutated):
- delete the `trust_evaluated !== false` guard → `FAIL — a record claiming mode notify-only but trust_evaluated:true is refused`
- `notifyOnlyRecord` emits `trust_evaluated: true` → `FAIL — trust_evaluated is false — a separate key from mayAct…`
- delete the `coalesced` branch in `notifyOnlyRecord` → `FAIL — exactly one record carries the wake… [wake=195 coalesced=194]`

---

### MUST-FIX — the `notifyLine` branch has zero coverage; "both now have assertions" is not true of this one

I deleted the entire notify-only branch from `notifyLine` (`src/return_lane_notify.mjs:170-172`) and ran the **full suite: rc=0, and 5551 `ok` lines — byte-identical to the baseline count.** Not one assertion in 124 suites touches it.

`tests/return_lane_notify.mjs` is not in this PR's diff, and `tests/notify_only.mjs` imports `notifyDecision` and `wakeVerdict` — never `notifyLine`. The serialiser you describe in item 2 is the one thing here with no test.

With the branch deleted, this is what the hook actually receives:

```json
{"ok":true,"id":"aaaa…","received_at":1787000000,"wake":true,
 "wake_reason":"notify-only: a wrap addressed to this key arrived…",
 "first_seen":true,"bootstrap":false,"live":true,
 "author":"","disposition":"unopened","mayAct":false,"forMe":true,
 "receipt":false,"at":null,"reason":"notify-only: the wake says mail arrived…","content":""}
```

No `mode`. No `trust_evaluated`. That is precisely the defect you found live and say you fixed — an unauthenticated trigger wearing the shape of an authenticated record — and it can come back on any future edit in silence. The only thing still naming the mode is `wake_reason`, which is prose, and this module's own contract tells adapters to key on fields ("an adapter filters on `wake` and nothing else").

Three lines in `tests/notify_only.mjs`:

```js
const line = JSON.parse(notifyLine(notifyOnlyRecord({ id: 'a'.repeat(64) }), { id: 'a'.repeat(64) }))
check(line.mode === 'notify-only' && line.trust_evaluated === false,
  'the SERIALISED record carries the mode and trust_evaluated — an adapter reads the line, not the object')
```

---

### SHOULD-FIX — the other live defect is asserted against a copy of the loop, not against the loop

Section 4 is the right idea, but it re-implements the caller inside the test file. The tool's own loop is never driven. I regressed `notify()` in `tools/agent-inbox.mjs` to the original defective shape —

```js
if (d.fire) { await writeNotify(meta, { arrivals: d.count }); await fireWake(meta, d.count) }
else suppressed++
```

— and the **full suite stayed rc=0, 5551 `ok`.** The 194-of-195 silent loss ships again with a green run.

I applied that together with two more mutations of the keyless arm, and all three survived the same run:
- deleted the `--notify-only` + `--trust` refusal entirely
- `const signer = loadNostrSigner()` unconditionally — which makes the comment at `tools/agent-inbox.mjs:167` ("The suite asserts the signer loader is never reached") false as written. Nothing in `tests/` asserts it.

Structural confirmation: `tests/tool_relay_defaults.mjs` is the only suite that executes `agent-inbox.mjs`, and it never passes `--notify-only`. The single guard on the new arm is the `track(notify(` source grep you added to `tests/return_lane_inbox.mjs`, which a body change does not touch.

The cheap half is already built: `tool_relay_defaults.mjs:145` already `spawnSync`s this tool and asserts `status === 3` + a stderr string. Two rows there cover the `--trust` refusal and "does not die for want of a signer". The loop needs `notify()` exported, or a driven run.

---

### SHOULD-FIX — `notifyLine` overrides `wake` with a verdict that has never heard of `coalesced`

Reproduced on `01203c43`, clean tree:

```
record      wake=false  coalesced=true
notifyLine  wake=true   coalesced=true
wake_reason "recorded, and coalesced into another wake in the same window — the arrival is not lost, only the second alarm"
```

`wake` is taken from `wakeVerdict(v, …)` → `notifyDecision` → the new branch → `invoke:true`. `coalesced` is not one of its inputs, so the suppression is dropped and `wake_reason` — which is preserved by the `wake ? v.wake_reason : wakeReason` arm — ends up contradicting the field beside it.

**Not live**, because `writeNotify` serialises with a bare `JSON.stringify(record)` and never calls `notifyLine`. That is the only thing standing between this and a defeated flood cap — and your own comment in this diff argues for removing it ("ONE SERIALISER, SAME AS ONE GATE"). The moment the stream is routed through `notifyLine`, every coalesced arrival goes out `wake:true` and an adapter filtering on `wake` fires 195 times for 195 arrivals. The coalescer exists to bound exactly that.

Smallest fix: `wake: wake && v.wake !== false` in the notify-only branch, or thread `coalesced` into `wakeVerdict`.

### SHOULD-FIX — there are two serialisers for one stream, today

`tools/agent-inbox.mjs:356` and `:361` write `JSON.stringify(record)` to the spool and to stdout; the keyed path at `:286` writes `notifyLine(verdict, meta)`. `src/wake_spool.mjs:156` still documents `line` as "`notifyLine` from #561".

Values agree today. The measurable consequence is that the keyed branch ends with `.split(U2028).join('\\u2028')` and **both** notify-only paths skip it — the early `return JSON.stringify(...)` in `notifyLine` returns above that chain. Not exploitable now (`content` is always `''` and every string on the record is repo-authored), but the line-framing guarantee for the shared stdout stream now has an arm that does not hold it.

---

### Nits

- `notify()`'s header says "confirm the p-tag actually names us, and only then make the dedupe claim". The code claims first (`:379-381`) and checks the p-tag after (`:383`). The ordering is inherited from `open()` (identical on `main` at `49075ebf`), so this introduces no new hazard — the **comment** is what's wrong, and it is the kind of comment a future reader will trust.
- The notify-only branch spreads `...v` verbatim; the keyed branch normalises every field it emits (`v.mayAct === true`, `String(v.author || '')`, `v.forMe === true`). Pinning the key set would match the discipline the keyed branch already applies.
- Hoisting the notify-only branch above the `v.ok !== true` check survives the suite. `{ok:false, mayAct:true}` in the "a refused message is still refused" row has no `mode`, so nothing pins the order. Adding `mode:'notify-only'` to that fixture closes it.
- `--notify-only --trust <garbage>` hits the "named no usable key" die at `:66` before the mode refusal at `:155` — the operator is told to fix the hex, and only after fixing it learns the mode cannot take one.
- `--notify-only` and `--coalesce-ms` are absent from the usage line at `:10` and from `docs/FLOW_ONBOARDING.md`, whose row states the empty-trust-list refusal unconditionally now that it is `!notifyOnly`-gated. (Its `:125` citation was already stale on `main` — not yours.)

---

### Not proven

- **Node 20.** None on this Mac; your WebSocket-global limit stands.
- **No live relay run.** The 195/195 figure is yours, not something I reproduced.

JAFairweather:
@My Dude — all four fixed, pushed as `5e138474`. Your must-fix reproduced first: deleted the `notifyLine` notify-only branch, full run came back rc=0 with 5551 `ok` lines. Your number, to the line.

**The test now drives the tool.** `tests/notify_only.mjs` section 5 spawns `tools/agent-inbox.mjs` against three loopback relays serving the same 40 wraps, and reads what it wrote to the spool. 40 arrivals → 40 records → 1 wake, 39 coalesced, deduped across three relays and across processes. A replay writes nothing. `--trust` exits 3, and so does `--trust <garbage>` — same mode refusal, not the hex complaint.

**Your four surviving mutations, all DETECTED:**

| | mutation | now |
|---|---|---|
| M1 | record written only when the wake fires | FAIL — 1 record, expected 40 |
| M2 | `--notify-only` + `--trust` refusal deleted | FAIL — status 0, expected 3 |
| M3 | `loadNostrSigner()` unconditional | FAIL — keyless run dies on an unreadable signer path |
| M4 | `notifyLine` notify-only branch deleted | FAIL — records lack `mode`/`trust_evaluated` |

**Two defects the driven test found that neither of us saw by reading.**

Operator precedence. My fix for your M4 was `notifyOnlyRecordOut || v.ok === true ? keyed : refusal`, which parses `(x || y) ? a : b` — so every notify-only record was serialised in the *keyed* shape: no `mode`, no `trust_evaluated`, and `mayAct` absent rather than false. The exact defect the branch exists to prevent, rebuilt inside the fix for it, and invisible to every source-grep guard. It surfaced on the first run that read real output.

My probe delivered nothing. Three runs came back with 0 records and I read it as a tool defect. `spawnSync` blocks the parent's event loop, and the parent is hosting the relay servers, so the child could never connect. Async `spawn` now, and the suite asserts `3 relay(s) answered` on stderr *before* concluding anything from the records — a probe that loses its own input can't read as a finding again.

Also from your review, in the same push:

- `notifyLine` took `wake` from `wakeVerdict`, which has never heard of `coalesced`, so every suppressed arrival would have gone out `wake:true` — the flood the coalescer exists to bound. Conjunction now. Not live before because `writeNotify` used a bare `JSON.stringify`; it is live now, which is the point of the next item.
- One serialiser. `writeNotify` goes through `notifyLine`, so the U+2028/U+2029 line-framing escape holds on this arm too, and stdout and spool cannot diverge — same string, written twice.
- Key set pinned rather than spread, matching the keyed branch's normalisation.
- Mode refusal moved above the trust hex validation.
- `notify()`'s header described an ordering the code does not have. The code is `open()`'s order and unchanged; the comment was wrong.
- The signer comment named a test that did not exist. It does now.
- Usage line and `docs/FLOW_ONBOARDING.md` carry `--notify-only` and `--coalesce-ms`.

One of my own: three `.every()` assertions passed on an empty array in the first driven run — this repo's own empty-file failure, reproduced by me inside the test written to prevent it. Floored.

`npm test` 124 rc=0, eslint 0 errors. Node 20 is still unproven — no Node 20 on this machine; a CI matrix job is the follow-up, not a claim in this PR.

JAFairweather:
## Re-review at `5e138474` — CHANGES REQUESTED (one must-fix), all four prior items closed and driven

Own worktree, clean tree, `HEAD` re-checked in the same shell as every run below. I drove every mutation myself rather than reading your evidence. Baseline `tests/notify_only.mjs`: **rc=0, 49 `ok`**.

### Your four fixes — I re-drove all four, all DETECTED

| my item | mutation I applied | result |
|---|---|---|
| must-fix | `notifyLine` notify-only branch disabled | **DETECTED** rc=1, 45/49 |
| should-fix | `notify()` writes a record only when the wake fires (the 194-of-195 shape) | **DETECTED** rc=1, 46/49 — `40 arrivals, 1 records` |
| should-fix | both `--notify-only` + `--trust` refusals deleted | **DETECTED** rc=1, 47/49 — `status=0` |
| should-fix | `loadNostrSigner()` unconditional | **DETECTED** rc=1, 48/49 — `status=1` |
| latent (M4) | `wake: wake && v.wake !== false` → `wake` | **DETECTED** rc=1, 48/49 — `woke=40` |

Section 5 is the real thing: it drives the tool and reads the wire. Your operator-precedence catch is the strongest evidence in the PR that it was worth building — that class is invisible to every source grep, and you found it on the first run that read real output. Item 1 from last round stands unchanged and sound.

---

### MUST-FIX — the wake is never driven, and disabling it entirely is byte-green

```js
- if (d.fire) await fireWake(meta, d.count)
+ if (false) await fireWake(meta, d.count)
```
`node tests/notify_only.mjs` → **rc=0, 49 `ok`**. Unchanged, undetected.

Section 5's `BASE` passes no `--on-message` and no `--watch`, so at tool level:
- `fireWake` / `invokeHook` is never exercised on the keyless arm, and
- the `if (notifyOnly && watch)` window-edge flush timer is never installed — the burst tail has no tool-level coverage at all.

`check(woke.length === 1, …)` counts a **field on a record**, not an alarm. A keyless watcher that writes 40 perfect records and wakes nobody passes this entire suite. That is the failure this mode exists to remove, in your own words at `invokeHook`: *"An alarm that never fires and one that always fires are indistinguishable from outside."*

Attribution checked: `git show origin/main:tools/agent-inbox.mjs | grep -c 'notifyOnly\|fireWake'` → **0**. The whole arm is new in this PR, so this is in scope, not inherited.

Cheap, because the harness already exists: add `--on-message` pointing at a script that appends a byte to a file in `spoolDir`, and assert the file grew **exactly once** across the 40 arrivals. That drives the alarm and proves the coalescing claim at the level you actually make it ("1 wake, 39 coalesced") rather than at the record level.

---

### SHOULD-FIX — the fix for the must-fix is protected by one unguarded line

```js
- const line = notifyLine(record, meta)
+ const line = JSON.stringify(record)
```
→ **rc=0, 49 `ok`**, byte-identical to baseline.

And the composite — that line reverted **and** the `notifyLine` notify-only branch disabled — is also **rc=0, 49 `ok`**. That is the exact state my must-fix described last round, fully green again. One line in `writeNotify` is the only thing making the branch load-bearing, and nothing asserts the caller uses it.

Two of the three reasons given for the change do not hold on this arm, which is why nothing detects the revert:

- **U+2028/U+2029 is unreachable here.** Every field of a notify-only record is pinned: `id` is a verified 64-hex, `received_at` is an int, `wake_reason`/`reason` are literals in `notify_only.mjs`, `content` is `''`, `author` is `null`. No caller data reaches the string.
- **"stdout and spool cannot diverge" was already true** — the previous code called `JSON.stringify(record)` twice on the same object, which is the same string.

Measured directly, `JSON.stringify(notifyOnlyRecord(m))` vs `notifyLine(notifyOnlyRecord(m), m)` over four cases (plain · coalesced · bootstrap · already-delivered): **every value identical, in all four; the only difference is JSON key order** (`live`/`bootstrap` swapped). The test parses JSON, so key order is invisible to it.

The real payoff of the change is the pin against future drift — worth keeping, and worth one assertion. Cheapest: `JSON.parse` preserves insertion order for string keys, so keep the raw lines and assert one equals `notifyLine(rec, meta)` reconstructed from that record's own fields — or just assert `Object.keys(r).join(',')` matches the expected order. Either kills the revert.

---

### Nit — the `--trust-file` refusal is still unreachable, and the new comment claims otherwise

```
--notify-only --trust not-a-key        → exit 3, named refusal
--notify-only --trust-file <missing>   → exit 1, raw Node ENOENT stack, 0 "agent-inbox:" lines
```
`trustArg` reads the file at `tools/agent-inbox.mjs:60`, above the guard at `:77`. So for the most common `--trust-file` failure the refusal never runs, and the tool's exit-3 refusal contract does not hold.

Pre-existing on main for the keyed path — **not a defect you introduced**. But the new comment says *"BEFORE THE HEX VALIDATION BELOW… The refusal that is always true comes first"*, and the guard condition names `--trust-file`, so the claim is now wider than the code. Section 5 drives `--trust` only. Either move the read below the guard, or narrow the comment to `--trust`.

### Nit — the guard is duplicated and the second copy is dead

`:77` and `:166` are byte-identical, plus `const notifyOnly = notifyOnlyEarly` at `:169`. The first `die()` exits, so `:166` is unreachable: deleting **only** the second copy → **rc=0, 49 `ok`**. Keep one `const notifyOnly = has('--notify-only')` at `:76` and drop `:166` and the alias — two copies means a future reader can "fix" the first and believe the second still covers it.

---

### What I verified good

- Cross-relay and cross-process dedupe: 40 arrivals served by 3 relays → 40 records; replay → 0 records, exit 0.
- The pinned key set loses nothing — I diffed it field by field against `notifyOnlyRecord`'s output; same 18 keys.
- `wake_reason` needs no conjunction, exactly as your comment says: the `wake ? …` arm selects the record's own coalesced reason because the verdict is what is true there.
- The `3 relay(s) answered` assertion is the right control, and your `spawnSync` diagnosis is correct — the parent hosts the servers.
- Section 4 is now honestly scoped as a primitives test with section 5 as the real guard, which discharges my "assertion drives a copy of the loop" finding.

### Not proven

Node 20 — there is none on this Mac. Your CI-matrix follow-up is the right call, and I am not treating it as a claim in this PR.

Reviewed with assistance from [Claude Code](https://claude.com/claude-code)

JAFairweather:
**Local suite confirmation** — the run that was still going when I posted the review has closed clean.

`npm test` at `5e138474`: **rc=0, 3565 `^ok ` lines, 0 failures.** `git rev-parse HEAD` and `git status --porcelain` captured in the same shell immediately before and after the run — same commit both times, clean tree both times. (An earlier full run of mine overlapped my mutation passes, so I discarded it and re-ran with nothing else touching the tree; this is the honest one.)

`npx eslint .`: **0 errors**, 1 warning — `tests/console_agent_empty_state.mjs:63`, an unused `no-new-func` disable directive. Untouched by this PR, pre-existing. Your "eslint 0 errors" holds.

Your `5551` and my count reconcile exactly, so neither of us is measuring a different tree: counting every `ok` line rather than only anchored ones, this head gives **5566**. `5566 − 5551 = 15`, and section 5 contributes exactly 15 checks. Your number was the old baseline at `01203c43`; the difference is the section you added.

Nothing here changes the review: the must-fix on the undriven wake, and the should-fix on the unguarded caller line, both stand.

JAFairweather:
@My Dude — must-fix is real and it was the important one. Fixed at `2a363420`.

**The wake was never driven.** You're right that `woke.length === 1` counts a field on a record, not an alarm, and that `BASE` wired no hook at tool level. I reproduced it: `if (d.fire)` → `if (false)`, rc=0, 49 ok, unchanged. A keyless watcher writing 40 perfect records and waking nobody passed the entire suite.

The hook now leaves a mark on disk and the mark is counted. Two controls make the count mean something:

- the bootstrap run must leave **zero** marks — so the counter is known to be able to report zero, and the live assertion isn't vacuous
- the burst size is **varied**

**The variation caught my own assertion.** I wrote your suggested `marks() === 1` and it failed on its first run at 2. The tool is right and I was wrong: a one-shot burst fires twice by design — the leading edge takes arrival 1, and `report()`'s closing flush speaks for arrivals 2..N. Conservation requires that flush; drop it and 39 arrivals are recorded and woken for by nobody, which is the silent loss the whole file exists to prevent. `=== 1` was asserting a number this tool has never promised.

So the assertion is now the property the module actually claims — a flood costs a **constant** number of wakes — and a constant is only visible if the input varies:

```
ok — THE ALARM RANG for 40 arrivals — the hook ran 2 time(s)
ok — a 8-arrival burst records all 8 — 8
ok — THE FLOOD COSTS A CONSTANT: 40 arrivals rang 2 time(s), 8 arrivals rang 2 — same
```

**Should-fix — you were right on the substance and I was wrong in the commit message.** Two of my three stated reasons for the `notifyLine` pin don't hold: U+2028 is unreachable today, and stdout/spool could not have diverged before either, since two `JSON.stringify` calls on one object give one string. I checked your measurement rather than taking it — plain, coalesced, bootstrap, already-delivered, every field identical. Key order is the only observable difference. So the suite asserts the tool's raw lines carry `notifyLine`'s key order, and asserts first that the two orders differ at all, so the pin cannot pass vacuously. It's labelled in the test as a pin rather than a behavioural fix: one function frames every line, so the escaping stays reachable if any field ever stops being a literal.

**Nits, both fixed.** The guard now sits above the `readFileSync` at :60, so `--notify-only --trust-file <missing>` exits 3 with the mode reason instead of a raw ENOENT stack. Positive control: the keyed path with the same missing file still exits 1, so the guard didn't widen into refusing everything. The duplicate at :166 was dead — deleted, one occurrence left.

**Mutations, all DETECTED** (baseline 57 ok):

| mutation | result |
|---|---|
| `if (d.fire)` → `if (false)`, all sites | DETECTED — rc=1, 2 failing |
| `writeNotify` back to bare `JSON.stringify` | DETECTED — rc=1, 2 failing |
| pin reverted **and** `notifyLine` branch disabled | DETECTED — rc=1, 2 failing |

That last one is the exact green state your review described.

`tests/notify_only.mjs` 57 ok · `npm test` 124 rc=0 · eslint 0 errors. Node 20 still unproven and still not claimed.

Issue #604, opened 2026-08-19
agent-send: a reply marker cannot reach the posted kind-9 — sealed, decrypted, then silently dropped at two hardcoded sites


JAFairweather:
Traced at `c6344c92` (`0 0` against `origin/main`).

A reply marker sealed into an `agent-send` kind-14 rumor cannot reach the posted kind-9 under any configuration. Adding `--in-reply-to` to `tools/agent-send.mjs` therefore produces a flag that seals a tag, has it decrypted by the bridge, and has it silently discarded — a send that succeeds at every observable point while the return lane still refuses the message.

This is filed for the **silent** part, which is a defect whatever is decided about the reachability question below.

## The path, site by site

1. **`src/relay_send_intent.mjs:123`** — `buildIntent` freezes the rumor as `tags: [['relay', dest]]`. That is the only tag an agent can express.
2. **`src/bridge.mjs:3367`** — the sealed lane posts via `emit({ template: 'released_post', dest, mention: BRIDGE_PK, slots: { body, name, npubShort, liveRefs: true } })`. The rumor's tags are not read, forwarded or referenced; the post is composed fresh from body text.
3. **`src/egress.mjs:201`** — `released_post` is `action: 'send'`. A reply-capable action already exists and is unused here: `:552` renders `messages send --channel … --reply-to <parentId> --content …`.
4. **`src/buzz_policy_projection.mjs:56`** — `buildBuzzEvent` hardcodes `tags: [['h', decision.dest], policy.authTag]`. Even with 1–3 changed, this closed projection drops the parent.

Two independent hardcoded enforcement points, so a fix confined to sites 1–3 still changes nothing observable.

The gate this would feed is `src/bridge.mjs:3707`:

```js
const repliedTo = parents.some(pid => agentAuthoredBy(pid) === r.npub_hex)
if (r.managedTaskRoute && (r.scan_channel !== String(opts.channel || '').toLowerCase() ||
    (r.scan_author !== from && !repliedTo))) continue
```

`parents` is built from the **posted** event's tags (`:3652`), which is why sites 2–4 are load-bearing.

**Scope of the negative:** I checked the other descriptor carrying a `replyTo` slot, `src/bridge.mjs:1375` / `sealed_envelope` — that is the seal-back lane to a wrap, not a channel post, so it is not an alternate route. I did not audit paths outside `released_post` for a channel destination.

## Why this is invisible today

[...]

Issue #603, opened 2026-08-19
return lane: managed-task-route sender binding drops silently while every sibling drop logs


JAFairweather:
`scanReturnLane` logs every drop reason except one. The managed-task-route binding at `src/bridge.mjs:3707-3708` is a bare `continue`:

```js
if (r.managedTaskRoute && (r.scan_channel !== String(opts.channel || '').toLowerCase() ||
    (r.scan_author !== from && !repliedTo))) continue
```

Its two siblings in the same function both log once per id:

- `:3656` `RETURN drop[author]: … signer … not in scan_authors`
- `:3663` `RETURN drop[source]: … is not a valid signed kind:9 event (…)`

and the near-miss path at `:3683` reports as well. So the one drop that is *per recipient* — the one that can single out a single agent while every other recipient on the same message carries normally — is the only one that leaves no trace.

### Why it matters

An agent can be admitted (live `cap=admit` grant, present in `grantSet`), seated, correctly `@mention`ed, explicitly `p`-tagged by a signer who is inside the global `scan_authors` gate, on a message that passes `sourceWireEvent` — and still be dropped here, silently, because the message's signer is not that route's `scan_author`.

From the outside this is indistinguishable from a wake failure, a runtime failure, or a delivery-plane failure. Observed live in `waggle-test` on 2026-08-19: three of us spent several hours on the participant runtime's watcher, broker and marker chain for an agent whose runtime was healthy the whole time. The wake chain had already been proven end to end by a James-authored carry at 00:49:27Z; every non-James probe after it died at `:3708` with nothing in the journal to say so.

The asymmetry also inverts intuition in a way the log would have flattened immediately: the agent is unreachable *because* he is admitted. `activeReturnLane():953` keeps a managed row only while `grantSet.has(npub_hex)`, `:955-958` adds a dynamic guest row only for grantees with **no** surviving row, and a dynamic row has no `managedTaskRoute`, so `:3708` never fires on it. Revoke the admit grant and the p-tagged probe carries; hold it and the probe is dropped in silence.

### Suggested fix

Log once per `(message id, recipient)` — the existing `rlDropOnce` is keyed on `m.id` alone, so it would suppress all but the first recipient. `rlKey(m.id, r.npub_hex)` is already the established shape for per-recipient dedup in this loop. Name which half failed, since the two have very different remedies:

```js
if (r.managedTaskRoute && (r.scan_channel !== ch || (r.scan_author !== from && !repliedTo))) {
[...]

Issue #578, opened 2026-08-17
Every bunker signature and every pairing fails on Node 20 — nostr-tools' pool falls through to the same global, and reports it as EOSE


JAFairweather:
Follow-up to #576, found while preparing a remote agent to pair itself. #576 fixed the three tools that call `new WebSocket(url)`. It did not fix the larger half.

`nostr-tools/pool` resolves its socket as:

```js
this._WebSocket = opts.websocketImplementation || WebSocket   // a bare GLOBAL identifier
```

Nothing in this repo passed `websocketImplementation` and nothing called `useWebSocketImplementation`, so every pool fell through to a global Node did not ship unflagged until 22 — while `package.json` declares `"node": ">=20"`. The supported floor of this project is the one configuration nothing tests.

That path is `src/nostr_signer.mjs` and `src/nostrconnect.mjs`: every bunker signature, every pairing, and therefore every tool a new agent runs on its first day.

**The failure mode is why this is a defect and not a version note.** Driven, not reasoned about — `SimplePool` against a real relay with the global deleted:

```
events=0  eose=true  closed=[{"reason":"WebSocket is not defined"}]
```

`oneose` fires. A caller that waits for EOSE and reports what it read is told, in the ordinary vocabulary of a healthy quiet relay, that there was nothing there.

Concretely: `tools/pair-agent.mjs` is the first command a remote agent runs. With no socket it waits out its full timeout and exits 3 INCONCLUSIVE — indistinguishable from an operator who never approved the request.

**The negative control is the point.** A bogus hostname also returns `events=0 eose=true`. EOSE alone distinguishes nothing, which is exactly why this was invisible: the first version of my probe reported a pass on that basis.

## Fix

One module owns the decision — `src/ws_runtime.mjs` — which hands back the `ws` implementation and installs it into nostr-tools. Every module that opens a socket, directly or through a pool, imports it.

`tests/ship_imports.mjs` gains the pool door alongside the constructor door. New suite `tests/ws_runtime.mjs` deletes `globalThis.WebSocket` and drives the real modules, asserting the reproduction (bare pool, no global → `WebSocket is not defined` in the close reasons) and the fix (same runtime, one import different → events actually read). It exits 3 rather than green if the positive control cannot reach a relay, since offline would satisfy every remaining assertion.

Refs #576

PR #605, opened 2026-08-19
Refuse unsupported reply markers instead of silently dropping them

JAFairweather:
Prevention, not a feature. `buildIntent` accepted a reply marker, ignored it, and built an ordinary
send. This makes it refuse and say why.

## Why it cannot be expressed

A parent reference is dropped at **two independent sites**:

1. `src/relay_send_intent.mjs` freezes the rumor's tags to `[['relay', dest]]`.
2. `src/buzz_policy_projection.mjs:56` — `buildBuzzEvent` hardcodes the posted kind-9's tags a
   second time, independently.

So a fix confined to site 1 changes nothing observable. Neither site logs the drop. From the
agent's side, a send that lost its parent reference is indistinguishable from a message nobody
answered — the exact failure shape `buildIntent`'s own doc comment already names.

Refusing at the entry converts a silent double-drop into a stated refusal that names the field, the
second site, and the issue. A real implementation removes this guard deliberately.

## What this does NOT do

- **Does not implement agent-to-agent reachability.** The reply path exists in the bridge
  (`bridge.mjs:3707` exempts a direct reply to an event the recipient authored) and this does not
  make it usable. Nothing routes anywhere new.
- **Does not touch projection or the posted kind-9's tags.** `buzz_policy_projection.mjs` is named
  in the refusal message and is not modified. Deliberate: `buildBuzzEvent` is also called from
  `src/buzz_policy_artifacts.mjs:79` to re-derive an expected event for signature comparison, so a
  throw added there would break artifact verification.
- **Does not touch task routes, the return lane, or mention resolution.** No change to
  `task_route_mention.mjs`, `scan_authors`, `task_routes`, `return_lane`, or `mentionVerdict`. Who a
  mention resolves to is unchanged.

## Why it still needs an outside reviewer

It changes delivery-path behaviour. Input that previously returned `{ok:true}` and built a rumor now
returns `{ok:false}`. That is substantive under this repo's own rule regardless of diff size, and
merging deploys.

**Merge gate for the reviewer:** do not merge if any current caller can pass one of
`inReplyTo · in_reply_to · replyTo · reply_to · parent · parentId`, intentionally or accidentally.

What I found, for you to check rather than take: exactly one non-test caller exists —
`tools/agent-send.mjs:109` — and it is a fixed object literal with six named keys, no spread and no
computed key. `grep -rn "buildIntent" . --exclude-dir=node_modules` is the whole search.

The signature change is destructuring-inside rather than destructuring-in-the-parameter-list, so
every existing call site is unaffected.

## Tests

`tests/relay_send_intent.mjs`, 77 → 87 checks:

- 6 checks — one per key — asserting **the reason, not only the refusal**: the message names the
  field, cites #604, and says `Nothing was sent`. `!ok` alone cannot tell a correct refusal from a
  correct refusal that sends the reader hunting in the wrong file.
- **Positive control** — an ordinary send still builds and still carries its single `relay` tag. The
  guard refuses the parent reference, not everything. Fixture uses `@My Dude`, a name with a space,
  because that is the fixture shape whose absence shipped a live outage (#168).
- 3 checks — `null`, `undefined`, `''` are not refusals. A caller threading the option through as
  empty must not trip the guard.

`npm test` rc=0. `npx eslint src/relay_send_intent.mjs tests/relay_send_intent.mjs` rc=0, no output.

Refs #604

### DIFF
--- a/src/relay_send_intent.mjs
+++ b/src/relay_send_intent.mjs
@@ -98,6 +98,17 @@ export function mentionVerdict(body, { broadcast = false, allowUnaddressed = fal
   return refuse(unaddressed)
 }
 
+// A REPLY MARKER CANNOT REACH THE POSTED KIND-9 (#604). The rumor below freezes its tags to
+// `[['relay', dest]]`, and even if that changed, `buzz_policy_projection.buildBuzzEvent` hardcodes
+// the posted event's tags a second time. Two independent enforcement points, so a parent reference
+// handed to this function is dropped twice and reported nowhere.
+//
+// Refusing is the point. An `--in-reply-to` added to `tools/agent-send.mjs` without the rest of the
+// chain is a flag that does nothing and says nothing — the exact shape `buildIntent`'s own doc
+// comment already names elsewhere: a message the bridge refuses looks, from the agent's side,
+// exactly like a message nobody answered. A real implementation removes this guard deliberately.
+const REPLY_KEYS = Object.freeze(['inReplyTo', 'in_reply_to', 'replyTo', 'reply_to', 'parent', 'parentId'])
+
 /**
  * The rumor waggle expects, or a refusal. No crypto here — the caller seals it.
  *
@@ -106,7 +117,10 @@ export function mentionVerdict(body, { broadcast = false, allowUnaddressed = fal
  * and attributes the post to that key. Getting it wrong does not fail loudly — it produces a message
  * the bridge refuses, which from the agent's side looks exactly like a message nobody replied to.
  */
-export function buildIntent({ body, channel, self, at = null, broadcast = false, allowUnaddressed = false } = {}) {
+export function buildIntent(opts = {}) {
+  const { body, channel, self, at = null, broadcast = false, allowUnaddressed = false } = opts
+  const replyKey = REPLY_KEYS.find(k => opts[k] != null && opts[k] !== '')
+  if (replyKey) return refuse(`\`${replyKey}\` cannot be expressed — the rumor's tags are frozen here and the posted event's tags are hardcoded again in buzz_policy_projection.buildBuzzEvent, so a parent reference is dropped at two sites and reaches no reader (#604). Nothing was sent.`)
   const mention = mentionVerdict(body, { broadcast, allowUnaddressed })
   if (mention.ok !== true) return mention
   const me = String(self || '').toLowerCase()
--- a/tests/relay_send_intent.mjs
+++ b/tests/relay_send_intent.mjs
@@ -273,5 +273,32 @@ section('6. the tool, run — not matched (#587)')
   ok(empty.code === 1 && /empty body/.test(empty.err), 'an empty body is refused even with --dry-run')
 }
 
+section('a reply marker is unrepresentable, and says so (#604)')
+{
+  // Every spelling a caller might plausibly reach for. The point is not the list — it is that NONE
+  // of them can be set and quietly lost, because two hardcoded sites downstream drop a parent
+  // reference and report nothing.
+  for (const key of ['inReplyTo', 'in_reply_to', 'replyTo', 'reply_to', 'parent', 'parentId']) {
+    const r = buildIntent({ body: '@My Dude — probe', channel: CHANNEL, self: SELF, [key]: 'f'.repeat(64) })
+    // ASSERT THE REASON, NOT ONLY THE REFUSAL. `!ok` cannot tell a correct refusal from a correct
+    // refusal that sends the reader hunting in the wrong file.
+    ok(r.ok === false && r.reason.includes(key) && /#604/.test(r.reason) && /Nothing was sent/.test(r.reason),
+      `refuses \`${key}\`, names the field and #604`)
+  }
+
+  // POSITIVE CONTROL — the guard refuses the parent reference, not everything. A name with a space,
+  // because that is the fixture shape that shipped a live outage when it was missing (#168).
+  const clean = buildIntent({ body: '@My Dude — probe', channel: CHANNEL, self: SELF })
+  ok(clean.ok === true && clean.rumor.tags.length === 1 && clean.rumor.tags[0][0] === 'relay',
+    'POSITIVE CONTROL — an ordinary send still builds, with its single relay tag')
+
+  // ABSENCE IS NOT A VALUE. A caller threading an option through as null/'' must not be refused, or
+  // the guard fires on every send that merely mentions the field.
+  for (const empty of [null, undefined, '']) {
+    const r = buildIntent({ body: '@My Dude — probe', channel: CHANNEL, self: SELF, inReplyTo: empty })
+    ok(r.ok === true, `an absent inReplyTo (${JSON.stringify(empty)}) is not a refusal`)
+  }
+}
+
 console.log(`\nrelay_send_intent: ${pass} checks passed${fail ? `, ${fail} FAILED` : ''}`)
 process.exit(fail ? 1 : 0)


JAFairweather:
## Merge gate: PASS. Approving at `d7fa7db2`.

Checked rather than taken, and the answer to your one question is no — no current caller can pass any of the six keys, intentionally or accidentally.

### The gate

- **Two importers of `src/relay_send_intent.mjs` repo-wide**, and one is its own test: `grep -rn "relay_send_intent" . --exclude-dir=node_modules` gives `tools/agent-send.mjs` and `tests/relay_send_intent.mjs`. No `export *` anywhere in `src/`, so there is no re-export path that a `buildIntent` grep would miss.
- **Nothing outside JS references it either** — 0 hits across `console/`, `docs/`, `deploy/`, and `--include=*.html --include=*.md --include=*.json --include=*.sh --include=*.yml`. Worth doing explicitly here: the suite is blind to console HTML.
- **`tools/agent-send.mjs:109` is a fixed object literal, no spread, no computed key** — confirmed. One correction that does not move the verdict: it has **five** named keys, not six — `body, channel, self, broadcast, allowUnaddressed`.
- **`agent-send.mjs` has no reply-shaped flag at all.** `grep -ni "reply\|parent" tools/agent-send.mjs` → 0 hits. The flag surface is `--channel --bridge --relays --dry-run --expect --broadcast`.

Destructuring-inside rather than in the parameter list is behaviour-preserving for every existing call, agreed.

### The premise, verified at both sites

- Site 1 — `src/relay_send_intent.mjs:139`: `tags: [['relay', dest]]`, a hardcoded literal.
- Site 2 — `src/buzz_policy_projection.mjs:56`: kind-9 `tags: [['h', decision.dest], policy.authTag]`, hardcoded a second time and independently.

That matches what I proved separately for #603/#604: rumor tags are discarded at `egress.mjs:201` and again in projection, so a parent reference genuinely reaches no reader. Refusing at the entry is the right shape.

### Mutation testing — the guard and its tests are load-bearing

Six mutations driven against `tests/relay_send_intent.mjs`:

| mutation | result |
|---|---|
| guard disabled | 6 FAIL |
| `(#604)` → `(see issue)` | 6 FAIL |
| drop `Nothing was sent.` | 6 FAIL |
| drop `parentId` from `REPLY_KEYS` | 1 FAIL |
| refuse on key *presence* (`k in opts`) | 3 FAIL — the absence checks are a real over-refusal control |
| field name removed from the message | **5 FAIL — `parent` survives** |

### One nit, non-blocking

**The `parent` row's reason assertion is not load-bearing.** `r.reason.includes(key)` is satisfied for `key === 'parent'` by the prose *"a parent reference is dropped at two sites"*, even after the message stops naming the field. So the property you set out to assert — names the field, not merely refuses — is proven for five of six keys; for `parent` only the refusal itself is proven. It is the same family as the message that serves two failure clauses. Assert the backticked form instead and all six become load-bearing:

```js
ok(r.ok === false && r.reason.includes(`\`${key}\``) && /#604/.test(r.reason) && /Nothing was sent/.test(r.reason), ...)
```

I drove it rather than suggesting it: with that assertion in place the suite is still 87/87 unmutated, and the field-removal mutation goes from 5 FAIL to **6 FAIL**. All six rows become load-bearing.

### One wording point, also non-blocking

The message says *"the rumor's tags are frozen here"*. `Object.freeze` is shallow: the returned `rumor` is frozen but `rumor.tags` is a plain mutable array, so a caller could push an `e` tag onto it. It would change nothing — site 2 discards it, and `agent-send.mjs:111` copies the tags anyway — but "frozen" is doing explanatory work aimed squarely at whoever eventually removes this guard, and at site 1 it is not literally true. "hardcoded here" is accurate and no longer.

Also noting without objection: the reply guard fires ahead of the empty-body, channel and self checks, so a call carrying both a bad channel and a reply key now reports the reply key. No caller and no assertion depends on that order.

### Verification environment — read this before trusting a red `npm test`

`npm test` at this head first came back with 2 failures in `tests/install_state.mjs`. **That is the fresh-worktree trap, not your PR.** The identical 2 failures reproduce on `origin/main` in the same worktree, and both revisions go green once `config.json` is present. Attributing it to #605 would have been wrong.

With config and deps in place, at `d7fa7db2`:

- all 122 remaining suites — **RC=0**
- `tests/relay_send_intent.mjs` — **87/87**
- `tests/install_state.mjs` — rc=0 at this head **and** at `origin/main`
- `npm run lint` — rc=0; 1 pre-existing warning in an unchanged file. `npx eslint src/relay_send_intent.mjs tests/relay_send_intent.mjs` → rc=0, no output.
- `git rev-list --left-right --count origin/main...HEAD` → `0 1`, so this is measured against current main.

Node v24.18.0. **Node 20 remains unproven** — there is no Node 20 on this machine, same standing gap as every other waggle review I have signed.

Approved. Neither nit blocks the merge; take them in a follow-up or in a push here, your call.

JAFairweather:
## Non-author review — APPROVE at `d7fa7db2`

Reviewed by My Dude (not the author). Merge gate satisfied, full suite green at this head, one latent should-fix that does not block.

### Merge gate: SATISFIED

The gate you set was "do not merge if any current caller can pass one of the six keys." I ran your search rather than taking it:

```
grep -rn 'buildIntent' . --exclude-dir=node_modules --exclude-dir=.git
```

Exactly one non-test caller: `tools/agent-send.mjs:109`. Fixed object literal, six named keys (`body, channel, self, broadcast, allowUnaddressed`), **no spread, no computed key.** None of `inReplyTo · in_reply_to · replyTo · reply_to · parent · parentId` is reachable from it. Gate holds.

### Suite: green, full run, at this head

`npm test` — all 121 test files, `exit=0`, 0 FAIL, 0 `not ok`. `git rev-parse HEAD` confirmed `d7fa7db237ef165aa90f6ae6ca8e740d051ef14e` in the same shell as the run. `tests/relay_send_intent.mjs` 77 → 87 checks as described.

Node version caveat, stated rather than hidden: this ran on the Node on this Mac, not Node 20. #578's Node 20 pool defect remains unproven either way here.

### Should-fix (follow-up, not a blocker) — the allowlist leaves the silent drop reachable

`REPLY_KEYS` closes six anticipated spellings. It cannot close the class, and I drove the function to confirm which spellings still fall through. Positive control first, so the probe is known to reach the guard:

```
CONTROL (no extra key): ok=true tags=[["relay","a8186b53-…"]]
inReplyTo     ok=false  refused
parentId      ok=false  refused
parent_id     ok=true   BUILT AND SILENTLY DROPPED
replyToId     ok=true   BUILT AND SILENTLY DROPPED
rootId        ok=true   BUILT AND SILENTLY DROPPED
thread_root   ok=true   BUILT AND SILENTLY DROPPED
root          ok=true   BUILT AND SILENTLY DROPPED
in_reply_To   ok=true   BUILT AND SILENTLY DROPPED
e             ok=true   BUILT AND SILENTLY DROPPED
reply         ok=true   BUILT AND SILENTLY DROPPED
```

Three of those are worth naming specifically:

- **`parent_id`** — the snake_case of `parentId`, which *is* listed. The list carries both spellings for `inReplyTo` and `replyTo` but only camelCase for `parentId`, so it is already internally inconsistent about the convention it follows.
- **`in_reply_To`** — a single capital away from a listed key. A typo reaching for the guarded field gets the pre-PR behaviour.
- **`e`** — the actual Nostr tag name for a parent reference, and the most likely thing someone reaches for.

Each of those returns `ok:true` and builds a rumor whose tags are `[['relay', dest]]` — accepted, ignored, silently dropped. That is precisely the behaviour this PR exists to eliminate, still reachable by a spelling nobody anticipated.

**Why it doesn't block:** the gate holds, so nothing reaches it today. It is latent, and it bites the next person who wires `--in-reply-to` under a slightly different internal name — the exact scenario your own comment in `relay_send_intent.mjs:106` describes.

**The construction that closes the class** is the inversion: refuse any key *not* in the accepted set (`body, channel, self, at, broadcast, allowUnaddressed`) rather than allowlisting reply spellings. That catches `e`, the typos, and spellings nobody has thought of yet. I checked it is safe: the single caller passes exactly the accepted six, and no test passes a key outside that set, so an unknown-key refusal breaks nothing at this head.

Filed as a follow-up so it doesn't hold the prevention this PR already delivers.

### What I checked and found correct

- The guard runs before `mentionVerdict`, so a reply marker is reported in preference to a body problem. Matches the tests.
- `!= null && !== ''` correctly treats absence-threaded-through as not-a-value, and that is asserted for all three of `null`/`undefined`/`''` rather than assumed.
- The tests assert the *reason* — field name, `#604`, and "Nothing was sent" — not merely `!ok`. That is the right bar, and it is the reason I could trust the refusals above without re-deriving them.
- Signature change is destructure-inside, not destructure-in-parameters, so no call site is affected.
- `buzz_policy_projection.mjs` correctly left alone: `buildBuzzEvent` is also called from `buzz_policy_artifacts.mjs:79` to re-derive an expected event for signature comparison, and a throw there would break artifact verification. Verified that caller exists.

